PR #118, opened 2019-11-27
Analytics configuration and record event api

desokroshan:
*Issue #, if available:*

*Description of changes:*
Must address the following : 
1. addressing https://github.com/aws-amplify/amplify-android/pull/106#discussion_r350387943
2. https://github.com/aws-amplify/amplify-android/pull/106#discussion_r350371745
3. Adding implementation for new methods in AnalyticsCategory.java 


aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -72,61 +85,179 @@ public void disable() {
...
+        analyticsClient.submitEvents();
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public String getPluginKey() {
-        return null;
+        return "amazonPinpointAnalyticsPlugin";
```
jd-lacey: I believe this should be "awsPinpointAnalyticsPlugin" as that is the name of the key that is added to amplifyconfiguration.json by `amplify add analytics` command.  This causes an NPE in `configure` because it's trying to fetch the configuration using the wrong key name.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointProperties.java
```diff
@@ -0,0 +1,74 @@
...
+
+        Builder add(String key, String value) {
+            attributes.put(key, value);
+            return this;
+        }
+
+        Builder add(String key, Double value) {
+            metrics.put(key, value);
+            return this;
+        }
```
jd-lacey: These methods need to be public.

PR #153, opened 2019-12-03
[Analytics] Add support for global event properties

desokroshan:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
+++ b/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
@@ -24,7 +24,6 @@ import com.amplifyframework.analytics.BasicAnalyticsEvent;
 import com.amplifyframework.analytics.pinpoint.test.R;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.AmplifyConfiguration;
-import com.amplifyframework.hub.HubException;
 import com.amplifyframework.logging.Logger;
 import com.amplifyframework.testutils.Sleep;
 
@@ -41,6 +40,8 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Validates the functionality of the {@link AmazonPinpointAnalyticsPlugin}.
@@ -142,12 +143,10 @@ public class AnalyticsPinpointInstrumentedTest {
     /**
      * Registers a global property and ensures that all recorded events have the global property.
      * @throws AnalyticsException Caused by incorrect usage of the Analytics API.
-     * @throws HubException thrown if an error condition is encountered while registering global properties
-     *                      and Amplify fails to publish the error message to the hub.
      * @throws JSONException Caused by unexpected event structure.
      */
     @Test
-    public void testRegisterGlobalProperties() throws AnalyticsException, HubException, JSONException {
+    public void testRegisterGlobalProperties() throws AnalyticsException, JSONException {
         // Register a global property
         registerGobalProperty();
 
@@ -164,20 +163,19 @@ public class AnalyticsPinpointInstrumentedTest {
                 new JSONObject(analyticsClient.getAllEvents().get(1).get("attributes").toString());
 
         assertEquals(2, analyticsClient.getAllEvents().size());
-        assertEquals(true, eventAttributes.has("Property"));
-        assertEquals(true, eventAttributes.has("GlobalProperty"));
-        assertEquals(false, event2Attributes.has("Property"));
-        assertEquals(true, event2Attributes.has("GlobalProperty"));
+        assertTrue(eventAttributes.has("Property"));
+        assertTrue(eventAttributes.has("GlobalProperty"));
+        assertFalse(event2Attributes.has("Property"));
+        assertTrue(event2Attributes.has("GlobalProperty"));
     }
 
     /**
      * Registers a global property and then Unregisters the global property
      * and ensures that it is respected in events recorded thereafter.
-     * @throws HubException thrown if an error condition is encountered while registering global properties
-     *                      and Amplify fails to publish the error message to the hub.
+     * @throws AnalyticsException Caused by incorrect usage of the Analytics API.
      */
     @Test
-    public void testUnregisterGlobalProperties() throws HubException {
+    public void testUnregisterGlobalProperties() throws AnalyticsException {
         // Register a global property
         registerGobalProperty();
 
@@ -189,10 +187,10 @@ public class AnalyticsPinpointInstrumentedTest {
         Amplify.Analytics.recordEvent("amplify-test-event-without-property");
 
         assertEquals(1, analyticsClient.getAllEvents().size());
-        assertEquals(false, analyticsClient.getAllEvents().get(0).has("attributes"));
+        assertFalse(analyticsClient.getAllEvents().get(0).has("attributes"));
     }
 
-    private void registerGobalProperty() throws HubException {
+    private void registerGobalProperty() throws AnalyticsException {
         // Register a global property
         Amplify.Analytics.registerGlobalProperties(PinpointProperties.builder()
                 .add("GlobalProperty", "globalVal")
--- a/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
+++ b/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
@@ -28,9 +28,14 @@ import com.amplifyframework.logging.Logger;
 import com.amplifyframework.testutils.Sleep;
 
 import com.amazonaws.mobileconnectors.pinpoint.analytics.AnalyticsClient;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.util.HashSet;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
@@ -47,6 +52,8 @@ public class AnalyticsPinpointInstrumentedTest {
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-analytics");
     private static final int EVENT_FLUSH_TIMEOUT = 60;
     private static final int EVENT_FLUSH_WAIT = 2;
+    private static AmazonPinpointAnalyticsPlugin plugin;
+    private static AnalyticsClient analyticsClient;
 
     /**
      * Configure the Amplify framework.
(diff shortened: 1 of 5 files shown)


aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -59,6 +66,19 @@ public static void setUp() throws AmplifyException {
         configuration.populateFromConfigFile(context, R.raw.amplifyconfiguration);
         Amplify.addPlugin(new AmazonPinpointAnalyticsPlugin());
         Amplify.configure(configuration, context);
+        plugin = (AmazonPinpointAnalyticsPlugin) Amplify
```
kvasukib: AmazonPinpointAnalyticsPlugin plugin = new AmazonPinpointAnalyticsPlugin();
Amplify.configure(plugin);

// Now you can reference it using the reference `plugin`

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -138,6 +140,50 @@ public void testAutoFlush() throws AnalyticsException {
...
     }
 
+    /**
+     * Registers a global property and ensures that all recorded events have the global property.
+     * Unregisters a global property and ensures that it is respected in events recorded thereafter.
+     * @throws AnalyticsException Caused by incorrect usage of the Analytics API.
+     * @throws JSONException Caused by unexpected event structure.
+     */
+    @Test
+    public void testGlobalProperties() throws AnalyticsException, JSONException {
```
kvasukib: if possible refactor this test into multiple smaller tests where each tests asserts on one major condition.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -120,16 +120,29 @@ public void recordEvent(@NonNull AnalyticsEvent analyticsEvent)
...
-        throw new UnsupportedOperationException("This operation has not been implemented yet.");
+    public void registerGlobalProperties(@NonNull Properties properties) throws AnalyticsException {
+        for (Map.Entry<String, Property<?>> entry : properties.get().entrySet()) {
+            if (entry.getValue() instanceof StringProperty) {
+                analyticsClient.addGlobalAttribute(entry.getKey(), ((StringProperty) entry.getValue()).getValue());
+            } else if (entry.getValue() instanceof DoubleProperty) {
+                analyticsClient.addGlobalMetric(entry.getKey(), ((DoubleProperty) entry.getValue()).getValue());
+            } else {
+                throw new AnalyticsException("Invalid property type detected.",
+                        "AmazonPinpointAnalyticsPlugin support only StringProperty or DoubleProperty." +
```
kvasukib: `supports`

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -120,16 +120,29 @@ public void recordEvent(@NonNull AnalyticsEvent analyticsEvent)
...
+                                "Refer tio the docuemntation for details.");
```
kvasukib: `to`

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategory.java
```diff
@@ -94,13 +94,17 @@ public void recordEvent(@NonNull final AnalyticsEvent analyticsEvent)
     }
 
     @Override
-    public void registerGlobalProperties(@NonNull Properties properties) {
-        throw new UnsupportedOperationException("This operation is currently not supported.");
+    public void registerGlobalProperties(@NonNull Properties properties) throws AnalyticsException {
+        if (enabled) {
```
kvasukib: Will there be any error reported if analytics is not enabled?

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -120,16 +120,29 @@ public void recordEvent(@NonNull AnalyticsEvent analyticsEvent)
...
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public void unregisterGlobalProperties(@NonNull Set<String> keys) {
-        throw new UnsupportedOperationException("This operation has not been implemented yet.");
+        for (String key: keys) {
+            analyticsClient.removeGlobalAttribute(key);
```
kvasukib: We need some sort of synchronization primitives to protect against multiple threads doing register and unregister global properties.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -120,16 +120,29 @@ public void recordEvent(@NonNull AnalyticsEvent analyticsEvent)
...
+                                "Refer tio the docuemntation for details.");
+            }
+        }
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public void unregisterGlobalProperties(@NonNull Set<String> keys) {
```
kvasukib: If register throws AnalyticsException, I think unregister should throw exceptions for failure conditions too. Also the 

```
if (enabled)
```

protection is missing here.

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
desokroshan: There are just two tests here registering and unregistering global properties. Unregistering is better tested with some global properties registered, so thought of joining them together.

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategory.java
desokroshan: No we don't report an error currently. Analytics APIs are a no-op if Analytics is disable. I think this allows them to conditionally enable and disable calls without error being reported.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -120,16 +120,29 @@ public void recordEvent(@NonNull AnalyticsEvent analyticsEvent)
      * {@inheritDoc}
      */
     @Override
-    public void registerGlobalProperties(@NonNull Properties properties) {
-        throw new UnsupportedOperationException("This operation has not been implemented yet.");
+    public void registerGlobalProperties(@NonNull Properties properties) throws AnalyticsException {
```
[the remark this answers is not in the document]
desokroshan: registering global properties is synchronous call that does not make a network call or any background processing, so we don't need a callback, but we can consider logging and notifying through hub rather than throwing an exception in case of failure. I'll check ios behavior with Michael.

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategoryBehavior.java
```diff
@@ -68,8 +70,10 @@
      * Examples of global properties would be `selectedPlan`, `campaignSource`
      *
      * @param properties Map of global properties wrapped as an instance of {@link Properties}
+     * @throws HubException thrown if an error condition is encountered while registering global properties
+     * and Amplify fails to publish the error message to the hub.
      */
-    void registerGlobalProperties(@NonNull Properties properties);
+    void registerGlobalProperties(@NonNull Properties properties) throws HubException;
```
jamesonwilliams: If you'll keep a checked exception on this API, then this should be `throws AnalyticsException` instead of `throws HubException`, since this is on an Analytics API.

If a `HubException` occurs in the implementation you should wrap it inside of an `AnalyticsException`, before returning it, here. Like:
```
try {
    Amplify.Hub.publish(....);
} catch (HubException publishFailure) {
    throw new AnalyticsException("Failed to register global properties." publishFailure, "Recovery suggestion TODO...");
}
```

The interface Javadoc could talk about how registration errors are communicated:
```
/**
 * ...
 * If registration fails, a (what kind?) of message is published via {@link Hub}.
 */
```

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -57,8 +65,19 @@ public static void setUp() throws AmplifyException {
         Context context = ApplicationProvider.getApplicationContext();
         AmplifyConfiguration configuration = new AmplifyConfiguration();
         configuration.populateFromConfigFile(context, R.raw.amplifyconfiguration);
-        Amplify.addPlugin(new AmazonPinpointAnalyticsPlugin());
+        plugin = new AmazonPinpointAnalyticsPlugin();
+        Amplify.addPlugin(plugin);
         Amplify.configure(configuration, context);
+        analyticsClient = plugin.getAnalyticsClient();
```
jamesonwilliams: A better test pattern is to allow "injection" of a class' dependency via its constructor. So,

Instead of doing:
```
final class Container {
    private final Inside inside;
    ...
    Container() {
        this.inside = new DefaultInsideImpl();
    }

    @VisibleForTesting
    Inside getInside() {
        return inside;
    }
    ...
}
```
See if you can make this look more like:
```
final class Container {
    private final Inside inside;

    Container() {
        this(new DefaultInsideImpl());
    }

    @VisibleForTesting
    Container(Inside inside) {
        this.inside = inside;
    }
    ...
}
```

Then in your test, you'd have like:
```
@Before
public void setup() {
    this.analyticsClient = // Some method of obtaining one -- I forget how this works, its a factory right?
    this.plugin = new AmazonPinpointAnalyticsPlugin(analyticsClient);
}
```

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -138,6 +139,66 @@ public void testAutoFlush() throws AnalyticsException {
...
+
+        // Unregister global property
+        Set<String> globalPropertyKeys = new HashSet<>();
+        globalPropertyKeys.add("GlobalProperty");
+        Amplify.Analytics.unregisterGlobalProperties(globalPropertyKeys);
+
+        Amplify.Analytics.recordEvent("amplify-test-event-without-property");
+
+        assertEquals(1, analyticsClient.getAllEvents().size());
+        assertEquals(false, analyticsClient.getAllEvents().get(0).has("attributes"));
```
jamesonwilliams: `assertEquals(false, something)` can be simplified as `assertFalse(something)`.

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
desokroshan: analyticsClient is created only when the Analytics is configured. So I don't see a way to obtain analyticsClient before the Analytics plugin is configure other than may be to recreate the configuration steps of Analytics inside the the test setup which feels redundant. Will discuss offline.

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
desokroshan: Done.

PR #162, opened 2019-12-03
Add changelog

kvasukib:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,14 @@
+# Amplify for Android (Beta) - CHANGELOG
+
+A declarative library for application development using cloud services.
+
+## 0.9.0
+
+Initial release! Includes Core features, plus support for these categories:
+
+- Analytics
+- API
+- DataStore
+- Hub
+- Logging
+- Storage


CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+A declarative library for application development using cloud services.
+
+## 0.9.0
+
+Initial release! Includes Core features, plus support for these categories:
+
+- Analytics
+- API
+- DataStore
+- Hub
```
royjit: Hub is implemented? I saw few places this todo like this `//TODO: Dispatch to hub`

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+- Logging
```
raphkim: Are these currently supported?

CHANGELOG.md
kvasukib: Hub as an independent category is supported and implemented. However Hub is not used by other categories.

CHANGELOG.md
kvasukib: We have plugins for those categories which customers can use.

PR #163, opened 2019-12-03
Enable integration tests in CircleCI

desokroshan:
*Issue #, if available:*

*Description of changes:*


.circleci/config.yml
```diff
@@ -16,6 +16,42 @@ commands:
...
+      install aws cli and configure android aws release profile
+    steps:
+      - run:
+          name: install aws cli
+          command: |
+            sudo pip install awscli
+      - run:
+          name: configure aws profile
+          command: |
+            aws configure --profile amplify_sdk_test set region us-east-1
```
kvasukib: Would recommend testing with us-west-1 or us-west-2 as it is closer.

.circleci/config.yml
```diff
@@ -34,3 +71,39 @@ jobs:
           command: bash gradlew publishToMavenLocal
       - store_test_results:
           path: test-results
+
+  integrationtest:
+    working_directory: ~/amplify-android-home
+    docker:
+      - image: circleci/android@sha256:5cdc8626cc6f13efe5ed982cdcdb432b0472f8740fed8743a6461e025ad6cdfc
+    environment:
+      JVM_OPTS: -Xmx3200m
```
kvasukib: Minor: Would be nice to have this as a ENV variable or a configurable parameter to the workflow run instance especially when we get failures due to size limits.

.circleci/config.yml
```diff
@@ -34,3 +71,39 @@ jobs:
...
+    steps:
+      - checkout
+      - generate_gradle_wrapper
+      - configure_aws
+      - setup_emulator
+      - run:
+          name: Copy configuration files
+          command: |
+            aws s3 cp s3://amplify-circleci-android-payload/awsconfiguration.json  aws-analytics-pinpoint/src/androidTest/res/raw/awsconfiguration.json --profile amplify_sdk_test
```
kvasukib: Minor: Later, would be nice to have a map of source and destination and iterate over map and execute the `cp` command for each entry of the map.

.circleci/config.yml
```diff
@@ -34,3 +71,39 @@ jobs:
...
+  version: 2
+  build_test:
+    jobs:
+      - build
+      - integrationtest:
+          filters:
+            branches:
+              only:
+                - master
+                - roshan/circleCI
```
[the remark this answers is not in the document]
desokroshan: Yep, just to make sure integ tests pass on this branch before we merge. Once it passes, I'll make a final commit remove it before merging

.circleci/config.yml
desokroshan: Sounds good. Will udpate

.circleci/config.yml
desokroshan: Yes. A future enhancement is to not have separate config files for each of the module but minimum possible number of config files. I'll create an issue with these enhancements.

PR #164, opened 2019-12-03
[DataStore] Remove API name logic

raphkim:
*Description of changes:*
### Core
* Plugin.configure() no longer requires plugin configuration parameter to be `@NonNull`
  * Plugins of certain categories, such as DataStore and Hub, don't require a configuration
* Null-check responsibility is now transferred from category layer -> plugin layer

### API
* Unit tested changes from #148 
* Made every method return nullable
  * It could always return null (e.g. if provided apiName was not found inside the map), so it is just annotation change
  * Any error related to apiName will now return null after invoking `ResultListener.onError(ApiException)` (as opposed to how it used to be runtime exception)

### DataStore
* Removed every logic related to `apiName`
  * No longer a field inside configuration file
  * Always uses API methods without apiName
    * DataStore is now UNUSABLE if API plugin contains multiple GraphQL API's


aws-api/build.gradle
```diff
@@ -35,6 +35,7 @@ android {
     testOptions {
         unitTests {
             includeAndroidResources = true
+            returnDefaultValues = true
```
jamesonwilliams: **Let's not add this.**

> Caution: Setting the returnDefaultValues property to true should be done with care. The null/zero return values can introduce regressions in your tests, which are hard to debug and might allow failing tests to pass. Only use it as a last resort.

maybe you added this before you put `@RunWith(RobolectricTestRunner.class)` on your test class? That should be enough, no?

core/src/main/java/com/amplifyframework/core/plugin/Plugin.java
```diff
@@ -40,9 +39,9 @@
      * Configure the plugin with customized configuration object.
      * @param pluginConfiguration plugin-specific configuration data
      * @param context An Android Context
-     * @throws AmplifyException when configuration for a plugin was not found
+     * @throws AmplifyException an error is encountered during configuration.
      */
-    void configure(@NonNull JSONObject pluginConfiguration, Context context) throws AmplifyException;
+    void configure(JSONObject pluginConfiguration, Context context) throws AmplifyException;
```
jamesonwilliams: Use `@Nullable` to indicate that the config may be `null` and the developer should check for that.

aws-datastore/src/test/java/com/amplifyframework/datastore/network/AppSyncApiTest.java
```diff
@@ -76,16 +72,16 @@ public void setup() {
      * @throws JSONException On bad request JSON found in API category call
      */
     @Test
-    public void validateBasSyncQueryGen() throws JSONException {
+    public void validateBaseSyncQueryGen() throws JSONException {
```
jamesonwilliams: :-D thank you

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginConfigurationTest.java
```diff
@@ -37,68 +33,32 @@
 @RunWith(RobolectricTestRunner.class)
 public final class AWSDataStorePluginConfigurationTest {
     /**
-     * If no configuration is specified, it's okay, sync mode will default to
+     * If configuration is null, it's okay, sync mode will default to
      * {@link AWSDataStorePluginConfiguration.SyncMode#LOCAL_ONLY}.
```
jamesonwilliams: As a matter for another PR (its unrelated to this PR) - this behavior is backwards :-(

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -130,44 +133,77 @@ public void configure(@NonNull JSONObject pluginConfigurationJson, Context conte
...
             @NonNull Class<T> modelClass,
             @NonNull ResultListener<GraphQLResponse<Iterable<T>>> responseListener
     ) {
-        final String apiName = getSelectedApiName(EndpointType.GRAPHQL);
+        final String apiName;
+        try {
+            apiName = getSelectedApiName(EndpointType.GRAPHQL);
+        } catch (ApiException exception) {
+            responseListener.onError(exception);
+
```
kvasukib: remove newline.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -130,44 +133,77 @@ public void configure(@NonNull JSONObject pluginConfigurationJson, Context conte
...
     }
 
+    @Nullable
     @Override
     public <T> GraphQLOperation<T> query(
             @NonNull GraphQLRequest<T> graphQLRequest,
             @NonNull ResultListener<GraphQLResponse<Iterable<T>>> responseListener
     ) {
-        final String apiName = getSelectedApiName(EndpointType.GRAPHQL);
+        final String apiName;
```
kvasukib: I noticed this code snippet gets repeated in multiple methods. Would it be better to extract it to a private method?

aws-api/build.gradle
raphkim: you are right!

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
```diff
@@ -31,39 +30,22 @@
...
+    private AWSDataStorePluginConfiguration(final SyncMode syncMode) {
         this.syncMode = syncMode;
-        this.apiName = apiName;
     }
 
     static AWSDataStorePluginConfiguration fromJson(JSONObject pluginJson) throws DataStoreException {
         // If no sync mode is specified, we just use the default (no sync) and continue
-        if (!pluginJson.has("syncMode")) {
-            return new AWSDataStorePluginConfiguration(SyncMode.LOCAL_ONLY, null);
+        if (pluginJson == null || !pluginJson.has("syncMode")) {
```
kvasukib: If pluginJson is empty, would `has` return false and not throw any exception?

core/src/main/java/com/amplifyframework/core/plugin/Plugin.java
raphkim: This will be overrided with either `@NonNull` or `@Nullable` depending on the plugin, so I decided to leave it without.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginConfigurationTest.java
raphkim: This PR also contains the logic to make AWS DataStore plugin's configuration parameter to be nullable, so I felt like including this test was necessary.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
raphkim: if pluginJson is null, it would throw NPE :)

raphkim: The idea is to now allow null objects to be passed as pluginConfiguration parameter to prevent crashes from happening on category level

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: i thought about doing that, but something just didn't feel right about nesting private method the third time, not that there's anything conventionally wrong with it.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: okay

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: also this code snippet is both setting the value of a string and returning null operation if an error is thrown, so the utility code wouldn't simplify it much

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
```diff
@@ -371,14 +395,16 @@
      * The subscription may be canceled by calling
      * {@link GraphQLOperation#cancel()}.
      * Requires that only one API is configured in your
-     * `amplifyconfigurationl.json`. Otherwise, throws an IllegalStateException.
+     * `amplifyconfigurationl.json`. Otherwise, emits an ApiException to
```
TrekSoft: amplifyconfiguration.json (no l)

TrekSoft: A bunch of the others have this typo as well since it was probably copy/pasted

raphkim: good catch. fixed

PR #166, opened 2019-12-04
[DataStore] Create RemoteModelState class

jamesonwilliams:
Provides and observe() signature though which an actor can obtain
all current values of all models in an App Sync endpoint, by means
of k-many base syncs, for k, the number of models managed by the system.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/network/RemoteModelState.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/network/RemoteModelState.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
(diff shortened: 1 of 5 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/network/RemoteModelState.java
```diff
@@ -0,0 +1,146 @@
...
+import io.reactivex.Single;
+import io.reactivex.SingleEmitter;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.schedulers.Schedulers;
+
+@SuppressWarnings("unused") // one sec
+final class RemoteModelState {
+    private final AppSyncEndpoint endpoint;
+    private final ModelProvider modelProvider;
+    private final ConfiguredApiProvider configuredApiProvider;
```
raphkim: In PR #164 , I removed all reference to api name inside DataStore. If that gets merged in, we are going to need to remove it here too

raphkim: LGTM otherwise

aws-datastore/src/main/java/com/amplifyframework/datastore/network/RemoteModelState.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        @Override
+        public void onResult(GraphQLResponse<Iterable<ModelWithMetadata<T>>> resultFromEndpoint) {
+            if (resultFromEndpoint.hasErrors()) {
+                emitter.onError(new DataStoreException(String.format(
+                    "A model sync failed: %s", resultFromEndpoint.getErrors()
+                ), "Check your schema."));
+            } else if (!resultFromEndpoint.hasData()) {
+                emitter.onError(new DataStoreException("Empty response from AppSync.", "Report to AWS team."));
```
kvasukib: Nice! very precise!

PR #177, opened 2019-12-06
[DataStore] Add Data Hydrator

jamesonwilliams:
DataHydrator consumes RemoteModelState's observe(). For each model
emitted on the Observable, the Hydrator populates the local data store
with the model data, and the metadata about that model.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/network/DataHydration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/network/DataHydration.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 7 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/network/DataHydration.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    Completable hydrate() {
+        // Observe the remote model states,
+        return remoteModelState.observe()
+            .subscribeOn(Schedulers.io())
+            .observeOn(Schedulers.io())
+            // For each model state, provide it as an input to a completable operation
+            .flatMapCompletable(modelWithMetadata -> Completable.mergeArray(
+                // And save both the item and the metadata
+                saveItemToStorage(modelWithMetadata), saveMetadataToStorage(modelWithMetadata)
```
TrekSoft: Maybe I'm overthinking this or don't understand the mechanism but would we have to worry about perhaps the item failing to save but the metadata succeeding and then the item can never be updated because according to the metadata table it thinks it has the latest version?

jamesonwilliams: I changed the logic so that the metadata will only be updated if the data write succeeds

jamesonwilliams: Update: I changed this so that the saves are chained. I only try to write the metadata if the data saves successfully.

Issue #169, opened 2019-12-04
Fix IAM auth in REST API


royjit:
IAM authType is not currently supported in REST API.

royjit:
PR submited

jamesonwilliams:
Updated PR https://github.com/aws-amplify/amplify-android/pull/178

jamesonwilliams:
#178 is in.

PR #178, opened 2019-12-06
[API] IAM support for REST api request

royjit:
* Issues: #169 
Changes to support signing a Rest Request with IAM credentials.

- Made `EndpointType` public
- New signed class for ApiGateway IAM
- Removed hard coded header `content-type`

### DIFF
--- a/aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
+++ b/aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
@@ -16,8 +16,12 @@
 package com.amplifyframework.api.aws;
 
 import com.amplifyframework.AmplifyException;
+import com.amplifyframework.api.ApiException;
 import com.amplifyframework.api.rest.RestOptions;
 import com.amplifyframework.api.rest.RestResponse;
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.testutils.EmptyConsumer;
+import com.amplifyframework.testutils.LatchedConsumer;
 import com.amplifyframework.testutils.SynchronousApi;
 
 import org.json.JSONException;
@@ -26,24 +30,24 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 
 /**
  * Validates the functionality of the {@link AWSApiPlugin} for REST operations.
- *
  */
 public final class RestApiInstrumentationTest {
-
     private static SynchronousApi api;
 
     /**
      * Configure the Amplify framework, if that hasn't already happened in this process instance.
-     * @throws AmplifyException Exception is thrown if configuration fails.
+     * @throws AmplifyException if configuration fails
      */
     @BeforeClass
     public static void onceBeforeTests() throws AmplifyException {
         AmplifyTestConfigurator.configureIfNotConfigured();
         api = SynchronousApi.singleton();
+        TestAWSMobileClient.initialize();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
```diff
@@ -15,37 +15,63 @@
...
+     *
+     * @throws Exception Exception is thrown if configuration fails.
      */
     @BeforeClass
-    public static void onceBeforeTests() throws AmplifyException {
+    public static void onceBeforeTests() throws Exception {
         AmplifyTestConfigurator.configureIfNotConfigured();
+
+        final CountDownLatch latch = new CountDownLatch(1);
+        AWSMobileClient.getInstance().initialize(getApplicationContext(), new Callback<UserStateDetails>() {
```
kvasukib: Would be beneficial to move this to a core test configurator with the pattern of latched response listener so we can reuse this for all other categories.

aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
```diff
@@ -58,62 +60,92 @@
...
+     * @param endpointType Endpoint type for the api
      */
-    public AppSyncSigV4SignerInterceptor(@NonNull ApiKeyAuthProvider apiKeyProvider) {
-        this.apiKeyProvider = Objects.requireNonNull(apiKeyProvider);
-        this.credentialsProvider = null;
-        this.cognitoUserPoolsAuthProvider = null;
-        this.oidcAuthProvider = null;
-        this.awsRegion = null;
-        this.authType = AuthorizationType.API_KEY;
+    public AppSyncSigV4SignerInterceptor(@NonNull ApiKeyAuthProvider apiKeyProvider,
```
kvasukib: A builder would be better for this class.

jamesonwilliams: I agree ... the `AppSyncSigV4SignerInterceptor` is tech debt that we carried over from the App Sync SDK. At the time we brought it in, it was argued "we won't modify it, we'll just leave it like it was in the App Sync code base."

If we are going to modify it, we should pretty much re-write it. They way it constructs, with varying groups of `null` params, is fairly insane.

I think what we should do is have 1 interceptor implementation per auth type. Then, maybe there is some factory that sits above them, and can select an appropriate interceptor, based on the configured auth type. But all of the details for each auth type would be delegated to their respective interceptor implementations.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
royjit: Will handle this as part of another PR

aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
```diff
@@ -118,4 +146,29 @@ public void getRequestWithApiKey() throws JSONException {
...
+    /**
+     * Test whether we can make api Rest call in IAM as auth type.
+     */
+    @Test
+    public void getRequestWithIAM() {
+        final RestOptions options = RestOptions.builder().addPath("items").build();
+        LatchedRestResponseListener responseListener = new LatchedRestResponseListener();
+        Amplify.API.get("iamAPI", options, responseListener);
+        RestResponse response = responseListener.awaitTerminalEvent().awaitSuccessResponse();
+        assertTrue("Should return a non null data", response.getData() != null);
```
jamesonwilliams: I think this might already be inside of `awaitSuccessResponse`, can you check?

If not, this is more simply worded in jUnit as `assertNotNull(response.getData())`;.

aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
jamesonwilliams: @jamesonwilliams these are good points, let's do this in a single, focused refactor PR.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/RestApiInstrumentationTest.java
jamesonwilliams: I updated the current PR to include this change

PR #179, opened 2019-12-07
[DataStore] Rx-ify the DataStore contract

jamesonwilliams:
Don't use the ResultListener callback on the DataStore APIs.
Instead, return an Rx Observable for the query() method, and return
Completables for the save() and delete() methods.

observe() is already returning an Rx Observable.


jamesonwilliams:
Holding off on this, for now, until some more discussion occurs. Will re-open this PR if we go this direction.

PR #183, opened 2019-12-10
[CircleCI] Update scripts for pulling remote configs

jamesonwilliams:
*Issue #, if available:*

*Description of changes:*


lawmicha:
> Very nice looking PR as always! Perhaps the instruction should be added locally in the repo as well since it contains some very specific steps?

if by instructions, you mean provisioning the backend with amplify CLI? this would definitely be really helpful for any external developer that does not have access to our test account to run the tests, and also provide them an e2e experience, almost replacing the need for a sample app

.circleci/copy-configs
```diff
@@ -0,0 +1,45 @@
...
+
+# Mac OS X's antiquated version of bash doesn't have associative arrays.
+# So, use two simple arrays and join on index.
+# sources are remote URLs, targets are local file paths
+readonly sources=(
+    "s3://$config_bucket/awsconfiguration.json"
+    "s3://$config_bucket/amplifyconfiguration-analytics.json"
+    "s3://$config_bucket/amplifyconfiguration-core.json"
+    "s3://$config_bucket/amplifyconfiguration-api.json"
+    "s3://$config_bucket/amplifyconfiguration-datastore.json"
```
lawmicha: any ideas around process for when API key expires?

jamesonwilliams:
> Very nice looking PR as always! Perhaps the instruction should be added locally in the repo as well since it contains some very specific steps?

Thanks @raphkim ! I thought about doing that. The solution here is still not usable by the broader community, unfortunately. I should document the setup steps in our internal Amazon wiki, for now. I would like to get to a state where any developer on the internet would be able to setup a test endpoint and run the tests against their own configuration(s), in the future.

.circleci/copy-configs
jamesonwilliams: I don't have any immediate ideas for that, no. So long as we use the maximum expiration time on these test keys, though, it should be pretty low-effort to have on-call rotate them once per year. Since they aren't used for production, I think it should be acceptable (not desirable) for our test APIs to experience some occasional downtime.

jamesonwilliams:
> > Very nice looking PR as always! Perhaps the instruction should be added locally in the repo as well since it contains some very specific steps?
> 
> if by instructions, you mean provisioning the backend with amplify CLI? this would definitely be really helpful for any external developer that does not have access to our test account to run the tests, and also provide them an e2e experience, almost replacing the need for a sample app

100% - this is where we need to get to, ultimately. Either a non-interactive `amplify` command that can set everything up, or even better, some Gradle target that will do it. We may want to use Cloud Formation templates directly.

PR #188, opened 2019-12-10
Adds field for denoting owner based authorization

TrekSoft:
AppSync has this weird policy where only if a model has owner based authorization, you must pass the owner name as a parameter to the subscribe function. This adds an optional annotation to models so the CLI can set it to true in the same circumstances it is adding this parameter.

https://aws-amplify.github.io/docs/cli-toolchain/graphql#authorizing-subscriptions


core/src/main/java/com/amplifyframework/core/model/annotations/ModelConfig.java
```diff
@@ -40,4 +40,10 @@
      * @return the plural version of the name.
      */
     String pluralName() default "";
+
+    /**
+     * Specifies whether this model has owner based authorization.
```
kvasukib: Would be beneficial is there is an example to demonstrate what this means OR add more description about it?

TrekSoft: Done

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -688,19 +700,15 @@ private String selectApiName(Set<String> apiClients) throws ApiException {
...
         if (clientDetails == null) {
-            responseListener.onError(
-                    new ApiException(
-                            "No client information for API named " + apiName,
-                            "Check your amplify configuration to make sure there " +
-                            "is a correctly configured section for " + apiName
-                    )
+            throw new ApiException(
+                    "No client information for API named " + apiName,
+                    "Check your amplify configuration to make sure there " +
```
kvasukib: amplifyconfiguration.json

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -221,6 +222,14 @@ private AppSyncGraphQLRequestFactory() { }
...
             Class<T> modelClass,
             SubscriptionType type
+    ) throws ApiException {
+        return buildSubscription(modelClass, type, null);
+    }
+
+    public static <T extends Model> GraphQLRequest<T> buildSubscription(
+            Class<T> modelClass,
+            SubscriptionType type,
+            CognitoUserPoolsAuthProvider cognitoAuth
```
kvasukib: Would there be other overloaded versions of `buildSubscription` with other AuthTypes such as IAM, APIKey and OIDC?

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -230,19 +239,60 @@ private AppSyncGraphQLRequestFactory() { }
...
+                if (username == null) {
+                    throw new ApiException(
+                            "Attempted to subscribe to a model with owner based authorization without a username",
+                            "Make sure that a user is logged in before subscribing to a model with owner based auth"
+                    );
+                }
+
+                return new GraphQLRequest<>(
+                        doc.toString(),
+                        Collections.singletonMap("owner", username),
```
kvasukib: `Immutable.of(Collections.singletonMap("owner", username))` if the map can be immutable.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
TrekSoft: We're going to do a deep dive with Richard on this after Christmas break but for now are just handling the simple case of doing this with Cognito since that's the only provider the documentation talks about: 

```
Passing in the current user can be done dynamically in your code by using Auth.currentAuthenticatedUser() in JavaScript, AWSMobileClient.default().username in iOS, or AWSMobileClient.getInstance().getUsername() in Android.
```

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
TrekSoft: I think it's better to leave it like this since the user could be passing in a custom configuration not through the JSON file.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
TrekSoft: Done

TrekSoft:
Failing integration test can be ignored - it passes locally and it seems to be giving the error expected but for some reason when CirclCI runs it the format is slightly different than the test expects.

PR #191, opened 2019-12-13
[DataStore] Implement conditional save for Storage Engine

raphkim:
*Issue #, if available:*

*Description of changes:*

Implemented conditional save on storage engine layer. Saving with predicate will only write to database if the data being overwritten meets the conditions.

* Overloaded `save` method in `LocalStorageAdapter`
  * predicate parameter is `@Nullable` (internal use-case)
  * inherited to `SQLiteStorageAdapter`, `InMemoryStorageAdapter`
* Overloaded `save` method in `DataStoreCategoryBehavior`
  * predicate parameter is `@NonNull`
  * inherited to `DataStoreCategory`, `AWSDataStorePlugin`
* Re-factored logic involved with binding values to pre-compiled SQL statement
  * De-coupled the processing of an object for fields and binding procedure. Field extraction from object is now a separate utility method, and binding of values is a separate utility method.
  * This way, additional values can be appended to `update` statement without being confined to precompiled `insert` statement
* Minor update to predicate parser
  * Store arguments as `Object` type to remember their data types
  * `BETWEEN` operator native to SQL is used

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
@@ -29,7 +29,6 @@ import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.predicate.QueryField;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.DataStoreException;
-import com.amplifyframework.datastore.storage.GsonStorageItemChangeConverter;
 import com.amplifyframework.datastore.storage.StorageItemChange;
 import com.amplifyframework.testmodels.commentsblog.AmplifyModelProvider;
 import com.amplifyframework.testmodels.commentsblog.Blog;
@@ -158,7 +157,7 @@ public final class SQLiteStorageAdapterInstrumentedTest {
         final BlogOwner blogOwner = BlogOwner.builder()
             .name("Alan Turing")
             .build();
-        assertEquals(blogOwner, saveModel(blogOwner));
+        saveModel(blogOwner);
 
         final Cursor cursor = sqliteStorageAdapter.getQueryAllCursor("BlogOwner");
         assertNotNull(cursor);
@@ -183,7 +182,7 @@ public final class SQLiteStorageAdapterInstrumentedTest {
             .name("Tony Danielsen")
             .wea(null)
             .build();
-        assertEquals(blogOwner, saveModel(blogOwner));
+        saveModel(blogOwner);
 
         final Cursor cursor = sqliteStorageAdapter.getQueryAllCursor("BlogOwner");
         assertNotNull(cursor);
@@ -336,7 +335,7 @@ public final class SQLiteStorageAdapterInstrumentedTest {
         final BlogOwner blogOwner = BlogOwner.builder()
             .name("Alan Turing")
             .build();
-        assertEquals(blogOwner, saveModel(blogOwner));
+        saveModel(blogOwner);
 
         Iterator<BlogOwner> result = queryModel(BlogOwner.class);
         assertNotNull(result);
@@ -575,26 +574,24 @@ public final class SQLiteStorageAdapterInstrumentedTest {
         deleteModel(raphael);
 
         // Get the BlogOwner record from the database
-        Iterator<BlogOwner> blogOwnerterator = queryModel(BlogOwner.class);
-        assertFalse(blogOwnerterator.hasNext());
+        Iterator<BlogOwner> blogOwnerIterator = queryModel(BlogOwner.class);
+        assertFalse(blogOwnerIterator.hasNext());
 
         // Get the Blog record from the database
         Iterator<Blog> blogIterator = queryModel(Blog.class);
         assertFalse(blogIterator.hasNext());
     }
 
-    private <T extends Model> T saveModel(@NonNull T model) throws DataStoreException {
-        return saveModel(model, null);
+    private <T extends Model> void saveModel(@NonNull T model) {
+        saveModel(model, null);
     }
 
-    private <T extends Model> T saveModel(
-            @NonNull T model, @Nullable QueryPredicate predicate) throws DataStoreException {
+    private <T extends Model> void saveModel(
+            @NonNull T model, @Nullable QueryPredicate predicate) {
         LatchedResultListener<StorageItemChange.Record> saveListener =
                 LatchedResultListener.waitFor(SQLITE_OPERATION_TIMEOUT_MS);
         sqliteStorageAdapter.save(model, StorageItemChange.Initiator.DATA_STORE_API, predicate, saveListener);
-        return saveListener.awaitResult()
-                .<T>toStorageItemChange(new GsonStorageItemChangeConverter())
-                .item();
+        saveListener.awaitTerminalEvent();
     }
 
     private <T extends Model> Iterator<T> queryModel(@NonNull Class<T> modelClass) {
@@ -610,13 +607,11 @@ public final class SQLiteStorageAdapterInstrumentedTest {
     }
 
     @SuppressWarnings("UnusedReturnValue")
-    private <T extends Model> T deleteModel(@NonNull T model) throws DataStoreException {
+    private <T extends Model> void deleteModel(@NonNull T model) {
         LatchedResultListener<StorageItemChange.Record> deleteListener =
             LatchedResultListener.waitFor(SQLITE_OPERATION_TIMEOUT_MS);
         sqliteStorageAdapter.delete(model, StorageItemChange.Initiator.DATA_STORE_API, deleteListener);
-        return deleteListener.awaitResult()
-            .<T>toStorageItemChange(new GsonStorageItemChangeConverter())
-            .item();
+        deleteListener.awaitTerminalEvent();
     }
 }
 
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
@@ -277,6 +277,57 @@ public final class SQLiteStorageAdapterInstrumentedTest {
     }
 
     /**
+     * Test save with predicate. Conditional write is useful for making sure that
+     * no data is overwritten with outdated assumptions.
+     * @throws DataStoreException from possible underlying DataStore exceptions
+     */
+    @Test
+    public void saveModelWithPredicateUpdatesConditionally() throws DataStoreException {
+        final BlogOwner john = BlogOwner.builder()
+                .name("John")
+                .build();
+        final BlogOwner jane = BlogOwner.builder()
+                .name("Jane")
+                .build();
+        final BlogOwner mark = BlogOwner.builder()
+                .name("Mark")
+                .build();
+        saveModel(john);
+        saveModel(jane);
+        saveModel(mark);
+
+        // Only update John and Jane
+        final QueryPredicate predicate = BlogOwner.NAME.beginsWith("J");
+        final BlogOwner newJohn = john.copyOfBuilder()
+                .name("John Doe")
+                .build();
+        final BlogOwner newJane = jane.copyOfBuilder()
+                .name("Jane Doe")
+                .build();
+        final BlogOwner newMark = mark.copyOfBuilder()
+                .name("Mark Doe")
+                .build();
+        saveModel(newJohn, predicate);
+        saveModel(newJane, predicate);
+        saveModel(newMark, predicate); // Should not update
+
+        Iterator<BlogOwner> blogOwners = queryModel(BlogOwner.class);
+        assertNotNull(blogOwners);
+        Set<BlogOwner> actualBlogOwners = new HashSet<>();
+        while (blogOwners.hasNext()) {
+            actualBlogOwners.add(blogOwners.next());
+        }
+        assertEquals(
+                new HashSet<>(Arrays.asList(
+                        newJohn,
+                        newJane,
+                        mark
+                )),
+                actualBlogOwners
+        );
+    }
+
+    /**
      * Test querying the saved item in the SQLite database.
      * @throws DataStoreException from possible underlying DataStore exceptions
      */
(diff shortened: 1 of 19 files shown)


raphkim:
Temporarily closing to confirm compatibility with sync-engine

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
```diff
@@ -89,7 +89,7 @@ public String toString() {
      * Returns the selection arguments for the converted query string.
      * @return the selection arguments for the converted query string.
      */
-    public List<String> getSelectionArgs() {
+    public List<Object> getSelectionArgs() {
```
jamesonwilliams: Is it the duty of this `SQLPredicate`'s to prepare SQL statements? If so, SQLite will need them as `String`, right? Should we not just return `String` (or `List<String>`, here)?

The generic `Object` thing comes from the specification of predicates in the `.model.` sub-package right? But, whatever component is responsible for interpreting that into SQL, that component should stop using generic `Object` and convert to a form that a SQL database can act on.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
```diff
@@ -75,6 +75,22 @@ void initialize(
             @NonNull StorageItemChange.Initiator initiator,
             @NonNull ResultListener<StorageItemChange.Record> itemSaveListener);
 
+    /**
+     * Save an item into local storage only if the data being overwritten meets the
```
jamesonwilliams: `overwritten` --> `written`. 

On this API, `@Nullable` predicate is okay, since this is an internal interface.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -191,6 +191,18 @@ public CategoryType getCategoryType() {
             new ResultConversionListener<>(saveItemListener, this::toDataStoreItemChange));
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public <T extends Model> void save(
```
jamesonwilliams: Should the version at L187 just call this one, now?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
```diff
@@ -276,6 +276,57 @@ public void saveModelWithMaliciousInputs() throws DataStoreException {
...
+        saveModel(newJohn, predicate);
+        saveModel(newJane, predicate);
+        saveModel(newMark, predicate); // Should not update
+
+        Iterator<BlogOwner> blogOwners = queryModel(BlogOwner.class);
+        assertNotNull(blogOwners);
+        Set<BlogOwner> actualBlogOwners = new HashSet<>();
+        while (blogOwners.hasNext()) {
+            actualBlogOwners.add(blogOwners.next());
+        }
```
jamesonwilliams: Should we update `queryModel` to perform this `Set` building, itself? So it can just return `Set<T>` directly?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
```diff
@@ -276,6 +276,57 @@ public void saveModelWithMaliciousInputs() throws DataStoreException {
         assertEquals(blogOwner, result.next());
     }
 
+    /**
+     * Test save with predicate. Conditional write is useful for making sure that
+     * no data is overwritten with outdated assumptions.
+     * @throws DataStoreException from possible underlying DataStore exceptions
+     */
+    @Test
+    public void saveModelWithPredicateUpdatesConditionally() throws DataStoreException {
```
jamesonwilliams: Nice looking test method!

core/src/main/java/com/amplifyframework/util/Immutable.java
```diff
@@ -80,4 +81,18 @@ private Immutable() {
...
+     * @param mutableArray A possibly null, possibly empty array
+     * @param <T> The type of items in the array
+     * @return An immutable copy of the provided array
+     */
+    public static <T> T[] of(final T[] mutableArray) {
+        if (mutableArray == null) {
+            return null;
+        }
+
+        return Arrays.copyOf(mutableArray, mutableArray.length);
```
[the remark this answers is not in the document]
raphkim: Ah, I guess that naming it an immutable array doesn't make sense since there is no such thing as immutable array in Java.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -48,6 +48,21 @@
             @NonNull T item,
             @NonNull ResultListener<DataStoreItemChange<T>> saveItemListener);
 
+    /**
+     * Saves an item into the DataStore if the data being overwritten meets
+     * the provided conditions. This is useful for making sure that no data
+     * is overwritten with an outdated/incorrect assumption.
```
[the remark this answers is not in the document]
raphkim: It applies the predicate to existing data in database, rather than the data being saved. So predicate wouldn't do anything for creation is what my understanding is.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
raphkim: so the SQLite library has this thing, where assigning variable values to compiled update/insert/delete statement requires binding each value along with its specific type. That's why I changed it to object so that it can remember the data type to be converted to equivalent SQLite datatype. 

Meanwhile, for query, the selection arguments must be provided as `String[]`, which is why I initially had it as List<String> (since I implemented query before this).

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
raphkim: Now that I've made `DataStoreCategoryBehavior`'s predicates into `@NonNull`, I can't call the it with `null` predicate anymore, so I'll just leave it directly calling the method from storage adapter.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
raphkim: Hmm, a nice idea, but I think I'll leave it out of this PR since this contains some other refactoring changes already haha

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
raphkim: thanks!

raphkim:
Reopening with changes

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
raphkim: just emphasizing that predicate is being applied to data already inside storage, not the new data being written :)

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -48,6 +48,21 @@
...
+     * @param item An item to save
+     * @param predicate Predicate condition to apply for conditional write
+     * @param saveItemListener
+     *        An optional listener which will be callback'd when the save succeeds or fails
+     * @param <T> The time of item being saved
+     */
+    <T extends Model> void save(
+            @NonNull T item,
+            @Nullable QueryPredicate predicate,
```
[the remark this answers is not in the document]
raphkim: Agreed. I will leave the internal ones (`SQLiteStorageAdapter` and `LocalStorageAdapter`) as is, but update the publicly exposed ones (`DataStoreCategoryBehavior`, `DataStoreCategory`, and `AWSDataStorePlugin`)

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
```diff
@@ -46,9 +49,20 @@ private StorageItemChange(
...
             @NonNull T item,
             @NonNull Class<T> itemClass) {
+        this(changeId, initiator, type, null, item, itemClass);
+    }
+
+    private StorageItemChange(
+            @NonNull UUID changeId,
+            @NonNull Initiator initiator,
+            @NonNull Type type,
+            @Nullable QueryPredicate predicate,
```
jamesonwilliams: Alternately, instead of using `null` to mean "a predicate which doesn't check anything", we can have a

```
// A factory class to build some different types of query predicates.
final class QueryPredicates {
    private QueryPredicates() {}

    // Builds a query predicate that doesn't filter out anything.
    static QueryPredicate matchAll() {
        return new QueryPredicate() {}
    }
}
```

And then pass in a:
```
QueryPredicates.matchAll()
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
```diff
@@ -134,6 +156,9 @@ public boolean equals(Object thatObject) {
         if (type != that.type) {
             return false;
         }
+        if (predicate != that.predicate) {
```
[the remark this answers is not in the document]
raphkim: Nice catch! Thanks for pointing it out :)

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
raphkim: This is a good idea, but `QueryPredicate` interface/class is sensitive in that GraphQL implementation also relies on it, so I think refactoring it should involve a separate PR with David's input as well.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -667,30 +715,44 @@ private void bindPreCompiledInsertStatementWithJavaFields(
...
             final SQLiteStatement compiledSqlStatement = sqlCommand.getCompiledSqlStatement();
             compiledSqlStatement.clearBindings();
-            bindPreparedSQLStatementWithValues(model, sqlCommand);
+
+            // fieldValues is a list of field values that are
+            // applicable when binding to insert statement.
+            final List<Object> fieldValues = new ArrayList<>(extractFieldValuesFromModel(model));
+            if (sqlCommand.hasSelectionArgs()) {
+                // These are additional values to bind in the case of updating SQLite table.
+                fieldValues.addAll(sqlCommand.getSelectionArgs());
```
kvasukib: Is there an order of insertion be maintained?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -667,30 +715,44 @@ private void bindPreCompiledInsertStatementWithJavaFields(
...
+            }
+            bindPreCompiledStatementWithFieldValues(compiledSqlStatement, fieldValues);
+
             switch (modelConflictStrategy) {
                 case OVERWRITE_EXISTING:
-                    compiledSqlStatement.executeUpdateDelete();
+                    // executeUpdateDelete returns the number of rows affected.
+                    writeSuccess = compiledSqlStatement.executeUpdateDelete() > 0;
```
kvasukib: if there is a public constant in SQLIteStatement which would represent 0 and -1, would prefer using it. Else, ignore the comment,

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/OperatorInterfaceAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+        JsonDeserializer<QueryOperator>,
+        JsonSerializer<QueryOperator> {
+
+    private static final String TYPE = "type";
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public QueryOperator deserialize(JsonElement json, Type type,
```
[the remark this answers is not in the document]
raphkim: It just happened to be part of the method being overridden, but I didn't need it for this purpose.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -274,17 +297,21 @@ public synchronized void initialize(
...
                         .item(item)
                         .itemClass((Class<T>) item.getClass())
                         .type(StorageItemChange.Type.SAVE)
+                        .predicate(predicate)
                         .initiator(initiator)
                         .build()
                         .toRecord(storageItemChangeConverter);
-                itemChangeSubject.onNext(record);
+                if (writeSuccess) {
+                    // Do not publish item change if write did not go through.
```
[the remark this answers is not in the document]
raphkim: I personally think that it shouldn't be an error if write doesn't go through because the condition isn't met, but I think this is worth discussing.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: Yes. the indexes should match.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: i didnt see any in the document

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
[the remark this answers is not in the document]
raphkim: Upon further discussion, I think that it makes sense to throw an error!

PR #194, opened 2019-12-16
[SQLiteStorageAdapter] Minor refactoring

kvasukib:

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
@@ -248,7 +248,7 @@ public final class SQLiteStorageAdapter implements LocalStorageAdapter {
         threadPool.submit(() -> {
             try {
                 final ModelSchema modelSchema =
-                    modelSchemaRegistry.getModelSchemaForModelClass(item.getClass().getSimpleName());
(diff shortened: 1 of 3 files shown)


raphkim:
#191 contains semi-major refactoring of `SQLiteStorageAdapter`, so can we merge that first?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -475,10 +481,19 @@ private CreateSqlCommands getCreateCommands(@NonNull Set<Class<? extends Model>>
...
+            cursor.close();
+        }
+    }
+
+    private <T> void bindAllFieldsWithValues(@NonNull T object,
+                                             SQLiteStatement preCompiledInsertStatement,
+                                             Cursor cursor,
+                                             Map<String, SQLiteColumn> columns)
+            throws IllegalAccessException, DataStoreException {
+        final Iterator<Field> fieldIterator = FieldFinder.findFieldsIn(object.getClass()).iterator();
```
raphkim: `findFieldsIn` method relies on annotation, and I think this is the only place in `SQLiteStorageAdapter` that relies on annotation rather than model schema. This should be replaced with simple field iterator, since I already have logic to ignore irrelevant/unsupported fields inside the loop.

kvasukib: Rebased with your changes.

PR #196, opened 2019-12-20
[DataStore] Implement in-memory predicate evaluator

raphkim:
*Issue #, if available:*

*Description of changes:*
Implemented in-memory predicate evaluator.


core/src/main/java/com/amplifyframework/core/model/query/predicate/NotEqualQueryOperator.java
```diff
@@ -40,6 +40,16 @@ public Object value() {
...
     }
 
+    /**
+     * Returns true if the the provided field value does
+     * not equal the value associated with this operator.
+     * @param field the field value to operate on
+     * @return evaluated result of the operator
+     */
+    public boolean evaluate(Object field) {
+        return !field.equals(value);
```
kvasukib: This would rely on the fact that `field` 's class implements `equals` method. What happens when the class doesn't implement the equals method?

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateGroup.java
```diff
@@ -103,6 +103,38 @@ public static QueryPredicateGroup not(QueryPredicateGroup predicate) {
...
+                return false;
+            case AND:
+                for (QueryPredicate predicate : predicates) {
+                    if (!predicate.evaluate(object)) {
+                        return false;
+                    }
+                }
+                return true;
+            case NOT:
+                return !predicates.get(0).evaluate(object);
```
kvasukib: If predicates is empty, would cause a crash. Would be better to guard with a check.

kvasukib:
The integration tests have failed.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BeginsWithQueryOperator.java
```diff
@@ -40,6 +40,17 @@ public Object value() {
...
     }
 
+    /**
+     * Returns true if the the provided field value begins
+     * with the value associated with this operator.
+     * @param field the field value to operate on
+     * @return evaluated result of the operator
+     */
+    @Override
+    public boolean evaluate(String field) {
```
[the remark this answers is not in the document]
raphkim: `QueryPredicateOperation#evaluate()` method already has the logic to return false on every exception that isn't `ClassCastException` (which implies that field type was not evaluable). Each implementation of `QueryOperator` now has package private accessor, so it will always be called by `QueryPredicateOperation`.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BetweenQueryOperator.java
```diff
@@ -19,17 +19,18 @@
 
 /**
  * Represents a between condition with a starting and ending value for comparison.
```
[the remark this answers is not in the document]
raphkim: Initially implemented it as exclusive-between, but it appears that both SQLite and DynamoDB follows inclusive policy:
* https://www.techonthenet.com/sqlite/between.php
* https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html

Will be fixed

core/src/main/java/com/amplifyframework/core/model/query/predicate/NotEqualQueryOperator.java
raphkim: `.equals` is implemented by every instance of `Object`
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateGroup.java
raphkim: good point

raphkim:
> The integration tests have failed.

CircleCI integ test is currently inconsistent. It passed upon retry.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BeginsWithQueryOperator.java
```diff
@@ -20,14 +20,14 @@
 /**
  * Represents a begins with condition with a target value for comparison.
  */
-public final class BeginsWithQueryOperator extends QueryOperator {
-    private Object value;
+public final class BeginsWithQueryOperator extends QueryOperator<String> {
```
jamesonwilliams: If the constructor can be package-only access, can the class, too?

jamesonwilliams: If the constructor can be package-only access, can the class, too?

core/src/main/java/com/amplifyframework/core/model/query/predicate/BetweenQueryOperator.java
```diff
@@ -18,18 +18,19 @@
...
 
 /**
- * Represents a between condition with a starting and ending value for comparison.
+ * Represents an inclusive between condition with a starting and ending value for comparison.
+ * @param <T> Comparable data type of the field
  */
-public final class BetweenQueryOperator extends QueryOperator {
-    private Object start;
-    private Object end;
+public final class BetweenQueryOperator<T extends Comparable<T>> extends QueryOperator<T> {
```
jamesonwilliams: Nice.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BetweenQueryOperator.java
```diff
@@ -18,18 +18,19 @@
...
+    private T start;
+    private T end;
 
     /**
      * Constructs a between condition.
      * @param start the value to be used for the beginning of the comparison range
      * @param end the value to be used for the end of the comparison range
      */
-    public BetweenQueryOperator(Object start, Object end) {
+    BetweenQueryOperator(T start, T end) {
```
jamesonwilliams: What's going on with null-ability in this class? Is it allowed for `start`, `end`, and the `field` argument to evaluate?

core/src/main/java/com/amplifyframework/core/model/query/predicate/ContainsQueryOperator.java
```diff
@@ -40,6 +40,17 @@ public Object value() {
...
     }
 
+    /**
+     * Returns true if the the provided field value contains
+     * the value associated with this operator.
+     * @param field the field value to operate on
+     * @return evaluated result of the operator
+     */
+    @Override
+    public boolean evaluate(String field) {
```
jamesonwilliams: Hm, this one I'm not as sold on.
Strings can contain other strings. But collections can contain items. So the choice of `String` here is somewhat arbitrary. Couldn't `ContainsQueryOperation` just easily have `implements QueryOperator<Collection<T>>` and ask `collection.contains( /* (T) */ fieldForEvalution)`?

core/src/main/java/com/amplifyframework/core/model/query/predicate/EqualQueryOperator.java
```diff
@@ -40,6 +40,16 @@ public Object value() {
...
     }
 
+    /**
+     * Returns true if the the provided field value equals
+     * the value associated with this operator.
+     * @param field the field value to operate on
+     * @return evaluated result of the operator
+     */
+    public boolean evaluate(Object field) {
+        return field.equals(value);
```
jamesonwilliams: what's up with the nullness here? If `value` is non-`null`, then it maybe better to reverse this check and write it as `value.equals(field)`. That way, if `field` is null, you'll get a returned value of `false`, not a `NullPointerException`.

core/src/main/java/com/amplifyframework/core/model/query/predicate/Evaluable.java
```diff
@@ -0,0 +1,29 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model.query.predicate;
+
+/**
+ * Interface for evaluable item.
+ * @param <T> Type of target value being evaluated against
+ */
+public interface Evaluable<T> {
```
jamesonwilliams: If this is only ever used by the `QueryOperator`s, you could consider baking the `evalute(T target)` method right into that interface.

core/src/main/java/com/amplifyframework/core/model/query/predicate/NotEqualQueryOperator.java
jamesonwilliams: Josha Block [offers this guidance on when to override `equals()`](http://thefinestartist.com/effective-java/08) in his [Effective Java 2E](https://www.amazon.com/Effective-Java-2nd-Joshua-Bloch/dp/0321356683):
```
Appropriate time to override equals method:
  - a class has a notion of logical equality that differs from mere object identity
  - a superclass has not already overridden equals to implement the desired behavior
```

A common source of programming errors in Java are classes which do not correctly implement `equals()` and `hashCode()`. In fact, we had [one such problem in the CLI code-gen](https://github.com/aws-amplify/amplify-cli/commit/5b74dfa0b0badc9a6f4d8eb8f1bbdc03e3819ecc), before re:Invent.

TLDR: we should assume that any `Object` passed into this method has a sensibly implemented `equals()`.

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryField.java
```diff
@@ -40,84 +40,89 @@ public static QueryField field(String fieldName) {
...
      */
-    public QueryPredicateOperation contains(Object value) {
-        return new QueryPredicateOperation(fieldName, new ContainsQueryOperator(value));
+    public QueryPredicateOperation<Object> eq(Object value) {
+        return new QueryPredicateOperation<>(fieldName, new EqualQueryOperator(value));
     }
 
     /**
-     * Generates a new equality comparison object to compare this field to the specified value.
+     * Generates a new not equals comparison object to compare this field to the specified value.
```
jamesonwilliams: :-)

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryField.java
```diff
@@ -40,84 +40,89 @@ public static QueryField field(String fieldName) {
...
+        return new QueryPredicateOperation<>(fieldName, new BetweenQueryOperator<>(start, end));
+    }
+
+    /**
+     * Generates a new beginsWith comparison object to compare this field to the specified value.
      * @param value the value to be compared
-     * @return an operation object representing the not equal condition
+     * @return an operation object representing the beginsWith condition
+     */
+    public QueryPredicateOperation<String> beginsWith(String value) {
```
jamesonwilliams: This one is also a little presumptive, I think. It is true that some strings begin with the contents of other strings. e.g., `"foo"` _begins with_ `"fo"`.

However, it is also true that an ordered collection can exhibit this property. For example,
```
STATES_BY_ORDER_OF_STATEHOOD.beginsWith(Arrays.asList(DELAWARE, PENNSYLVANIA))
```
could conceivably evaluate to `true`, as could:
```
STATES_BY_ORDER_OF_STATEHOOD.beginsWith(DELAWARE)
```
Where `PENNSYLVANIA` and `DELAWARE` are instances of some user-defined `State`.

(Even though I was born in the U.S., I don't actually know this stuff. I cheated and [looked up the domain details](https://en.wikipedia.org/wiki/List_of_U.S._states_by_date_of_admission_to_the_Union), here. :0)

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateGroup.java
```diff
@@ -100,7 +101,49 @@ public QueryPredicateGroup or(QueryPredicate predicate) {
...
+     *          not contain any predicate element
+     */
+    @Override
+    public boolean evaluate(Object object) throws
+            IllegalArgumentException,
+            IllegalStateException {
+        if (predicates.isEmpty()) {
+            throw new IllegalStateException("A predicate group " +
+                    "must contain at least one predicate element");
+        }
```
jamesonwilliams: If this is true, can you require this in the constructor, instead of waiting until the `evaluate(...)`, here?

Alternately, if you only need >=1 predicates in a specific scenario, can you define that specific scenario in this error message?

Lastly, should this throw a `AmplifyException`, instead? (Not sure)?

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateOperation.java
```diff
@@ -76,8 +79,35 @@ public QueryPredicateGroup or(QueryPredicate predicate) {
...
+            T fieldValue = (T) objectField.get(object);
+            return operator.evaluate(fieldValue);
+        } catch (ClassCastException castException) {
+            throw new IllegalArgumentException(field + " field inside " +
+                    "provided object cannot be evaluated by the operator " +
+                    "type: " + operator.type().name(),
+                    castException);
+        } catch (Exception exception) {
+            return false;
+        }
```
jamesonwilliams: This is just logic to extract a value of a field in an Object, by its java member name, right? Even though it is ~short, it is a little bit disjoint from the goal of the class it is in. Also, it is probably re-usable, by other code we have which needs this functionality, too. TLDR, let's get this is in some little utility class like:
```
final class FieldValue {
    private FieldValue() {}
 
    public extract(final Object fromObject, final String aJavaMemberByName) {
        ... java.lang.reflect.* stuff is encapsulated here, does not leak outside this class ....
    }
}
```

core/src/test/java/com/amplifyframework/core/model/query/predicate/OperatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Test the accuracy of Equal operator evaluation.
+     */
+    @Test
+    public void testEqualOperator() {
+        final EqualQueryOperator numericalOperator = new EqualQueryOperator(123);
+
+        Assert.assertTrue(numericalOperator.evaluate(123));
+        Assert.assertFalse(numericalOperator.evaluate(1234));
+        Assert.assertFalse(numericalOperator.evaluate("123"));
```
jamesonwilliams: Most of the request of the project is using `import static` for the jUnit and Mockito test methods. For all unit tests, can you switch these to just `assertTrue(...)` and `assertFalse(...)`, etc.?

core/src/test/java/com/amplifyframework/core/model/query/predicate/OperatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * Test the accuracy of Not Equal operator evaluation.
+     */
+    @Test
+    public void testNotEqualOperator() {
+        final NotEqualQueryOperator numericalOperator = new NotEqualQueryOperator(123);
+
+        Assert.assertTrue(numericalOperator.evaluate(124));
+        Assert.assertTrue(numericalOperator.evaluate("123"));
```
jamesonwilliams: Ah, gee. Is the `String` `"123"` equal to the `int` or `Integer` `123?` I don't know. This looks a conscious design decision, more than a black-and-white kind of thing. But, [by default, anyway, they __are not__ equal](https://ideone.com/f47Fbt).

core/src/test/java/com/amplifyframework/core/model/query/predicate/PredicateTest.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Tests the creation of a simple single operation predicate.
+     */
+    @Test
+    public void testSingleQueryPredicateOperation() {
+        QueryPredicateOperation<?> op = Person.ID.eq("1234");
+
+        assert (op.field().equals("id"));
+        assert (op.operator().getClass().equals(EqualQueryOperator.class));
+        assert (((EqualQueryOperator) op.operator()).value().equals("1234"));
```
jamesonwilliams: Should use jUnit's `assert<Whatever>(...)` family of methods instead of these `assert`s.

[Some discussion around this is provided on StackOverflow](https://stackoverflow.com/questions/2966347/assert-vs-junit-assertions).

raphkim: This part of the test was written by David, so I didn't touch it (simply moved the file + added additional tests)

core/src/test/java/com/amplifyframework/core/model/query/predicate/OperatorTest.java
raphkim: Personally as a user, I would be more interested in the comparison of values rather than the technicality of data type. For example, it would be very inconvenient to have to check if `Person.age` is stored as `Integer` vs `int` before doing a predicate statement.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BeginsWithQueryOperator.java
raphkim: Yep it definitely can be and should be.

core/src/main/java/com/amplifyframework/core/model/query/predicate/BetweenQueryOperator.java
raphkim: These query operators will be evaluated by `QueryPredicateOperation`'s evaluate call, which will return `false` upon encountering an exception (except select few). `NullPointerException` thrown by these operators will just cause them to be evaluated as false in the end.

core/src/main/java/com/amplifyframework/core/model/query/predicate/ContainsQueryOperator.java
raphkim: This is something I was discussing with others, and I'm not decided yet if I would rather create a separate operator for this purpose. It would be nice if there is an interface for `Container` to group String and Collection together hmm

core/src/main/java/com/amplifyframework/core/model/query/predicate/EqualQueryOperator.java
raphkim: Oops. Good catch

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateGroup.java
raphkim: I decided to make it throw a runtime exception because it figured predicate evaluation is a simple boolean logic operation that shouldn't be try-catched every time.

Also I agree that constructor should guarantee non-empty list of `predicates` instead of error-checking here. Will be fixed.

core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateOperation.java
raphkim: Good idea. I believe there's already a utility class for Field somewhere in the project haha

core/src/test/java/com/amplifyframework/core/model/query/predicate/OperatorTest.java
raphkim: Will do

core/src/main/java/com/amplifyframework/core/model/query/predicate/ContainsQueryOperator.java
jamesonwilliams: I think we'll probably have to have `StringContainsQueryOperator implements QueryOperator<String>`, and `CollectionContainsQueyrOperator implements QueryOperator<Collection<?>>`, as two separate things. Otherwise, the implementation will have a `String value`, and a `Collection<?> value`, and would need to conditionally use one or the other, which feels like an anti-pattern.

raphkim: For now, it will only operate on `String`, since current storage engine implementation doesn't support predicate on `Collection` yet

core/src/test/java/com/amplifyframework/core/model/query/predicate/OperatorTest.java
raphkim: Oh I thought you were talking about comparing `int 123` and `new Integer(123)`, which should be equal. `String "123"` and `int 123` should NOT be equal as you say, and the test operator is using `NotEqualQueryOperator`, so it will evaluate to true when compared.

core/src/main/java/com/amplifyframework/core/model/query/predicate/Evaluable.java
raphkim: Also used by `QueryPredicate`

raphkim:
Regarding comments about supporting evaluation of `beginsWith` and `contains` operators with `Collection`: #220

core/src/main/java/com/amplifyframework/core/model/query/predicate/ContainsQueryOperator.java
jamesonwilliams: Okay, sounds good

PR #198, opened 2019-12-20
[DataStore] Implement conditional delete for storage engine

raphkim:
**Note:**
Stub implementation in `InMemoryStorageAdapter` does not deal with predicates at the moment. Apply changes once #196 is merged.


aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -116,23 +116,34 @@ public void initialize(
...
+    ) {
+        delete(item, initiator, null, itemDeletionListener);
+    }
+
     @SuppressWarnings("unchecked") // item.getClass() -> Class<?>, but type is T. So cast as Class<T> is OK.
     @Override
     public <T extends Model> void delete(
             @NonNull final T item,
             @NonNull final StorageItemChange.Initiator initiator,
+            @Nullable final QueryPredicate predicate,
```
jamesonwilliams: `@NonNull`. We have a version of this method that doesn't require a predicate. So, the user should just use that, instead of passing `null` to this version.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -383,44 +383,76 @@ public synchronized void initialize(
...
+                            .initiator(initiator)
+                            .build()
+                            .toRecord(storageItemChangeConverter);
+                    itemChangeSubject.onNext(record);
+                    itemDeleteListener.onResult(record);
+                } else {
+                    //TODO: implement ConditionCheckFailedException to be cause of this error
+                    itemDeleteListener.onError(new DataStoreException("Predicate condition not met.",
+                            AmplifyException.TODO_RECOVERY_SUGGESTION));
+                }
```
jamesonwilliams: I'm not sure if it is appropriate for us to fire `onError(...)`, here. It is an expected outcome for a predicate not to match, and so it is a normal operation of the system for a delete to not occur, in this case.

Perhaps what we need is a way to dispatch different types of success results. For example, maybe we need:

```
interface LocalStorageAdapter {
    ...
    public <T extends Model> void delete(
            @NonNull T item,
            @NonNull StorageItemChange.Initiator initiator,
            @Nullable QueryPredicate predicate,
            @NonNull ResultListener<Optional<StorageItemChange.Record>> itemDeleteListener);
    ...
}
```
So that:
```
new ResultListener(/* on result */ result -> {
    if (result.isPresent()) {
        // deletion went through.
    } else {
        // predicate did not match.
    }
}, /* on error */ error -> {
    // we tried to delete the thing, after a predicate match, but that failed!
});
```

Or, alternately, just the existing contract:
```
interface LocalStorageAdapter {
    ...
    public <T extends Model> void delete(
            @NonNull T item,
            @NonNull StorageItemChange.Initiator initiator,
            @Nullable QueryPredicate predicate,
            @NonNull ResultListener<StorageItemChange.Record> itemDeleteListener);
    ...
}
```
But we find a way to communicate a "no-op" in the existing argument. Like maybe we set the `entry` field as `null`, and provide a method to check it?

```
new ResultListener(changeRecord -> {
    if (changeRecord.hasEntry()) {
        // Change went through ...
    } else {
        // Predicate did not match.
    }
}, error -> {
    // Failure to delete, even though we tried
});
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
```diff
@@ -582,15 +582,52 @@ public void deleteModelCascades() throws DataStoreException {
...
+
+        // Delete everybody but Mark
+        final QueryPredicate predicate = BlogOwner.NAME.ne(mark.getName());
+        deleteModel(john, predicate);
+        deleteModel(jane, predicate);
+        deleteModel(mark, predicate); // Should not be deleted
+
+        Iterator<BlogOwner> blogOwners = queryModel(BlogOwner.class);
+        assertNotNull(blogOwners);
+        assertEquals(mark, blogOwners.next());
```
jamesonwilliams: What guarantees that `mark` is the first result on the `Iterator`, here?

We should really rework `queryModel` so that it returns a `Set`/`List` directly and then we can just ask like `results.contains(mark)` and be done.

raphkim: good point.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: I initially had it implemented as a no-op, but I discussed with Francisco and he told me that condition check failure is a valid error thrown in this case. 

DynamoDB's "conditional write" also appears to have similar mechanism: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
raphkim: Feels like unnecessary duplicate code to me. `LocalStorageAdapter` interface already defines predicate to be nullable, so there wouldn't be a point in changing it just for this implementation of the interface.

raphkim: Besides, once predicate evaluator is merged, I will need to add id-check logic on both sides if I do that, which would result in even more duplicate code.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
jamesonwilliams: Okay, cool! Yes, I've receives those conditional checks as 400 series from Dynamo before. Good to be consistent with other AWS systems.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
jamesonwilliams: Okay, yea. Since `LocalStorageAdapter` isn't user-facing / public, I don't think it's a big deal, either way.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterInstrumentedTest.java
jamesonwilliams: Thanks for making this change in another PR, @raphkim !

PR #200, opened 2019-12-21
[Core] Tighten Type Bounds on Error Consumers

jamesonwilliams:
The error callbacks had been left as Consumer&lt;Throwable&gt;. But, each
category is meant to return a category-specific exception type, e.g.
ApiException. So, the error callback should be a Consumer&lt;ApiException&gt;.

To achieve this, the ResultListener and StreamListener are reworked so
that the error consumer accepts Consumer&lt;E&gt; where &lt;E extends AmplifyException&gt;.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/network/MutationAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/network/MutationAdapter.java
@@ -15,16 +15,18 @@
 
 package com.amplifyframework.datastore.network;
 
+import com.amplifyframework.api.ApiException;
 import com.amplifyframework.api.graphql.GraphQLResponse;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.ResultListener;
 import com.amplifyframework.core.model.Model;
+import com.amplifyframework.datastore.DataStoreException;
 
 final class MutationAdapter {
     @SuppressWarnings("checkstyle:all") private MutationAdapter() {}
 
-    static <T extends Model> ResultListener<GraphQLResponse<String>> instance(
-            ResultListener<GraphQLResponse<ModelWithMetadata<T>>> responseListener,
+    static <T extends Model> ResultListener<GraphQLResponse<String>, ApiException> instance(
+            ResultListener<GraphQLResponse<ModelWithMetadata<T>>, DataStoreException> responseListener,
             Class<T> itemClass,
             ResponseDeserializer responseDeserializer) {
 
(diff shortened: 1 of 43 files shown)


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
```diff
@@ -97,7 +100,7 @@ public void onStateChanged(int transferId, TransferState state) {
                     @SuppressWarnings("checkstyle:MagicNumber")
                     @Override
                     public void onProgressChanged(int transferId, long bytesCurrent, long bytesTotal) {
-                        int percentage = (int) (bytesCurrent / bytesTotal * 100);
+                        // int percentage = (int) (bytesCurrent / bytesTotal * 100);
```
desokroshan: Should we just get rid of this?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageUploadFileOperation.java
```diff
@@ -101,7 +104,7 @@ public void onStateChanged(int transferId, TransferState state) {
                     @SuppressWarnings("checkstyle:MagicNumber")
                     @Override
                     public void onProgressChanged(int transferId, long bytesCurrent, long bytesTotal) {
-                        int percentage = (int) (bytesCurrent / bytesTotal * 100);
+                        // int percentage = (int) (bytesCurrent / bytesTotal * 100);
```
desokroshan: Should we remove this for now?

jamesonwilliams:
> LGTM.
> Seems like description is incomplete.

Ah yea, GitHub interpreted `Consumer&lt;Throwable&gt`; and `Consumer&lt;ApiException&gt`; as HTML tags, and so just says `Consumer`.

aws-datastore/src/main/java/com/amplifyframework/datastore/network/MutationAdapter.java
```diff
@@ -36,6 +38,9 @@
             responseListener.onResult(responseDeserializer.deserialize(result.getData(), itemClass));
         };
 
-        return ResultListener.instance(resultConsumer, responseListener::onError);
+        //noinspection CodeBlock2Expr Keep down line lenfth a bit
```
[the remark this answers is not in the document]
jamesonwilliams: whoops, thanks

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
jamesonwilliams: I'll just leave these like they were, for this PR, and remove my added `//`.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageUploadFileOperation.java
jamesonwilliams: Will update to leave the line as it had been originally. Maybe it shouldn't be here, but it's not really the point of this current PR.

PR #202, opened 2019-12-23
[RxBindings] Prototype of an Rx adapter module

jamesonwilliams:
The Rx Bindings express Amplify category behaviors in an Rx-idiomatic
way. That is, instead of:

StorageDowloadOperation<?> downloadFile(
        String key,
        ResultListener<StorageDownloadResult> resultListener);

We might say, instead:

Single<StorageDownloadResult> downloadFile(String key);

*Issue #, if available:*

*Description of changes:*


raphkim:
I can see the additional option being useful to those familiar with Rx, but wouldn't it be a bit daunting to some people who just want to set up a single workflow that we tell them to use?

Maybe Rx adapter and binding should all be a separate project of its own?

jamesonwilliams:
> I can see the additional option being useful to those familiar with Rx, but wouldn't it be a bit daunting to some people who just want to set up a single workflow that we tell them to use?
> 
> Maybe Rx adapter and binding should all be a separate project of its own?

Hey @raphkim - this work is not intended to replace our existing Rx-free APIs. I agree that this could be daunting for the average user, so that's why I've built it as an optional add-on module. The average user should continue to use the plain callback-style APIs, which _now_ look like:
```
Amplify.API.query(Person.class,
    response -> Log.i(TAG, "Got a response: " + response),
    failure -> Log.e(TAG, "Failed to query Person.class.", failure)
);
```

As is, this would be distributed as its own `.aar`, and its own Maven artifact. The advantage to keeping it in the same Gradle project (as a distinct module), is that it'll be easier to make sure everything builds together. For example, if we make a change to one of the category behaviors, the Rx Bindings would stop building, if they're in the same project.

TrekSoft:
Hmm, yeah I feel like it might better belong as a separate (maybe even unofficial project under your own username) since a big part of Amplify is that we are an opinionated framework giving one way of doing things.

jamesonwilliams:
This work has been moved into my personal developer account, as [`amplify-android-rxbindings`](https://github.com/jamesonwilliams/amplify-android-rxbindings). It will not receive official AWS support at this time. We can re-evaluate later, based on user feedback.

PR #204, opened 2019-12-25
Use Retrolambda for broader consumer compatibility

jamesonwilliams:
To reduce friction in using the library, Retrolambda is used to convert
class files to Java class file version 50. before producing aars. This
prevents the need for the user to elect Java 1.8 source and target
compatibility in their consuming project. This configuration may also be
more compatible with older versions of the Android Studio and Android
Build Tools, that may be found in a user's environment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,8 +21,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:3.5.3'
         classpath 'com.amazonaws:aws-devicefarm-gradle-plugin:1.3'
+        classpath 'me.tatarka:gradle-retrolambda:3.7.1'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -19,10 +19,12 @@ buildscript {
     repositories {
         google()
         jcenter()
+        mavenCentral()
```
desokroshan: jcenter is a superset of mavenCentral and shouldn't be needed.
https://bintray.com/bintray/jcenter/me.tatarka:gradle-retrolambda

build.gradle
```diff
@@ -33,6 +35,7 @@ allprojects {
...
```
desokroshan: Same as above.

desokroshan:
Does this actually obviate the need to set source and target compatibility to 1.8? Found the following on plugin's Github : 
https://github.com/evant/gradle-retrolambda#android-studio-setup

jamesonwilliams:
@desokroshan
> Does this actually obviate the need to set source and target compatibility to 1.8?

Amplify source _itself_ will still need to set source and target compatibility, since it uses Java 1.8 features, and we need to compile these to bytecode.

Retrolambda then modifies that bytecode, to downgrade from version 52 (the output of a Java 8 compiler) to version 50 (the output of a Java 6 compiler.)

As a result, a consuming project will see our library as being Java 6 compatible. A project that consumes our library artifact will not need to be able to understand Java 8 constructs in bytecode, since we will no longer use them.

So, the change that is being made here is to allow _consumers_ of Amplify to not specify the compat flags.

PR #206, opened 2019-12-26
[API] Accept Lambdas for REST Behaviors

jamesonwilliams:
The `ResultListener` callback is removed from the public contract.
Previously, the syntax to perform a REST behavior was verbose:
```
Amplify.API.post(request, new ResultListener<RestResponse, ApiException>() {
    @Override
    public void onResult(RestResponse response) {
        Log.i("API", "POST response: " + response),
    }

    @Override
    public void onError(ApiException failure) {
        Log.e("API", "No response for POST.", failure);
    }
});
```
By splitting the callbacks into the `Consumer<RestResponse>` and
`Consumer<ApiException>`, the behavior may accept lambas:

```
Amplify.API.post(request,
    response -> Log.i("API", "POST response: " + response),
    failure -> Log.e("API", "No response for POST.", failure)
);
```

See GitHub issue: https://github.com/aws-amplify/amplify-android/issues/212

### DIFF
--- a/core/src/main/java/com/amplifyframework/api/RestBehavior.java
+++ b/core/src/main/java/com/amplifyframework/api/RestBehavior.java
@@ -21,7 +21,7 @@ import androidx.annotation.Nullable;
 import com.amplifyframework.api.rest.RestOperation;
 import com.amplifyframework.api.rest.RestOptions;
 import com.amplifyframework.api.rest.RestResponse;
-import com.amplifyframework.core.ResultListener;
+import com.amplifyframework.core.Consumer;
 
 /**
  * REST behaviors which include the family of HTTP verbs (GET, POST, etc.).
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/amplifyframework/api/RestBehavior.java
```diff
@@ -31,222 +31,258 @@
     /**
      * This is a helper method for easily invoking GET HTTP request.
      * Requires that only one API is configured in your
-     * `amplifyconfiguration.json`. Otherwise, emits an ApiException
-     * to the provided `ResultListener`.
+     * `amplifyconfiguration.json`. Otherwise, emits an {@link ApiCategory}
```
TrekSoft: Should be ApiException vs. ApiCategory

jamesonwilliams: Whoops, nice catch, thanks - updating, one sec...

jamesonwilliams:
> Re-approving since David's approval got overridden

Around the time that the project switched from private to public, Tim changed some configurations, to tighten up our process. One of the changes is that a new review is required each time a revision is made to a PR.

I hit the "request review again" (recycle) icon, next to @TrekSoft's name, after I updated the PR.

PR #217, opened 2020-01-03
[API] Fixes #187: Subscription blocking + web socket connection error not correctly reported

### LINKED ISSUE #187: [API] Subscribe is currently a blocking operation and times out on error
All operations should be asynchronous but `SubscriptionOperation` currently has the blocking method `requestSubscription` in its [`start`](https://github.com/aws-amplify/amplify-android/blob/a213c0e2965aaccbe8da2560424e1efd89006fd4/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java#L84) method.

Furthermore, if it gets an error, it fails to countdown the latch, immediately returning the error to the user. So, as a result, it gives a ["Subscription not acknowledged" error message 30 seconds later](https://github.com/aws-amplify/amplify-android/blob/a213c0e2965aaccbe8da2560424e1efd89006fd4/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java#L372).

TrekSoft:
*Issue #, if available:*
#187 

*Description of changes:*
Makes subscription a non blocking operation and handles the case of the web socket being unable to establish a connection correctly. Also updates the version of the AWS SDK as it was required for linting.


aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -77,23 +77,32 @@
...
                         exception,
                         AmplifyException.TODO_RECOVERY_SUGGESTION
                 ));
+                return null;
             }
 
             try {
                 connectionAcknowledgement.await(CONNECTION_ACKNOWLEDGEMENT_TIMEOUT, TimeUnit.SECONDS);
-            } catch (InterruptedException interruptedException) {
+            } catch (InterruptedException interruptedException) { }
```
jamesonwilliams: We have enough of these `{}` constructs now that maybe we should change the checkstyle to allow them. However, I think this will currently upset the checkstyle unless you drop `}` onto the next line, or add an `@SuppressWarnings("checkstyle:WhitespaceAround")` (I think it is `WhitespaceAround` that's the issue?)

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -195,6 +200,10 @@ private void processJsonMessage(WebSocket webSocket, String message) throws ApiE
                     );
                     connectionAcknowledgement.countDown();
                     break;
+                case CONNECTION_ERROR:
+                    connectionFailure = message;
+                    connectionAcknowledgement.countDown();
```
jamesonwilliams: This variable was named `connectionAcknowledgement` since it was in response to a `connection_ack` message type. If we use it for two distinct (though closely related purposes), perhaps it should be renamed as `connectionResponsesPending`, or something.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -77,23 +77,32 @@
...
+            if (connectionAcknowledgement.getCount() != 0) {
                 responseListener.onError(new ApiException(
                         "Subscription timed out waiting for acknowledgement",
-                        interruptedException,
                         AmplifyException.TODO_RECOVERY_SUGGESTION
                 ));
+                return null;
+            } else if (connectionFailure != null) {
+                responseListener.onError(
+                    new ApiException(connectionFailure, "Check if you are authorized to make this subscription")
```
jamesonwilliams: Since this is in response to a bug report, and is the primary intent of the PR, we should add some test to cover this scenario.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
TrekSoft: Makes sense

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
TrekSoft: Alright - also in running the integration tests I'm seeing a problem with a normal subscription now so trying to figure out the issue.

jamesonwilliams: It might be the race condition we were talking about earlier, where a couple of different tests try use subscriptions for the same model type, but the order of execution of the tests matters, or something?

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -47,12 +49,14 @@ private SubscriptionOperation(
             @NonNull final OkHttpClient client,
             @NonNull final GraphQLRequest<T> graphQLRequest,
             @NonNull final GraphQLResponse.Factory responseFactory,
-            @NonNull final StreamListener<GraphQLResponse<T>, ApiException> subscriptionListener) {
+            @NonNull final StreamListener<GraphQLResponse<T>, ApiException> subscriptionListener,
```
raphkim: is this in-line with the recent changes that Jameson made to accommodate lambdas?

build.gradle
```diff
@@ -48,7 +48,7 @@ task clean(type: Delete) {
 }
 
 ext {
-    awsSdkVersion = '2.16.5'
+    awsSdkVersion = '2.16.6'
```
raphkim: Perhaps we should use `2.16.+` so that each minor version update doesn't break our compiler?

jamesonwilliams: The build failure is caused by the [`NewerVersionsAvailable`](https://android.googlesource.com/platform/tools/base/+/24973b0/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/GradleDetector.java#210) lint. It would ordinarily be just a warning, but we have warnings [configured as errors](https://github.com/aws-amplify/amplify-android/blob/master/configuration/android-lint.gradle#L18). Projects that don't have this configuration tend to have lots of unaddressed warnings. All in all, I think this is the right setting.

I think there are ~three options here, and each has pros and cons:
1. The default, use `2.16.6`, and leave the `NewerVersionsAvailable` lint as a warning. Almost certainly, the warning will go un-addressed.
2. Use `2.16.6`, but make the `NewerVersionsAvailable` an error, as it is now. With this, we always know the exact version being used, and that it is the latest. But, sometimes our build randomly breaks.
3. Use `2.16.+`. With this, the build always works, and it always used the latest. But, it is harder to tell what version was used for any particular build. Things change somewhat "silently", under the hood. Of course, that's happens when a dependency publishes a new version without updating the point release, which is ~common. So, all this would really say is "AWS SDK point releases are not a big deal."

TLDR I think I like the `2.16.+`, too.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
jamesonwilliams: My PR also changes this line, so this will probably generate a merge conflict. One of us (whoever loses the "push first!" race) will have to update.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
```diff
@@ -45,6 +47,9 @@
     private static final String PERSON_API_NAME = "personApi";
     private static final String PROJECT_API_NAME = "projectApi";
     private static final String BLOG_API_NAME = "blogApi";
+    private static final String NOTES_WITH_AUTH_API_NAME = "notesWithAuthApi";
+
+    private static final int ONE_SECOND_OF_MILLISECONDS = 1000;
```
jamesonwilliams: Instead of defining this constant, you could consider to use `TimeUnit.SECONDS.toMilliseconds(1)` as an in-line constant - or atlernately, use a semantic label, like:
```
private static final int MAX_SUBSCRIPTION_FAILURE_TIME_MS = TimeUnit.SECONDS.toMillseconds(1);
```
which sort of achieves two purposes of documentation - how long it is, and what it purpose is

aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
```diff
@@ -118,6 +123,12 @@ public void subscribeReceivesMutationEvent() {
         SynchronousApi.Subscription<Person> subscription =
             api.onCreate(PERSON_API_NAME, Person.class);
 
+        // Give the subscription time to complete the connection before running the mutation.
+        // This should happen within 1 second - if not, it's good that this fails so we can investigate.
+        try {
+            Thread.sleep(ONE_SECOND_OF_MILLISECONDS);
+        } catch (InterruptedException exception) { }
```
jamesonwilliams: Ah crumbs. I think this is why I had left this part of the method as blocking. If the subscription call returns before getting acknowledgement, we have this new issue - how to do we convey the acknowledgement to the user? It used to be synonymous with getting a subscription ID back.

The user would also need to sleep for a random amount of time, like this test does.

If that part is now async, we can provide the information with a callback. Like:
```
/**
 * Subscribe to mutations that occur for a given model class.
 * @param clazz Type of model being subscribed
 * @param subscriptionType type of events to subscribe to
 * @param onSubscriptionStarted Called when a subscription begins; a subscription ID is passed
 * @param onNextItem Called when a new subscription item is available
 * @param onSubscriptionFailed Called when the subscription terminates due to a failure
 * @param onSubscriptionCompleted Called when a subscription terminates gracefully
 */
<T extends Model> void subscribe(
    Class<T> clazz,
    SubscriptionType subscriptionType,
    Consumer<String> onSubscriptionStarted,
    Consumer<T> onNextItem,
    Consumer<ApiException> onSubscriptionFailed,
    Action onSubscriptionCompleted
);
```

Am (irrelevant) note on mechanics, here -- you can use the [`Sleep#milliseconds(long)`](https://github.com/aws-amplify/amplify-android/blob/master/testutils/src/main/java/com/amplifyframework/testutils/Sleep.java#L30) test utility to clean things up a bit and get rid of this exception block.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
jamesonwilliams: Updated PR

aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
jamesonwilliams: Updated PR to do this

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
jamesonwilliams: Updated PR to account for this

build.gradle
jamesonwilliams: (There was a thread of discussion around this, I think I might have accidentally deleted it.)

TLDR sounds good, let's do that in another PR

PR #218, opened 2020-01-04
[API] Accept Lambdas in GraphQL Behaviors

jamesonwilliams:
To reduce verbosity, the `ResultListener` and `StreamListener` are
removed from the contracts of the GraphQL behaviors. Instead, the
GraphQL behaviors are reworked to accept "decomposed" versions of those
callbacks, using `Consumer` and `Action`, directly.

For example, a query in Amplify 0.9 would have been written as:

```
Amplify.API.query(Person.class,
    new ResultListener<GraphQLResponse<Iterable<Person>>, ApiException>() {
        @Override
        public void onResponse(GraphQLResponse<Iterable<Person>> response) {
            Log.i(TAG, "Got results!");
        }

        @Override
        public void onError(ApiException failure) {
            Log.e(TAG, "Failed.", failure);
        }
    }
);
```

But is now expressed as:

```
Amplify.API.query(Person.class,
    response -> Log.i(TAG, "Got a response: " + response),
    failure -> Log.e(TAG, "Failed to get a GraphQL response.", failure)
);
```

*Issue #, if available:*

*Description of changes:*


aws-datastore/src/test/java/com/amplifyframework/datastore/network/MockAppSyncEndpoint.java
```diff
@@ -61,24 +60,24 @@ static Configurator configure(AppSyncEndpoint mock) {
...
-                final int argumentPositionForResultListener = 2;
-                final ResultListener<GraphQLResponse<Iterable<ModelWithMetadata<T>>>, DataStoreException> listener =
-                    invocation.getArgument(argumentPositionForResultListener);
+                // Get a handle to the response consumer that is passed into the sync() method
+                // Response consumer is the third param, at index 2 (@0, @1, @2, @3).
+                final int argumentPositionForResponseConsumer = 2;
+                final Consumer<GraphQLResponse<Iterable<ModelWithMetadata<T>>>> consumer =
+                    invocation.getArgument(argumentPositionForResponseConsumer);
 
                 // Call its onResult(), and pass the mocked items inside of a GraphQLResponse wrapper
```
raphkim: not exactly onResult() anymore ;)

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
```diff
@@ -23,14 +23,13 @@
...
-import com.amplifyframework.core.StreamListener;
+import com.amplifyframework.core.Action;
+import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
-import com.amplifyframework.hub.HubCategory;
 
 /**
- * GraphQL behaviors which include operations like query/subscribe/mutate operations.
+ * GraphQL behaviors include operations like query, subscribe, mutate, etc.
```
raphkim: Is there any other GraphQL behavior besides those three specified?

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
```diff
@@ -23,14 +23,13 @@
 import com.amplifyframework.api.graphql.GraphQLResponse;
 import com.amplifyframework.api.graphql.MutationType;
 import com.amplifyframework.api.graphql.SubscriptionType;
-import com.amplifyframework.core.ResultListener;
-import com.amplifyframework.core.StreamListener;
+import com.amplifyframework.core.Action;
+import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
-import com.amplifyframework.hub.HubCategory;
```
raphkim: Are we no longer using the Hub? I thought that we would be publishing everything to hub by default on top of the local callback to caller thread?

aws-datastore/src/test/java/com/amplifyframework/datastore/network/MockAppSyncEndpoint.java
jamesonwilliams: Ah good catch, I'll update it.

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
jamesonwilliams: Will update.

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
jamesonwilliams: That was a plan, originally. I think it sounds like a fine idea. The only problem is that we never actually implemented any code to _use_ Hub. We just wrote the Hub and left a bunch of documentation saying that it was getting used, when it wasn't. That was a ~dangerous state, so I've been going through and removing the false documentation. If/when we add Hub integration, we can go back and update the documentation to mention Hub again.

PR #227, opened 2020-01-08
[DataStore] Refactor in-memory storage adapter

raphkim:
*Description of changes:*
In-memory storage adapter's current implementation does not behave as it should (e.g. it does not update/overwrite existing models).

* ~~Refactored it to store models in a HashMap for better performance~~
  * Reverted this change to not mess up data hydration test
* Distinguished update and insert
* Invoke error on conditional check failure

### DIFF
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
@@ -28,10 +28,8 @@ import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.DataStoreException;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 import io.reactivex.disposables.Disposable;
 import io.reactivex.subjects.PublishSubject;
@@ -42,12 +40,12 @@ import io.reactivex.subjects.PublishSubject;
  */
 public final class InMemoryStorageAdapter implements LocalStorageAdapter {
 
-    private final Map<String, Model> items;
+    private final List<Model> items;
     private final PublishSubject<StorageItemChange.Record> changeRecordStream;
     private final GsonStorageItemChangeConverter storageItemChangeConverter;
 
     private InMemoryStorageAdapter() {
-        this.items = new HashMap<>();
+        this.items = new ArrayList<>();
         this.changeRecordStream = PublishSubject.create();
         this.storageItemChangeConverter = new GsonStorageItemChangeConverter();
     }
@@ -87,11 +85,12 @@ public final class InMemoryStorageAdapter implements LocalStorageAdapter {
             @NonNull final Consumer<StorageItemChange.Record> onSuccess,
             @NonNull final Consumer<DataStoreException> onError
     ) {
-        if (items.containsKey(item.getId())) {
+        final int index = indexOf(item);
+        if (index > -1) {
             // Update
-            Model savedItem = items.remove(item.getId());
+            Model savedItem = items.get(index);
             if (predicate == null || predicate.evaluate(savedItem)) {
-                items.remove(item.getId());
+                items.remove(index);
                 save(item, initiator, predicate, onSuccess, onError);
                 return;
             }
@@ -100,7 +99,7 @@ public final class InMemoryStorageAdapter implements LocalStorageAdapter {
                     "Verify the saved models."));
         } else {
             // Insert
-            items.put(item.getId(), item);
+            items.add(item);
             StorageItemChange.Record save = StorageItemChange.<T>builder()
                     .item(item)
                     .itemClass((Class<T>) item.getClass())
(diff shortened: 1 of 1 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -81,21 +81,36 @@ public void initialize(
...
+            // Update
+            Model savedItem = items.get(index);
+            if (predicate == null || predicate.evaluate(savedItem)) {
+                items.remove(index);
+                save(item, initiator, predicate, onSuccess, onError);
+                return;
+            }
+            onError.accept(new DataStoreException(
+                    "Conditional check failed.",
+                    "Verify the saved models."));
```
jamesonwilliams: Maybe, instead, `"Verify that there is a model in the datastore that matches the provided predicate."` ?

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -81,21 +81,36 @@ public void initialize(
...
-                .type(StorageItemChange.Type.SAVE)
-                .predicate(predicate)
-                .initiator(initiator)
-                .build()
-                .toRecord(storageItemChangeConverter);
-        changeRecordStream.onNext(save);
-        onSuccess.accept(save);
+        final int index = indexOf(item);
+        if (index > -1) {
+            // Update
```
jamesonwilliams: You can cut out this inner body, and make it a helper method called `update(..)`, perhaps? This will help keep the line length of the `save(...)` method down. However, if they re-use tons of the same params, it might end up being too minor of a savings.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -81,21 +81,36 @@ public void initialize(
...
+            if (predicate == null || predicate.evaluate(savedItem)) {
+                items.remove(index);
+                save(item, initiator, predicate, onSuccess, onError);
+                return;
+            }
+            onError.accept(new DataStoreException(
+                    "Conditional check failed.",
+                    "Verify the saved models."));
+        } else {
+            // Insert
```
jamesonwilliams: Same as above, consider breaking this stuff out into a helper method called `create`. The semantics would become like "A save is either an update, or a create, depending on whether or not `index` is non-zero."
```
} else {
  create(...);
}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -144,22 +159,23 @@ public void initialize(
...
-                        .predicate(predicate)
-                        .initiator(initiator)
-                        .build()
-                        .toRecord(storageItemChangeConverter);
-                changeRecordStream.onNext(deletion);
-                onSuccess.accept(deletion);
-                return;
-            }
+        final int index = indexOf(item);
+        if (index > -1 && (predicate == null || predicate.evaluate(item))) {
```
jamesonwilliams: You could invert this expression, and put the error handling in the body.

Then, do your happy path back -4 indentation to the left.

e.g.
```
if (unlikeBadThing) {
    announce("Hey no fair...");
    return;
}
// otherwise
performBusiness();
asUsual();
etc();
```

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
raphkim: yea that's the conclusion i've reached: that they share too many parameters to actually take advantage of this

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
raphkim: good idea. Also caught a bug where i checked the predicate condition against item to delete rather than on the already stored item.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -160,23 +160,30 @@ public void initialize(
...
-                    .toRecord(storageItemChangeConverter);
-            changeRecordStream.onNext(deletion);
-            onSuccess.accept(deletion);
-        } else {
+        if (index < 0) {
             onError.accept(new DataStoreException(
-                    "Item does not exist or conditional check failed.",
-                    "Verify the saved models."));
+                    "This item was not found in the datastore: " + item.toString(),
+                    "Use save() function to create models to store."
```
jamesonwilliams: I think you need a `return;` here, too, afteryou fire `onError.accept(...)`.

raphkim: AHA! pushed the change right before this comment ;)

PR #230, opened 2020-01-13
Add User-Agent header to all outbound network requests

jamesonwilliams:

### DIFF
--- a/core/src/main/java/com/amplifyframework/util/UserAgent.java
+++ b/core/src/main/java/com/amplifyframework/util/UserAgent.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -0,0 +1,132 @@
...
+     * Gets a String to use as the value of a User-Agent header.
+     * @return A value for a User-Agent header.
+     */
+    @SuppressLint("SyntheticAccessor")
+    @NonNull
+    public static String string() {
+        if (instance == null) {
+            instance = new UserAgent.Builder()
+                .libraryName("amplify-android")
+                .libraryVersion("0.9")
```
desokroshan: I am wondering if there is a way to read the version number from a version file so user-agent is updated automatically for future versions without requiring manual intervention.

jamesonwilliams: `BuildConfig.VERSION_NAME` is one possible candidate for this. It would mean we'd be updating the `build.gradle`'s central `versionName` field, instead of this `.java` file. Which would at least be a step in the right direction.

Perhaps there's a way to read the pom artifact version at runtime, too. Agree that it warrants a closer look. I'll do a quick investigation.

configuration/checkstyle-rules.xml
```diff
@@ -26,6 +26,8 @@
     <!-- Check that source files have a license header. -->
     <module name="Header">
         <property name="headerFile" value="configuration/java.header" />
+        <!-- TODO: re-instate checks for this line. 2019/2020 requires RegexpHeader -->
+        <property name="ignoreLines" value="2" />
```
raphkim: #231

core/src/main/java/com/amplifyframework/util/UserAgent.java
jamesonwilliams: ^ Did the thing

PR #233, opened 2020-01-16
[Analytics] Support Auto session tracking initial commit

desokroshan:
*Issue #, if available:*

*Description of changes:*


aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+import android.app.Activity;
+import android.app.Application;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.os.Bundle;
+import android.util.Log;
+
+import com.amazonaws.mobile.auth.core.internal.util.ThreadUtils;
```
jamesonwilliams: It would be better to use a utility that exists in Amplify, directly. This utility is from a dependency which may/may not be removed in the future (when we do the auth category), and also it is marked `.internal.` (meaning, "don't consume this from another library.")

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+    private boolean inForeground = false;
+    /** Tracks the lifecycle of activities that have not stopped (including those restarted). */
+    private WeakHashMap<Activity, String> activityLifecycleStateMap = new WeakHashMap<>();
+
+    /**
+     * Constructor. Registers to receive activity lifecycle events.
+     * @param application The Android Application class.
+     */
+    public ApplicationLifeCycleHelper(final Application application) {
+        application.registerActivityLifecycleCallbacks(this);
```
jamesonwilliams: A problem here is that you have a way to register the callbacks, but no way to unregister the callbacks.

Someday, `Amplify` will probably have some kind of `teardown()` method. And it would call through and we'd want to do:
```
application.unregisterActivityLifecycleCallbacks(this);
```

So, either:
1. Just stash a reference to the `Application` into an instance variable, and provide two new methods, `start()`, and `stop()`. These methods would call `application.registerActivityLifecycleCallbacks(this)` and `application.unregisterActivityLifecycleCallbacks(this)`.
2. Don't bother passing an `Application` at all. Just document in the class Javadoc for `ApplicationLifecycleHelper` that you intend for instances of this class to registered elsewhere, by someone else making calls to `registerActivityLifecycleCallbacks(Application.ActivityLifecycleCallbacks)` and `unregisterActivityLifecycleCallbacks(Application.ActivityLifecycleCallbacks)`.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+        final ScreenOffReceiver screenOffReceiver = new ScreenOffReceiver();
+        application.registerReceiver(screenOffReceiver, new IntentFilter(ACTION_SCREEN_OFF));
```
jamesonwilliams: Same issue here, there's no "undo" method, if you do this in the constructor. So, it probably shouldn't happen in the constructor.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+ * Any activity lifecycle callbacks can easily be overriden if additional handling
+ * is needed. Just be sure to call through to the super method so that this class
+ * will still behave as intended.
+ **/
+public abstract class ApplicationLifeCycleHelper implements Application.ActivityLifecycleCallbacks {
+    private static final String LOG_TAG = ApplicationLifeCycleHelper.class.getSimpleName();
+    private static final String ACTION_SCREEN_OFF = "android.intent.action.SCREEN_OFF";
+    private boolean inForeground = false;
+    /** Tracks the lifecycle of activities that have not stopped (including those restarted). */
+    private WeakHashMap<Activity, String> activityLifecycleStateMap = new WeakHashMap<>();
```
jamesonwilliams: `ActivityBase` and `AppCompatActivity` have existing solutions to track/obtain the current lifecycle. Checkout the AndroidX Lifecycle Components, with which V7 App Compat Activity is integrated.

However, I can see an advantage to having something that works with every single Activity, ever. And for that, I guess you might need to have your own tracker.

A good name for this class might be `ApplicationForegroundTracker`, instead of `ApplicationLifecycleHelper`.

Anyway, even if you implement your own tracker, you can still use as much of AndroidX Lifecycle here as possible. Instead of using strings, you should use `enum`s, which provide a bounded set of possible states the map value can be in. And, as it turns out, that `enum` is already defined by AndroidX Lifecycle --

So this map would (probably) become:
```
private Map<Activity, Lifecycle.State> activityStates = new WeakHashMap<>();
```
(Ref: [`LifecycleState`](https://developer.android.com/reference/android/arch/lifecycle/Lifecycle.State))

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Called back when your application enters the Foreground.
+     */
+    protected abstract void applicationEnteredForeground();
+
+    /**
+     * Called back when your application enters the Background.
+     */
+    protected abstract void applicationEnteredBackground();
```
jamesonwilliams: Instead of doing this as an `abstract class`, and requiring someone to extend it, I suggest to use the "observer" design pattern. If you do that, this class would become a `final class`, and would provide a way to be "observed."

A classic Java implementation of this would be like:
```
final class ApplicationForegroundTracker {
    private final Set<Observer> observers;

    ApplicationForegroundTracker() {
        this.observers = new HashSet<>();
    }

    void registerForegroundObserver(Observer observer) {
        synchronized (observers) {
            this.observers.add(ObjectsCompat.requireNonNull(observer));
        }
    }

    void unregisterForegroundObserver(Observer observer) {
        synchronized (observers) {
            observers.remove(Objects.requireNonNull(observer));
        }
    }

    // Somewhere in your code, when its time to notify that you have a new state ...
    // You can call this either like notify(observer::onApplicationEnteredForeground);
    // or like notify(observer::onApplicationEnteredBackground);
    void notify(NotificationStrategy strategy) {
        Set<Observer> threadSafeObserversCopy = new HashSet<>();
        synchronized (observers) {
            threadSafeObserversCopy.addAll(observers);
        }

        for (Observer observer: threadSafeObserversCopy) {
            strategy.apply(observer);
        }      
    }

    interface Observer {
        void onApplicationEnteredForeground();
        void onApplciationEnteredBackground();
    }

    // An internal interface to define a function which calls back an observer.
    // This is a strategy design pattern. One strategy might be to call 
    // {@link Observer#onApplicationEnteredForeground()}. Another strategy might be
    // to call {@link Observer#onAppliactionEnteredBackground()}.
    // This strategy is used to execute a small bit of code inside of the notification routine,
    // where a lot of the same setup is used regardless of the notification strategy.
    interface NotificationStrategy {
        void apply(Observer observer);
    }
}
```

Anyway, then you'd use it like:
```
final ApplicationForegroundTracker tracker = new AppliactionForegroundTracker();
final ApplicationForegroundTracker.Observer observer = new ApplicationForegroundTracker.Observer() {
    @Override public void onApplicationEnteredForeground() {
        analyticsSession.resume(); // or whatever
    }
    @Overridd public void onApplicactionEnteredBackground() {
        analyticsSession.pause(); // or whatever
    }
};

tracker.addForegroundObserver(observer);
...
tracker.removeForegroundObserver(observer);
```

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Called from onActivityCreated and onActivityStarted to handle when the application enters
+     * the foreground.
+     */
+    private void handleOnCreateOrOnStartToHandleApplicationEnteredForeground() {
+        // if nothing is in the activity lifecycle map indicating that we are likely in the background, and the flag
+        // indicates we are indeed in the background.
+        if (activityLifecycleStateMap.size() == 0 && !inForeground) {
```
jamesonwilliams: When this class first runs, you won't know the right value of `inForeground` for sure. Did this code start executing when the initial state was in foreground? Or in background?

So, I propose that you change this from a `boolean` to an `enum`, that looks like:
```
enum ForegroundState {
    FOREGROUND,
    BACKGROUND,
    UNKNOWN;
}
```

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
desokroshan: I have updated to use LifecycleState and updated the naming as well.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/ApplicationLifeCycleHelper.java
desokroshan: Made it a final class. We want to provide this feature automatically where session are recorded for every time app goes to back ground after it is brought to foreground. Enduser should not have to write any code for it. Will discuss offline

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java
```diff
@@ -77,20 +75,19 @@ void stopSessionTracking(final Application application) {
...
     public void onActivityStarted(final Activity activity) {
         Log.d(LOG_TAG, "onActivityStarted " + activity.getLocalClassName());
-        handleOnCreateOrOnStartToHandleApplicationEnteredForeground();
         activityLifecycleStateMap.put(activity, Lifecycle.State.STARTED);
     }
 
     @Override
     public void onActivityResumed(final Activity activity) {
         Log.d(LOG_TAG, "onActivityResumed " + activity.getLocalClassName());
+        handleOnCreateOrOnStartToHandleApplicationEnteredForeground();
```
jamesonwilliams: This method name is long to the point of being unreadable. Maybe just inline that method, here? Or, name it like `checkForApplicationEnteredForeground()`?  As a bonus, this would match your `checkForApplicationBackground()` method.

desokroshan: I have some more clean-ups lined up where I have this change, cleaning up comments and removing Screen off receiver because it turn out screen turn off also causes running activity to stop thus giving a onStop callback. This PR is WIP and there is no way to make it a Draft PR once its marked ready for review :/

core/src/main/java/com/amplifyframework/util/ThreadUtils.java
```diff
@@ -0,0 +1,37 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.util;
+
+import android.os.Handler;
+import android.os.Looper;
+
+public final class ThreadUtils {
```
jamesonwilliams: In our call, we agreed that this can be removed.

jamesonwilliams:
> I missed this in the PR you had for the initial commit but believe it's important enough to mention now for fixing before GA - in the registerGlobalProperties method you are taking a custom Properties object which is literally just a Map so I think that the Properties object should be removed and the method signature updated to just be Map<String, Property> which would be more in line with iOS and also more in line with the unregisterGlobalProperties method besides being a simpler, easier interface for customers.

I agree that the mismatch is weird.
```
    // @param properties Map of global properties wrapped as an instance of {@link Properties}
    void registerGlobalProperties(@NonNull Properties properties) throws AnalyticsException;
```
```
    // @param keys a collection of property names to unregister
    void unregisterGlobalProperties(@NonNull Set<String> keys);
```

If I re-call, I was saying that we should use data models for our types, instead of just having lists and maps of strings everywhere.

But, it didn't get carried through to the Javadoc or to the unregister API.

_Alternately_, we should use lists and maps of strings everywhere, and be consistent. :-) 

Right now we're doing a mix of both, and that's probably worse yet.

___To be sure, we should address this discussion in a separate PR from the current PR. The current PR is about auto session tracking, not global properties APIs.___

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Tracks when the host application enters or leaves foreground.
+ * The constructor registers to receive activity lifecycle events.
+ **/
+public final class AutoSessionTracker implements Application.ActivityLifecycleCallbacks {
+    private static final String LOG_TAG = AutoSessionTracker.class.getSimpleName();
+    private final SessionClient sessionClient;
+    private final AnalyticsClient analyticsClient;
+    private boolean inForeground = false;
+    private int activityCount = 0;
```
jamesonwilliams: trivial comment: move assignments into constructor

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Called when the application enters the background.
+     */
+    void applicationEnteredBackground() {
+        Log.d(LOG_TAG, "Application entered the background.");
+        sessionClient.stopSession();
+        analyticsClient.submitEvents();
+
```
jamesonwilliams: nit: an extra empty line here

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java
```diff
@@ -0,0 +1,146 @@
...
+            applicationEnteredForeground();
+        }
+    }
+
+    /**
+     * Called from onActivityStopped to check if the application receded to the background.
+     */
+    private void checkIfApplicationEnteredBackground() {
+        // If the App is in the foreground and there are no longer any activities that have not been stopped.
+        if ((activityCount == 0) && inForeground) {
```
jamesonwilliams: nit: remove extra set of parens around `activityCount == 0`.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AutoSessionTracker.java
jamesonwilliams: Minor: naming, maybe, `foregroundActivityCount`?

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointClientFactory.java
```diff
@@ -47,7 +46,7 @@
     private PinpointClientFactory() {
     }
 
-    static AnalyticsClient create(Context context,
+    static PinpointManager create(Context context,
```
jamesonwilliams: The class that creates a `PinpointManager` should probably be called a `PinpointManagerFactory`, not a `PinpointClientFactory`, I guess.

aws-analytics-pinpoint/src/test/java/com/amplifyframework/analytics/pinpoint/AutoSessionTrackerTest.java
```diff
@@ -0,0 +1,185 @@
...
+        callbacks.onActivityCreated(activity, bundle);
+        callbacks.onActivityStarted(activity);
+        callbacks.onActivityResumed(activity);
+
+        // Activity stopped when home button is pressed.
+        callbacks.onActivityPaused(activity);
+        callbacks.onActivityStopped(activity);
+
+        // Then: Make sure startSession and stopSession are invoked in that order.
+        InOrder inOrder = inOrder(sessionClient);
```
jamesonwilliams: I've never seen this `InOrder`, but it looks quite useful. Nice find, and nice application.

aws-analytics-pinpoint/src/test/java/com/amplifyframework/analytics/pinpoint/AutoSessionTrackerTest.java
```diff
@@ -0,0 +1,185 @@
...
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests the {@link AutoSessionTracker}.
+ */
+@RunWith(RobolectricTestRunner.class)
+public final class AutoSessionTrackerTest {
```
jamesonwilliams: For some of these, you have "app is in goreground" was a "when" step, but it is more like a "given."

For example, "Given that the app had already been opened, when the user hits the home button, then the session is stopped."

This will change your "then" logic slightly too, since I think you'll only want to verify `stopSession()`, in that case, not _both_ `startSession()` and `stopSession()`, back to back.

Although, that `InOrder` is cool!! I'm surprised I've never seen it before. Nice find.

aws-analytics-pinpoint/src/test/java/com/amplifyframework/analytics/pinpoint/AutoSessionTrackerTest.java
desokroshan: Came across while going through the documentation 😄

PR #243, opened 2020-01-21
OkHttp 4.3.0 -> 4.3.1

jamesonwilliams:


jamesonwilliams:
@raphkim Just a minor hygeine thing. We should always do things like this as we see them, as opposed to letting them accrue as technical debt, over a period of time. Some good invariants for the project should include:
- Dependencies are always the latest
- Tests are passing, with those latest deps

Issue #240, opened 2020-01-20
No option to get the URL of item in S3 bucket using key.


zubairzahoor:
I have files(pictures) in my S3 bucket and I am displaying them on screen. I don't want to use the `Amplify.Storage.downloadFile(...)`
as it takes a lot of time to download and then render. I want to use a picture loading library like glide and to do that I need to get the URL of the picture (probably by using the key). 
Currently there is no such option in android.

zubairzahoor:
I guess AmazonS3 SDK has something like this.
```
AmazonS3 s3 = AmazonUtil.getS3Client(this);
URI uri = s3.generatePresignedUrl(bucket, key, expiry);
```

zubairzahoor:
In JS you can easily get the publically accessible URL by using
```
Storage.get('test.txt')
    .then(result => console.log(result))
    .catch(err => console.log(err));
```
Any workaround for doing the same in android?

raphkim:
This is a feature that currently isn't included yet in preview release, but we certainly hope to include it in our upcoming releases. 

Temporary workaround for this usecase would be to utilize `getEscapeHatch()` method to obtain a s3 client with which you can call [`generatePresignedUrl()`](https://github.com/aws-amplify/aws-sdk-android/blob/7e04c0269e59f8eb1db9c98b59a0197b63501aba/aws-android-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L3401).

zubairzahoor:
@raphkim Thanks for the response but is it really applicable in my case. I have to access a 'PROTECTED' resource in s3 and there is no configurable option for AmazonS3Client for achieving that. S3ClientOptions does not let me change the access level like StorageDownloadFileOptions does.
How do I call generatePresignedUrl() for a protected image(resource)?

raphkim:
Amplify Storage internally constructs a "service key", which is a combination of `StorageDownloadFileOptions`, `AWSMobileClient#getIdentityId()`, and S3 key. This service key can be constructed using [`S3RequestUtils`](https://github.com/aws-amplify/amplify-android/blob/master/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java) and be passed directly as a key along with the bucket name to `generatePresignedUrl()`. I imagine the sample code for temporary workaround would look like the following:

```
StorageDownloadFileOptions options = StorageDownloadFileOptions.defaultInstance(); // This is your download file options
[...]

PR #248, opened 2020-01-29
[Storage] Support getUrl()

raphkim:
*Issue #, if available:*
#240

*Description of changes:*
Add the items related to obtaining pre-signed URL for an S3 object.


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageGetUrlOperation.java
```diff
@@ -0,0 +1,97 @@
...
+                    getRequest().getAccessLevel(),
+                    identityId,
+                    getRequest().getKey(),
+                    getRequest().getTargetIdentityId()
+            );
+
+            try {
+                this.url = storageService.getPresignedUrl(serviceKey, getRequest().getExpires());
+                onResult.accept(StorageGetUrlResult.fromUrl(url));
+            } catch (Exception exception) {
```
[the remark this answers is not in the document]
raphkim: I did this because I honestly couldn't figure out how `AmazonS3Client#generatePresignedUrl()` could ever fail. As long as the credentials are correctly setup in the first try-catch, I don't see a case where PresignedUrl generation would fail. I only try-caught it because it says it can throw `AmazonClientException`.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/request/AWSS3StorageGetUrlRequest.java
```diff
@@ -0,0 +1,81 @@
...
+    private final String targetIdentityId;
+    private final int expires;
+
+    /**
+     * Constructs a new AWSS3StorageGetUrlRequest.
+     * @param key key for item to obtain URL for
+     * @param accessLevel Storage access level
+     * @param targetIdentityId The user id for the user this URL should be generated for
+     *                         (to override it from assuming the currently logged in user)
+     * @param expires The number of seconds before the URL expires
```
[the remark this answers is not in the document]
raphkim: i got the name directly from iOS but I'm also not a fan of "expires". I'll leave it as is for platform parity, but we should come up with a better name for it collectively. I was thinking if we could just pass a `Date` directly with the name `expiration` because that's how the S3 client takes it anyways.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -141,6 +148,36 @@ public AmazonS3Client getEscapeHatch() {
...
+    public StorageGetUrlOperation<?> getUrl(
+            @NonNull String key,
+            @NonNull StorageGetUrlOptions options,
+            @NonNull Consumer<StorageGetUrlResult> onSuccess,
+            @NonNull Consumer<StorageException> onError) {
+        AWSS3StorageGetUrlRequest request = new AWSS3StorageGetUrlRequest(
+                key,
+                options.getAccessLevel() != null ? options.getAccessLevel() : defaultAccessLevel,
+                options.getTargetIdentityId(),
+                options.getExpires() != 0 ? options.getExpires() : DEFAULT_EXPIRE_IN_SECONDS
```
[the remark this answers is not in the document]
raphkim: It does, but that happens if `null` is passed as expiration `Date` object. I just think it's rather dirty to check if `expires` is zero and and pass a null object to the backend.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageGetPresignedUrlOperation.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param onError Notified upon URL generation error
+     */
+    public AWSS3StorageGetPresignedUrlOperation(
+            @NonNull StorageService storageService,
+            @NonNull ExecutorService executorService,
+            @NonNull AWSS3StorageGetPresignedUrlRequest request,
+            @NonNull Consumer<StorageGetUrlResult> onSuccess,
+            @NonNull Consumer<StorageException> onError
+    ) {
+        super(request);
```
jamesonwilliams: Later there is `getRequest().weBelieveItsNotNull()` so if super doesn’t require non-bull, this constructor should

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/StorageService.java
```diff
@@ -32,6 +33,14 @@
  */
 public interface StorageService {
 
+    /**
+     * Generate pre-signed download URL for an object.
+     * @param serviceKey key to uniquely specify item to generate URL for
+     * @param expires Number of seconds before URL expires
+     * @return A pre-signed URL
+     */
+    URL getPresignedUrl(@NonNull String serviceKey, int expires);
```
jamesonwilliams: Should it be called `generatePresignedUrl` to make clear that the URL doesn’t already exist. Otherwise right now this looks like an accessory method to a cached variable

core/src/main/java/com/amplifyframework/storage/result/StorageGetUrlResult.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Creates a new StorageGetUrlResult containing the presigned URL
+     * of requested object.
+     * @param url The presigned URL
+     * @return A StorageGetUrlResult
+     */
+    @NonNull
+    public static StorageGetUrlResult fromUrl(@NonNull URL url) {
+        return new StorageGetUrlResult(Objects.requireNonNull(url));
```
jamesonwilliams: Frequently these are ordered towards the top to sit near the constructor(s), if any.

core/src/main/java/com/amplifyframework/storage/options/StorageGetUrlOptions.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        /**
+         * Constructs and returns a new immutable instance of the
+         * StorageGetUrlOptions, using the configurations that
+         * have been provided the current instance of the Builder.
+         * @return A new immutable instance of StorageGetUrlOptions
+         */
+        public StorageGetUrlOptions build() {
+            return new StorageGetUrlOptions(this);
```
jamesonwilliams: Do you want to validate anything about the provided values before you build the instance? I.e. a range on the int, null-ness is the provided objects?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/StorageService.java
raphkim: The storage service isn't exposed to the customer anyways so I think it's okay. I gave it the `get` prefix to be more aligned with `getUrl()` public API.

core/src/main/java/com/amplifyframework/storage/options/StorageGetUrlOptions.java
raphkim: To be revisited in `StorageOptions` PR

core/src/main/java/com/amplifyframework/storage/result/StorageGetUrlResult.java
raphkim: Fixed.

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
```diff
@@ -93,6 +97,48 @@ private static StorageCategoryConfiguration buildConfiguration() {
...
+     *         URL from storage service
+     */
+    @Test
+    public void testGenerateUrlGetsPresignedUrl() throws StorageException {
+        final String fromRemoteKey = RandomString.string();
+        final URL urlFromRemoteKey;
+        try {
+            // URL instance cannot be mocked so just make one
+            // https://{random-host}:0/{fromRemoteKey}
+            urlFromRemoteKey = new URL(
```
jamesonwilliams: It would probably work without this random URL if you use Robolectric, which will provide an implementation of andorid's URL. (But, Robolectric is slow. So I do like your solution, too.)

You could pull this out to a `RandomURL.url()` test util! Like:
```
public static URL url() {
    try {
        return new URL(... params ...);
    } catch (MalformedURLException exception) {
        throw new RuntimeException(exception);
    }
}
```

PR #250, opened 2020-01-29
Category tweaks

jamesonwilliams:


core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -167,35 +162,56 @@ public final P getPlugin(@NonNull final String pluginKey) {
...
+    private void validateConfigurationState(ConfigurationState targetState) throws AmplifyException {
+        synchronized (configurationState) {
+            if (!configurationState.get().equals(targetState)) {
+                throw new AmplifyException(
+                    "Amplify was already configured.",
+                    "Be sure to only call Amplify.configure once"
+                );
+            }
+        }
+    }
```
raphkim: We discussed a bug about this :)

Should have `validateConfigured()` and `throwIfConfigured()` separately to clearly indicate two validation cases

jamesonwilliams: Fixing! Thanks for the catch.

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -167,35 +176,46 @@ public final P getPlugin(@NonNull final String pluginKey) {
...
-                    "Make sure you added it with Amplify.addPlugin and then called Amplify.config");
+    protected final P getSelectedPlugin() throws IllegalStateException {
+        synchronized (configurationState) {
+            if (!ConfigurationState.CONFIGURED.equals(configurationState.get())) {
+                throw new IllegalStateException(
+                    "Category " + getCategoryType() + " is not configured. " +
+                    "Ensure that you have configured the category before trying to use it."
+                );
+            }
         }
```
raphkim: isn't this just 
```
if (!isConfigured()) {
    throw ...
}
```
?

jamesonwilliams: Yes! I'll do it this way instead.

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -146,17 +153,17 @@ public final void removePlugin(@NonNull P plugin) {
      * Retrieve a plugin by its key.
      * @param pluginKey A key that identifies a plugin implementation
      * @return The plugin object associated to pluginKey, if registered
-     * @throws IllegalStateException If there is no plugin with the given key
+     * @throws IllegalStateException If the requested plugin does not exist
      */
     @NonNull
-    public final P getPlugin(@NonNull final String pluginKey) {
+    public final P getPlugin(@NonNull final String pluginKey) throws IllegalStateException {
```
TrekSoft: What's the benefit to specifying the runtime exception in the method header? Also, that won't force calling methods to catch/throw it will it?

jamesonwilliams: My intention here was just to provide clear documentation on what can happen. Since `IllegalStateException` derives from `RuntimeException`, the caller still won't need to `try`-`catch` it explicitly.

_However_, I didn't honestly know the best practice on this. Should one put runtime exceptions on the method signature, or not? Apparently, the guidance is to _not_ do this. [Here's a StackOverflow post about it](https://stackoverflow.com/a/12269443/695787), which contains this quote from Oracle documentation:

> Having to add runtime exceptions in every method declaration would reduce a program's clarity.

Anyway, these are all over the code-base at this point. I'll do a separate CR to remove them all in one pass.

PR #251, opened 2020-01-29
[DataStore] Use Hub notifications to communicate Sync Engine events

jamesonwilliams:
Previously, DataStore instrumentation tests were using Sleep to wait an
arbitrary amount of time for synchronization tasks to complete. See the
new statement in Sleep.java about why waiting for a random amount of
time is a bad strategy for synchronization. Instead, DataStore
synchronization events are now published to the Hub.

Some test utilities are added to help await Hub events, in tests.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
@@ -21,18 +21,19 @@ import androidx.test.core.app.ApplicationProvider;
 
 import com.amplifyframework.AmplifyException;
 import com.amplifyframework.api.ApiException;
+import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.testmodels.commentsblog.Blog;
 import com.amplifyframework.testmodels.commentsblog.BlogOwner;
-import com.amplifyframework.testutils.Sleep;
+import com.amplifyframework.testutils.HubAccumulator;
 import com.amplifyframework.testutils.SynchronousApi;
 import com.amplifyframework.testutils.SynchronousDataStore;
 
+import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Test;
-
-import java.util.concurrent.TimeUnit;
 
 import static org.junit.Assert.assertEquals;
 
@@ -44,12 +45,12 @@ import static org.junit.Assert.assertEquals;
  */
 public final class AWSDataStorePluginInstrumentedTest {
     private static final String DATABASE_NAME = "AmplifyDatastore.db";
-    private static final long DATA_STORE_OP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(2);
-    private static final long NETWORK_OP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(2);
     private static Context context;
     private static AWSDataStorePlugin awsDataStorePlugin;
     private static SynchronousApi api;
     private static SynchronousDataStore dataStore;
+    private static HubAccumulator outboundModelEventAccumulator;
+    private static HubAccumulator inboundModelEventAccumulator;
 
     /**
      * Enable strict mode for catching SQLite leaks.
@@ -75,6 +76,30 @@ public final class AWSDataStorePluginInstrumentedTest {
         context = ApplicationProvider.getApplicationContext();
         api = SynchronousApi.singleton();
         dataStore = SynchronousDataStore.singleton();
+        outboundModelEventAccumulator = HubAccumulator.create(HubChannel.DATASTORE, event ->
+            DataStoreChannelEventName.PUBLISHED_TO_CLOUD.toString().equals(event.getName())
+        );
+        inboundModelEventAccumulator = HubAccumulator.create(HubChannel.DATASTORE, event ->
+            DataStoreChannelEventName.RECEIVED_FROM_CLOUD.toString().equals(event.getName())
+        );
+    }
+
+    /**
+     * Before each test, clear and restart the hub event accumulators.
+     */
+    @Before
+    public void beforeEachIndividualTest() {
+        outboundModelEventAccumulator.stop().clear().start();
+        inboundModelEventAccumulator.stop().clear().start();
+    }
+
+    /**
+     * After each test, stop and clear the hub event accumulators.
+     */
+    @After
+    public void afterEachIndividualTest() {
+        outboundModelEventAccumulator.stop().clear();
+        inboundModelEventAccumulator.stop().clear();
     }
 
     /**
@@ -91,8 +116,8 @@ public final class AWSDataStorePluginInstrumentedTest {
             .build();
         dataStore.save(localCharley);
 
-        // Wait a bit. TODO: this is lame; how to tell deterministically when sync engine has sync'd?
-        Sleep.milliseconds(NETWORK_OP_TIMEOUT_MS + DATA_STORE_OP_TIMEOUT_MS);
+        // Wait for a Hub event telling us that our Charley model got published to the cloud.
+        outboundModelEventAccumulator.takeOne();
 
         // Try to get Charley from the backend.
         BlogOwner remoteCharley = api.get(BlogOwner.class, localCharley.getId());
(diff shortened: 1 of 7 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
```diff
@@ -91,8 +116,8 @@ public void blogOwnerSavedIntoDataStoreIsThenQueriableInRemoteAppSyncApi() throw
             .build();
         dataStore.save(localCharley);
 
-        // Wait a bit. TODO: this is lame; how to tell deterministically when sync engine has sync'd?
-        Sleep.milliseconds(NETWORK_OP_TIMEOUT_MS + DATA_STORE_OP_TIMEOUT_MS);
+        // Wait for a Hub event telling us that our Charley model got published to the cloud.
+        outboundModelEventAccumulator.takeOne();
```
undefobj: What's a Charley model?

undefobj: Oh I see, `.name("Charley Crockett")`
I was thinking this was a new data structure 😂

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SyncEngine.java
```diff
@@ -137,6 +140,11 @@ private void startModelSubscriptions() {
                         AmplifyException.TODO_RECOVERY_SUGGESTION
                     );
             }
+
+            // Notify Hub that we just updated the local storage.
+            HubEvent<? extends Model> receivedFromCloudEvent =
+                HubEvent.create(DataStoreChannelEventName.RECEIVED_FROM_CLOUD, mutation.model());
+            Amplify.Hub.publish(HubChannel.DATASTORE, receivedFromCloudEvent);
```
undefobj: One thought here is do we want to have this on a channel for DataStore or API? We've already seen customers asking about using the API category and DataStore together, so there might be merit to having everything going over the API channel for events from AppSync. OTOH we could always publish to both or switch it later. Food for thought.

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SyncEngine.java
```diff
@@ -151,7 +159,12 @@ private void startDrainingChangeJournal() {
                 .flatMapSingle(this::publishToNetwork)
                 .flatMapSingle(storageItemChangeJournal::remove)
                 .subscribe(
-                    processedChange -> LOG.info("Change processed successfully! " + processedChange),
+                    processedChange -> {
+                        LOG.info("Change processed successfully! " + processedChange);
```
undefobj: Is this just for local debugging and should we remove?

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -0,0 +1,82 @@
...
+import com.amplifyframework.hub.HubEvent;
+
+import java.util.Objects;
+
+/**
+ * An enumeration of the names of events relating the the {@link DataStoreCategory},
+ * that are published via {@link HubCategory#publish(HubChannel, HubEvent)} on the
+ * {@link HubChannel#DATASTORE} channel.
+ */
+public enum DataStoreChannelEventName {
```
undefobj: I like this but do we have a similar structure on iOS that we could align with names? @palpatim @drochetti

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * An item in the local storage has been published to the cloud.
+     * An event that uses this value for {@link HubEvent#getName()} will contain a model object
+     * in the {@link HubEvent#getData()}.
+     */
+    PUBLISHED_TO_CLOUD("published_to_cloud"),
```
undefobj: Another thing that we are looking at is "local only models" which do not get sync'd to the cloud. A few customers have been asking about this and @drochetti and Manuel are doing a design. In that case we might want to still publish to hub for state management in your app, so it's worth keeping that in mind for the list of possible events - e.g. `PUBLISHED_TO_LOCAL`

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Accumulates {@link HubEvent}s received on a {@link HubChannel}, into an in-memory
+ * buffer. These may be queried by tests in a synchronous way to check if they have arrived.
+ */
+@SuppressWarnings({"unused", "UnusedReturnValue"})
+public final class HubAccumulator {
+    private final HubChannel channel;
+    private final HubEventFilter filter;
```
undefobj: I'm not sure if we ever got the filter implemented fully in Hub on Android, just an FYI

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
jamesonwilliams: Haha we should invent it!!

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SyncEngine.java
jamesonwilliams: Hm, I should change it from `.info` to `.debug`. Since the default is to only print info's or better, debug would effectively turn if off (unless somebody changes the log level.)

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
jamesonwilliams: For Hub, we did, it's [more or less just a single `if (...)` check](https://github.com/aws-amplify/amplify-android/blame/e9536f2df8f7f7803faefa4cf5926103b351dab0/core/src/main/java/com/amplifyframework/hub/AWSHubPlugin.java#L66-L73), in the `publish(...)` method.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
jamesonwilliams: Okay, cool. Yea, as you note, the `DataStoreChangeEventName` is really part of the API surface itself, since customers will engage with these names to understand the events on the Hub. I suspect this file will see a bunch more useful stuff come into it.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
jamesonwilliams: Good point - since this is API surface, we should get the names and behavior aligned across platforms.

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SyncEngine.java
jamesonwilliams: Since this event was initiated by the datastore, I thought to put it out onto the datastore channel.

I think there will be fewer hub events for the API category, overall. Although, there might be some good reasons to use Hub for subscription events, paging on list APIs, retries on recoverable error, things like this.

jamesonwilliams:
> Looks pretty straightforward overall to me, just some comments based on strategic alignment like we talked about earlier today. It's not immediately clear to me what `HubAccumulator` is needed for in the tests though, I'm guessing you're thinking about high volume messages?

So much of the code in AWS Mobile is async, and the easiest was to write tests is basically as check-list of synchronous statements. Did A work? Ok, did B work? Ok, etc. So, I've been investing time in writing all of these little gizmos on Android, `Await.java`, `HubAccumulator.java`, etc., that encapsulate some best practices for using latches. It's not stuff we want in our production source code, but if we have to have it in tests, we might as well do it fully-dressed.

The `HubAccumulator` is a test utility which decouples the "when stuff shows up on hub" from the "when I want stuff to be available on hub." A listener can query the accumulator, in a synchronous way, for an event. If it hasn't shown up, it waits. If it has, it just returns.

undefobj:
> > Looks pretty straightforward overall to me, just some comments based on strategic alignment like we talked about earlier today. It's not immediately clear to me what `HubAccumulator` is needed for in the tests though, I'm guessing you're thinking about high volume messages?
> 
> So much of the code in AWS Mobile is async, and the easiest way to write tests is basically as a check-list of synchronous statements: Did A work? Ok, did B work? Ok, etc.
> 
> So, I've been investing time in writing all of these little _test utility_ gizmos on Android, `Await.java`, `HubAccumulator.java`, etc., that encapsulate some best practices for using latches. It's not stuff we want in our production source code, but if we have to have it in tests, we might as well do it fully-dressed.
> 
> The `HubAccumulator` is a test utility which decouples the "when stuff shows up on hub" from the "when (I wish!) stuff is available on hub." A listener can query the accumulator, in a synchronous way, for an event. If it hasn't shown up, it waits. Or if it already has (sometime in the past), it just returns.

Got it. This sounds like a series of patterns that maybe would make a good lunch and learn/talk in the future so that we could encode as standard practice or generalize :-)

PR #253, opened 2020-01-30
Use a centralized thread pool for all of Amplify

jamesonwilliams:


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -162,7 +162,7 @@ public synchronized void initialize(
         Objects.requireNonNull(onSuccess);
         Objects.requireNonNull(onError);
 
-        threadPool.submit(() -> {
+        threadPool.execute(() -> {
```
raphkim: Nice catch. TIL the difference between the two.

core/src/main/java/com/amplifyframework/core/async/AmplifyExecutors.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final String THREAD_GROUP_NAME = "Amplify";
+    private static final ThreadGroup THREAD_GROUP = threadGroup();
+    private static final ExecutorService STANDARD = standardExecutorService();
+    private static final ScheduledExecutorService PERIODIC = scheduledExecutorService();
+
+    @SuppressWarnings("checkstyle:all") private AmplifyExecutors() {}
+
+    /**
+     * Gets a fixed thread pool {@link ExecutorService} with as many threads
+     * as there are processors on the host.
```
raphkim: Maybe have another one where we can set pool size?

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -74,7 +74,7 @@
     private static final AtomicBoolean CONFIGURATION_LOCK = new AtomicBoolean(false);
 
     // An executor on which categories may be initialized.
-    private static final ExecutorService INITIALIZATION_POOL = Executors.newFixedThreadPool(CATEGORIES.size());
```
raphkim: I think it makes sense to keep this logic? What if the host processor only has 1 core (hypothetically speaking), and it began initializing datastore first? All other categories will be blocked until datastore is initialized right?

jamesonwilliams:

PR #254, opened 2020-01-30
[Storage] Refactor and component test

raphkim:
*Issue #, if available:*

*Description of changes:*
- Added component test for Amplify Storage plugin
- Refactor some structures
- Add `@Nullable` and `@NonNull` annotations where applicable

**Functional changes!!!**
- Since `targetIdentityId` used to always overwrite `identityId` in a request, it now combines the logic and sends it as one ID
- `identityId` from mobile client is now obtained at plugin configuration and reused rather than obtaining a new one everytime per request. Identity ID per initialized mobile client instance should be constant


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -138,7 +161,7 @@ public void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context
     @NonNull
     @Override
     public AmazonS3Client getEscapeHatch() {
-        return storageService.getClient();
+        return ((AWSS3StorageService) storageService).getClient();
```
jamesonwilliams: Our `StorageService` is already scoped into this S3 plugin, so I think it'd be fair to just bake an S3 gizmo right into it. Maybe we should use a non-`final` `AWSS3StorageService`?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -190,13 +220,17 @@ public AmazonS3Client getEscapeHatch() {
...
                 key,
                 local,
-                options.getAccessLevel() != null ? options.getAccessLevel() : defaultAccessLevel,
-                options.getTargetIdentityId(),
+                options.getAccessLevel() != null
+                        ? options.getAccessLevel()
+                        : defaultAccessLevel,
+                options.getTargetIdentityId() != null
+                        ? options.getTargetIdentityId()
+                        : userIdentityId,
```
jamesonwilliams: Although this ternary is simple logic, it is repeated enough times that we might as well have a `getTargetIdentityId()` and `getUserIdentityId()`, that we can in-line to all of these request creations.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/IdentityIdProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.amplifyframework.storage.s3;
+
+/**
+ * Interface for providing identity ID of the authorized
+ * requester.
+ */
+public interface IdentityIdProvider {
+    /**
+     * Returns the identity ID.
```
jamesonwilliams: We should figure out what an Identity ID actually is (lol), and update the Javadoc ...

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
```diff
@@ -69,63 +69,49 @@ public AWSS3StorageDownloadFileOperation(
...
+                    getRequest().getTargetIdentityId(),
+                    getRequest().getKey()
+            );
+
+            this.file = new File(getRequest().getLocal()); //TODO: Add error handling if path is invalid
 
             try {
-                identityId = AWSMobileClient.getInstance().getIdentityId();
+                transferObserver = storageService.downloadToFile(serviceKey, file);
+            } catch (Exception exception) {
```
jamesonwilliams: Can you `catch (AMoreSpecificException specificProblem)`?

Ref: [PMD's `AvoidCatchingGenericException` lint documentation](https://pmd.github.io/latest/pmd_rules_java_design.html#avoidcatchinggenericexception)

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageListOperation.java
```diff
@@ -64,35 +65,21 @@ public AWSS3StorageListOperation(
...
-                    onSuccess.accept(result);
-                } catch (Exception exception) {
-                    onError.accept(new StorageException(
-                        "Something went wrong with your AWS S3 Storage list operation",
-                        exception,
-                        "See attached exception for more information and suggestions"
-                    ));
-                }
+                onSuccess.accept(StorageListResult.fromItems(listedItems));
             } catch (Exception exception) {
```
jamesonwilliams: Any change to `catch (SpecificThingOne fineGrainedCatch)`, `catch (SpecificThingTwo fineGrainedThignTwo)` ? Will help create a less generic error message in the `StorageException` that you create

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageRemoveOperation.java
```diff
@@ -64,29 +63,16 @@ public AWSS3StorageRemoveOperation(
...
-                    onSuccess.accept(StorageRemoveResult.fromKey(getRequest().getKey()));
-                } catch (Exception exception) {
-                    onError.accept(new StorageException(
-                        "Something went wrong with your AWS S3 Storage remove operation",
-                        exception,
-                        "See attached exception for more information and suggestions"
-                    ));
-                }
+                onSuccess.accept(StorageRemoveResult.fromKey(getRequest().getKey()));
             } catch (Exception exception) {
```
jamesonwilliams: Here's another one

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageUploadFileOperation.java
```diff
@@ -69,68 +69,55 @@ public AWSS3StorageUploadFileOperation(
...
-                    @Override
-                    public void onError(int transferId, Exception exception) {
-                        onError.accept(new StorageException(
-                            "Something went wrong with your AWS S3 Storage upload file operation",
-                            exception,
-                            "See attached exception for more information and suggestions"
-                        ));
-                    }
-                });
             } catch (Exception exception) {
```
jamesonwilliams: One more

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
```diff
@@ -23,29 +25,46 @@
  * A utility to help form requests to S3.
  */
 public final class S3RequestUtils {
-    private S3RequestUtils() { }
 
-    @SuppressWarnings("JavadocMethod") // TODO: Add missing documentation
+    @SuppressWarnings("WhitespaceAround") // Looks better this way
+    private S3RequestUtils() {}
```
jamesonwilliams: In a separate CR, we should update the checkstyle rules once and for all so that can have empty one-line `{}` bodies on lambdas and constructors

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
```diff
@@ -23,29 +25,46 @@
...
+
+    /**
+     * Amplify Storage implementation with S3 integrates access level
```
jamesonwilliams: This Java doc is AWSOME. Thanks.

core/src/main/java/com/amplifyframework/storage/StorageItem.java
```diff
@@ -0,0 +1,98 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Date;
+
+/**
+ * Used to store the data on each item in a storage.
+ */
+public final class StorageItem {
```
jamesonwilliams: Might as well add `hashCode()`, `equals()`, and `toString()` on this to be safe. I see you are using it in collections (needs hash code) and using it in a jUnit assertion (needs equals).

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
```diff
@@ -0,0 +1,235 @@
...
+        // Since we use a mock TransferObserver, it has no internal logic
+        // to know to call back the listener! So, we simulate the success
+        // callback, as part of our "happy path" test.
+        doAnswer(invocation -> {
+            TransferListener listener = invocation.getArgument(0);
+            listener.onStateChanged(0, TransferState.COMPLETED);
+            return null;
+        })
+                .when(observer)
+                .setTransferListener(any(TransferListener.class));
```
jamesonwilliams: Ugh, yea. Does checkstyle not like it as:
```
    })
    .alignedToSameCol()
```
?
I think I have been getting around this by dropping the first argument to a newline and adding +4 ...
```
doAnswer(
    invocation -> {
    })
    .when(observer)
    .setTransfer(...)
```

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
```diff
@@ -0,0 +1,235 @@
...
+     *         Storage category configuration.
+     */
+    @Before
+    public void setup() throws AmplifyException {
+        this.storage = new StorageCategory();
+        this.storageService = mock(StorageService.class);
+        StorageService.Factory storageServiceFactory = (context, region, bucket) -> storageService;
+        IdentityIdProvider identityIdProvider = RandomString::string;
+        this.storage.addPlugin(new AWSS3StoragePlugin(storageServiceFactory, identityIdProvider));
+        this.storage.configure(buildConfiguration(), mock(Context.class));
```
jamesonwilliams: Ah yea, since you _are_ using `RobolectricTestRunner.class` at this point, you might as well pass in a context from that framework, e.g. `ApplicationProvider.getApplicationContext()` from `dependency.androidx.test.core`

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    private static StorageCategoryConfiguration buildConfiguration() {
+        StorageCategoryConfiguration configuration = new StorageCategoryConfiguration();
+        JSONObject storageJson;
+        try {
+            storageJson = new JSONObject()
```
jamesonwilliams: I realized later, this might as well be:
```
try {
    configuration.populateFromJson(new JSONObject()
````
without all the extra ceremony (currently) to keep the `JSONObject` at method scope.

File: aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java [hunk 14]
```diff
@@ -0,0 +1,235 @@
...
+        // Since we use a mock TransferObserver, it has no internal logic
+        // to know to call back the listener! So, we simulate the success
+        // callback, as part of our "happy path" test.
+        doAnswer(invocation -> {
+            TransferListener listener = invocation.getArgument(0);
+            listener.onStateChanged(0, TransferState.COMPLETED);
+            return null;
+        })
+                .when(observer)
+                .setTransferListener(any(TransferListener.class));
```
jamesonwilliams: You could probably pull this out to a helper method, like:
```
mockTransferListenerResult(TransferState.COMPLETED);

private void mockTransferListenerResult(TransferState transferState) {
    doAnswer(invocation -> {
        TransferListener listener = invocation.getArgument(0);
        listener.onStateChanged(0, transferState);
        return null;
    }).when(observer)
      .setTransferListener(any(TransferListener.class));
}
```

It'd be cool to add a few more tests that test what happens when the transfer state is `FAILED`, or whatever -- other states than `COMPLETE`.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
raphkim: I agree with this, but it is non-trivial to look for all possible ways that `TransferUtility` could throw a runtime exception due to its complexity. For example, it throws `IllegalArgumentException` for multiple reasons and it becomes kind of pointless to catch it and give a generic message for it (and repeat the process for other types of runtime exceptions it could potentially throw).

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
raphkim: I personally dont even like the fact that `StorageService` directly interacts with `TransferObserver`. Considering that our ultimate goal is to not rely on `TransferUtility`, we should look into re-factoring it in the near future

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
raphkim: Storage operations, as currently implemented, does not handle other transfer states. Not sure if that's by design.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
raphkim: I thought about it, but that required passing in `options` parameter to the helper method also, which kind of looked ugly.

PR #255, opened 2020-01-31
Use latest AWS SDK release 2.16.7

jamesonwilliams:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ ext {
     minSdkVersion = 15
     targetSdkVersion = 29
 
-    awsSdkVersion = '2.16.6'
+    awsSdkVersion = '2.16.7'
     dependency = [
         androidx: [
             v4support: 'androidx.legacy:legacy-support-v4:1.0.0',


raphkim:
Thoughts on just using `2.16.+`?

jamesonwilliams:
> Thoughts on just using `2.16.+`?

I can see the appeal, although, I guess it isn't a best practice.

Android Studio will warn:
> "Avoid using '+' in version numbers; can lead to unpredictable and unrepeatable builds."

[Here's some guy going on a rant about the topic.](https://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/)

The Brazil build system has a convention where `2.x` (or whatever) will mean basically `2.+`. What happens in the Brazil ecosystem, at least, is that people accidentally make incompatible changes that break stuff, frequently. And worse, `2.x` usually doesn't point to the latest minor anyway, since there's often no single threaded owner for those versions.

In Gradle, we don't have the "isn't actually the latest" problem. But we might still have the second problem. Even in highly public projects like our dependencies, there's enough churn that people break things in interesting ways with some frequency (example: right now, we have a hack for robolectric, to ignore autovalue.)

PR #257, opened 2020-02-05
Fix the error that caused configuration to fail unless mobile client …

raphkim:
…was initialized

*Issue #, if available:*

*Description of changes:*
In the last PR, I cached `identityId` in the configuration step. This required obtaining it from `AWSMobileClient`, which may not always be initialized at the time of plugin configuration. 

I reverted my change so that it is obtained each time a method is invoked rather than at configuration step. 

### DIFF
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
@@ -69,7 +69,6 @@ public final class AWSS3StoragePlugin extends StoragePlugin<AmazonS3Client> {
     private final ExecutorService executorService;
 
     private StorageService storageService;
-    private String userIdentityId;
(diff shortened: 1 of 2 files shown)


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -315,6 +313,10 @@ public AmazonS3Client getEscapeHatch() {
         return operation;
     }
 
+    private String getUserIdentityId() {
+        return identityIdProvider.getIdentityId();
```
jamesonwilliams: Maybe just inline this instead of using the getUserIdentityId()?

raphkim: That was how I initially had it, but I just force pushed this change because it looked ugly lol

PR #259, opened 2020-02-11
[Storage] Begin adding integration tests

raphkim:
*Issue #, if available:*

*Description of changes:*
* Add integration test for storage access level
* Add integration test for pause, resume, cancel
* Re-organize test configuration in S3 bucket
* Add `testconfiguration.json` for credentials and other testing configuration that does not belong in either `awsconfiguration.json` or `amplifyconfiguration.json`
* Add `AmplifyTestBase.java` to read from `testconfiguration.json`
  * TODO: Make other instrumentation tests extend it
* Make S3 operations publish to `Amplify.Hub`
  * TODO: Decide on the actual structure of the published events
* Abstract out `StorageOptions` to remove duplicate code between different types of storage options

~~**CURRENTLY BLOCKED BY #260**~~ (merged)


aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,224 @@
...
+        largeFileKey = getS3Key(DEFAULT_ACCESS_LEVEL, largeFileName);
+        smallFileKey = getS3Key(DEFAULT_ACCESS_LEVEL, smallFileName);
+    }
+
+    /**
+     * Cleans up the S3 bucket for files that were uploaded
+     * during testing processes.
+     */
+    @After
+    public void cleanUp() {
```
lawmicha: I just remembered that I may have set up a TTL on all objects in the bucket. Clean up may be best effort but could be test execution may be terminated for some reason. You can do the same for the bucket if this isn’t sharing the same resource as iOS

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,224 @@
...
+    @SuppressWarnings("unchecked")
+    public void testUploadFileIsCancelable() throws Exception {
+        final CountDownLatch canceled = new CountDownLatch(1);
+
+        // Begin uploading large file
+        StorageUploadFileOperation<?> op = Amplify.Storage.uploadFile(
+                largeFileName,
+                largeFile.getAbsolutePath(),
+                options,
+                onResult -> fail("Upload finished before being successfully cancelled."),
```
lawmicha: Is there onProgress on Amplify.Storage.download|upload’s?

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: Good idea!

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: There isn't at the moment. Progress will only be accessible via Hub in Android

raphkim: Speaking of which, it might be a good idea to add progress directly to the operation? I'll look into that idea.

aws-storage-s3/build.gradle
```diff
@@ -30,5 +30,11 @@ dependencies {
     testImplementation dependency.mockito
     testImplementation dependency.robolectric
     testImplementation dependency.androidx.test.core
+
+    androidTestImplementation project(path: ':testutils')
+    androidTestImplementation dependency.androidx.annotation
+    androidTestImplementation dependency.androidx.test.runner
+    androidTestImplementation dependency.androidx.test.junit
+    androidTestImplementation dependency.aws.mobileclient
```
TrekSoft: We already have mobile client as a dependency above - do we still need to include it as a test implementation?

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,258 @@
...
+    /**
+     * Test downloading with public access without signing in.
+     *
+     * @throws Exception if download is unsuccessful
+     */
+    @Test
+    public void testDownloadUnauthenticatedPublicAccess() throws Exception {
+        final StorageAccessLevel accessLevel = StorageAccessLevel.PUBLIC;
+
+        upload(accessLevel);
```
TrekSoft: Rather than upload the upload file during each download test (which would result in download tests giving failures if there's only a problem with upload), it'd probably make sense to just upload one file per access level during setup and then clean them up once at the end of all the tests.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * Tests that large file (multi-part) downloads successfully.
+     *
+     * @throws Exception if uploading test-file fails
+     */
+    @Test
+    public void testDownloadLargeFile() throws Exception {
+        latchedUpload(largeFile);
```
TrekSoft: Similar comment here - I think it'd be better to just upload the large file once during setup vs. uploading it here and then also two more times below.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,224 @@
...
+                largeFileName,
+                largeFile.getAbsolutePath(),
+                onResult -> {
+                    assertEquals(largeFileName, onResult.getKey());
+                    assertS3ObjectExists(largeFileKey);
+                    completed.countDown();
+                },
+                onError -> fail("Upload was not successful.")
+        );
+        completed.await(DEFAULT_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
```
TrekSoft: One improvement that could be made here is that rather than have to wait for the timeout if it fails you have a way of immediately returning.

core/src/main/java/com/amplifyframework/storage/options/StorageUploadFileOptions.java
```diff
@@ -127,15 +105,19 @@ public static StorageUploadFileOptions defaultInstance() {
         private String contentType;
         private Map<String, String> metadata;
 
+        private Builder() {
+            this.metadata = new HashMap<>();
+        }
```
TrekSoft: You think it'd be better to have an empty map vs. a null if there is no metadata?

aws-storage-s3/build.gradle
raphkim: oof good catch. Will remove.

lawmicha:
Is there a read me file which outlines how to provision the s3 bucket for testing? These tests should be able to be run by the local developer given some instructions and we should not be pulling down integration test resources to do so

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
raphkim: I agree. I will do the upload portion of the tests before class rather than before each test suite.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
raphkim: yep that's a much better idea

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: Wouldn't the test immediately terminate upon encountering `fail("Upload was not successful.")`?

core/src/main/java/com/amplifyframework/storage/options/StorageUploadFileOptions.java
raphkim: IIRC it was already like this, but I removed it in one of my recent PRs. This is just a re-addition. Besides, `TransferUtility` simply calls overloaded method with a `new ObjectMetadata()` upon calling upload without metadata anyways.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
TrekSoft: No because the completed latch won't count down in that case and the test waits for the latch to either countdown or timeout.

core/src/main/java/com/amplifyframework/storage/options/StorageUploadFileOptions.java
TrekSoft: Ok sounds good

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
TrekSoft: Oooh, sorry I'm an idiot - I missed that fail was a function not a log statement. Never mind.

TrekSoft:
@lawmicha I think that's a larger issue we have to figure out across Android

raphkim:
Storage tests should now be passing, but CI/CD is blocked by other flaky tests. Should be resolved in #263

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,232 @@
...
+     * Upload the required resources in cloud prior to running the tests.
+     * @throws Exception if there is a problem while creating or uploading
+     *         the files.
+     */
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        // Randomly write a file to upload
+        uploadFile = createTempFile(UPLOAD_NAME, UPLOAD_SIZE);
+
+        // Upload to each access level
```
jamesonwilliams: Can you use a control structure -- maybe,

```
for (StorageAccessLevel level : StorageAccessLevel.values()) {
    // do upload ... , level ...
}
```

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/StorageInstrumentationTestBase.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Abstract test base for Storage instrumented tests. This test contains
+ * basic static methods that involve direct interactions with low-level
+ * {@link AmazonS3Client} and {@link AWSMobileClient}.
+ */
+public abstract class StorageInstrumentationTestBase extends AmplifyTestBase {
```
jamesonwilliams: Let's get rid of both the `AmplifyTestBase` and `StorageInstrumentationTestBase`. This is a well-intentioned idea, at first glance -- an attempt to re-use some code. However -- and I've seen this more times than I can count -- here's what tends to happen with this pattern. The base classes grow and grow, and they have all kinds of stuff put into them. Because, all of the things that are being added are to support tests! And that's the name of the base class. But, it becomes very hard to maintain.

So, a better way to organize these things is to create a ton of small utility classes, and then pick and choose them when you're writing your test class(es).

You could split this class out into the following static utilities, perhaps:
```
File tempFile = TempFile.create() // Gets a random name?
File tempPotatoFile = TempFile.create("potato.txt");
File tempPotatoFile30bytes = TempFile.create("potato.txt", 30 /* bytes */);
```

```
S3Assert.exists("bucketName");
```

For the other async APIs, the pattern we have been using is to create a test-only synchronous version, for convenience. You already have a `latchedUploadAndConfirm` method. So, if you rework this into like a :
```
final class SynchronousStorage {
    StorageUploadFileResult uploadFile(StorageOptions options) throws StorageException {
        return Await.result(...);
    }
```

You can have that stuff off in a single-responsibility component.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+        return singleton;
+    }
+
+    private static void configureAmplify(Context context) {
+        // Configure Amplify, and wait for Storage to be ready.
+        Await.result((onResult, onError) -> {
+            // Listen for initialization success messages
+            Amplify.Hub.subscribe(HubChannel.STORAGE, event -> {
+                // When we get one, end the Await call by firing a result
+                if (InitializationStatus.SUCCEEDED.toString().equals(event.getName())) {
```
jamesonwilliams: I know I have this stuff in the DataStore. But, we really shouldn't need it. We should have the source code itself enqueue operations to execute _after_ receiving this event. That way, the caller doesn't need to wait for this event. The caller can just start enqueing operations.

This is a much larger architectural problem, that is not really related to your work. I'm just worried about this pattern spreading into more tests (even though I started it.)

At minimum, we should leave a note at the top of all this like, "TODO: wait for initialization in source code, not in test. Otherwise, a customer would have to use this await block, too...."

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+                Amplify.configure(context);
+            } catch (AmplifyException configurationFailure) {
+                // If the configuration fails before even initialization begins, kill the Await with onError.
+                onError.accept(new RuntimeException("Configuration failed.",
+                        configurationFailure));
+            }
+        });
+    }
+
+    private static void setUpCredentials(Context context) {
```
jamesonwilliams: We have this code repeated enough times in our code base now that we shuold make a `TestMobileClient` utility or something like that (if we don't already have one??) So that we can just do like:
```
SynchronousMobileClient.initialize();
```
And have that work in a blocking way.

testutils/src/main/java/com/amplifyframework/testutils/AmplifyTestBase.java
```diff
@@ -0,0 +1,110 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.InputStream;
+import java.util.Scanner;
+
+/**
+ * Test base for Amplify instrumentation tests. Any configurable information should
+ * be stored inside "testconfiguration.json", and obtained via static
+ * {@link AmplifyTestBase#getPackageConfigure(String)} method.
```
jamesonwilliams: Hm. Is this intended to be _separate_ from the `amplifyconfiguration.json`? Like, you're just using JSON config files to store runtime properties for the tests, but they're not in anyway related to `amplifyconfiguration.json`, right?

That *might* be alright. We should not do this as a base class though. I'm wondering if YAGNI here. Seems like a lot of centralized infrastructure. Also, these `testconfiguration.json` are going to end up having lots of small details about various parts of the test system.

I think it would be better to have tens, hundres, of small `.json` snippet files littered into a plugin project's `src/test/resources/some-bit-of-config-for-a-unit-test.json` (for unit tests) or `src/androidTest/assets/somebit-of-config-for-an-instrumentation-test.json`.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+            largeFile.getName(),
+            largeFile.getAbsolutePath(),
+            options,
+            onResult -> completed.countDown(),
+            onError -> fail("Upload is not successful.")
+        );
+
+        // Listen to Hub events to pause when progress has been made
+        SubscriptionToken pauseToken = Amplify.Hub.subscribe(HubChannel.STORAGE, hubEvent -> {
+            if ("uploadProgress".equals(hubEvent.getName())) {
```
jamesonwilliams: For datastore, I created a single file where I keep all of the event names that are possibly published by datastore. I think we should try to do the same for the different categories.

Also - we need to figure out how we are going to organize *category* events vs. *plugin* events. And how to namespace them!! Anyway, I would at least put this `uploadProgress` string into an enum called like `StorageCategoryEventNames` (or whatever is appropriate to match the existing version of this for DataStore.)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+                Float progress = progressEvent.getData();
+                if (progress != null && progress > 0) {
+                    op.cancel();
+                }
+            }
+        });
+        subscriptions.add(progressToken);
+
+        // Listen to Hub events for cancel
+        SubscriptionToken cancelToken = Amplify.Hub.subscribe(HubChannel.STORAGE, hubEvent -> {
```
jamesonwilliams: You want to be sure to setup all of your subscriptions _before_ doing the Storage operation. Right now, you have a race condition. One pony starts running, executing the code to do the S3 upload. And another pony starts running, forming these Hub subscriptions. It might be that the first pony dispatches the Hub event, before the second pony is ready to receive it. But, if you setup a subscription first, it is guaranteed to get whatever event(s) it is waiting for. (If they are sent ever, that is.)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Instrumentation test for operational work on upload.
+ */
+public final class AWSS3StorageUploadTest extends StorageInstrumentationTestBase {
+
+    // TODO: This is a temporary work-around to resolve a race-condition
+    // TransferUtility crashes if a transfer is paused and instantly resumed.
+    private static final int SLEEP_DURATION_IN_MILLISECONDS = 300;
+
+    private static final StorageAccessLevel DEFAULT_ACCESS_LEVEL = StorageAccessLevel.PUBLIC;
```
jamesonwilliams: I this coded somewhere in the source? Can you do `@VisibleForTesting` on it, and re-use it here in the test code? Even if this becomes a `public` field in source, that's probably not the end of the world. If a user finds out the default is `PUBLIC`: so what.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+        final CountDownLatch completed = new CountDownLatch(1);
+        final CountDownLatch resumed = new CountDownLatch(1);
+
+        // Begin uploading large file
+        StorageUploadFileOperation<?> op = Amplify.Storage.uploadFile(
+            largeFile.getName(),
+            largeFile.getAbsolutePath(),
+            options,
+            onResult -> completed.countDown(),
+            onError -> fail("Upload is not successful.")
```
jamesonwilliams: This probably isn't going to work. This will crash the thread on which the callback happens. But it _won't_ crash the jUnit thread. The `completed.countDown()` approach does work.

But, this whole thing can be replaced by my `Await` utility (this is 100% what that class is for.)

If you put `uploadFile` into a `SynchronousStorage` utility, then have:
```
    StorageUploadFileResult uploadFile(String name, String path, StorageOptions options) throws StorageException {
    return Await.result((onResult, onError) -> {
        Amplify.Storage.uploadFile(..., onResult, onError);
    });
}
```

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+    public void testUploadFileIsCancelable() throws Exception {
+        final CountDownLatch canceled = new CountDownLatch(1);
+
+        // Begin uploading large file
+        StorageUploadFileOperation<?> op = Amplify.Storage.uploadFile(
+            largeFile.getName(),
+            largeFile.getAbsolutePath(),
+            options,
+            onResult -> fail("Upload finished before being successfully cancelled."),
+            onError -> fail("Upload failed for a different reason.")
```
jamesonwilliams: These probably won't work the way you expect. They will crash the thread on which these callbacks occur, not the jUnit thread which is ready to intercept any thrown exceptions and report them. See notes about `Await` below.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -0,0 +1,238 @@
...
+        subscriptions.add(pauseToken);
+
+        // Listen to Hub events to resume when operation has been paused
+        SubscriptionToken resumeToken = Amplify.Hub.subscribe(HubChannel.STORAGE, hubEvent -> {
+            if ("uploadState".equals(hubEvent.getName())) {
+                HubEvent<String> stateEvent = (HubEvent<String>) hubEvent;
+                TransferState state = TransferState.getState(stateEvent.getData());
+                if (TransferState.PAUSED.equals(state)) {
+                    Amplify.Hub.unsubscribe(pauseToken); // So it doesn't pause on each progress report
+                    Sleep.milliseconds(SLEEP_DURATION_IN_MILLISECONDS); // TODO: This is kind of gross
```
jamesonwilliams: Is this so that you can be sure and get a progress event with a time in the future? I agree its kind of gross but it'll be helpful for us to remember why it's here too

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,232 @@
...
+    private void testDownload(
+            File downloadTo,
+            StorageAccessLevel accessLevel,
+            String identityId
+    ) throws Exception {
+        StorageDownloadFileOptions options = StorageDownloadFileOptions.builder()
+                .accessLevel(accessLevel)
+                .targetIdentityId(identityId)
+                .build();
+        Await.<StorageDownloadFileResult, StorageException>result(
```
jamesonwilliams: There a bunch of these similar structures throughout these new tests, just rip them all out and stuff them into a re-usable `SynchronousStorage` test utility. It should probably go into `aws-storage-s3` not `testutils` (cause I'm not sure what other type of category/plugin would ever use it.)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: I intentionally didn't use `Await` because the code needs to be unblocked. If you look closely, you will notice that I am not using the result, but the actual storage operation so that I can pause and resume it.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: I would have liked to also, but I need the reference to operation object from storage upload so that I can call methods on it.

jamesonwilliams:
> @lawmicha I think that's a larger issue we have to figure out across Android

@lawmicha & @TrekSoft  - Completely agree. Right now, instrumentation tests can only be run by a handful of Amazon Web Services employees. We need a solution that works for arbitrary open-source contributors. Have been thinking about this problem in the background a little ... someone  should throw together a one-pager on the topic.

At a high level, we could:
1. Use scripts to arrange and tear down resources , before every test run. A shell script that calls `amplify` would be easy for arrangement. The cleanup might be a little harder.
2. Directly leverage CloudFormation templates that the `amplify` CLI uses, for the same effect.
3. Offer _public_ test endpoints. There's a knee-jerk reaction against this idea, but I invite is to really get clear on _what_ our concerns with it are. www.amazon.com is a public endpoint, after all.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
jamesonwilliams: Ah crap, yea I see what you mean. I guess you could do:

```
final AtomicReference<Cancelable> cancelableContainer = new AtomicReference<>();
Amplify.Hub.subscribe(channel, () -> {
   ....
   Cancelable cancelable = cancelableContainer.get();
   if (cancelable != null) cancelable.cancel();
   ...
}

cancelable.set(Amplify.Storage.doStuff(...));
```

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
jamesonwilliams: Oops, sorry. Sounds good!

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: I described it at the top when declaring the constant variable! maybe i should copy and paste the comment here too

testutils/src/main/java/com/amplifyframework/testutils/AmplifyTestBase.java
raphkim: Ah yea, I initially had it as `res/raw/credentials.json` inside aws-storage-s3 module, but I just got the idea to do this from our SDK repo (which also uses `testconfiguration.json`). I can make the changes to make them local to each module again.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
jamesonwilliams: Ah okay, I just missed it. Once is good enough, I just need to pay more attention. :-D

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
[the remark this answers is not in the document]
raphkim: Maybe so, but it feels like the way I have it now is more readable and easier to understand the purpose of (albeit more verbose for sure)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /**
+     * Signs out by default and sets up the file to download object to.
+     * @throws Exception if an error is encountered while creating file
+     */
+    @Before
+    public void setUp() throws Exception {
+        signOut();
+        downloadFile = createTempFile(destination);
```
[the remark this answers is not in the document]
raphkim: well upload file is instantiated in `@BeforeClass` so that it can only be uploaded one time to reduce heavy workload. Technically download file can be created once also, but this kind of just serves to show that each there is a file being downloaded in each test case (as opposed to one uploaded file in the entire test suite). This is no big deal and I did it without any major reason though, so I can change it if you do not like it this way.

core/src/main/java/com/amplifyframework/storage/options/StorageUploadFileOptions.java
```diff
@@ -127,15 +105,19 @@ public static StorageUploadFileOptions defaultInstance() {
...
         /**
          * Configures the storage access level for the new
          * StorageUploadFileOptions instance.
          * @param accessLevel Storage access level
          * @return Current Builder instance for fluent chaining
          */
         @NonNull
-        public Builder accessLevel(@NonNull StorageAccessLevel accessLevel) {
-            this.accessLevel = Objects.requireNonNull(accessLevel);
+        public Builder accessLevel(@Nullable StorageAccessLevel accessLevel) {
```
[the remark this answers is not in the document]
raphkim: > So, if someone doesn't provide an accessLevel, I think our business rule is to use public, by default, right?

The agreed business rule is to have an overridable default access level in StorageCATEGORY configuration (rather than on plugin configuration level). We do not have this implemented yet, and I planned on getting to it eventually on a separate PR. In such case, I will build it into the `options.getDefaultOptions()` logic and change this back to `@NonNull`.

core/src/main/java/com/amplifyframework/storage/options/StorageUploadFileOptions.java
```diff
@@ -156,8 +138,8 @@ public Builder targetIdentityId(@NonNull String targetIdentityId) {
          * @return Current Builder instance for fluent chaining
          */
         @NonNull
-        public Builder contentType(@NonNull String contentType) {
-            this.contentType = Objects.requireNonNull(contentType);
+        public Builder contentType(@Nullable String contentType) {
+            this.contentType = contentType;
```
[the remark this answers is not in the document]
raphkim: I... am not sure to be honest. In fact, the content type doesn't seem to be used anywhere in the code at the moment.

Will refactor storage service + upload operation to take content-type into account.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: Oh i meant default for testing. It could have been anything, but I didn't want to sign in for testing functionality since access level was already tested in a separate class.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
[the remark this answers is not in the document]
jamesonwilliams: Yea, I agree. This has pros and cons; your Descriptive And Meaningful Phrasing (DAMP) way of doing it without control flow has advantages too.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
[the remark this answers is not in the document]
jamesonwilliams: Either way - this one wasn't isn't a big deal.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
```diff
@@ -160,4 +139,38 @@ public void resume() {
...
+            }
+        }
+
+        @Override
+        public void onProgressChanged(int transferId, long bytesCurrent, long bytesTotal) {
+            final float progress;
+            if (bytesTotal != 0) {
+                progress = (float) bytesCurrent / bytesTotal;
+            } else {
+                progress = 1f;
```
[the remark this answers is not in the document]
raphkim: I had no trouble uploading and downloading an empty file. It shouldn't be treated as an error I believe.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageUploadFileOperation.java
```diff
@@ -166,4 +141,38 @@ public void resume() {
...
+        @Override
+        public void onProgressChanged(int transferId, long bytesCurrent, long bytesTotal) {
+            final float progress;
+            if (bytesTotal != 0) {
+                progress = (float) bytesCurrent / bytesTotal;
+            } else {
+                progress = 1f;
+            }
+            Amplify.Hub.publish(HubChannel.STORAGE,
+                    HubEvent.create("uploadProgress", progress));
```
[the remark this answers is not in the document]
raphkim: This is pretty much here to enable my test, and I agree we need more information. We were emitting zero event from the hub previously, so please treat this as a placeholder until we agree on the actual event format.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
```diff
@@ -160,4 +139,38 @@ public void resume() {
             }
         }
     }
+
+    private final class DownloadTransferListener implements TransferListener {
+        @Override
+        public void onStateChanged(int transferId, TransferState state) {
+            Amplify.Hub.publish(HubChannel.STORAGE,
+                    HubEvent.create("downloadState", state.name()));
+            if (TransferState.COMPLETED == state) {
```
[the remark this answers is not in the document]
raphkim: good idea!

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestUtils.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Test utility for comparing file content.
+ */
+final class TestUtils {
```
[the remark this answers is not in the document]
raphkim: `FileAssert` is really the only thing we need and the rest were just helper methods. I'll just make one class for File assertion and have the rest be private static methods inside it.

lawmicha:
> > @lawmicha I think that's a larger issue we have to figure out across Android
> 
> @lawmicha & @TrekSoft - Completely agree. Right now, instrumentation tests can only be run by a handful of Amazon Web Services employees. We need a solution that works for arbitrary open-source contributors. Have been thinking about this problem in the background a little ... someone should throw together a one-pager on the topic.
> 
> At a high level, we could:
> 
> 1. Use scripts to arrange and tear down resources , before every test run. A shell script that calls `amplify` would be easy for arrangement. The cleanup might be a little harder.
> 2. Directly leverage CloudFormation templates that the `amplify` CLI uses, for the same effect.
> 3. Offer _public_ test endpoints. There's a knee-jerk reaction against this idea, but I invite is to really get clear on _what_ our concerns with it are. [www.amazon.com](http://www.amazon.com) is a public endpoint, after all.

I think offering the Cloudformation templates would be pretty neat. I'm not sure what that looks like though. Currently I just throw in the steps in the readme like https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/README.md

Even as an internal developer, we should be provisioning a backend to be run tests against that is not touching the integration test resources. provisioning our own backend would act as staging, before we finalize the tests and how the backend should actually provisioned and deployed for the integration test run to run against.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
```diff
@@ -64,9 +66,9 @@ private static String getAccessLevelPrefix(
             @NonNull String identityId
     ) {
         if (accessLevel.equals(StorageAccessLevel.PRIVATE) || accessLevel.equals(StorageAccessLevel.PROTECTED)) {
```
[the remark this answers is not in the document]
raphkim: It's not immediately obvious if public > protected or vice versa, so I would rather not use a numerical comparator for this

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestConfiguration.java
raphkim: Ah okay, I'll replace it with simple `Amplify.configure()`. I thought it was necessary to include this.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: fixed

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,239 @@
...
+ * correctly with regards to the provided storage access level.
+ */
+public final class AWSS3StorageDownloadAccessLevelTest extends StorageInstrumentationTestBase {
+
+    private static String userNameOne;
+    private static String userNameTwo;
+    private static Map<String, String> identityIds = new HashMap<>();
+
+    private static final String UPLOAD_NAME = "upload-test-" + System.currentTimeMillis();
+    private static final long UPLOAD_SIZE = 100L;
```
jamesonwilliams: Should order these constants above the members, in the file

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,239 @@
...
+    private static File uploadFile;
+
+    private final String destination = "download-test-" + System.currentTimeMillis();
```
jamesonwilliams: this can be `static` and organized up to be next to the other constants, too.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadAccessLevelTest.java
```diff
@@ -0,0 +1,239 @@
...
+     * Upload the required resources in cloud prior to running the tests.
+     * @throws IOException if there is a problem while creating the files
+     * @throws StorageException if there is a problem with uploading the files
+     * @throws SynchronousAWSMobileClient.MobileClientException from failure to sign in with
+     *         Cognito User Pools or obtain an identity ID
+     */
+    @BeforeClass
+    public static void setUpBeforeClass() throws IOException,
+            StorageException,
+            SynchronousAWSMobileClient.MobileClientException {
```
jamesonwilliams: You might want to call `cleanup()` inside of here, too, since `AfterClass` is not guaranteed to run. (If the process crashes during a test, AfterClass won't ever run, for example.)

jamesonwilliams:
This body of work got pretty large, and this isn't necessarily an "atomic commit" at this point. But in the interest of getting things done we decided to get it in (its green in CI/CD) and iterate on this base.

testutils/src/main/java/com/amplifyframework/testutils/RandomTempFile.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public RandomTempFile(@NonNull String filename, long byteSize) throws IOException {
+        super(TEMP_DIR + File.separator + filename);
+        createFile(byteSize);
+    }
+
+    private void createFile(long byteSize) throws IOException {
+        deleteOnExit();
+        RandomAccessFile raf = new RandomAccessFile(this, "rw");
+        raf.setLength(byteSize);
```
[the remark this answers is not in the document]
raphkim: oh good catch. I would likely have to close it.

testutils/src/main/java/com/amplifyframework/testutils/SynchronousAWSMobileClient.java
```diff
@@ -0,0 +1,217 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raphkim: 2020

testutils/src/main/java/com/amplifyframework/testutils/SynchronousStorage.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raphkim: 2020

raphkim:
This PR is too big with too many QOL/refactoring changes that I decided to split it into multiple PRs. Please see #281 #285 #286

PR #260, opened 2020-02-11
Update copy-configs

raphkim:
*Issue #, if available:*

*Description of changes:*
- Add `testconfiguration.json` so that #259 can pass CI/CD
- The S3 configuration bucket was also re-organized accordingly

### DIFF
--- a/.circleci/copy-configs
+++ b/.circleci/copy-configs
@@ -16,20 +16,48 @@ readonly config_bucket='amplify-circleci-android-payload'
 # So, use two simple arrays and join on index.
 # sources are remote URLs, targets are local file paths
 readonly sources=(
-    "s3://$config_bucket/awsconfiguration-analytics.json"
-    "s3://$config_bucket/amplifyconfiguration-analytics.json"
-    "s3://$config_bucket/amplifyconfiguration-core.json"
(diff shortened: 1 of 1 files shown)


.circleci/copy-configs
```diff
@@ -16,20 +16,48 @@ readonly config_bucket='amplify-circleci-android-payload'
...
+    "s3://$config_bucket/DataStore/awsconfiguration.json"
+
+    # Storage
+    "s3://$config_bucket/Storage/amplifyconfiguration.json"
+    "s3://$config_bucket/Storage/awsconfiguration.json"
 )
 readonly targets=(
-    "aws-analytics-pinpoint/src/androidTest/res/raw/awsconfiguration.json"
-    "aws-analytics-pinpoint/src/androidTest/res/raw/amplifyconfiguration.json"
+    "testutils/src/main/res/raw/testconfiguration.json"
```
jamesonwilliams: the test utilities live “under” the category tests, and you can only read one config at a time since amplify is a process-wide singleton.

So it should be the category tests which initiate a configuration, not this utility code.

The category configuration should contain information only related to what is being tested. This file might need to include information duplicated from a category test, if we were to include it.

PR #262, opened 2020-02-13
[DataStore] Split SQLiteStorageAdapter tests to reduce load

raphkim:
*Issue #, if available:*

*Description of changes:*
Split `SQLiteStorageAdapterInstrumentationTest` into three tests (separated by method being tested) for both abstraction + load reduction.

I noticed that the CI/CD test often fails near the end due to process unexpectedly dying where it shouldn't. It may be because the instrumentation test has very heavy workload PER test-suite (i.e. setting up `SQliteStorageAdapter` instance from scratch and terminating it after) and contains several test suites. This PR is mainly an experiment to test my hypothesis.


raphkim:
Closing in favor of #263

PR #263, opened 2020-02-13
Patch flaky tests for CI/CD

raphkim:
This is an extension of PR #262 that also includes timeout extension for API instrumentation tests

Content of #262 copied here:

> Split `SQLiteStorageAdapterInstrumentationTest` into three tests (separated by method being tested) for both abstraction + load reduction.
>
> I noticed that the CI/CD test often fails near the end due to process unexpectedly dying where it shouldn't. It may be because the instrumentation test has very heavy workload PER test-suite (i.e. setting up `SQliteStorageAdapter` instance from scratch and terminating it after) and contains several test suites. This PR is mainly an experiment to test my hypothesis.

Timeout duration changes made:
 - ~~API connection error response threshold: 1s -> 2s~~ (removed)
 - API synchronous operation timeout: 5s -> 10s
 - Storage Engine timeout: 1s -> 2s


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/StorageAdapterInstrumentedTestBase.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Test the functionality of {@link SQLiteStorageAdapter} operations.
+ */
+@SuppressWarnings("DesignForExtension") // Utility methods shouldn't be overwritten
+public abstract class StorageAdapterInstrumentedTestBase {
```
jamesonwilliams: If they are utility methods, then I recommend to put them into utility classes, not into an abstract base.

Abstract classes are good candidates for enforcing policy on an interface. But, that’s not really what’s needed here. Here, you’re just trying to re-use some utility code in a few places.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesonwilliams: Change to 2020 in all new files

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/StorageAdapterInstrumentedTestBase.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final String DATABASE_NAME = "AmplifyDatastore.db";
+
+    private SQLiteStorageAdapter sqliteStorageAdapter;
+    private Context context;
+
+    /**
+     * Enable strict mode for catching SQLite leaks.
+     */
+    @BeforeClass
+    public static void enableStrictMode() {
```
jamesonwilliams: A final utility class `StrictMode` with a single static `enable` method?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/StorageAdapterInstrumentedTestBase.java
```diff
@@ -0,0 +1,202 @@
...
+     * Drop all tables and database, terminate and delete the database.
+     * @throws DataStoreException from possible underlying DataStore exceptions
+     */
+    @After
+    public void tearDown() throws DataStoreException {
+        sqliteStorageAdapter.terminate();
+        context.deleteDatabase(DATABASE_NAME);
+    }
+
+    <T extends Model> void saveModel(@NonNull T model) throws DataStoreException {
```
jamesonwilliams: These methods should go in a class called `SynchronousStorageAdapter` that lives in the test sources. It would be analogous to the existing `SynchronousApi` and `SynchronousDataStore`, but it would pose a synchronous version of the storage adapter interface.

If you do it that way, then any test class may use this functionality, not just those who (must) extend this base class.

testutils/src/main/java/com/amplifyframework/testutils/SynchronousApi.java
```diff
@@ -45,6 +46,10 @@
  * performing various operations.
  */
 public final class SynchronousApi {
+
+    private static final long EXTENDED_TIMEOUT_IN_MILLISECONDS =
+            TimeUnit.SECONDS.toMillis(10); // 5 seconds is insufficient
```
jamesonwilliams: this seems dubious. If we have >5 second round trips to AppSync, we have bigger problems. I suspect there’s something else going on here aside from the timeout being too short. Did you measure the response time of the actual network call?

raphkim: Actually, the problem existed mainly for the tests that were authenticated via IAM. I believe the most significant delay stemmed from IAM authentication rather than the AppSync itself. I can isolate the extended timeout just for just those tests if you would prefer it that way?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/StorageAdapterInstrumentedTestBase.java
raphkim: The biggest reason for using abstract class (it can just be a non-final class to be honest) here was to allow re-initialization of `SqliteStorageAdapter` per test case without having to copy and paste the same code into each individual test. The "utility methods" make a direct reference to this instance, so I can't move them to a static utility class either.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/StorageAdapterInstrumentedTestBase.java
raphkim: sure thing

raphkim:
@jamesonwilliams Thanks for the review!

I addressed the minor comments, but I will address the rest separately (refining API op timeout and refactoring utility structure for DataStore tests) on a different PR. The main goal of this PR is to unblock CI/CD so that other PRs are not blocked in the pipeline.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
raphkim: will do.

testutils/src/main/java/com/amplifyframework/testutils/SynchronousApi.java
jamesonwilliams: Ahh okay! Hm, you could. Or at least a comment. A self-documenting approach would be to name the timeout like IAM_PLUS_APPSYNC_TIMEOUT or something

PR #265, opened 2020-02-16
[aws-datastore] Synchronous adapter utility for instrumentation tests

jamesonwilliams:
A recent commit introduced a base class to be re-used by three similar
instrumentation tests. Instead, this functionality is reworked into
utility code which may be composed into the various storage
instrumentation tests.


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -35,10 +39,27 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
-/**
+/**f
```
raphkim: F

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SynchronousStorageAdapter.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * Save a model into the storage adapter.
+     * @param model Model to save
+     * @param <T> Type of model being saved
+     * @throws DataStoreException On any failure to save model into storage adapter
+     */
+    <T extends Model> void save(@NonNull T model) throws DataStoreException {
+        //noinspection ConstantConditions
```
raphkim: Something I noticed as weird while I was working on the same thing, but why did we do this? `save(T, QueryPredicate)` takes in two non-null values, but we are passing in `null` here and suppressing the warning with `ConstantConditions`. Same thing in other operations.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SynchronousStorageAdapter.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Save a model.
+     * @param model Model to save
+     * @param predicate An existing instance of the model in the storage adapter must meet these criteria
+     *                  in order for the save to succeed. If null, no criteria are considered
+     * @param <T> Type of model being saved
+     * @throws DataStoreException On any failure to save the model
+     */
+    <T extends Model> void save(@NonNull T model, @NonNull QueryPredicate predicate)
```
raphkim: Let's make predicate `@Nullable` since the overloaded method calls this with `null` predicate.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -84,7 +84,8 @@
     private static final int DATABASE_VERSION = 1;
 
     // Name of the database
-    private static final String DATABASE_NAME = "AmplifyDatastore.db";
+    @VisibleForTesting @SuppressWarnings("checkstyle:all") // Keep logger first
```
raphkim: I was thinking the same thing! 👍

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
raphkim: image

jamesonwilliams:
> Wait I was already half way done with this. Not fair lol

Haha sorry! I should have guessed you'd be on top of this. :-)

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
jamesonwilliams: Hahah :-D

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SynchronousStorageAdapter.java
jamesonwilliams: I think ultimately what we need to do is to dis-allow this `null` `QueryPredicate` everywhere, entirely. Right now, when we use a `null` `QueryPredicate`, it means "we don't want to do any filtering; allow everything." So, we could make that explicit by having a special type of non-null query predicate. Maybe called like `AllowAllQueryPredicate` or `NoOpQueryPredicate`. We could make a little utility class, with a factory method: `QueryPredicates.matchAll()`. The thing that this call returns would be one of these special non-null query predicates, that just matches everything.

"Matching everything" would still need to be coded differently for in-memory, SQL, and GraphQL. Right now we have this code, I think it just checked `null == queryPredicate` instead of like `queryPredicate instanceof AllowAllQueryPreciate`, (or, however we might name this new type of predicate.)

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SynchronousStorageAdapter.java
jamesonwilliams: (See discussion above.)

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
TrekSoft: Loool

jamesonwilliams:
> Typo is minor enough, but do you plan to fix?
> 
> Also, we should definitely discuss regarding `Predicate.MatchAll` at one point. Long story short, I thought its existence may be confusing to some people regarding certain aspects, but I think it makes sense to not expose it and just use it internally for this purpose.

Ooh, whoops. I was distracted by the image, I didn't notice what you were referring to. Yes, I'll fix the typo!

PR #268, opened 2020-02-17
[rxbindings] Add an RxJava2 facade

jamesonwilliams:
RxAmplify exposes all of the functionalities of the Amplify library, but
using RxJava2 primitives: Observable, Single, Completable.


core/src/main/java/com/amplifyframework/hub/AWSHubPlugin.java
```diff
@@ -68,8 +65,7 @@
                         !subscription.getHubEventFilter().filter(hubEvent)) {
                     continue;
                 }
-
-                mainHandler.post(() -> subscription.getHubSubscriber().onEvent(hubEvent));
```
jamesonwilliams: This line is problematic when writing a unit test that exercises the `AWSHubPlugin`. It is also a strange choice, to callback on the main thread, when the subscription did not necessarily happen on the main thread. So, while this line doesn't appear to have to do with the purpose of the PR at first glance -- it's towards making this class test-able.

rxbindings/src/main/java/com/amplifyframework/rx/RxApi.java
```diff
@@ -0,0 +1,25 @@
...
+package com.amplifyframework.rx;
+
+import com.amplifyframework.api.ApiCategoryBehavior;
+
+/**
+ * An Rx-idiomatic expression of Amplify's {@link ApiCategoryBehavior}s.
+ */
+@SuppressWarnings("WeakerAccess") // This is a public API.
+public interface RxApi extends RxRestApi, RxGraphQlApi {
+}
```
raphkim:
```java
public interface RxApiCategoryBehavior extends RxRestBehavior, RxGraphQLBehavior {}
```

Also would be nice to agree on `GraphQL` vs `GraphQl` when naming our java classes. 

On somewhat-related note, there is also similar inconsistency in datastore storage engine (`Sql` vs `SQL`, `SQLite` vs `Sqlite`)

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -0,0 +1,262 @@
...
+ */
+final class RxApiBinding implements RxApi {
+    private final ApiCategoryBehavior api;
+
+    RxApiBinding() {
+        this(Amplify.API);
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    RxApiBinding(ApiCategoryBehavior api) {
```
raphkim: Is this meant to be for plugin or category?

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStore.java
```diff
@@ -0,0 +1,137 @@
...
+import com.amplifyframework.datastore.DataStoreItemChange;
+
+import io.reactivex.Completable;
+import io.reactivex.Observable;
+
+/**
+ * An Rx-idiomatic expression of the behaviors in {@link DataStoreCategoryBehavior}.
+ */
+@SuppressWarnings("unused") // These are all public APIs
+public interface RxDataStore {
```
raphkim: `RxDataStoreCategoryBehavior`

rxbindings/src/main/java/com/amplifyframework/rx/RxHub.java
```diff
@@ -0,0 +1,56 @@
...
+import com.amplifyframework.hub.HubException;
+
+import io.reactivex.Completable;
+import io.reactivex.Observable;
+
+/**
+ * An Rx-idiomatic version of Amplify's {@link HubCategoryBehavior}.
+ */
+@SuppressWarnings("unused") // These are public APIs
+public interface RxHub {
```
raphkim: `RxHubCategoryBehavior`

rxbindings/src/main/java/com/amplifyframework/rx/RxStorage.java
```diff
@@ -0,0 +1,149 @@
...
+import com.amplifyframework.storage.result.StorageRemoveResult;
+import com.amplifyframework.storage.result.StorageUploadFileResult;
+
+import io.reactivex.Single;
+
+/**
+ * An Rx idiomatic expression of the facilities in Amplify's {@link StorageCategoryBehavior}.
+ */
+@SuppressWarnings("unused")
+public interface RxStorage {
```
raphkim:
```java
public interface RxStorageCategoryBehavior {
```

rxbindings/src/main/java/com/amplifyframework/rx/RxStorageBinding.java
```diff
@@ -0,0 +1,114 @@
...
+            @NonNull String key, @NonNull String local, @NonNull StorageUploadFileOptions options) {
+        return toSingle((onResult, onError) -> storage.uploadFile(key, local, options, onResult, onError));
+    }
+
+    @NonNull
+    @Override
+    public Single<StorageRemoveResult> remove(@NonNull String key) {
+        return toSingle((onResult, onError) -> {
+            storage.remove(key, onResult, onError);
+            return new NoOpCancelable(); // StorageRemoveOperation (above) is not Cancelable right now!
```
raphkim: Is it cancelable in iOS?

rxbindings/src/test/java/com/amplifyframework/rx/RxAdaptersTest.java
```diff
@@ -0,0 +1,185 @@
...
+     * as canceled.
+     */
+    @Test
+    public void singleIsCancelable() {
+        SimpleCancelable cancelable = new SimpleCancelable();
+        TestObserver<?> observer = RxAdapters.toSingle(((onResult, onError) -> cancelable))
+            .test();
+        observer.cancel();
+        assertTrue(observer.isCancelled());
+        assertTrue(cancelable.isCanceled());
```
raphkim: Ugh, we should just use `Cancellable` and `Cancelled` everywhere since that's what RX went with when naming [`Cancellable` interface](http://reactivex.io/RxJava/javadoc/io/reactivex/functions/Cancellable.html).

rxbindings/src/main/java/com/amplifyframework/rx/RxAmplify.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @SuppressWarnings("checkstyle:all") private RxAmplify() {}
+
+    // Breaking the rules, here. Don't look.
+    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxApi API = new RxApiBinding();
+    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxDataStore DataStore = new RxDataStoreBinding();
+    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxHub Hub = new RxHubBinding();
+    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxStorage Storage = new RxStorageBinding();
+    // Logging currently has no callback/async behaviors
+    // Analytics currently has no callback/async behaviors
```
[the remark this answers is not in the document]
jamesonwilliams: Yes we will need to - but in a separate PR. In that PR, we will also need to change this binding, to update.

rxbindings/src/main/java/com/amplifyframework/rx/RxApi.java
jamesonwilliams: there is this guidance - 

https://stackoverflow.com/a/15526526/695787

But it’s complicated since the acronyms are embedded in other words. I guess it should be `GraphQL` and `Sqlite` in pascal-case, and `graphql` and `sqlite` in camel-case? Does that sound right?

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
jamesonwilliams: For the category’s APIs. This stuff should work regardless of which plugin is implementing the behaviors.

raphkim: Then I think `ApiCategory` is a better fit than `ApiCategoryBehavior`, since `RxApiBinding` simulates the category, not just a plugin

rxbindings/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+   Copyright 2020 nosemaj.org. All Rights Reserved.
```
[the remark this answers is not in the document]
jamesonwilliams: Yes! Good catch.

rxbindings/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+### Gradle
+Take an `implementation` dependency on this library. In your module's
+`build.gradle`:
+```
+dependencies {
+    # ...
+    implementation 'com.amplifyframework:core:0.9.0'
+    implementation 'com.amplifyframework:rxbindings:0.9.0'
```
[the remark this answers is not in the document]
jamesonwilliams: Updating

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
jamesonwilliams: Hmm, yes you might be right. I think I agree. Let me give it a try anyway and see if there are any negative side effects to that proposal.

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStore.java
jamesonwilliams: Eh, yea, I think you're probably right.

We had talked at one point about dropping the `CategoryBehavior` from even core. They're descriptive, but when you think about it from a customer standpoint, they have like:
```
DataStoreCategoryBehavior ds = Amplify.DataStore;
```
But the customer would probably rather just write like:
```
DataStore ds = Amplify.DataStore
```

rxbindings/src/main/java/com/amplifyframework/rx/RxStorageBinding.java
jamesonwilliams: Looks like it is, yea: https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Storage/AWSS3StoragePlugin/Operation/AWSS3StorageRemoveOperation.swift#L32

rxbindings/src/test/java/com/amplifyframework/rx/RxAdaptersTest.java
jamesonwilliams: Yea, I agree. American English is the only dialect of English that prefers the spelling `Cancelable`. But, I think I researched this one day and determined that it was a non-American who added `Cancellable` into Java. So, I guess they win by virtue of getting their first!

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
jamesonwilliams: It worked, I just had to play some small games in the unit tests to deal with the `final` categories.

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStore.java
jamesonwilliams: I went with the + `CategoryBehavior` versions for now. We can come back and remove them from all places at the same time, maybe.

rxbindings/src/test/java/com/amplifyframework/rx/RxAdaptersTest.java
jamesonwilliams: For another PR...

rxbindings/src/main/java/com/amplifyframework/rx/RxStorage.java
jamesonwilliams: Done

rxbindings/src/main/java/com/amplifyframework/rx/RxHub.java
jamesonwilliams: Done

raphkim:
Also I agree about {Category}CategoryBehavior vs just {Category}. I think it is less awkward with the latter, but we initially decided on the former merely for the sake of having descriptive names. I am okay with renaming it as long as we are in sync with iOS side about it!

jamesonwilliams:
> Looking great! It definitely gives the users a lot of power, but I'm still concerned that this "optional module" may be a source of confusion to those seeking to use Amplify lightly. Do you have any feedback regarding this @palpatim @undefobj ?

At the moment this project is not part of any public documentation, so I'm more concerned that no-one would even find it, to begin with! Let's worry about it when we see our first issue come in.

PR #272, opened 2020-02-19
[aws-datastore] Align names of SyncEngine components to iOS, JavaScript

jamesonwilliams:


aws-datastore/src/main/java/com/amplifyframework/datastore/network/Orchestrator.java
```diff
@@ -0,0 +1,88 @@
...
+     * Start performing sync operations between the local storage adapter
+     * and the remote GraphQL endpoint.
+     * @return A Completable operation to start the sync engine orchestrator
+     */
+    @NonNull
+    public Completable start() {
+        return Completable.fromAction(() -> {
+            storageObserver.startObservingStorageChanges();
+            subscriptionProcessor.startSubscriptions();
+            //syncProcessor.hydrate().blockingAwait(); This crashes right now ...
```
undefobj: I'm guessing because this is the routine that we want to coordinate all of these actions in order on startup

aws-datastore/src/main/java/com/amplifyframework/datastore/network/MutationProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * The {@link MutationProcessor} observes the {@link MutationOutbox}, and publishes its items to an
+ * {@link AppSyncEndpoint}.
+ *
+ * The responses to these mutations are themselves forwarded to the Merger (TODO: write a merger.)
+ */
+final class MutationProcessor {
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
+
+    private final AppSyncEndpoint appSyncEndpoint;
```
undefobj: Minor nit - maybe this should be `GraphQLEndpoint` since we want this to be agnostic to AppSync

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SubscriptionProcessor.java
```diff
@@ -0,0 +1,126 @@
...
+                    break;
+                default:
+                    throw new DataStoreException(
+                        "Unknown mutation type = " + mutation.type(),
+                        AmplifyException.TODO_RECOVERY_SUGGESTION
+                    );
+            }
+
+            // Notify Hub that we just updated the local storage.
+            HubEvent<? extends Model> receivedFromCloudEvent =
```
undefobj: Will this get really noisy? I had suggested notifications from subscriptions to Hub in the past and the pushback was that the messages might get really verbose if you get a lot of subscription updates.

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SubscriptionProcessor.java
```diff
@@ -0,0 +1,126 @@
...
+                () -> LOG.warn("Reading from subscriptions buffer is completed.")
+            )
+        );
+    }
+
+    /**
+     * Stop any active subscriptions, and stop draining the mutation buffer.
+     */
+    void stopAllSubscriptionActivity() {
+        disposable.clear();
```
undefobj: should we also clear `buffer` here?

aws-datastore/src/main/java/com/amplifyframework/datastore/network/MutationProcessor.java
jamesonwilliams: In #271 I make an attempt to start sorting out the "this is appsync" from the "this is generic" -- However, I think I kept this as named (just) `AppSync`, since it has functionalities like Base/Delta sync and deliberately named mutations like Create Update Delete encoded into the interface.

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SubscriptionProcessor.java
jamesonwilliams: Eh, yea. Good catch, and good call out. One of these events for every subscription event is probably going to get unwieldly ...

Maybe we can add an off/on switch in the config - default is on, and then if the customer is like "whoa, are you kidding me?" our documentation does mention a way to shut it down?

aws-datastore/src/main/java/com/amplifyframework/datastore/network/SubscriptionProcessor.java
jamesonwilliams: The thing doing the buffering is the `ReplaySubject`; this `disposable` represents a subscription to that subject. In Rx1, they called the `Disposable` as `Subscription`, which I liked better. But anyway, TLDR, yea I'd be open to a better name for this ... maybe `subscriptionToBuffer`, instead of the (what does it mean??) `disposable`?

PR #278, opened 2020-02-25
Proguard fix for empty classes.jar in mavenLocal()

jamesonwilliams:
With minifyEnabled true, and inadequate configuration of the
proguard-rules.pro, we experienced empty classes.jar inside of the built
.aar for each module. A resolution to this issue was to use Proguard's
example library configuration files, for each library.

The proguard-rules.pro for each module are copied from the Android SDK's
ProGuard distribution, at ${sdk_base}/tools/proguard/examples/library.pro.

Refer: https://stackoverflow.com/q/47652924/695787
Refer: https://stackoverflow.com/a/27052696/695787

### DIFF
--- a/aws-analytics-pinpoint/proguard-rules.pro
+++ b/aws-analytics-pinpoint/proguard-rules.pro
@@ -1,21 +1,79 @@
-# Add project specific ProGuard rules here.
-# You can control the set of applied configuration files using the
-# proguardFiles setting in build.gradle.
 #
-# For more details, see
-#   http://developer.android.com/guide/developing/tools/proguard.html
+# This ProGuard configuration file illustrates how to process a program
+# library, such that it remains usable as a library.
+# Usage:
+#     java -jar proguard.jar @library.pro
+#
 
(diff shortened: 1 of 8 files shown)


raphkim:
Setting minify enable to true actually was a source of several issues in TransferUtility due to how it interacts with gson and enums. 

Does this account for that too by any chance? See https://medium.com/@hanru.yeh/gson-will-crash-at-enum-fields-in-custom-class-with-proguard-bbcf5ad1b623

jamesonwilliams:
> Setting minify enable to true actually was a source of several issues in TransferUtility due to how it interacts with gson and enums.
> 
> Does this account for that too by any chance? See https://medium.com/@hanru.yeh/gson-will-crash-at-enum-fields-in-custom-class-with-proguard-bbcf5ad1b623

Actually, no. I think we should disable minification for now and come back to it. There is some more nuance here than this PR solves.

jamesonwilliams:
I brought this back to life in https://github.com/aws-amplify/amplify-android/pull/283

PR #279, opened 2020-02-25
Workaround for Proguard causing empty classes.jar in mavenLocal()

jamesonwilliams:
With minifyEnabled true, and inadequate configuration of the
proguard-rules.pro, we experienced empty classes.jar inside of the built
.aar for each module.

As a work-around, disable minification. It can be re-enabled in the
future. The references below were identified while researching the
problems.

Refer: https://stackoverflow.com/q/47652924/695787
Refer: https://stackoverflow.com/a/27052696/695787

*Issue #, if available:*

*Description of changes:*


jamesonwilliams:
#283 is a fix for this issue, so this #279 workaround is no longer needed

PR #281, opened 2020-02-26
[testutils] Add synchronous storage and mobile client to testutils

raphkim:
*Issue #, if available:*

*Description of changes:*
- Re-organize module into packages to prevent overcrowding it
  - Added `random` package for generating random objects
  - Added `sync` package for synchronously executing Amplify commands
- Added `SynchronousMobileClient` to assist in initialization and sign-in (authored by @jamesonwilliams )
- Added `SynchronousStorage` to execute storage operations synchronously


testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousMobileClient.java
```diff
@@ -0,0 +1,223 @@
...
+import com.amazonaws.mobile.config.AWSConfiguration;
+
+import java.util.Objects;
+
+/**
+ * A wrapper around the {@link AWSMobileClient} which turns its asynchronous methods
+ * into synchronous ones.
+ */
+@SuppressWarnings("unused")
+public final class SynchronousMobileClient {
```
jamesonwilliams: There is an existing class called `TestMobileClient` (I think) and this is supposed to be a rename and iteration on top of that existing file

raphkim: https://github.com/aws-amplify/amplify-android/blob/master/aws-api/src/androidTest/java/com/amplifyframework/api/aws/TestAWSMobileClient.java

Is this what you are referring to? I believe `SynchronousMobileClient` class is more extensive and it should replace the old `TestAWSMobileClient` in other categories

raphkim: Oops I misunderstood your intent. Removing that file and replacing its usage with the new `SynchronousMobileClient` now

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousDataStore.java
```diff
@@ -13,14 +13,15 @@
...
-package com.amplifyframework.testutils;
+package com.amplifyframework.testutils.sync;
 
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.datastore.DataStoreException;
 import com.amplifyframework.datastore.DataStoreItemChange;
+import com.amplifyframework.testutils.Await;
```
jamesonwilliams: If you have a sync package, this should probably go there too, since that’s what this util is all about

raphkim: I considered `Await` as almost a primitive tool used by the the items inside sync package, so I didn't want to group it in there. I feel like things like `Await`, `Latch`, and `Sleep` all basic building blocks of test utilities, so they should probably be just stay where they are. Wasn't sure what to do with the rest.

PR #280, opened 2020-02-25
Make HubAccumulator more reliable

raphkim:
*Issue #, if available:*

*Description of changes:*
- `HubAccumulator` now waits proportionally as long as the number of events it is waiting on.
- Potentially resolve a race-condition in `HubAccumulator`
- Use consistent 2 seconds timeout duration throughout datastore instrumented tests

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #282, opened 2020-02-26
Use try-with-resources statement to assure that cursor is closed

raphkim:
Potentially fix SQLite leak by making sure that cursor closes with try-with-resources statement

Merged PR #280:
> - ~~`HubAccumulator` now waits proportionally as long as the number of events it is waiting on.~~
> - ~~Potentially resolve a race-condition in `HubAccumulator`~~
> - Use consistent 2 seconds timeout duration throughout datastore instrumented tests

### DIFF
--- a/testutils/src/main/java/com/amplifyframework/testutils/Latch.java
+++ b/testutils/src/main/java/com/amplifyframework/testutils/Latch.java
@@ -27,7 +27,7 @@ import java.util.concurrent.TimeUnit;
  */
 @SuppressWarnings({"SameParameterValue", "WeakerAccess"})
 final class Latch {
-    private static final long REASONABLE_WAIT_TIME_MS = TimeUnit.SECONDS.toMillis(5);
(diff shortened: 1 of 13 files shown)


testutils/src/main/java/com/amplifyframework/testutils/Latch.java
```diff
@@ -27,7 +27,7 @@
  */
 @SuppressWarnings({"SameParameterValue", "WeakerAccess"})
 final class Latch {
-    private static final long REASONABLE_WAIT_TIME_MS = TimeUnit.SECONDS.toMillis(5);
+    static final long REASONABLE_WAIT_TIME_MS = TimeUnit.SECONDS.toMillis(5);
```
jamesonwilliams: can this remain `private`? Who else needs to know about this value?

raphkim: This was for my hub accumulator change, but I'll revert this change along-side the hub accumulator

PR #285, opened 2020-02-26
Refactor storage options to abstract out common properties

raphkim:
*Issue #, if available:*

*Description of changes:*
Portion of #259 that deals with `StorageOptions`.
- Abstracted out `StorageOptions` to reduce duplicate code
- ~~`StorageOptions` is now sufficient for building S3 service key, so added utility method in `S3RequestUtils` class~~

### DIFF
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
@@ -18,6 +18,8 @@ package com.amplifyframework.storage.s3.utils;
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.storage.StorageAccessLevel;
+import com.amplifyframework.storage.StorageException;
+import com.amplifyframework.storage.options.StorageOptions;
 
 import java.util.Locale;
 
@@ -26,8 +28,9 @@ import java.util.Locale;
  */
 public final class S3RequestUtils {
 
-    @SuppressWarnings("WhitespaceAround") // Looks better this way
-    private S3RequestUtils() {}
+    private static final char BUCKET_SEPARATOR = '/';
+
+    @SuppressWarnings("checkstyle:all") private S3RequestUtils() {}
 
     /**
      * Amplify Storage implementation with S3 integrates access level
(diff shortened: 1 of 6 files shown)


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
```diff
@@ -64,9 +83,30 @@ private static String getAccessLevelPrefix(
...
-            return accessLevel.name().toLowerCase(Locale.US) + "/";
+            return accessLevel.name().toLowerCase(Locale.US);
         }
     }
+
+    @NonNull
+    private static String getAccessLevelPrefix(@NonNull StorageOptions options) {
+        StorageAccessLevel accessLevel = options.getAccessLevel();
+        if (accessLevel == null) {
+            throw new RuntimeException(new StorageException(
```
jamesonwilliams: Will the run time exception get caught for sure? Or should you just throw a storage exception directly, to be sure/safe? Same for all in this file

core/src/main/java/com/amplifyframework/storage/options/StorageOptions.java
```diff
@@ -0,0 +1,108 @@
...
+
+import com.amplifyframework.core.async.Options;
+import com.amplifyframework.storage.StorageAccessLevel;
+
+/**
+ * Storage options interface requires that every
+ * storage operation will inspect the provided options
+ * instance for access level and target ID.
+ */
+public abstract class StorageOptions implements Options {
```
jamesonwilliams: I think this could be non-public, since only the sub classes get used outside of this package. Right?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/utils/S3RequestUtils.java
raphkim: These utility methods aren't yet used anywhere to be honest. I just remembered that I went through the trouble of abstracting out options to make it easier to test, but I realized that I didn't actually use it anywhere. I think this will be used later on in one of the refactors once we are done with the testing shenanigans, so we'll see exactly how this utility gets used.

core/src/main/java/com/amplifyframework/storage/options/StorageOptions.java
raphkim: ah shoot you are right. I'm pretty sure I fixed it before, but not sure why its public again now.

raphkim: OH it's because I added the bit for S3RequestUtils... I'll just get rid of that since I don't plan on using it just yet.

PR #287, opened 2020-02-28
[aws-datastore] Wait for initialization before attempting operations

jamesonwilliams:
The Category `initialize(...)` is called on a background thread, and
the Amplify framework does not wait around for it to complete. However,
a user may attempt to use a category before `initialize(...)` has completed.

As a work-around to this problem, for DataStore only, listen for an
event that signifies initialization is complete. Only after this, begin
executing DataStore operations.

This approach was verified from a sample app, using the code in the below-referenced Gist.

Refer: https://gist.github.com/jamesonwilliams/032bfaa23020a04b4b5561bc85b33f00#file-simple-datastore-consumer-md


undefobj:
Ok we are getting closer. I built this branch and published it to maven local, however when I call `Amplify.configure(getApplicationContext()` I'm getting an error of `Unknown/bad category type: dataStore`. It's in the `configure` call because when I comment out that line the try/catch doesn't throw an exception.

```
public class MainActivity extends AppCompatActivity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        AWSDataStorePlugin dataStorePlugin = AWSDataStorePlugin.forModels(AmplifyModelProvider.getInstance());
        try {
            Amplify.addPlugin(dataStorePlugin);
//            Amplify.addPlugin(new AWSApiPlugin());
            Log.i("DATASTORE", "Initialized");
        } catch (AmplifyException e){
            Log.e("INITIALIZATION", e.getLocalizedMessage());
        }

        Button button = (Button) findViewById(R.id.button2);
        button.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View view) {
                save();
            }
        });

    }

    private void save(){
        Post post = Post.builder()
                .title("My test")
                .rating(10)
                .status(PostStatus.ACTIVE)
                .build();

        Amplify.DataStore.save(post,
                save -> Log.i("Datastore", "Saved" + save),
                failure -> Log.i("DataStore", "Failed" + failure)
        );
    }
}
```

And here is by config json:

```
{
  "UserAgent": "aws-amplify-cli/2.0",
  "Version": "1.0",
  "api": {
    "plugins": {
      "awsAPIPlugin": {
        "friendly_name_API_KEY": {
          "endpointType": "GraphQL",
          "endpoint": "https://xyz.appsync-api.us-west-2.amazonaws.com/graphql",
          "region": "us-west-2",
          "authorizationType": "API_KEY",
          "apiKey": "da2-abcdefghijklmnopqr"
        }
      }
    }
  },
  "datastore": {}
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -56,13 +60,17 @@
     // local storage adapter, and a remote API
     private final Orchestrator orchestrator;
 
+    // Keeps track of whether of not the category is initialized yet
+    private final CountDownLatch categoryInitializationsPending;
```
undefobj: Minor nit but maybe since this is effectively the latch variable we're using for concurrency management in category initialization we could have explicit naming of `categoryInitializationLatch`?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -314,20 +329,24 @@ public Cancelable observe(
...
             @NonNull QueryPredicate selectionCriteria,
+            @NonNull Consumer<Cancelable> onObservationStarted,
             @NonNull Consumer<DataStoreItemChange<T>> onDataStoreItemChange,
             @NonNull Consumer<DataStoreException> onObservationFailure,
             @NonNull Action onObservationCompleted) {
         onObservationFailure.accept(new DataStoreException("Not implemented yet, buster!", "Check back later!"));
-        return () -> {};
+    }
+
+    private void afterInitialization(@NonNull final Runnable runnable) {
```
undefobj: I like this refactor out of the individual overrides, basically lifecycle checks. I wonder do we need a separate `beforeInitialization` to ensure everything is in a good state? Could that be used for the race condition issue we saw the other day on app startup instead of requiring a singleton?

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -101,25 +98,47 @@ public final boolean isConfigured() {
...
-            try {
-                plugin.initialize(context);
-                result = InitializationResult.success();
-            } catch (AmplifyException pluginInitializationFailure) {
-                result = InitializationResult.failure(pluginInitializationFailure);
+    @NonNull
+    @WorkerThread
+    public final synchronized CategoryInitializationResult initialize(@NonNull Context context) {
+        final Map<String, InitializationResult> pluginInitializationResults = new HashMap<>();
+        if (!State.CONFIGURED.equals(state.get())) {
```
undefobj: I really like the new `State` enum here but I will admit that readability of the if-not-equals-get is a bit much and someone could introduce a bug in the future if they're not careful. Not sure I have an answer just calling it out.

jamesonwilliams:
> Ok we are getting closer. I built this branch and published it to maven local, however when I call `Amplify.configure(getApplicationContext()` I'm getting an error of `Unknown/bad category type: dataStore`. It's in the `configure` call because when I comment out that line the try/catch doesn't throw an exception.
> 
> ```
> public class MainActivity extends AppCompatActivity {
> 
>     @Override
>     protected void onCreate(Bundle savedInstanceState) {
>         super.onCreate(savedInstanceState);
>         setContentView(R.layout.activity_main);
> 
>         AWSDataStorePlugin dataStorePlugin = AWSDataStorePlugin.forModels(AmplifyModelProvider.getInstance());
>         try {
>             Amplify.addPlugin(dataStorePlugin);
> //            Amplify.addPlugin(new AWSApiPlugin());
>             Amplify.configure(getApplicationContext());
>             Log.i("DATASTORE", "Initialized");
>         } catch (AmplifyException e){
>             Log.e("INITIALIZATION", e.getLocalizedMessage());
>         }
> 
>         Button button = (Button) findViewById(R.id.button2);
>         button.setOnClickListener(new View.OnClickListener() {
>             @Override
>             public void onClick(View view) {
>                 save();
>             }
>         });
> 
>     }
> 
>     private void save(){
>         Post post = Post.builder()
>                 .title("My test")
>                 .rating(10)
>                 .status(PostStatus.ACTIVE)
>                 .build();
> 
>         Amplify.DataStore.save(post,
>                 save -> Log.i("Datastore", "Saved" + save),
>                 failure -> Log.i("DataStore", "Failed" + failure)
>         );
>     }
> }
> ```
> 
> And here is by config json:
> 
> ```
> {
>   "UserAgent": "aws-amplify-cli/2.0",
>   "Version": "1.0",
>   "api": {
>     "plugins": {
>       "awsAPIPlugin": {
>         "friendly_name_API_KEY": {
>           "endpointType": "GraphQL",
>           "endpoint": "https://xyz.appsync-api.us-west-2.amazonaws.com/graphql",
>           "region": "us-west-2",
>           "authorizationType": "API_KEY",
>           "apiKey": "da2-abcdefghijklmnopqr"
>         }
>       }
>     }
>   },
>   "datastore": {}
> }
> ```

Looks like the `amplifyconfiguration.json` has `datastore` with a lower-case `s` instead of an upper-case `S`.

core/src/main/java/com/amplifyframework/core/category/Category.java
jamesonwilliams: The enum is a decent way to handle the state machine, however, I would have liked to avoid the state machine. Hopefully this can be revisited ...

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: Hm, yea we could. Right now, `afterInitialization` is just only a method in this implementation class, not anything fancier. And there logically _is_ some `beforeInitialization`, stuff (basically, setup a listener). Right now, the `configure(...)` is being used for both configuration and for this "before" initialization step.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: The naming here is awkward, agreed. The logic here was that it's a count that starts at 1 and then goes to 0, so it's like, how can I name this _count_ in a reasonable way? _Quantity of things that still need some operation..._, etc.

PR #288, opened 2020-02-28
Use "master" as the default version name

jamesonwilliams:
During development, it is useful to publish development versions of the libraries to a local maven repository, via `publishToMavenLocal`. In the `master` branch, the library version is currently `0.9.0`.

As a result, it can be confusing to know what version of the library is getting consumed, from a consuming application that uses both `mavenLocal()` and `jcenter()`. Both could provide a `0.9.0`.

To avoid confusion, the version name in the master branch is changed to `master`. Development consumers may access the development version of the library through the local maven repository using:

    implementation 'com.amplifyframework:core:master'

etc.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,7 +13,7 @@ org.gradle.jvmargs=-Xmx3g
 org.gradle.parallel=true
 
 GROUP=com.amplifyframework
-VERSION_NAME=0.9.0
+VERSION_NAME=master
 VERSION_CODE=1
 
 POM_URL=https://github.com/aws-amplify/amplify-android


raphkim:
Appears to have failed build-test in CI/CD.

Also how does this affect publishing it in maven?

jamesonwilliams:
> Appears to have failed build-test in CI/CD.
> 
> Also how does this affect publishing it in maven?

Passed when I re-ran the tests.

Our workflow for non-local maven publishing will be the same. We would have had to update 0.9.0 to a new release version, before. Same here, we'll have to change it from master, before we cut a release tag/branch. We have some process to define around that, but definitely, we would have always needed to update this string, regardless.

PR #291, opened 2020-03-03
[Storage] Make list return Amplify, not service, key

raphkim:
*Issue #, if available:*

*Description of changes:*
`Storage.list()` currently returns S3 service key (which is not supposed to be exposed) rather than the Amplify key (e.g. "public/key" rather than "key"). It should return the key directly so that the user can use that key directly for other Amplify.Storage operations.

Also added access level test for list operation.

### DIFF
--- a/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageListAccessLevelTest.java
+++ b/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageListAccessLevelTest.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


TrekSoft:
This is intentional behavior. Without returning the full path, it's ambiguous to the customer what files are being returned or how to access them.

jamesonwilliams:
> This is intentional behavior. Without returning the full path, it's ambiguous to the customer what files are being returned or how to access them.

Whichever behavior you guys decide on -- we can probably still carry these tests forward, with updates, if needed.

jpignata:
Out of curiosity, what do the iOS and JavaScript libraries do? A quick skim suggested to me this prefix is stripped: [iOS](https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Storage/AWSS3StoragePlugin/Service/Storage/AWSS3StorageService%2BListBehavior.swift#L69), [JavaScript](https://github.com/aws-amplify/amplify-js/blob/master/packages/storage/src/Providers/AWSS3Provider.ts#L411)

TrekSoft:
Yeah sorry we talked about this offline and agreed this behavior is fine and matches iOS.

PR #302, opened 2020-03-11
[Predictions] Add Predictions category in core

raphkim:
*Issue #, if available:*

*Description of changes:*
Add Predictions category and relevant models in core module.


core/src/main/java/com/amplifyframework/predictions/models/TextFormatType.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Specifies the type of text detection to perform.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum TextFormatType implements IdentifyAction {
+    FORM,
+    TABLE,
+    PLAIN,
+    ALL;
```
jamesonwilliams: We should get JavaDocs on these, indicating why a user would choose to select one of them. What are the implications of choosing `FORM` vs. `TABLE`, vs. `PLAIN`? What does `ALL` mean - it does all of the first three? What are the result outputs if I select each one of these - do they differ?

core/src/main/java/com/amplifyframework/predictions/models/Table.java
```diff
@@ -0,0 +1,133 @@
...
+    public Integer getRows() {
+        return rows;
+    }
+
+    /**
+     * Gets the column count.
+     * @return the column count
+     */
+    @NonNull
+    public Integer getColumns() {
```
jamesonwilliams: A simple `int`, in all places?

core/src/main/java/com/amplifyframework/predictions/models/Table.java
```diff
@@ -0,0 +1,133 @@
...
+
+package com.amplifyframework.predictions.models;
+
+import androidx.annotation.NonNull;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A representation of table.
```
jamesonwilliams: > A table is a two-dimensional grid of cells that is `rows`-high, and `columns`-wide. This data type is used <by what> as part of the <which?> operation for the purpose of <TODO: why?>".

core/src/main/java/com/amplifyframework/predictions/models/SyntaxToken.java
```diff
@@ -0,0 +1,184 @@
...
+
+import androidx.annotation.NonNull;
+
+import java.util.Objects;
+
+/**
+ * Class that holds the syntax detection results for a
+ * string of text for the predictions category.
+ */
+public final class SyntaxToken {
```
jamesonwilliams: Javadoc makes it look like it should be named as one of the `Predictions*Result`  classes. But this is just a single token, in a collection of tokens, inside of a result, right?

core/src/main/java/com/amplifyframework/predictions/models/SpeechType.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Different types of parts of speech.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum SpeechType {
+    ADJECTIVE,
```
jamesonwilliams: Can we get javadocs on all of these? Some of them, like `ADVERB`, are already clear. However, there are some that will be less familiar to the average user, and some that are just ambigious. For example, I don't know what a `DETERMINER` or a `COORDINATING_CONJUNCTION` are, in the natural language domain. But, a simple text above them with an example of one of the things would go a long way.

For example,,
```
/**
 * Modifies or describes a noun.
 * For example, in the phrase "A fast car", "fast" is an adjective which modifies the noun "car."
 */
ADJECTIVE,
```

core/src/main/java/com/amplifyframework/predictions/models/Sentiment.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Gets a map of associated sentiments and their confidence
+     * scores.
+     * @return the map of associated sentiments and scores
+     */
+    @Nullable
+    public Map<SentimentType, Float> getSentimentScores() {
+        return sentimentScores;
```
jamesonwilliams: Should arrange for this to be an `Immutable.of(...)`, probably by doing making it immutable in the `Builder`'s  `build()`.

core/src/main/java/com/amplifyframework/predictions/models/Sentiment.java
```diff
@@ -0,0 +1,112 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Class that holds the sentiment detection results for a
+ * string of text for the predictions category.
```
jamesonwilliams: > The sentiment of a piece of text, as identified by the \<TODO: which behavior?\> behavior.

core/src/main/java/com/amplifyframework/predictions/models/Selection.java
```diff
@@ -0,0 +1,132 @@
...
+
+package com.amplifyframework.predictions.models;
+
+import android.graphics.Rect;
+import androidx.annotation.NonNull;
+
+import java.util.Objects;
+
+/**
+ * Class to represent a selection from an image.
```
jamesonwilliams: > A selected portion of an image. The image is described as a bounding box, and the selection inside of it is a described by a polygon. The selection can be marked active or inactive, while retaining the selection definition, by toggling isSelect.

Something like that?

core/src/main/java/com/amplifyframework/predictions/models/Pose.java
```diff
@@ -0,0 +1,88 @@
...
+
+package com.amplifyframework.predictions.models;
+
+import androidx.annotation.NonNull;
+
+import java.util.Objects;
+
+/**
+ * Class to represent a pose by using three main
+ * positions: pitch, roll, and yaw.
```
jamesonwilliams: Maybe include a link to like https://en.wikipedia.org/wiki/Aircraft_principal_axes#Principal_axes which describes the domain entity here in more detail?

core/src/main/java/com/amplifyframework/predictions/models/Polygon.java
```diff
@@ -0,0 +1,64 @@
...
+     * @return A polygon instance
+     */
+    @NonNull
+    public static Polygon fromPoints(@NonNull List<Point> points) {
+        Objects.requireNonNull(points);
+        if (points.size() < MINIMUM_POINTS_REQUIRED) {
+            throw new IllegalArgumentException("A polygon must " +
+                    "contain at least three points.");
+        }
+        return new Polygon(points);
```
jamesonwilliams: I don't think we need to validate this, in code, here. But, as a math thought for the day - is do the following coordinates in 3-space define a "polygon"? [0,1,0], [0,3,0], [0,2,0]?

Apparently!

https://gamedev.stackexchange.com/questions/34225/can-a-straight-line-be-called-a-polygon

core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
```diff
@@ -0,0 +1,647 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Enum of different types of language.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum LanguageType {
```
jamesonwilliams: Can we just use `java.util.Locale`?

Do we actually support all of the things in this enum?

If we actually must use this enum, can we use the `SCREAMING_SNAKE("abbrev")` format?

If we must use this list, can we sort it?

If we must use this list, can we provide a `fromLocale(...)` helper, which finds the correct Amazon-defined language-type, from something more familiar to the user?

core/src/main/java/com/amplifyframework/predictions/models/LandmarkType.java
```diff
@@ -0,0 +1,36 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Types of detectable landmark.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum LandmarkType {
```
jamesonwilliams: The list is specific to facial features, it isn't generic like the name of the enum. How about we just call this `FacialFeature`. `FacialFeature.LEFT_EYE`, etc.

core/src/main/java/com/amplifyframework/predictions/models/Landmark.java
```diff
@@ -0,0 +1,110 @@
...
+import androidx.annotation.NonNull;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Class that holds the landmark detection results
+ * for the predictions category.
+ */
+public final class Landmark {
```
jamesonwilliams: The landmarks actually only have to do with faces. Can we call this `FacialFeatures`, or something? And have it contain `FacialFeature type;` and `List<Point> facialFeatureLocations`? 

Also, how do the points correspond to the type? Are the `List<Point>` actually a `Polygon`, describing the region of the picture that contains the facial feature?

core/src/main/java/com/amplifyframework/predictions/models/LabelType.java
```diff
@@ -0,0 +1,34 @@
...
+import androidx.annotation.NonNull;
+
+/**
+ * Specifies the type of label detection to perform.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum LabelType implements IdentifyAction {
+    ALL,
+    MODERATION,
+    LABELS;
```
jamesonwilliams: Can you add JavaDoc describing what these do?

core/src/main/java/com/amplifyframework/predictions/models/LabelMetadata.java
```diff
@@ -0,0 +1,108 @@
...
+        return score;
+    }
+
+    /**
+     * Gets the list of parents' names.
+     * @return the list of parents
+     */
+    @Nullable
+    public List<String> getParents() {
+        return parents;
```
jamesonwilliams: return empty list instead of  null list, empy list would mean "there are no parents"

core/src/main/java/com/amplifyframework/predictions/models/Label.java
```diff
@@ -0,0 +1,135 @@
...
+     */
+    @NonNull
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Builder for {@link Label}.
+     */
+    public static class Builder {
```
jamesonwilliams: add `final` - `public static final class Builder`

core/src/main/java/com/amplifyframework/predictions/models/IdentifyActionType.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    DETECT_CELEBRITY,
+
+    /**
+     * Identify a specific label from an image.
+     */
+    DETECT_LABELS,
+
+    /**w
+     * Identify an entity from an image.
```
jamesonwilliams: What's an entity?

core/src/main/java/com/amplifyframework/predictions/models/IdentifyActionType.java
```diff
@@ -0,0 +1,50 @@
...
+ * Predictions category.
+ */
+public enum IdentifyActionType implements IdentifyAction {
+    /**
+     * Identify a famous celebrity from an image.
+     */
+    DETECT_CELEBRITY,
+
+    /**
+     * Identify a specific label from an image.
```
jamesonwilliams: A label like, some text? Or like, a tag, describing what's happening in the picture? If the first one, how is it different from detect text?

core/src/main/java/com/amplifyframework/predictions/models/IdentifiedText.java
```diff
@@ -0,0 +1,164 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Objects;
+
+/**
+ * Abstract class representing the text that is identified
+ * from inside an image.
+ */
+abstract class IdentifiedText {
```
jamesonwilliams: Any getters in this class, and in the builder, should be marked `final`, since you can't make the `abstract` class itself as `final`.

core/src/main/java/com/amplifyframework/predictions/models/GenderType.java
```diff
@@ -0,0 +1,36 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Types of genders that can be detected.
+ */
+public enum GenderType {
```
jamesonwilliams: To best recognize and advocate for our trans and non-binary developer friends, maybe: `GenderArchetype`?

core/src/main/java/com/amplifyframework/predictions/models/GenderAttribute.java
```diff
@@ -0,0 +1,107 @@
...
+ * Class that holds the gender detection results
+ * for the predictions category.
+ */
+public final class GenderAttribute {
+    private final GenderType gender;
+    private final Float score;
+
+    private GenderAttribute(
+            @NonNull GenderType gender,
+            @NonNull Float score
```
jamesonwilliams: Should this field be called `confidence`? What is the permissible range? Can you add to the docs, whatever it is? Is this in the range [0,100]?

core/src/main/java/com/amplifyframework/predictions/models/EntityMetadata.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    @NonNull
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Builder for {@link EntityMetadata}.
+     */
+    public static class Builder {
```
jamesonwilliams: `public static final class` -> adds `final`

core/src/main/java/com/amplifyframework/predictions/models/EmotionType.java
```diff
@@ -0,0 +1,32 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Types of emotions that can be detected.
+ */
+@SuppressWarnings("JavadocVariable")
+public enum EmotionType {
```
jamesonwilliams: How do these relate to the sentiments? Can we add a text into the doc?

core/src/main/java/com/amplifyframework/predictions/models/GenderAttribute.java
```diff
@@ -0,0 +1,107 @@
...
+
+import androidx.annotation.NonNull;
+
+import java.util.Objects;
+
+/**
+ * Class that holds the gender detection results
+ * for the predictions category.
+ */
+public final class GenderAttribute {
```
jamesonwilliams: Let's be progressive and call this just `Gender`.

core/src/main/java/com/amplifyframework/predictions/models/GenderType.java
jpignata: Maybe `GenderBinaryType`?

core/src/main/java/com/amplifyframework/predictions/models/GenderAttribute.java
raphkim: I was thinking of [0,1], but makes sense to make it 0-100 because thats what the AWS services output. And I agree confidence is a better name.

core/src/main/java/com/amplifyframework/predictions/models/GenderType.java
raphkim: Will go with John's suggestion :)

core/src/main/java/com/amplifyframework/predictions/PredictionsCategoryBehavior.java
```diff
@@ -0,0 +1,219 @@
...
+import com.amplifyframework.predictions.result.InterpretResult;
+import com.amplifyframework.predictions.result.SpeechToTextResult;
+import com.amplifyframework.predictions.result.TextToSpeechResult;
+import com.amplifyframework.predictions.result.TranslateTextResult;
+
+import java.net.URL;
+
+/**
+ * Predictions category defines an abstract high-level behavior
+ * of machine learning tools.
```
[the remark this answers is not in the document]
raphkim: fixed

core/src/main/java/com/amplifyframework/predictions/PredictionsCategory.java
```diff
@@ -0,0 +1,164 @@
...
+import com.amplifyframework.predictions.result.IdentifyResult;
+import com.amplifyframework.predictions.result.InterpretResult;
+import com.amplifyframework.predictions.result.SpeechToTextResult;
+import com.amplifyframework.predictions.result.TextToSpeechResult;
+import com.amplifyframework.predictions.result.TranslateTextResult;
+
+import java.net.URL;
+
+/**
+ * Defines the Client API consumed by the application.
```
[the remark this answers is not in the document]
raphkim: yep

core/src/main/java/com/amplifyframework/predictions/PredictionsCategory.java
```diff
@@ -0,0 +1,164 @@
...
+    public CategoryType getCategoryType() {
+        return CategoryType.PREDICTIONS;
+    }
+
+    @NonNull
+    @Override
+    public PredictionsTranslateTextOperation<?> translateText(
+            @NonNull String text,
+            @NonNull LanguageType language,
+            @NonNull LanguageType targetLanguage,
```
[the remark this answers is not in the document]
raphkim: I agree. This asymmetry was bothering me too, but just rolled with what iOS had at the time. Changed to `fromLanguage` and `toLanguage`

core/src/main/java/com/amplifyframework/predictions/PredictionsCategory.java
```diff
@@ -0,0 +1,164 @@
...
+import com.amplifyframework.predictions.result.SpeechToTextResult;
+import com.amplifyframework.predictions.result.TextToSpeechResult;
+import com.amplifyframework.predictions.result.TranslateTextResult;
+
+import java.net.URL;
+
+/**
+ * Defines the Client API consumed by the application.
+ * Internally routes the calls to the Predictions Category
+ * plugins registered.
```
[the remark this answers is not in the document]
raphkim: fixed

core/src/main/java/com/amplifyframework/predictions/models/GenderAttribute.java
raphkim: haha we can do that

core/src/main/java/com/amplifyframework/predictions/models/EmotionType.java
raphkim: emotion and sentiments are different features that are detectable from a given text

core/src/main/java/com/amplifyframework/predictions/models/EntityMetadata.java
raphkim: fixed

core/src/main/java/com/amplifyframework/predictions/PredictionsCategoryBehavior.java
```diff
@@ -0,0 +1,219 @@
...
+     * Transcribe a text output from the audio input.
+     * @param speech The url of the audio to be transcribed
+     * @param options Parameters to specific plugin behavior
+     * @param onSuccess Triggered upon successful conversion
+     * @param onError Triggered upon encountering error
+     * @return The predictions operation object that can be used to directly access
+     *          the ongoing conversion operation
+     */
+    @NonNull
+    PredictionsSpeechToTextOperation<?> convertSpeechToText(
```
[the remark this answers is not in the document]
raphkim: Yea something that's been bothering me as well. Will remove :)

core/src/main/java/com/amplifyframework/predictions/models/AgeRange.java
```diff
@@ -0,0 +1,64 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.models;
+
+/**
+ * Class to represent a range of age.
+ */
+public final class AgeRange {
```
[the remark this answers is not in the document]
raphkim: I agree. One of the things that's been bothering me as well, but I'll leave it for now.

core/src/main/java/com/amplifyframework/predictions/result/InterpretResult.java
```diff
@@ -0,0 +1,188 @@
...
+import com.amplifyframework.predictions.models.LanguageDetectionResult;
+import com.amplifyframework.predictions.models.Sentiment;
+import com.amplifyframework.predictions.models.SyntaxToken;
+
+import java.util.List;
+
+/**
+ * The result of the call to interpret text.
+ */
+public final class InterpretResult implements Result {
```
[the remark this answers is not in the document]
raphkim: Depends on the plugin unfortunately. AWS Comprehension provides result for all of these, whereas tensorflow can only return detected sentiment.

core/src/main/java/com/amplifyframework/predictions/result/IdentifyDocumentTextResult.java
```diff
@@ -0,0 +1,240 @@
...
+import com.amplifyframework.predictions.models.IdentifiedLine;
+import com.amplifyframework.predictions.models.IdentifiedWord;
+import com.amplifyframework.predictions.models.Selection;
+import com.amplifyframework.predictions.models.Table;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * The result of the call to identify document text from an image.
```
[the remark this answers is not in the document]
raphkim: sounds way better.

core/src/main/java/com/amplifyframework/predictions/models/Polygon.java
raphkim: haha i was thinking this exact thing when I wrote this

core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
raphkim: I... gave up 1/3 of the way into doing exactly what you mentioned here manually... Not sure about using `Locale` though.

core/src/main/java/com/amplifyframework/predictions/models/LandmarkType.java
raphkim: I agree. I was thinking about this too, but was conservative about wildly renaming feature names from iOS

core/src/main/java/com/amplifyframework/predictions/models/AgeRange.java
```diff
@@ -0,0 +1,64 @@
...
+    private final int low;
+    private final int high;
+
+    /**
+     * Constructs a new instance of AgeRange.
+     * @param low minimum value of the range
+     * @param high maximum value of the range
+     * @throws IllegalArgumentException if low > high
+     */
+    public AgeRange(int low, int high) {
```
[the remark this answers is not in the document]
raphkim: I don't feel that adding a constraint like that is necessary. If a backend ML outputs something unreasonable, that should simply be displayed as such rather than us throwing an error because of it.

core/src/main/java/com/amplifyframework/predictions/models/AgeRange.java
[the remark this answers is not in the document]
raphkim: :(

core/src/main/java/com/amplifyframework/predictions/PredictionsException.java
```diff
@@ -0,0 +1,50 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions;
+
+import androidx.annotation.NonNull;
+
+import com.amplifyframework.AmplifyException;
+
+public final class PredictionsException extends AmplifyException {
```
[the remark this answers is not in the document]
raphkim: Yea not sure why build didn't catch that. Fixed.

core/src/main/java/com/amplifyframework/predictions/models/BooleanFeature.java
```diff
@@ -0,0 +1,79 @@
...
+        /**
+         * Sets the feature type and return this builder.
+         * @param type the type of feature
+         * @return this builder instance
+         */
+        @NonNull
+        public Builder type(@NonNull String type) {
+            this.type = Objects.requireNonNull(type);
+            return this;
+        }
```
jamesonwilliams: What's an example? Reading this, I'm not sure how to use it. What types of boolean features are there?

core/src/main/java/com/amplifyframework/predictions/models/BoundedKeyValue.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Holds the key-value detection results
+ * for the predictions category.
+ */
+public final class BoundedKeyValue extends ImageFeature<Map.Entry<String, String>> {
```
jamesonwilliams: It would be better not to use `Map.Entry`, here, unless you're working with a `Map`.

Android's `Pair` could be a candidate: https://developer.android.com/reference/android/util/Pair

Actually, you could have this just be `BoundedKeyValue<K,V>`, and then pass the key and vaule types into `ImageFeature`, somehow?

Wait, but also - why does a bounded key value pair extend an "Image" thing?

core/src/main/java/com/amplifyframework/predictions/models/Celebrity.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private Celebrity(final Builder builder) {
+        super(builder);
+    }
+
+    @Override
+    @NonNull
+    public String getType() {
+        return FEATURE_TYPE;
+    }
```
jamesonwilliams: Should these return an enumerated value? Like:
```
@Override
@NonNull
public FeatureType getType() {
    return FeatureType.CELEBRITY;
}
```

core/src/main/java/com/amplifyframework/predictions/models/Celebrity.java
```diff
@@ -0,0 +1,85 @@
...
+    public static final class Builder extends ImageFeature.Builder<Builder, Celebrity, String> {
+        /**
+         * Sets the name of celebrity. This is the same as
+         * setting the feature value for this builder.
+         * @param name the name of celebrity
+         * @return this builder instance
+         */
+        @NonNull
+        public Builder name(@NonNull String name) {
+            return super.feature(name);
```
jamesonwilliams: Cool

core/src/main/java/com/amplifyframework/predictions/models/CelebrityDetails.java
```diff
@@ -0,0 +1,179 @@
...
+public final class CelebrityDetails {
+    private final Celebrity celebrity;
+    private final Pose pose;
+    private final List<FacialFeature> facialFeatures;
+    private final List<URL> urls;
+
+    private CelebrityDetails(final Builder builder) {
+        this.celebrity = builder.getCelebrity();
+        this.pose = builder.getPose();
+        this.facialFeatures = builder.getfacialFeatures();
```
jamesonwilliams: nit, --> should be getFactialFeatures with the first F as capital

core/src/main/java/com/amplifyframework/predictions/models/EmotionType.java
jamesonwilliams: Please put it in Javadoc, though. I'm not asking for me -- I'm asking for the next person!

core/src/main/java/com/amplifyframework/predictions/models/Cell.java
```diff
@@ -0,0 +1,189 @@
...
+ */
+
+package com.amplifyframework.predictions.models;
+
+import androidx.annotation.NonNull;
+
+import java.util.Objects;
+
+/**
+ * A representation of table's individual cell.
```
jamesonwilliams: Hm wait. But why do we have tables for predictions? Is there a Table in the category types? How does this class related to the category behaviors?

core/src/main/java/com/amplifyframework/predictions/models/Feature.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public final float getConfidence() {
+        return confidence;
+    }
+
+    /**
+     * Compares an feature to another. The features are
+     * sorted by their types' alphabetic order, and by
+     * decreasing order of their confidence score for those
+     * within the same type.
```
jamesonwilliams: Would it make sense to sort only by confidence? (Just a question.)

core/src/main/java/com/amplifyframework/predictions/models/ImageEntity.java
```diff
@@ -0,0 +1,262 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Holds the image entity detection results
+ * for the predictions category.
+ */
+public final class ImageEntity {
```
jamesonwilliams: The components of this class look like they have to do with a human / some humans / or maybe the Celebrity type?

But this class name doesn't encode that. Instead of ImageEntity, it should probably say something about people / celebrities (whatever type has age, pose, gender.)

core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
jamesonwilliams: Yea, I agree we don't want to do it manually. But, we also don't want to maintain this list. Either way, it would be good to have the platform / Java / a computer program generate/maintain this listing for us.

core/src/main/java/com/amplifyframework/predictions/models/Table.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A representation of 2-D table in given dimensions.
+ * A table can be detected from an image of a document
+ * and be organized into this data type.
+ */
+public final class Table {
```
jamesonwilliams: Since your `Cell` class is tightly related to this `Table`, it could make sense to make the `Cell` as a static inner class of the Table. That way, they're related by design and the reader doesn't need to go find one or the other.

core/src/main/java/com/amplifyframework/predictions/result/TranslateTextResult.java
```diff
@@ -0,0 +1,115 @@
...
+         * Throw if any of the properties is null.
+         * @return An instance of {@link TranslateTextResult}
+         */
+        @NonNull
+        public TranslateTextResult build() {
+            return new TranslateTextResult(this);
+        }
+
+        @NonNull
+        String getTranslatedText() {
```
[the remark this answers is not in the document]
raphkim: Yea my original PR had constructor take each parameter, but my refactoring involved a lot of builder inheritance (e.g. `Feature.Builder` and `TextFeature.Builder`) so I redid them all like this. 

The reason for having a getter for each builder was to make it so that `requireNonNull` check logic stays within the builder, although I can simply follow your suggestion and null-check inside the constructor.  I don't think it makes a big enough difference though.

core/src/main/java/com/amplifyframework/predictions/options/SpeechToTextOptions.java
```diff
@@ -0,0 +1,29 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.options;
+
+import com.amplifyframework.core.async.Options;
+
+public final class SpeechToTextOptions implements Options {
```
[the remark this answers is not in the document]
raphkim: Not sure how these are getting though haha. I plan on revisiting options separately anyways, but I can add in the javadoc for now

core/src/main/java/com/amplifyframework/predictions/models/Cell.java
raphkim: Based on the result data types, it seems that we support detecting a table from a document image. I put explanation for it in the `Table.java` class, but not sure how accurate that is because I haven't visited `identify` API yet.

core/src/main/java/com/amplifyframework/predictions/models/Feature.java
raphkim: It is sorted only by confidence as long as you are comparing two features of same type. It wouldn't really make sense to compare the confidence rating of sentiment detection with say... language detection. I was considering throwing a runtime exception if the feature type mismatches, but I just left it like this.

Confidence sorting would make sense if an operation returns a list of results (for example, if text interpretation returned an output like `[70% positive, 10% negative, 20% neutral],` (just an example), then a user can sort this list to find the predominant sentiment (first element of sorted list) or something like that.

core/src/main/java/com/amplifyframework/predictions/models/Table.java
raphkim: will do

core/src/main/java/com/amplifyframework/predictions/models/TextFeature.java
```diff
@@ -0,0 +1,66 @@
...
+     * Gets the target text and associated index.
+     * @return the target text
+     */
+    @NonNull
+    public final TargetText getTarget() {
+        return target;
+    }
+
+    abstract static class Builder<B extends Builder<B, R, T>, R extends TextFeature<T>, T>
+            extends Feature.Builder<B, R, T> {
```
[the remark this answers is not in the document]
raphkim: My head was definitely exploding as I was writing this as well. I did document these inside `Feature.Builder`, but I forgot to do it for the others oops.

core/src/main/java/com/amplifyframework/predictions/models/BooleanFeature.java
raphkim: i was thinking along the line of... 
```
BooleanFeature noseDetected = BooleanFeature.builder()
        .type("Nose")
        .feature(true)
        .build();
```

Although typing this, I kind of want to rename `Feature#getFeature()` to `Feature#getValue()`. Hmm.

core/src/main/java/com/amplifyframework/predictions/models/BoundedKeyValue.java
raphkim: It seems to be something that is returned by Image identification method. Not sure what it means to detect a key-value pair from a document image though...

core/src/main/java/com/amplifyframework/predictions/models/CelebrityDetails.java
raphkim: ah shoot. "Replace All" strikes again haha. will fix.

core/src/main/java/com/amplifyframework/predictions/PredictionsCategoryConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+public final class PredictionsCategoryConfiguration extends CategoryConfiguration {
+
+    //TODO: Place global (category-wise) configuration options here
+
+    /**
+     * Constructs a new PredictionsCategoryConfiguration.
+     */
+    public PredictionsCategoryConfiguration() {
+        super();
+    }
```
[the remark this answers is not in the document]
raphkim: yep, just a copy-paste aftermath haha.

raphkim:
I will begin opening PRs by API to break up the contents of this PR. See #331

core/src/main/java/com/amplifyframework/predictions/models/BoundedKeyValue.java
```diff
@@ -16,16 +16,15 @@
...
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 
 /**
  * Holds the key-value detection results
  * for the predictions category.
  */
-public final class BoundedKeyValue extends ImageFeature<Map.Entry<String, String>> {
+public final class BoundedKeyValue extends ImageFeature<Pair<String, String>> {
```
jamesonwilliams: Nice

PR #308, opened 2020-03-19
[aws-datastore] Perform base sync at startup and save model versions

jamesonwilliams:
Perform a base sync as part of starting up the DataStore.

ModelMetadata is used to record the version of particular models managed
by the DataStore. A table for this structure must be created while the
storage adapter is initializing.

Items coming back from the base sync must be applied with respect to the
topological ordering of their model classes. For example, given a
relationship:

 1 Blog - n Posts - n Comments

The blog must be created, so that a post can "belongsTo" it, and the
post must be created next, so that a comment can "belongsTo" that.


jamesonwilliams:
Combined #309 and #308 into #315.

PR #309, opened 2020-03-19
[aws-datastore] Separate create and update types for storage adapter

jamesonwilliams:
While publishing a local change, the sync engine needs to know if the
change was a create or an update.

The storage adapter knows whether or not a save resulted in the creation
of a new record, or an update to an existing record. So, the storage
adapter is able to communicate out this information to its listeners.


jamesonwilliams:
Combined #309 and #308 into #315.

PR #312, opened 2020-03-19
[Analytics] Add api to attach user information to the endpoint

desokroshan:
*Issue #, if available:*

*Description of changes:*


core/src/main/java/com/amplifyframework/analytics/AnalyticsUserProfile.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public static final class Location {
+        private Double latitude;
+        private Double longitude;
+        private String postalCode;
+        private String city;
+        private String region;
+        private String country;
+
+        public Location(final Double latitude,
```
jamesonwilliams: Important: add `equals(...)`, `toString()`, and `hashCode()` for this `Location` class.

How many of these fields are optional? Can you add `@Nullable` to getters and constructor params for any that aren't required? And `@NonNull` for those that must exist?

If `latitude` == `null` means "it wasn't provided", that would be a good thing to document also. Otherwise, use just the primitive `double` if your location class will always have lat/long.

The ideal situation is that all fields are `@NonNull`, and you use primitive `double`s. This way, there's no guess work for the user. If a `Location` was able to be constructed, that means its fields are populated, and the user can just use it without fear of `NullPointerException`: `location.getRegion()`, etc.

core/src/main/java/com/amplifyframework/analytics/AnalyticsUserProfile.java
```diff
@@ -0,0 +1,106 @@
...
+ * Represents user specific data such as name, email, plan, location etc.
+ */
+public final class AnalyticsUserProfile {
+    private String name;
+    private String email;
+    private String plan;
+    private Location location;
+    private Properties customProperties;
+
+    public AnalyticsUserProfile(final String name,
```
jamesonwilliams: Which fields are `@Nullable`, `@NonNull`?

Implement `hashCode()`, `toString()`, and `equals()` for any DTOs like this, that would be used in production.

Can this be called just a `UserProfile`? Do we need to prefix everything with `AnalyticsUserProfile`? Are there other common `UserProfile` classes that it would clash with, if we call it that?

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategory.java
```diff
@@ -60,21 +60,25 @@ public CategoryType getCategoryType() {
...
     public void disable() {
         synchronized (LOCK) {
             enabled = false;
+            getSelectedPlugin().disable();
         }
     }
 
     @Override
     public void enable() {
         synchronized (LOCK) {
```
jamesonwilliams: Which components is this lock guarding? Is it to synchronize access to the boolean? Or what's the intention?

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -79,8 +93,71 @@ public void enable() {
...
+     * @param location location details.
+     */
+    private void addLocation(EndpointProfileLocation endpointProfileLocation,
+                             AnalyticsUserProfile.Location location) {
+        endpointProfileLocation.setLatitude(location.getLatitude());
+        endpointProfileLocation.setLongitude(location.getLongitude());
+        endpointProfileLocation.setPostalCode(location.getPostalCode());
+        endpointProfileLocation.setCity(location.getCity());
+        endpointProfileLocation.setRegion(location.getRegion());
+        endpointProfileLocation.setCountry(location.getCountry());
```
jamesonwilliams: Looks like you want a builder. :-D If some of these fields are required, then consider a step builder. (That way, you could check if required fields are present at build time, not just runtime.)

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -79,8 +93,71 @@ public void enable() {
...
+        if (customProperties != null) {
+            for (Map.Entry<String, Property<?>> entry : customProperties.get().entrySet()) {
+                if (entry.getValue() instanceof StringProperty) {
+                    endpointProfile.addAttribute(entry.getKey(),
+                            Arrays.asList(((StringProperty) entry.getValue()).getValue()));
+                } else if (entry.getValue() instanceof DoubleProperty) {
+                    endpointProfile.addMetric(entry.getKey(),
+                            ((DoubleProperty) entry.getValue()).getValue());
+                } else {
+                    Amplify.Hub.publish(HubChannel.ANALYTICS, HubEvent.create("Analytics.identifyUser",
```
jamesonwilliams: This behavior seems reasonable, but I'm not sold on the details of the `HubEvent`.

1. Can you get the event name into a shared `enum`? The other categories currently enumerate the possible event names in a central location, and have a pattern for what the event names should look like. Analytics should do likewise.

2. Since something exceptional has occurred, the payload type of the event should probably be an `AnalyticsException`. "Refer to the documentation" isn't an adequate recovery suggestion. If you'll point a user at documentation, it should include a link URL. But, the issue here is probably more like they tried to supply a property type that isn't supported, and so they should try to use one of the supported property types, instead. The supported property types are (state them here.)

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -79,8 +93,71 @@ public void enable() {
...
+    /**
+     * Add user specific data from {@link AnalyticsUserProfile} to the endpoint profile.
+     * @param endpointProfile endpoint profile.
+     * @param userProfile user specific data to be added to the endpoint.
+     */
+    private void addUserProfileToEndpoint(EndpointProfile endpointProfile,
+                                          AnalyticsUserProfile userProfile) {
+        endpointProfile.addAttribute("email", Arrays.asList(userProfile.getEmail()));
+        endpointProfile.addAttribute("name", Arrays.asList(userProfile.getName()));
+        endpointProfile.addAttribute("plan", Arrays.asList(userProfile.getPlan()));
```
jamesonwilliams: The thing in the value position here is called a "singleton list," a list with one element. Java `Collections` utility provides a more efficient utility for constructing these things, `Collections.singletonList(ofOneItem)`.

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -191,6 +200,43 @@ public void testUnregisterGlobalProperties() throws AnalyticsException {
...
 
+    @Test
+    public void testIdentifyUser() {
+        // Create a location object
+        AnalyticsUserProfile.Location location = new AnalyticsUserProfile.Location(SEATTLE_LATITUDE,
+                SEATTLE_LONGITUDE,
+                "98122",
+                "Seattle",
+                "WA",
+                "USA");
```
jamesonwilliams: Instead of splitting the magic data between class-level constants, and inline literals, I would just keep it all locally defined. In other words, the instance of `AnalyticsUserProfile.Location` _itself_ is magic data, not just its constituent fields:

```
@SuppressWarning("checkstye:MagicNumber") // Seattle's location is magic. Snow-capped volcanos, Orcas swimming in the Puget Sound, and the smell of fresh roasted coffee wafting on the pier...
UserProfile.Location locationOfSeattle = UserProfile.Location.builder()
    .latitute(47.6154086)
    .longitude(-122.3349685)
    .zipcode("98122")
    .city("Seattle")
    .locality("WA") // locality? TODO: what is the generic term for this, that accounts for states/provinces?
    .country("USA")
    .build();
```

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -191,6 +200,43 @@ public void testUnregisterGlobalProperties() throws AnalyticsException {
...
+        assertEquals(endpointProfile.getAttribute("name").get(0), "test-user");
+        assertEquals(endpointProfile.getAttribute("plan").get(0), "test-plan");
+        assertEquals(endpointProfileLocation.getLatitude(), (Double) SEATTLE_LATITUDE);
+        assertEquals(endpointProfileLocation.getLongitude(), (Double) SEATTLE_LONGITUDE);
+        assertEquals(endpointProfileLocation.getPostalCode(), "98122");
+        assertEquals(endpointProfileLocation.getCity(), "Seattle");
+        assertEquals(endpointProfileLocation.getRegion(), "WA");
+        assertEquals(endpointProfileLocation.getCountry(), "USA");
+        assertEquals(endpointProfile.getAttribute("TestStringProperty").get(0), "TestStringValue");
+        assertEquals(endpointProfile.getMetric("TestDoubleProperty"), (Double) 1.0);
```
jamesonwilliams: Once you get `equals()` working correctly on these POJOs, you want to end up with something like a _single_ `assertEquals()` call, on the top-level object:
```
assertEquals(expectedComplexType, actualComplexType);
```

Internally, `assertEquals` will call `equals()` on the object. And _those_ should cascade through the fields, determining equality.

Or, closer to what you'll actually have:
```
assertEquals(
    // Expected data. Build one here, inline, that contains the field values you _expect_ to see.
    EndpoingProfile.builder()
        .addAttribute("name", Collections.singletonList("test-user"))
        .addAttribute("...", "... etc. "'")
       .build(),
    // The actual expirementally obtained copy ...
   endpointProfile
);
```

Also, as a note - in your `assertEquals(expected, actual)` calls, you currently have the expected and actual positions backwards. (Expected data should be first position, you have it as second position.) This makes a difference when the test fails, and jUnit goes to print out the error. It ordinarily would say something like "Expected \"Seatte\", but got \"Toledo\"." Yours would end up saying something like "Expected \"Toledo\", but got \"Seattle\"."

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategory.java
desokroshan: Its guarding the access to boolean flag `enabled`.

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    public static final class Location {
+        private Double latitude;
+        private Double longitude;
+        private String postalCode;
+        private String city;
+        private String region;
+        private String country;
```
jamesonwilliams: These ones can be `final`

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+        public String getRegion() {
+            return region;
+        }
+
+        @Nullable
+        public String getCountry() {
+            return country;
+        }
+
+        public static Builder builder() {
```
jamesonwilliams: `@NonNull`

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+            public Location build() {
+                if (latitude == null &&
+                    longitude == null &&
+                    postalCode == null &&
+                    city == null &&
+                    region == null &&
+                    country == null) {
+                    throw new IllegalArgumentException("At least on parameter must be non-null to " +
+                            "construct a Location object");
+                }
```
jamesonwilliams: Since the check is so unlikely to trigger, it may not be worth it to maintain this code. Might be better off just allowing an object with all `null`s`.

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+    public Location getLocation() {
+        return location;
+    }
+
+    @Nullable
+    public Properties getCustomProperties() {
+        return customProperties;
+    }
+
+    public static Builder builder() {
```
jamesonwilliams: `@NonNull`

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+        return "UserProfile{" +
+                "name='" + name + '\'' +
+                ", email='" + email + '\'' +
+                ", plan='" + plan + '\'' +
+                ", location=" + location +
+                ", customProperties=" + customProperties +
+                '}';
+    }
+
+    public static final class Builder {
```
jamesonwilliams: These Builder methods all should return `@NonNull`.

You can configure your IDE to warn when things don't have these annotations, actually. (I have this setup.)

Our code base is too far "off" right now to enable it with an automated tool at compile time. But it's still possible to get the suggestions in the IDE.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -93,32 +96,45 @@ public void enable() {
...
      * @param userProfile user specific data to be added to the endpoint.
      */
     private void addUserProfileToEndpoint(EndpointProfile endpointProfile,
-                                          AnalyticsUserProfile userProfile) {
-        endpointProfile.addAttribute("email", Arrays.asList(userProfile.getEmail()));
-        endpointProfile.addAttribute("name", Arrays.asList(userProfile.getName()));
-        endpointProfile.addAttribute("plan", Arrays.asList(userProfile.getPlan()));
+                                          UserProfile userProfile) {
+        if (userProfile.getEmail() != null) {
+            endpointProfile.addAttribute("email", Collections.singletonList(userProfile.getEmail()));
```
jamesonwilliams: Can you get these keys into an enum or StringDef, too? `PinpointProfileAttribute.EMAIL`, etc.? You could add a little helper method like:
```
private void addAttribute(EndpointProfile profile, @PinpointProfileAttribute String profileAttribute, String value) {
    profile.addAttribute(profileAttribute, Collections.singletonList(value));
}
```

^^ Above, the at-PinpointProfileAttribute is a StringDef

core/src/main/java/com/amplifyframework/analytics/AnalyticsCategoryBehavior.java
```diff
@@ -33,7 +34,7 @@
      * @param userId The unique identifier for the user
      * @param profile User specific data (e.g. plan, accountType, email, age, location, etc)
      */
-    void identifyUser(@NonNull String userId, @NonNull AnalyticsProfile profile);
+    void identifyUser(@NonNull String userId, @Nullable UserProfile profile);
```
jamesonwilliams: If you make this nullable, add a note to the javadoc about what happens when it is null / what the null means.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -79,8 +96,84 @@ public void enable() {
      * {@inheritDoc}
      */
     @Override
-    public void identifyUser(@NonNull String userId, @NonNull AnalyticsProfile profile) {
-        throw new UnsupportedOperationException("This operation has not been implemented yet.");
+    public void identifyUser(@NonNull String userId, @Nullable UserProfile userProfile) {
```
jamesonwilliams: Right now, if `userProfile` is `null`, then it will get passed down into `addUserProfileToEndpoint(...)`, where `userProfile.getSomething()` happens, and will generate a`NullPointerException`.

So, either you should guard `if (userProfile != null) { useIt(); }`, or you should dis-allow `null` by `requireNonNull(...)` immediately after receiving the value.

It's possible that `null` for htis param is supposed to mean "clear the identified user."?

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,313 @@
...
+                return this;
+            }
+
+            public Builder region(@NonNull String region) {
+                Objects.requireNonNull(region);
+                this.region = region;
+                return this;
+            }
+
+            public Builder country(@NonNull String country) {
```
jamesonwilliams: All of these things that return a `Builder` will all return `@NonNull`, too. Might as well add it everywhere you can on public/protected members. It makes a big help for Kotlin code that calls this stuff.

core/src/main/java/com/amplifyframework/analytics/AnalyticsUserProfile.java
desokroshan: Added null notations where ever applicable.

core/src/main/java/com/amplifyframework/analytics/AnalyticsUserProfile.java
desokroshan: Added null notations.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
desokroshan: I agree `EndpointProfileLocation` can benefit from Builder because of so many optional params. We discussed not to do it since EndpointProfileLocation is a part of pinpoint sdk.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
desokroshan: Added enum and updated the event payload to be an exception.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
desokroshan: Added null check for userProfile. userProfile = null means that the endpoint will have userId as the only information about the user and nothing more.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -18,35 +18,58 @@
...
+    @StringDef({
+        USER_NAME,
+        USER_EMAIL,
+        USER_PLAN
+    })
+    private @interface PinpointUserProfileAttribute {}
+    private static final String USER_NAME = "name";
+    private static final String USER_EMAIL = "email";
+    private static final String USER_PLAN = "plan";
     private final Application application;
```
jamesonwilliams: nit: to help break up the groupings, add an empty line after the constants, before the `Application` declaration.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -65,6 +88,14 @@ protected AnalyticsClient getAnalyticsClient() {
         return analyticsClient;
     }
 
+    /**
+     * Accessor method for pinpoint targeting client.
+     * @return returns pinpoint targeting client.
+     */
+    protected TargetingClient getTargetingClient() {
```
jamesonwilliams: This class is `final`, and can't be extended. So `protected` isn't of much value, here. I think maybe you want just default package-local access, here?

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,314 @@
...
+ * Represents user specific data such as name, email, plan, location etc.
+ */
+public final class UserProfile {
+    private final String name;
+    private final String email;
+    private final String plan;
+    private final Location location;
+    private final Properties customProperties;
+
+    public UserProfile(@NonNull Builder builder) {
```
jamesonwilliams: If you make this as `private UserProfile(Builder builder)`, then only you will be able to call this constructor, and you don't have to worry about the builder being null (you'll be able to ensure that, with your own code in this file.)

core/src/main/java/com/amplifyframework/analytics/UserProfile.java
```diff
@@ -0,0 +1,314 @@
...
+
+    public static final class Location {
+        private final Double latitude;
+        private final Double longitude;
+        private final String postalCode;
+        private final String city;
+        private final String region;
+        private final String country;
+
+        public Location(@NonNull Builder builder) {
```
jamesonwilliams: Same deal here, you can probably make this `private Location(Builder builder)`, and then as long as you always pass non-null from `build()`, you don't need to check `Objects.requireNonNull(...)`, since that's the only single place it'll ever be possible.

PR #315, opened 2020-03-20
Versions in datastore

jamesonwilliams:
1. Write model versions to a local store, while working over the results from a base sync.
2. Read and post model versions to AppSync when publishing updates/deleted to AppSync.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
@@ -15,6 +15,7 @@
 
 package com.amplifyframework.datastore.storage;
 
+import android.annotation.SuppressLint;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.util.ObjectsCompat;
@@ -35,7 +36,6 @@ import java.util.UUID;
  * a stored item is changed.
  * @param <T> The type of the item that has changed
  */
-@SuppressWarnings({"WeakerAccess", "unused"})
 public final class StorageItemChange<T extends Model> {
     private final UUID changeId;
     private final Initiator initiator;
@@ -45,21 +45,12 @@ public final class StorageItemChange<T extends Model> {
     private final Class<T> itemClass;
 
     private StorageItemChange(
-            @NonNull UUID changeId,
-            @NonNull Initiator initiator,
-            @NonNull Type type,
-            @NonNull T item,
-            @NonNull Class<T> itemClass) {
-        this(changeId, initiator, type, null, item, itemClass);
-    }
-
-    private StorageItemChange(
-            @NonNull UUID changeId,
-            @NonNull Initiator initiator,
-            @NonNull Type type,
-            @Nullable QueryPredicate predicate,
-            @NonNull T item,
-            @NonNull Class<T> itemClass) {
+            UUID changeId,
+            Initiator initiator,
+            Type type,
+            QueryPredicate predicate,
+            T item,
+            Class<T> itemClass) {
         this.changeId = changeId;
         this.initiator = initiator;
         this.type = type;
@@ -72,6 +63,7 @@ public final class StorageItemChange<T extends Model> {
      * Gets the ID of this change.
      * @return Unique ID for this change
      */
+    @NonNull
     public UUID changeId() {
         return changeId;
     }
@@ -80,6 +72,7 @@ public final class StorageItemChange<T extends Model> {
      * Gets an identification of the actor who initiated this change.
      * @return Identification of actor who initiated the change
      */
+    @NonNull
     public Initiator initiator() {
         return initiator;
     }
@@ -88,6 +81,7 @@ public final class StorageItemChange<T extends Model> {
      * Gets the type of change.
      * @return Type of change
      */
+    @NonNull
     public Type type() {
         return type;
     }
@@ -96,6 +90,7 @@ public final class StorageItemChange<T extends Model> {
      * Gets the predicate that was applied before this item change.
      * @return Predicate applied for this item change
      */
+    @Nullable
     public QueryPredicate predicate() {
         return predicate;
     }
@@ -106,6 +101,7 @@ public final class StorageItemChange<T extends Model> {
      * For deletions, this is the item as it was before the deletion.
      * @return Representation of item that changed
      */
+    @NonNull
     public T item() {
         return item;
     }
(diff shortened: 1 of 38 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
```diff
@@ -133,12 +131,13 @@ public T item() {
...
+    @NonNull
     public Record toRecord(@NonNull RecordFactory recordFactory) {
         return Objects.requireNonNull(recordFactory).toRecord(this);
     }
 
     @Override
-    public boolean equals(Object thatObject) {
+    public boolean equals(@Nullable Object thatObject) {
         if (this == thatObject) {
             return true;
```
TrekSoft: This looks like the default equals functionality - is there a need to override it that I'm missing here?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -85,20 +103,78 @@ void stopDrainingMutationOutbox() {
...
-     * we have to keep the mutation in the outbox, so that we can try to publish
-     * it again later, when network conditions become favorable again.
+     * Attempt to publish a change (update, delete, creation) over the network.
      * @param storageItemChange A storage item change to be published to remote API
-     * @return A single which completes with the successfully published item, or errors
+     * @return A single which completes with the successfully published item, or emits error
      *         if the publication fails
      */
-    @SuppressWarnings("checkstyle:MethodTypeParameterName") // The generics are complex, so use meaningful names
     private <MODEL extends Model, SIC extends StorageItemChange<MODEL>> Single<SIC> publishToNetwork(
```
TrekSoft: What's the idea behind using caps to distinguish between these two classes vs. a naming difference?

core/src/main/java/com/amplifyframework/core/model/ModelSchemaRegistry.java
```diff
@@ -57,8 +55,10 @@ public synchronized void load(@NonNull Set<Class<? extends Model>> models) throw
      *                        {@link Class#getSimpleName()} method.
      * @return the ModelSchema object for the given Model class.
      */
+    @NonNull
     public synchronized ModelSchema getModelSchemaForModelClass(@NonNull String classSimpleName) {
-        return modelSchemaMap.get(classSimpleName);
+        final ModelSchema result = modelSchemaMap.get(classSimpleName);
+        return Objects.requireNonNull(result);
```
TrekSoft: Do you think in this case it'd make sense to throw a custom error with an explanation if it is null since this isn't checking input provided by the customer but rather the result we're returning? Same for below on line 73?

core/src/main/java/com/amplifyframework/datastore/DataStoreItemChange.java
```diff
@@ -259,9 +263,14 @@ public String toString() {
      */
     public enum Type {
         /**
-         * An item has been saved into the DataStore.
+         * An item with a new ID has been saved into the DataStore.
+         */
+        CREATE,
```
TrekSoft: Why do we have what looks like the same enum here as well as in StorageItemChange.java in the DataStore project?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
jamesonwilliams: The [default `equals(...)`](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)) works like this:
> for any non-null reference values x and y, this method returns true if and only if x and y refer to the same object (x == y has the value true).

So, that'd be equivalent to an implementation like `return this == thatObject`.

But here, I say, "okay, if they're the same object, yea, they're equal. BUT ALSO.... they're equal if they're fields are equal, _even if they're not_ the same object."

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
jamesonwilliams: I could use `<M extends Model, S extends StorageItemChange<M>>`, I guess.

Since there are so many generic types, and several types around here start with an `M` (`Model`, `ModelMetadata`, `ModelWithMetadata`, etc.) I thought it'd be clearer to read if I was explicit with the `MODEL`, etc.

core/src/main/java/com/amplifyframework/core/model/ModelSchemaRegistry.java
jamesonwilliams: Hm, yes, I agree. I'll update this.

core/src/main/java/com/amplifyframework/datastore/DataStoreItemChange.java
jamesonwilliams: True, yea. The local storage module and the public DataStore API share a number of features. In this case, the enum is entirely the same.

I could use just one, but I'd need a name for it that made sense for both.

Logically speaking, one is talking about changes to the DataStore, and one is talking about changes to an internal system component (the storage adapter). Which theoretically could be different, but in practice, aren't really.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
TrekSoft: Argh, the diff made it look like there was nothing more to this method than what's shown above. Ignore this comment.

core/src/main/java/com/amplifyframework/datastore/DataStoreItemChange.java
TrekSoft: Ah, ok yeah if they have different meanings than that makes sense to have the two.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
TrekSoft: Ok, what's the difference between MODEL and Model? Would it make sense to give MODEL a name that reflects that difference?

jamesonwilliams: Update: I got rid of all this crap, in a subsequent update. It ends up being easier to read with just the `<T extends Model>` that we use eery where else.

PR #316, opened 2020-03-20
Add Android device details to User-Agent

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Currently the User-Agent header passed to HTTP servers includes unnecessary details such as the versions of the Linux kernel and Dalvik/ART VMs. This change removes these elements and adds the version of Android, device manufacturer, and device model to the User-Agent string. Details will be passed as they presented by Android with spaces in tact. These data points will aid in future support efforts and decisions.

For example, on my [test phone](http://phonedb.net/index.php?m=device&id=14161&c=nokia_7.1_dual_sim_td-lte_am_64gb__hmd_rooster), the constructed User-Agent is:

```
amplify-android/master (Android 9; HMD Global Nokia 7.1; en_US)
```

### DIFF
--- a/core/src/main/java/com/amplifyframework/util/UserAgent.java
+++ b/core/src/main/java/com/amplifyframework/util/UserAgent.java
@@ -16,6 +16,7 @@
 package com.amplifyframework.util;
 
 import android.annotation.SuppressLint;
+import android.os.Build;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -128,7 +122,7 @@ private static String sanitize(@Nullable String string) {
                 return "UNKNOWN";
             }
 
-            return string.replace(' ', '_');
+            return string;
```
TrekSoft: Do you know why this was here before and why we no longer need it?

jpignata: Perhaps a bit of a drive-by change. I'm not sure why we were normalizing the output this way. @jamesonwilliams?

I removed it as it isn't customary to replace spaces with underscores in [User-Agent header values](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) generally.

jamesonwilliams: Mohit Srivastava & Nikhil Dabhade had requested that we use the same format as is used in the SDK.

However, my guess is that not a ton of thought had gone into designing the SDK one, previously. More just that it had set a formatting precedence.

Anyway, the SDK had this `_` character, among other weird choices (several of which John is correcting in this PR.)

If the `_` is weird, let's get rid of it. The best outcome is a well-formatted and usable `User-Agent` string.

PR #322, opened 2020-03-25
[aws-api] User wants to provide multiple path segments

jamesonwilliams:
Previously, when a user passed multiple path segments via the
RestOptions path string:
```
Amplify.API.get(new RestOptions("path1/path2"), ok -> {}, fail -> {});
```
The forward-slash character (`/`) in the path string would be changed
into the string "`%2F`". So, if the endpoint were `https://domain.tld/api`, you'd end
up making a request to `https://domain.tld/api/path1%2Fpath2`.

After this change, the `/` characters in the provide path are preserved.
The system would end up making a request, instead, to
`https://domain.tld/api/path1/path2`.

Resolves: https://github.com/aws-amplify/amplify-android/issues/304

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
@@ -15,11 +15,15 @@
 
 package com.amplifyframework.api.aws.utils;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import com.amplifyframework.api.rest.HttpMethod;
 
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Map;
+import java.util.Objects;
 
 import okhttp3.Headers;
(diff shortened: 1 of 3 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
```diff
@@ -15,78 +15,74 @@
...
-        }
         URL url = new URL(endpoint);
         HttpUrl.Builder builder = new HttpUrl.Builder()
-                .scheme(url.getProtocol())
-                .host(url.getHost())
-                .addPathSegment(url.getPath().replaceFirst("/", ""))
-                .addPathSegment(urlPath);
+            .scheme(url.getProtocol())
+            .host(url.getHost())
+            .addPathSegment(url.getPath().replaceFirst("/", ""))
```
TrekSoft: What if the path is "path1/path2" as in your example? Wouldn't this remove that slash?

PR #323, opened 2020-03-25
[aws-api] Test that two subscriptions for the same thing have unique ACKs

jamesonwilliams:
Try to create two subscriptions over a websocket, using the same request
body.

Expect this to result in two subscriptions, each with a unique ID, and
each having received a `subscription_ack` over the websocket.

### DIFF
--- a/aws-api/src/androidTest/java/com/amplifyframework/api/aws/SubscriptionEndpointTest.java
+++ b/aws-api/src/androidTest/java/com/amplifyframework/api/aws/SubscriptionEndpointTest.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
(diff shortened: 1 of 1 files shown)


aws-api/src/androidTest/java/com/amplifyframework/api/aws/SubscriptionEndpointTest.java
```diff
@@ -0,0 +1,156 @@
...
+            executor.execute(() ->
+                subscriptionEndpoint.requestSubscription(
+                    request,
+                    onResult,
+                    item -> {
+                        final String message;
+                        if (item.hasErrors()) {
+                            message = "Subscription error: " + item.getErrors().toString();
+                        } else {
+                            message = "Unexpect subscription data: " + item.getData();
```
TrekSoft: *Unexpected

TrekSoft: Also, I must be missing something but why are you failing if it receives an item without errors?

jamesonwilliams: Ah thanks the for the catch -

And, for _this_ test, all that's important is the establishment of the subscription, and the receipt of the subscription IDs. Since I don't create any actual models, I don't expect to receive any subscription data. If I do receive any items on the subscription, it'd be unexpected.

PR #326, opened 2020-03-25
Remove Retrolambda

jamesonwilliams:
This project will require Java 1.8 in its consumers, and will no longer
use Retrolambda to support older consumers.

Consumers will need to declare the following block in their
`build.gradle`:

```
android {
    compileOptions {
    }
}
```

Amplify and its consumers are expected to use only the first-party
Android tools to provide Java 1.8 features.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,6 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.6.1'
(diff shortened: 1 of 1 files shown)


jpignata:
Solving this with documentation versus a dependency seems right to me given my testing and read of https://developer.android.com/studio/write/java8-support#migrate_from_retrolambda. :thumbsup:

TrekSoft:
We already have this in all the docs anyway, to my understanding so makes sense.

PR #329, opened 2020-03-26
[Analytcis] Minimize visibility of classes and methods

desokroshan:
*Issue #, if available:*

*Description of changes:*
Reduce the accessibility of the Configuration class and remove configuration key `AUTO_SESSION_TRACKING_INTERVAL` which is no longer being used.  


undefobj:
Does this mean we are removing auto tracking? Or was it never implemented?

desokroshan:
> Does this mean we are removing auto tracking? Or was it never implemented?

No, we implemented auto-session tracking but that does not require the concept of an interval. It automatically records session events when app is opened and closed. I added it initially in anticipation when the auto session tracking implementation wasn't completely thought through.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPluginConfiguration.java
```diff
@@ -18,32 +18,29 @@
 /**
  * Configuration options for Amplify Analytics Pinpoint plugin.
  */
-public final class AmazonPinpointAnalyticsPluginConfiguration {
+final class AmazonPinpointAnalyticsPluginConfiguration {
```
TrekSoft: Without specifying public, doesn't it default to package-private which leads consumers of Amplify not to be able to view this?

jamesonwilliams: Yes; this is the key question. Should a user have access to this object, or it better internal? Unless we have a use case in mind where the user should interact with this thing, lesser access is the better default.

TrekSoft: If we're making this change, so that the customer can only provide JSON configuration,  having a configuration object at all might be extraneous. We already don't have one in Storage and the one in API seems like an unnecessary extra step where we read the JSON into the configuration object only to then read the same data back out of the configuration object.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -324,14 +324,6 @@ public void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context
                                 .getLong(PinpointConfigurationKey.AUTO_FLUSH_INTERVAL.getConfigurationKey()));
             }
 
-            if (pluginConfiguration
```
jamesonwilliams: How come this is getting removed? Isn't this PR about changing visibility of class/feild members? But, this looks like a functional change.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPluginConfiguration.java
jamesonwilliams: In our design the Plugin implementations are a development entry point. Kind of like an Activity class in Android. So, you could easily build a functioning system by writing all of your code in this class. But, the implementation will only grows, as features / bugfixes come in. So, the best thing you can do is to defensively split stuff out of the plugin implementation from the get-go. A reasonable question to ask at each iteration is "how can I decouple my business logic from this plugin class?"

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
desokroshan: Oh I should have kept it separate but this does not change the functionality. This flag is not required since the session is defined by app being in foreground and background rather than by time interval. So concept of time interval doesn't make sense anymore.

PR #330, opened 2020-03-26
Auth Category to Staging Branch [Do Not Merge To Master]

TrekSoft:
Initial PR so that work can be reviewed as I go vs. one massive PR at the end.

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -25,8 +25,8 @@ import com.amplifyframework.AmplifyException;
 import com.amplifyframework.auth.AuthCodeDeliveryDetails;
 import com.amplifyframework.auth.AuthException;
 import com.amplifyframework.auth.AuthPlugin;
+import com.amplifyframework.auth.AuthSignedInStatus;
 import com.amplifyframework.auth.AuthState;
-import com.amplifyframework.auth.AuthUserState;
 import com.amplifyframework.auth.options.AuthSignInOptions;
 import com.amplifyframework.auth.options.AuthSignUpOptions;
 import com.amplifyframework.auth.result.AuthSignInResult;
@@ -218,7 +218,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
         AWSMobileClient.getInstance().signIn(username, password, null, new Callback<SignInResult>() {
             @Override
             public void onResult(SignInResult result) {
-                AuthUserState state;
+                AuthSignedInStatus state;
                 UserCodeDeliveryDetails oldDetails = result.getCodeDetails();
                 // Take information from Cognito specific object and wrap it in the new Amplify object
                 AuthCodeDeliveryDetails newDetails =
@@ -231,9 +231,9 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                         : null;
 
                 if (result.getSignInState().equals(SignInState.DONE)) {
-                    state = AuthUserState.SIGNED_IN;
+                    state = AuthSignedInStatus.SIGNED_IN;
                 } else {
-                    state = AuthUserState.SIGNED_OUT;
+                    state = AuthSignedInStatus.SIGNED_OUT;
                 }
 
                 onSuccess.accept(new AuthSignInResult(state, newDetails));
@@ -328,7 +328,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     //   - Cognito ID Token
     //   - Cognito Refresh Token
     @Override
-    public void currentAuthState(
+    public void fetchAuthState(
             @NonNull Consumer<AuthState> onSuccess,
             @NonNull Consumer<AuthException> onError
     ) {
@@ -338,76 +338,27 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                 public void onResult(UserStateDetails result) {
                     UserState state = result.getUserState();
 
-                    if (state.equals(UserState.SIGNED_OUT) ||
-                            state.equals(UserState.SIGNED_OUT_FEDERATED_TOKENS_INVALID) ||
-                            state.equals(UserState.SIGNED_OUT_USER_POOLS_TOKENS_INVALID)
-                    ) {
-                        onSuccess.accept(new AWSCognitoAuthState(AuthUserState.SIGNED_OUT));
-                    } else {
-                        final AWSCognitoAuthState session;
-                        final CountDownLatch latch = new CountDownLatch(2);
-
-                        if (state.equals(UserState.SIGNED_IN) || state.equals(UserState.GUEST)) {
-                            session = new AWSCognitoAuthState(
-                                    state.equals(UserState.SIGNED_IN)
-                                            ? AuthUserState.SIGNED_IN
-                                            : AuthUserState.GUEST,
-                                    null,
-                                    AWSMobileClient.getInstance().getIdentityId(),
-                                    null,
-                                    null,
-                                    null
+                    switch (state) {
+                        case SIGNED_OUT:
+                        case SIGNED_OUT_USER_POOLS_TOKENS_INVALID:
+                        case SIGNED_OUT_FEDERATED_TOKENS_INVALID:
+                            onSuccess.accept(
+                                AWSCognitoAuthState.builder().signedInStatus(AuthSignedInStatus.SIGNED_OUT).build()
                             );
-
-                            AWSMobileClient.getInstance().getTokens(new Callback<Tokens>() {
-                                @Override
-                                public void onResult(Tokens result) {
-                                    session.setAccessToken(result.getAccessToken().getTokenString());
-                                    session.setIdToken(result.getIdToken().getTokenString());
-                                    session.setRefreshToken(result.getRefreshToken().getTokenString());
-                                    latch.countDown();
-                                    if (latch.getCount() == 0) {
-                                        onSuccess.accept(session);
-                                    }
-                                }
-
-                                @Override
-                                public void onError(Exception exception) {
-                                    session.setAccessToken(null);
-                                    session.setIdToken(null);
-                                    session.setRefreshToken(null);
-                                    latch.countDown();
-                                    if (latch.getCount() == 0) {
-                                        onSuccess.accept(session);
-                                    }
-                                }
-                            });
-
-                            AWSMobileClient.getInstance().getAWSCredentials(new Callback<AWSCredentials>() {
-                                @Override
-                                public void onResult(AWSCredentials result) {
-                                    session.setAwsCredentials(result);
-                                    latch.countDown();
-                                    if (latch.getCount() == 0) {
-                                        onSuccess.accept(session);
-                                    }
-                                }
-
-                                @Override
-                                public void onError(Exception exception) {
-                                    session.setAwsCredentials(null);
-                                    latch.countDown();
-                                    if (latch.getCount() == 0) {
-                                        onSuccess.accept(session);
-                                    }
-                                }
-                            });
-                        } else {
+                            break;
+                        case SIGNED_IN:
+                        case GUEST:
+                            try {
+                                onSuccess.accept(buildCognitoAuthState(state));
+                            } catch (AuthException exception) {
+                                onError.accept(exception);
+                            }
+                            break;
+                        default:
                             onError.accept(new AuthException(
                                     "User is in an unknown authorization state",
                                     "This is a bug. Please report it to the AWS team for resolution"
                             ));
-                        }
                     }
                 }
 
@@ -433,5 +384,63 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     @Override
     public AWSMobileClient getEscapeHatch() {
         return AWSMobileClient.getInstance();
+    }
+
+    private AWSCognitoAuthState buildCognitoAuthState(UserState state) throws AuthException {
+        final AWSCognitoAuthState.Builder sessionBuilder;
+        final CountDownLatch latch = new CountDownLatch(2);
+
+        sessionBuilder = AWSCognitoAuthState.builder()
+            .signedInStatus(
+                state.equals(UserState.SIGNED_IN)
+                    ? AuthSignedInStatus.SIGNED_IN
+                    : AuthSignedInStatus.GUEST
+            )
+            .identityId(AWSMobileClient.getInstance().getIdentityId());
+
+        AWSMobileClient.getInstance().getTokens(new Callback<Tokens>() {
+            @Override
+            public void onResult(Tokens result) {
+                sessionBuilder.accessToken(result.getAccessToken().getTokenString());
+                sessionBuilder.idToken(result.getIdToken().getTokenString());
+                sessionBuilder.refreshToken(result.getRefreshToken().getTokenString());
+                latch.countDown();
+            }
+
+            @Override
+            public void onError(Exception exception) {
+                latch.countDown();
+            }
+        });
+
+        AWSMobileClient.getInstance().getAWSCredentials(new Callback<AWSCredentials>() {
+            @Override
+            public void onResult(AWSCredentials result) {
+                sessionBuilder.awsCredentials(result);
+                latch.countDown();
+            }
+
+            @Override
+            public void onError(Exception exception) {
+                latch.countDown();
+            }
+        });
+
+        try {
+            if (latch.await(SECONDS_BEFORE_TIMEOUT, TimeUnit.SECONDS)) {
+                return sessionBuilder.build();
+            } else {
+                throw new AuthException(
+                    "Failed to retrieve auth details within " + SECONDS_BEFORE_TIMEOUT + " seconds",
+                    "Check network connectivity"
+                );
+            }
+        } catch (InterruptedException exception) {
+            throw new AuthException(
+                    "Failed to retrieve auth details",
+                    exception,
+                    "See attached exception for more details"
+            );
+        }
     }
 }
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -100,11 +100,11 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
     @Override
     public void signUp(
-            String username,
-            String password,
-            AuthSignUpOptions options,
-            final Consumer<AuthSignUpResult> onSuccess,
-            final Consumer<AuthException> onException
+            @NonNull String username,
+            @NonNull String password,
+            @NonNull AuthSignUpOptions options,
+            @NonNull final Consumer<AuthSignUpResult> onSuccess,
+            @NonNull final Consumer<AuthException> onException
     ) {
         AWSMobileClient.getInstance().signUp(
             username,
@@ -121,7 +121,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                         details != null
                             ? new AuthCodeDeliveryDetails(
                                 details.getDestination(),
-                                details.getDeliveryMedium(),
+                                AuthCodeDeliveryDetails.DeliveryMedium.getEnum(details.getDeliveryMedium()),
                                 details.getAttributeName()
                             )
                             : null
@@ -140,10 +140,10 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
     @Override
     public void confirmSignUp(
-        String username,
-        String code,
-        final Consumer<AuthSignUpResult> onSuccess,
-        final Consumer<AuthException> onException
+        @NonNull String username,
+        @NonNull String code,
+        @NonNull final Consumer<AuthSignUpResult> onSuccess,
+        @NonNull final Consumer<AuthException> onException
     ) {
         AWSMobileClient.getInstance().confirmSignUp(username, code, new Callback<SignUpResult>() {
             @Override
@@ -155,7 +155,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                     details != null
                         ? new AuthCodeDeliveryDetails(
                             details.getDestination(),
-                            details.getDeliveryMedium(),
+                            AuthCodeDeliveryDetails.DeliveryMedium.getEnum(details.getDeliveryMedium()),
                             details.getAttributeName()
                         )
                         : null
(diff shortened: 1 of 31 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -0,0 +1,430 @@
...
+                                    null,
+                                    null
+                            );
+
+                            AWSMobileClient.getInstance().getTokens(new Callback<Tokens>() {
+                                @Override
+                                public void onResult(Tokens result) {
+                                    session.setAccessToken(result.getAccessToken().getTokenString());
+                                    session.setIdToken(result.getIdToken().getTokenString());
+                                    session.setRefreshToken(result.getRefreshToken().getTokenString());
```
jamesonwilliams: Maybe you want a utility method where you pass in a `Tokens`, and get back a fully populated `AWSCognitoAuthSession ` object. I guess, `AWSCognitoAuthSessionFactory` would be a good name for it?

```
...
    public static AWSCognitoAuthSession fromTokens(Tokens tokens) {
       AWSCognitoAuthSession session = new AWSCognitoAuthSession();
       ... set fields
       return session;
    }
```

This would be economical from a software craft standpoint if you will use the utility ~3+ times, probably?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
```diff
@@ -0,0 +1,97 @@
...
+        super(state);
+    }
+
+    public AWSCognitoAuthSession(
+            @NonNull AuthUserState state,
+            @Nullable AWSCredentials awsCredentials,
+            @Nullable String identityId,
+            @Nullable String accessToken,
+            @Nullable String idToken,
+            @Nullable String refreshToken
```
jamesonwilliams: If all of these things can be null, how do we get a predictable experience for the user? Right now, they'll have to null check anything they get back from a getter on this object. Can we make fewer of these things null?

Alternately, we could provided `hasFoo()`, methods, to help a little:
```
if (session.hasIdentityId()) {
    String idnetityId = session.getIdnetitytId();
}
```

Or, a functional style:
````
session.ifPresent(SessionToken.REFRESH, /* Action */ () -> {
   // This gets executed if there is a session token.
});
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public String getIdToken() {
+        return idToken;
+    }
+
+    public String getRefreshToken() {
+        return refreshToken;
+    }
+
+    public void setAwsCredentials(AWSCredentials awsCredentials) {
```
jamesonwilliams: Instead of mutation the session object, we might like to decouple these token/credentials.

You can "defer" the resolution of these things by using a provider delegate.

```
final class ImmutableSession {
    ImmutableSession(final Provider<AWSCredentials> credentialsProvider, final Provider<String> identityIdProvider) {
    }
```

This way, the session object isn't changing. But, the functions that are passed in to resolve fields may return different results.

```
new WhateverSession(/* provider */ () -> awsCredentials, /* provider */ () -> identityId);
```
Or maybe there's some implementation that's real crazy, not just returning a variable.

This is the pattern of the AWS credentials provider.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -0,0 +1,81 @@
...
+        private Map<String, String> userAttributes;
+        private Map<String, String> validationData;
+
+        public Builder() {
+            this.userAttributes = new HashMap<>();
+            this.validationData = new HashMap<>();
+        }
+
+        @NonNull
+        public Builder userAttributes(@NonNull Map<String, String> userAttributes) {
```
jamesonwilliams: This is like the Analytics API. When you provide a map, the user can do stuff like:
```
Map<String, String> userAttributes = new HashMap<>();
map.put("emailZZ", "jameson@domain.tld");
```

And they won't realize they mispelled anything until their app is out in the field and it fails at runtime.

So, data modeling can help.

`Map<String, Object>` is basically a `class`, where you name the field members in keys, and populate their values in values.

`Map<String, String>` is more or less the same thing, but with less intricate data types.

For example,

```
Map<String, String> userAttributes = new HashMap<>();
enum AttributeKeys {
    EMAIL("email"),
    NAME("name"); // .... etc.. ..
}
interface Whatever {
    // By the way, please make sure that, by convention, you use AttributeKeys!!!
    void useAttributes(Map<String, String> userAttributes);
}
```
Well, okay. But we could have done:
```
final class UserAttributes {
    private Email email;
    private Name name;
   
    UserAttributes(Email email, Name name) {
       this.email = email;
       this.name = name;
    }

    static final class Email {
         private final String validEmail;
         private Email(String validEmail) {
            this.validEmail = validEmail; 
         }
         public static Email parse(String anyString) {
             // Emails have a defined format with TLD at the end, an @ in the middle, etc....
             // Does this string?
             return new Email(/* validated */ anyString);
         }
         public String getUserName() {
             // reutrn the part before at symbol
         }
         public String getDomainName() {
             // return portion after at...
         }
    }

    static final class Name {
        private final String validatedName;
        private Name(String validatedName) {
            this.validatedName = validatedName);
        }
   
        public static Name parse(@Nullable String fullName) {
            // Validate the format of the string is actually a name....
            // For sake of argument, let's require a first and last name.
            return new Name(/*valid! */ string);
        }

        public String getFirstName() {
        
    }
}
```

Okay so anyway, my point is, that data modeling created a semantic experience for a user, that is a lot more fool proof. The user would make this stuff like:

````
// This can be a step builder, to catch all the required fields, and get that stuff at runtime.
UserAttributes.builder()
    .email(Email.parse("tony@koolbiz.uk"))
    .name(Name.parse("Tony Daniels"))
    .build();
````

Compare with:
```
Map<String, String> attrs = new HashMap<>();
attrs.put("Eamil", "whoops");
attrs.put("Eamil", "whoopsOverwrite");
```

Alot more of the stuff is left to reading the docs carefully and then validating at runtime.

File: core/src/main/java/com/amplifyframework/auth/AuthException.java [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Creates a new exception with a message, root cause, and recovery suggestion.
+     * @param message An error message describing why this exception was thrown
+     * @param throwable The underlying cause of this exception
+     * @param recoverySuggestion Text suggesting a way to recover from the error being described
+     */
+    public AuthException(
+            @NonNull final String message,
+            final Throwable throwable,
```
jamesonwilliams: Naming, `cause`

core/src/main/java/com/amplifyframework/auth/AuthCodeDeliveryDetails.java
```diff
@@ -0,0 +1,40 @@
...
+package com.amplifyframework.auth;
+
+public final class AuthCodeDeliveryDetails {
+    private String destination;
+    private String deliveryMedium;
+    private String attributeName;
+
+    public AuthCodeDeliveryDetails(String destination, String deliveryMedium, String attributeName) {
+        this.destination = destination;
+        this.deliveryMedium = deliveryMedium;
```
jamesonwilliams: Can this be an enum? Which are the different delivery mediums? Are they known to us? Do we only support a constrained set of them?

core/src/main/java/com/amplifyframework/auth/AuthCategoryConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+public final class AuthCategoryConfiguration extends CategoryConfiguration {
+    // Any category level properties would be defined here and populateFromJson would be overridden
+    // below to fill in these values from the JSON data.
+
+    /**
+     * Constructs a new AuthCategoryConfiguration.
+     */
+    public AuthCategoryConfiguration() {
+        super();
+    }
```
jamesonwilliams: This is the  behavior of the default constructor in Java, if you don't supply one 

That is to say, you could safely remove this.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -0,0 +1,146 @@
...
+    void handleSignInWithUIResponse(Intent intent);
+
+    /**
+     * TODO: Replace this with a generic sign in with social method (or whatever naming lands on)
+     *       which takes the provider as an enum.
+     * @param token Token retrieved from the social provider's authentication code.
+     * @param onSuccess Success callback
+     * @param onError Error callback
+     */
+    void signInWithFacebook(
```
jamesonwilliams: Why does Facebook get to be special?

We should figure out a way to spec out this API without naming particular providers.

Even this woudl be more extensible:
```
signInWith(SignInProvider provider, ...);
```
with:
```
enum SignInProvider {
    FACEBOOK; // For not, it's gotta be Facebook or nuthin, pal.
    // GOOGLE; // Someday!
}
```

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -0,0 +1,146 @@
...
+    void signInWithUI(
+            @NonNull Activity callingActivity,
+            @NonNull Consumer<String> onSuccess,
+            @NonNull Consumer<AmplifyException> onError);
+
+    /**
+     * Handles the response which comes back from {@link #signInWithUI(Activity, Consumer, Consumer)}.
+     * @param intent The app activity's intent
+     */
+    void handleSignInWithUIResponse(Intent intent);
```
jamesonwilliams: What happens on error?

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Retrieve the user's current state - Signed Out, Guest, or Signed In.
+     * Depending on how a plugin implements this, the resulting AuthSession can also be cast to a type specific
+     * to that plugin which contains the various security tokens as an escape hatch if you want to manually use
+     * them outside the plugin. Within Amplify this should not be needed as the other categories will automatically
+     * work as long as you are signed in.
+     * @param onSuccess Success callback
+     * @param onError Error callback
+     */
+    void currentAuthState(
```
jamesonwilliams:
```
void lookupAuthSession(Consumer<AuthSession> onSessionFound, Consumer<AuthException> onSessionUnavailable);
```

1. Start method names with a verb;
2. Match the subject of the method name to the result parameter

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: The problem with this is that I need to populate some of the AWSCognitoAuthSession object before this callback and also need to populate some of it potentially after this (depending on what order the AWSMobileClient calls come back)

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -0,0 +1,430 @@
...
+                                            ? AuthUserState.SIGNED_IN
+                                            : AuthUserState.GUEST,
+                                    null,
+                                    AWSMobileClient.getInstance().getIdentityId(),
+                                    null,
+                                    null,
+                                    null
+                            );
+
+                            AWSMobileClient.getInstance().getTokens(new Callback<Tokens>() {
```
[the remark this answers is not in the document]
TrekSoft: There's only one level of nesting here right? After currentUserState succeeds, I make a bunch of calls in parallel depending on the result.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
TrekSoft: This is a raw escape hatch data object where the user gets back whatever underlying AWS Credentials may or may not exist depending on the type of account and whether it's signed in or not. We can talk about it more on the call.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
TrekSoft: :P

TODO: Replace this with a generic sign in with social method (or whatever naming lands on)
            which takes the provider as an enum.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
TrekSoft: The triggering function above, signInWithUI, gets that error in it's error listener

core/src/main/java/com/amplifyframework/auth/AuthCategoryConfiguration.java
TrekSoft: Ok, this was copying existing code from another category so I guess I can fix that across all of them.

core/src/main/java/com/amplifyframework/auth/AuthCodeDeliveryDetails.java
TrekSoft: The answers to your last two questions are both no since AWSMobileClient just returns it as a string. But I guess we could do an enum with the methods we currently know and an unknown value and add more values as needed.

core/src/main/java/com/amplifyframework/auth/AuthException.java
TrekSoft: We'd have to rename this for all the exceptions across the platform - do you think it's worth it?

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @NonNull
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static final class Builder {
+        private Map<String, String> userAttributes;
+        private Map<String, String> validationData;
```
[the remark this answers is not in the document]
TrekSoft: Sounds good

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: So the thing with this is that it's intentionally left open because the customer needs to be able to pass a map of any key/value pairs. There's no limit on what they could be.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -0,0 +1,430 @@
...
+    public String getPluginKey() {
+        return AWS_COGNITO_AUTH_PLUGIN_KEY;
+    }
+
+    @Override
+    public void configure(
+            @NonNull JSONObject pluginConfiguration,
+            @NonNull Context context
+    ) throws AmplifyException {
+        final CountDownLatch latch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
TrekSoft: Alright I'll find out more about this from you on our call

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -0,0 +1,430 @@
...
+            AuthSignUpOptions options,
+            final Consumer<AuthSignUpResult> onSuccess,
+            final Consumer<AuthException> onException
+    ) {
+        AWSMobileClient.getInstance().signUp(
+            username,
+            password,
+            options.getUserAttributes(),
+            options.getValidationData(),
+            new Callback<SignUpResult>() {
```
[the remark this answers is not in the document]
TrekSoft: Hmm, I see what you're saying but I think especially since this is only a temporary implementation of the plugin wrapping AWSMobileClient for GA and the plan is to do a fast follow with a completely fresh new Auth implementation, it probably isn't worth doing now.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -0,0 +1,430 @@
...
+    ) {
+        try {
+            AWSMobileClient.getInstance().currentUserState(new Callback<UserStateDetails>() {
+                @Override
+                public void onResult(UserStateDetails result) {
+                    UserState state = result.getUserState();
+
+                    if (state.equals(UserState.SIGNED_OUT) ||
+                            state.equals(UserState.SIGNED_OUT_FEDERATED_TOKENS_INVALID) ||
+                            state.equals(UserState.SIGNED_OUT_USER_POOLS_TOKENS_INVALID)
```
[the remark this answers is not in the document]
TrekSoft: Ok

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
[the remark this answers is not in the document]
TrekSoft: And while there are a couple extra lines of code added this way vs. the way you recommend, it's actually faster to write because the IDE autocompletes it.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
TrekSoft: Add a builder to make it immutable and comment on the reason some things can be null

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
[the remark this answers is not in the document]
TrekSoft: Discussed in the call that given this is calling AWSMobileClient's configure method it makes sense to leave this here. Were going to try using the Await class but it's timeout error handling isn't suitable for this so leaving it as is.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -0,0 +1,128 @@
...
+        public Builder validationData(@NonNull Map<String, String> validationData) {
+            Objects.requireNonNull(validationData);
+            this.validationData.clear();
+            this.validationData.putAll(validationData);
+            return this;
+        }
+
+        @NonNull
+        public AuthSignUpOptions build() {
+            return new AuthSignUpOptions(userAttributes, validationData);
```
jamesonwilliams: Somewhere around here is a good opportunity to wrap these in `Immutable.of(...)`, so that they are, int he created object.

core/src/main/java/com/amplifyframework/auth/options/AuthSignInOptions.java
```diff
@@ -0,0 +1,99 @@
...
+        public Builder validationData(@NonNull Map<String, String> validationData) {
+            Objects.requireNonNull(validationData);
+            this.validationData.clear();
+            this.validationData.putAll(validationData);
+            return this;
+        }
+
+        @NonNull
+        public AuthSignInOptions build() {
+            return new AuthSignInOptions(validationData);
```
jamesonwilliams: `Immutable.of(validationData)`

core/src/main/java/com/amplifyframework/auth/AuthSignedInStatus.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.amplifyframework.auth;
+
+public enum AuthSignedInStatus {
+    /** User is signed in. **/
+    SIGNED_IN,
+    /** User is signed out. **/
+    SIGNED_OUT,
+    /** User is signed in as a guest. **/
```
jamesonwilliams: Or sign in has not been attempted?

core/src/main/java/com/amplifyframework/auth/AuthException.java
jamesonwilliams: I do, yea. We should fix the other categories in a separate PR.

core/src/main/java/com/amplifyframework/auth/AuthCodeDeliveryDetails.java
jamesonwilliams: Jithin spec'd out an enum for this - you could check the iOS PR / sync with him.

royjit: Email and SMS are the two values added in iOS which are also supported by [Cognito](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CodeDeliveryDetailsType.html#CognitoUserPools-Type-CodeDeliveryDetailsType-Destination).

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -0,0 +1,146 @@
...
+     * Launch a pre-built sign in UI flow. You should also put the {@link #handleSignInWithUIResponse(Intent)} method in
+     * your activity's onResume method to capture the response which comes back from the UI flow.
+     * @param callingActivity The activity in your app you are calling this from
+     * @param onSuccess Success callback
+     * @param onError Error callback
+     */
+    void signInWithUI(
+            @NonNull Activity callingActivity,
+            @NonNull Consumer<String> onSuccess,
+            @NonNull Consumer<AmplifyException> onError);
```
royjit, line 114: Shouldnt we return the same result here as well? AuthSignInResult?

TrekSoft: Yeah this is not the final method signature - it's still from the POC

core/src/main/java/com/amplifyframework/auth/AuthSignedInStatus.java
TrekSoft: No if sign in hasn't been attempted they'll always be in SIGNED_OUT state - you have to explicitly sign in as guest.

TrekSoft:
@jamesonwilliams I talked with @jpignata about the subject of tests and he agreed we wouldn't include tests at this point because it's still subject to change pending the ongoing architecture decisions. Once those are finalized, I'll add in tests.

core/src/main/java/com/amplifyframework/auth/AuthCodeDeliveryDetails.java
TrekSoft: Ah yeah whoops - I forgot about adding this - will do.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
jamesonwilliams: > The triggering function above, signInWithUI, gets that error in it's error listener

Should add this to doco

TrekSoft:
@jamesonwilliams - yes there's passwordless auth and custom auth which is why those are nullable

PR #332, opened 2020-03-26
Temporarily ignore broken test

raphkim:
*Issue #, if available:*

*Description of changes:*
Ignore `AWSDataStorePluginInstrumentedTest`, which seems to break circleCI consistently at the moment. Re-enable upon fix.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
@@ -89,13 +89,13 @@ public final class AWSDataStorePluginInstrumentedTest {
     public void blogOwnerSavedIntoDataStoreIsThenQueriableInRemoteAppSyncApi() throws DataStoreException, ApiException {
         // Start listening for model publication events on the Hub.
         HubAccumulator outboundModelEventAccumulator =
-                HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.PUBLISHED_TO_CLOUD);
+            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.PUBLISHED_TO_CLOUD);
         outboundModelEventAccumulator.clear().start();
 
         // Save Charley Crockett, a guy who has a blog, into the DataStore.
         BlogOwner localCharley = BlogOwner.builder()
-                .name("Charley Crockett")
-                .build();
+            .name("Charley Crockett")
+            .build();
         dataStore.save(localCharley);
 
         // Wait for a Hub event telling us that our Charley model got published to the cloud.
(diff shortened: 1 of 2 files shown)


raphkim:
ah it seems android studio did some fun stuff with the whitespace.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
```diff
@@ -46,6 +47,7 @@
  * which were defined by the schema in:
  * testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/schema.graphql.
  */
+@Ignore("Currently does not pass.") // TODO: fix the error; re-enable test
```
jamesonwilliams: Can you move this to just the specific `@Test` that's failing? Having this at the class level will prevent some well-functioning tests from being exercised.

raphkim: When i tried just ignoring the `blogOwnerSavedIntoDataStoreIsThenQueriableInRemoteAppSyncApi`, the other test also failed. I think this test failing initially crashes the whole test suite, which made us unable to catch it last time.

core/src/androidTest/java/com/amplifyframework/hub/HubInstrumentedTest.java
```diff
@@ -43,7 +43,7 @@
  */
 public final class HubInstrumentedTest {
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:core:test");
-    private static final long SUBSCRIPTION_RECEIVE_TIMEOUT_MS = 200;
+    private static final long SUBSCRIPTION_RECEIVE_TIMEOUT_MS = 300;
```
jamesonwilliams: I had tried changing this to `TimeUnit.SECONDS.toMilliseconds(1)` -- a full second -- and the test still failed in the same way. Unfortunately, I don't think this will solve the problem.

raphkim: Hmm that's weird. It passes consistently with 200ms locally, and it seems to have passed with 300ms in circleCI

jamesonwilliams: It  always passes at 200ms locally for me too.

When I had it at 1,000ms, it continued to fail for me (most of the time) when I had a PR up for it.

It looked like - after 1000ms - only two of the messages were received, instead of 10.

Even assuming that the publish/receive flows were all performed serially, 100ms would be a _horridly_ long time to pass off an in-memory data structure.

So anyway, I think the problem is elsewhere. Probably around threads available to the test process in the test farm?
Something like that? Or perhaps a different test is polluting the state of the test environment, and it just happens to surface as an issue while this test runs?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
jamesonwilliams: Hm, alright. We can come back to it.

PR #333, opened 2020-03-26
Make plugin configuration nullable

raphkim:
*Issue #, if available:*

*Description of changes:*
Make plugin configuration nullable and force the implementation to null-check if it is required

### DIFF
--- a/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
+++ b/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
@@ -305,7 +305,17 @@ public final class AmazonPinpointAnalyticsPlugin extends AnalyticsPlugin<Object>
      * {@inheritDoc}
      */
     @Override
-    public void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context context) throws AnalyticsException {
+    public void configure(
+            @Nullable JSONObject pluginConfiguration,
+            @NonNull Context context
+    ) throws AnalyticsException {
+        if (pluginConfiguration == null) {
+            throw new AnalyticsException(
+                    "Missing configuration for " + getPluginKey(),
+                    "Check amplifyconfiguration.json to make sure that there is a section for " +
+                            getPluginKey() + " under the analytics category."
+            );
+        }
 
         AmazonPinpointAnalyticsPluginConfiguration.Builder configurationBuilder =
                 AmazonPinpointAnalyticsPluginConfiguration.builder();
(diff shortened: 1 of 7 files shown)


aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AmazonPinpointAnalyticsPlugin.java
```diff
@@ -305,7 +305,17 @@ public String getPluginKey() {
      * {@inheritDoc}
      */
     @Override
-    public void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context context) throws AnalyticsException {
+    public void configure(
+            @Nullable JSONObject pluginConfiguration,
+            @NonNull Context context
+    ) throws AnalyticsException {
+        if (pluginConfiguration == null) {
+            throw new AnalyticsException(
```
jamesonwilliams: Hm, I think this means the annotation should say `@NonNull`. If you throw an exception whenever you see a null param, it can't really be null.

(In Java, any object _can_ be null, but the annotation means what it _ought to_ be, for the software to work as designed.)

core/src/main/java/com/amplifyframework/core/plugin/Plugin.java
```diff
@@ -54,14 +54,17 @@
...
+     * the documentation for details.
+     *
      * This hook provides a good opportunity to instantiate resources.
      * Any long-lived initialization should take place in {@link #initialize(Context)}, instead.
      * @param pluginConfiguration plugin-specific configuration data
      * @param context An Android Context
      * @throws AmplifyException an error is encountered during configuration.
      */
-    void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context context) throws AmplifyException;
+    void configure(@Nullable JSONObject pluginConfiguration, @NonNull Context context) throws AmplifyException;
```
jamesonwilliams: Can you remove `@Nullable` from this? Or, possible write it like:
```
void configure(/* maybe-@NonNull */ JSONObject pluginConfiguration ...)
```

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -107,7 +108,7 @@ public String getPluginKey() {
 
     @Override
     @SuppressWarnings("MagicNumber") // TODO: Remove once default values are moved to configuration
-    public void configure(@NonNull JSONObject pluginConfiguration, @NonNull Context context) throws StorageException {
+    public void configure(@Nullable JSONObject pluginConfiguration, @NonNull Context context) throws StorageException {
```
jamesonwilliams: It isn't; passing a null value here will result in a `StorageException`. This should say `@NonNull`.

raphkim:
`@NonNull` annotation implies that the parameter was already checked to be not null prior to being passed as a parameter, which means that it is inappropriate to use in our situation. When category calls for configuration of its registered plugins, it does not do any null-checking on the plugin configuration.

I figured that it would be best to just leave it not annotated in all places.

Issue #216, opened 2020-01-02
GsonVariablesSerializer should support ISO 8601 format for Date type


shiroyama:
My GraphQL schema uses `createdAt: AWSDateTime!` type, and the auto-generated POJO class has a `java.util.Date` type for `createdAt` field.
When Amplify Android try to make a GraphQL query, this POJO is converted to a query string with using `com.amplifyframework.api.aws.GsonVariablesSerializer` by default which converts `java.util.Date` to `yyyy-MM-dd` format transparently.

```
public final class GsonVariablesSerializer implements GraphQLRequest.VariablesSerializer {
    @Override
    public String serialize(Map<String, Object> variables) {
        return new GsonBuilder()
                .registerTypeAdapter(Date.class, new DateSerializer())
                .create()
                .toJson(variables);
    }

    class DateSerializer implements JsonSerializer<Date> {
        public JsonElement serialize(Date date, Type typeOfSrc, JsonSerializationContext context) {
            DateFormat df = new SimpleDateFormat("yyyy-MM-dd", Locale.getDefault());
            return new JsonPrimitive(df.format(new Date()));
        }
    }
}
```

I believe AWSDateTime is ISO 8601 compliant according to the document, so this should be `"yyyy-MM-dd'T'HH:mm:ssXXX"` in order to keep the degree of precision.
https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

To work around this, I know I can make `GraphQLRequest` manually as follows:

```
// make your own Serializer
class GsonVariablesSerializer : VariablesSerializer {
    override fun serialize(variables: Map<String, Any>): String {
        return GsonBuilder()
            .registerTypeAdapter(
                Date::class.java,
                DateSerializer()
            )
            .create()
            .toJson(variables)
    }

    internal inner class DateSerializer : JsonSerializer<Date?> {
        override fun serialize(
            date: Date?,
            typeOfSrc: Type,
            context: JsonSerializationContext
        ): JsonElement {
            val df: DateFormat = SimpleDateFormat(
                "yyyy-MM-dd'T'HH:mm:ssXXX",
                Locale.getDefault()
            )
            return JsonPrimitive(df.format(Date()))
        }
    }
}

// make manual request
val query = """
    mutation CreateMessage(
      ${'$'}id: ID!
      ${'$'}content: String!
      ${'$'}roomId: String!
      ${'$'}username: String!
      ${'$'}createdAt: AWSDateTime!
    ) {
[...]

PR #335, opened 2020-03-30
Fixes Date fields being converted to current date

TrekSoft:
*Issue #, if available:*
Not the main topic but was reported on issue #216 

*Description of changes:*
Currently when serializing a Date field we have a horrible bug (my bad!) where it ignores the actual value the user provided and sets it to the current date 🤦‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 A declarative library for application development using cloud services.
 
-## 0.10.0
+## 0.10.1
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 A declarative library for application development using cloud services.
 
+## 0.10.0
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 A declarative library for application development using cloud services.
 
+## 0.10.0
```
jpignata: Would this be for 0.10.1 since 0.10.0 is out already??

TrekSoft: Oof - yep that's what I meant to do. Let me update.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+
+### Bug Fixes
+
+- **API**
+  - Fixed Date fields being converted to the current date instead of their original value
+
```
raphkim: we already released `0.10.0` no? This should be `0.10.1` I believe.

TrekSoft:
Fixed

PR #336, opened 2020-03-31
Implement offline interpret

raphkim:
*Issue #, if available:*

*Description of changes:*


aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+     * @param onError notified when classification fails
+     */
+    void classify(
+            @NonNull String text,
+            @NonNull Consumer<InterpretResult> onSuccess,
+            @NonNull Consumer<PredictionsException> onError
+    ) {
+        final Sentiment sentiment;
+        try {
+            loadIfNotLoaded();
```
jamesonwilliams: The model loading in this category is analogous to the initialization work, in the DataStore.

I added some lifecycle to accomodate this kind of stuff.

`configure()` on the plugin reads the config file and instantiates some classes.

`intialize()` can be used for this kind of heavy-lifting stuff, though.

Checkout the `afterInitializion()` method in DataStore. We could externalize this utility into core. Then, both categories could use it.

Right now, the logic is like:
1. Try to load models at start up.
2. Before doing anything, try again.

And its sort of relaxed about sequencing/error handling.

With the new approach, it'd be more like:
1. Try to load model at initialization. If it fails, pass back error.
2. Defer all operations _until after_ that one initialization attempt succeeds/fails.
3. Only ever try to initialize at a single point in time during the operation of the system.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+            sentiment = fetchSentiment(text);
+        } catch (PredictionsException exception) {
+            onError.accept(exception);
+            return;
+        }
+
+        InterpretResult result = InterpretResult.builder()
+                .sentiment(sentiment)
+                .build();
+        onSuccess.accept(result);
```
jamesonwilliams: trivial -- when you have such a small amount of code for the payload, you can inline it:
```
onSuccess.accept(InterpretResult.builder()
    .sentiment(sentiment)
    .build());
```

Actually, you could probably even bring that whole thing up into the `try` block:
```
try {
    onSuccess.accept(InterpretResult.builder()
        .sentiment(fetchSentiment(text))
        .build());
} catch (PredictionsException sentimentDetectionFailure) {
    onError.accept(sentimentDetectionFailure);
}
```

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+
+        // Pre-process input text
+        try {
+            input = tokenizeInputText(text);
+            output = new float[1][labels.size()];
+        } catch (IllegalArgumentException exception) {
+            throw new PredictionsException(
+                    "Tensorflow Lite failed to make inference.",
+                    exception,
+                    "Verify that the assets are loaded."
```
jamesonwilliams: Which assets?

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+            return;
+        }
+
+        InterpretResult result = InterpretResult.builder()
+                .sentiment(sentiment)
+                .build();
+        onSuccess.accept(result);
+    }
+
+    private Sentiment fetchSentiment(String text) throws PredictionsException {
```
jamesonwilliams: should this be called `fetchPredominantSentiment`?

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+                        .value(sentimentType)
+                        .confidence(confidenceScore)
+                        .build();
+            }
+        }
+        return sentiment;
+    }
+
+    @SuppressWarnings("ConstantConditions")
+    private float[][] tokenizeInputText(String text) {
```
jamesonwilliams: This method looks like it's doing some interesting stuff. Can you add a little Javadoc above it, to orient the reader? "We need the input in format (what format?) But, it's in (what format?). So we (perform transformations?)."

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
```diff
@@ -0,0 +1,332 @@
...
+            return;
+        }
+        loadModel();
+        loadDictionary();
+        loadLabels();
+        loaded.set(true);
+    }
+
+    @WorkerThread
+    private synchronized void loadModel() throws PredictionsException {
```
jamesonwilliams: You have a number of different components that have a similar lifecycle: `load()`, `isLoaded()`.

You could make a `Loadable` interface:
```
// V = value type for thing that is loaded, e.g. a Map<String, Integer> dictionary, etc.
interface Loadable<V> {
    void load();
    void unload();
    boolean isLoaded();
    onLoaded(Consumer<V> valueConsumer);
    onUnloaded(Action unloadAction);
    @Nullable V value()
}
```

If you have somehting like this, you can externalize all of these details into small little template files and compartmentalize the load/validation logic into them.

Then the classification refers to them more like:
```
for (Loadable<?> dependency : LOADABLES) {
    if (!dependency.isLoaded()) {
        dependency.load();
    }
}
```

I guess you'd want this to happen as a down-stream result of the call-chain from the`Plugin<?>`'s `initialize()` lifecycle method.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/adapter/SentimentTypeAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    @NonNull
+    public static SentimentType fromTensorflow(@NonNull String sentiment) {
+        switch (sentiment.toLowerCase(Locale.US)) {
+            case "positive":
+                return SentimentType.POSITIVE;
+            case "negative":
+                return SentimentType.NEGATIVE;
+            default:
+                return SentimentType.UNKNOWN;
```
jamesonwilliams: In software, there's always a choice, between "Fail fast" and "degrade gracefully."

With exception handling, sometimes we see this as "bubble up the error," versus. "return null."

With object construction - here, this enum factory - we have the same sort of thing to consider.  We can either:
1. Throw an error if the input is unknown/invalid. Perhaps, `InvalidArgumentException`.
2. Degrade to an `UNKNOWN`, type, like you are.

I think the key thing to consider is this: Is the system useful/usable when the value is `UNKNOWN`? If not, maybe we should "fail fast."

raphkim: I think it is useful to return unknown if the model decided that it ran its course successfully and was ultimately unable to determine the specific value for a feature. It will serve as a "value" of feature itself.

build.gradle
```diff
@@ -20,7 +20,8 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.1'
+        classpath 'com.android.tools.build:gradle:3.6.0'
+        classpath 'me.tatarka:gradle-retrolambda:3.7.1'
```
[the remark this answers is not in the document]
raphkim: :(

build.gradle
```diff
@@ -44,7 +45,7 @@ task clean(type: Delete) {
 ext {
     buildToolsVersion = "29.0.2"
     compileSdkVersion = 29
-    minSdkVersion = 16
+    minSdkVersion = 15
```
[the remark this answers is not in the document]
raphkim: :(

settings.gradle
```diff
@@ -18,13 +18,13 @@ include ':core'
 // Plugin Modules
 include ':aws-analytics-pinpoint'
 include ':aws-api'
-include ':aws-storage-s3'
 include ':aws-datastore'
+include ':aws-predictions-tfl'
```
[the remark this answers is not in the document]
raphkim: haha I didn't give much thought into this to be honest. I can go with your suggestion

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/adapter/SentimentTypeAdapter.java
jamesonwilliams: Ah okay, yes! I agree. This is an important distinction. Unknown as an output of the prediction, versus unknown due to not having results from a prediction.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/TFLitePredictionsEscapeHatch.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.tensorflow.lite.Interpreter;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * An escape hatch to give low-level access to Tensorflow interpreter.
+ */
+public final class TFLitePredictionsEscapeHatch {
```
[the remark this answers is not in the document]
raphkim: yea I did this because we would need to return multiple types of AWS clients in the AWS plugin. If we directly use a map instead of encapsulating it, then it would need to be `Map<String, AmazonWebServiceClient>`, which I think is pretty ugly.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/TFLitePredictionsEscapeHatch.java
```diff
@@ -0,0 +1,51 @@
...
+    private final Map<String, Interpreter> interpreters;
+
+    TFLitePredictionsEscapeHatch(@NonNull Map<String, Interpreter> interpreters) {
+        this.interpreters = new TreeMap<>();
```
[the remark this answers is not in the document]
raphkim: It sorts the key haha. Definitely no good reason besides that. I can just use a `HashMap` I guess

aws-predictions-tfl/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    http://aws.amazon.com/apache2.0
+
+   or in the "license" file accompanying this file. This file is distributed
+   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+   express or implied. See the License for the specific language governing
+   permissions and limitations under the License.
+-->
+
+<manifest package="com.amplifyframework.predictions" />
```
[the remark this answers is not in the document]
raphkim: This is actually a good point! I have `adapter` package in each module (AWS and this), and the overlapping namespace was giving me trouble haha

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
raphkim: the point is to relax the error handling. This is because models are optional for this plugin, which means that this plugin needs to successfully configure itself even if all of the models are missing (and therefore is useless). 

Let's look at this scenario:
- I want to detect sentiment
  - I want an accurate result for this particularly, so I want the result to be union of both offline and online results
- I want to translate text
  - memory is precious, so I'll only use the cloud resource for this
- I set network mode to `auto` so that it will combine result if both results are available
- If offline mode doesn't deliver (due to missing model or something), then only online result should be delivered

This use case is achieved by only downloading the required TFLite model for sentiment detection. Both online plugin and offline plugin needs to configure itself upon `Amplify.configure()`. We can't just crash the offline plugin or even leave it hanging at this point waiting for successful load of every model. The configuration needs to go through so that when the customer decides to:
1. call sentiment detection: both online and offline plugins must return result
2. call translate: online plugin returns result, but offline plugin may fail

My solution was to try loading at initialization but ignore error to allow configuration to succeed. That way, we can actually handle each API and their respective model requirement separately.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/TFLitePredictionsPlugin.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @NonNull
+    @Override
+    public InterpretOperation<?> interpret(
+            @NonNull String text,
+            @NonNull Consumer<InterpretResult> onSuccess,
+            @NonNull Consumer<PredictionsException> onError
+    ) {
+        final InterpretOptions options = InterpretOptions.defaultInstance();
```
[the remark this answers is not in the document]
raphkim: I think name convention like this should be addressed as a whole over the entire project in a different PR :) I'll leave it as is for now, but I agree that a shorter name would have made it much cleaner.

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/adapter/SentimentTypeAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+package com.amplifyframework.predictions.adapter;
+
+import androidx.annotation.NonNull;
+
+import com.amplifyframework.predictions.models.SentimentType;
+
+import java.util.Locale;
+
+/**
+ * Utility to convert third-party {@link SentimentType} equivalent
```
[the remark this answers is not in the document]
raphkim: hmm, i think the wording itself is clear, but the fact that I linked the word made it seem confusing. Will fix

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
raphkim: eh, it's private method so I don't think it matters

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
raphkim: It's actually the wrong message here! Thanks for catching it :)

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
raphkim: It's some magic code from their official sample code... that's all I can say

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLitePredictionsService.java
```diff
@@ -0,0 +1,75 @@
...
+            @NonNull Consumer<InterpretResult> onSuccess,
+            @NonNull Consumer<PredictionsException> onError
+    ) {
+        textClassificationService.classify(text, onSuccess, onError);
+    }
+
+    /**
+     * Free up resources used by Tensorflow lite.
+     */
+    public void close() {
```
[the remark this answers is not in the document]
raphkim: `terminate` should do

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/request/TFLiteTextClassificationRequest.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public final class TFLiteTextClassificationRequest {
+    private final String text;
+
+    /**
+     * Constructs an instance of {@link TFLiteTextClassificationRequest}.
+     * @param text the text to classify
+     */
+    public TFLiteTextClassificationRequest(@NonNull String text) {
+        this.text = Objects.requireNonNull(text);
```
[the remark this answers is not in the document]
raphkim: Well, since `text` isn't verified to be non null at plugin level, I figured here would be the good place to assert this

aws-predictions-tfl/src/main/java/com/amplifyframework/predictions/service/TFLiteTextClassificationService.java
raphkim: addressed

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/TensorFlowPredictionsPlugin.java
```diff
@@ -0,0 +1,115 @@
...
+ * TensorFlow Lite to carry out tasks offline.
+ */
+public final class TensorFlowPredictionsPlugin extends PredictionsPlugin<TensorFlowPredictionsEscapeHatch> {
+    private static final String TFL_PREDICTIONS_PLUGIN_KEY = "tflPredictionsPlugin";
+
+    private final ExecutorService executorService;
+
+    private TensorFlowPredictionsService predictionsService;
+
+    public TensorFlowPredictionsPlugin() {
```
jamesonwilliams, line 50: Should get a missing Javadoc warning from Checkstyle on this, no?

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/TensorFlowPredictionsPlugin.java
```diff
@@ -0,0 +1,115 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * A plugin for Predictions category that uses models from
+ * TensorFlow Lite to carry out tasks offline.
+ */
+public final class TensorFlowPredictionsPlugin extends PredictionsPlugin<TensorFlowPredictionsEscapeHatch> {
+    private static final String TFL_PREDICTIONS_PLUGIN_KEY = "tflPredictionsPlugin";
```
jamesonwilliams, line 44: `"tensorFlowPrecitionsPlugin"`?

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/TensorFlowPredictionsPlugin.java
```diff
@@ -0,0 +1,115 @@
...
+    public String getPluginKey() {
+        return TFL_PREDICTIONS_PLUGIN_KEY;
+    }
+
+    @Override
+    public void configure(
+            JSONObject pluginConfiguration,
+            @NonNull Context context
+    ) throws AmplifyException {
+        this.predictionsService = new TensorFlowPredictionsService(context);
```
jamesonwilliams, line 65: As per other PR, you could consider using a `ContextProvider` `@FunctionalInterface` to work around the ordering of object construction, e.g.
```
public final class TensorFlowPredictionsPlugin {
    public TensortFlowPredictionsPlugin() {
        this.predictionsService = new TensorFlowPredictionsService(() -> context);   
    }
}
```
```
final class TensorFlowPredictionsService {
    private final ContextResovler contextResolver;
    TensorFlowPredictionsService(ContextResolver contextResolver) {
        this.contextResovler = contextResovler;
    }

    public void someCategoryBehaviorImpl() {
        // Oh, we need it now ...
        Context resolvedContext = contextResolver.resolveContext();
    }

    @FunctionalItnerface interface ContextResolver {
        Context resolveContext();
    }
}
```

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/adapter/SentimentTypeAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+import com.amplifyframework.predictions.models.SentimentType;
+
+import java.util.Locale;
+
+/**
+ * Utility to convert third-party sentiment feature equivalent
+ * into Amplify-compatible data structure (i.e. {@link SentimentType}).
+ */
+public final class SentimentTypeAdapter {
+    @SuppressWarnings("checkstyle:all") private SentimentTypeAdapter() {}
```
jamesonwilliams, line 29: (Suppression not needed anymore)

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/Loadable.java
```diff
@@ -0,0 +1,76 @@
...
+     * @return this loadable instance for chaining
+     */
+    Loadable<V, E> onUnloaded(Action onUnloaded);
+
+    /**
+     * Gets the loaded value. Null if the resource was
+     * not loaded or fully loaded yet.
+     * @return the loaded value
+     */
+    @Nullable
```
jamesonwilliams, line 74: If we have `isLoaded()`, technically the caller knows all they need to. We could make this `@NonNull`, and have it throw if there's no loaded value. So then the usage would be:
```
if (loadable.isLoaded()) {
    loadable.getValue(); // This works!
} else {
    loadable.getValue(); // This throw.
}
```

ugh also, we have to start using `get` in front of accessors. Can you scrub all of your work for this, haha? I hate it, but Kotlin interop, loves it. I'm more flexible than the computer programming language, so have to go with what it wants. :-D

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/Loadable.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    Loadable<V, E> onLoaded(Consumer<V> onLoaded);
+
+    /**
+     * Sets the consumer of an error which will be triggered
+     * upon encountering an error while loading/unloading.
+     * @param onError the consumer of the thrown exception
+     * @return this loadable instance for chaining
+     */
+    Loadable<V, E> onError(Consumer<E> onError);
```
jamesonwilliams, line 59: `onLoadError(...)`, or `onErrorLoading(...)`?

(Also, if you don't actually need all of these bells and whistles, feel free to strip some out. I may have gone overboard when I initially spec'd out this `Loadable` in the last round of PR feedback ...)

Ohhh. Or, maybe we should have:
```
void onLoaded(Consumer<V> loadedValue, Consumer<E> failureToLoad);

void onUnloaded(Action unloadAction, Consumer<E> failureToUnload);
```

^^ Note, technically the unload could fail too, so we should include an error consumer for it, as well. But, by the time we have these four hooks, we probably don't want four seaprate APIs on `Loadable`. We can distill success/failure actions into just the two, though: `onLoad`, `onUnload`.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationDictionary.java
```diff
@@ -0,0 +1,247 @@
...
+    private Consumer<Map<String, Integer>> onLoaded;
+    private Consumer<PredictionsException> onError;
+    private Action onUnloaded;
+    private boolean loaded;
+
+    /**
+     * Constructs a loader for text classification dictionary.
+     * @param context the Android context
+     */
+    public TextClassificationDictionary(@NonNull Context context) {
```
jamesonwilliams, line 82: You can have this just take a dependency on Android's `Assets` directly, if that's all you need.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
```diff
@@ -0,0 +1,188 @@
...
+     * interpreter.
+     * @param context the Android context
+     */
+    TensorFlowTextClassificationService(@NonNull Context context) {
+        this.interpreter = new TextClassificationModel(context);
+        this.dictionary = new TextClassificationDictionary(context);
+        this.labels = new TextClassificationLabels(context);
+
+        this.assets = Arrays.asList(interpreter, dictionary, labels);
+        this.loaded = new CountDownLatch(assets.size());
```
jamesonwilliams, line 70: You could probablly even wrap this stuff up into its _own_ `Loadable`, a `Loadable<List<Loadable<?, PredictionsException>>, PredictionsException>`. (Lol, that type.) But basically, a loadable that is able to load a list of loadables.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
```diff
@@ -0,0 +1,188 @@
...
+    // Percentage multiplier
+    private static final int PERCENT = 100;
+
+    private final TextClassificationModel interpreter;
+    private final TextClassificationDictionary dictionary;
+    private final TextClassificationLabels labels;
+    private final List<Loadable<?, PredictionsException>> assets;
+    private final CountDownLatch loaded;
+
+    private PredictionsException loadingError;
```
jamesonwilliams, line 56: Hm, alright. Yea. You could even set an initial value for this, like `"Not yet initialized"`, and then clear it back to `null`, after intiailization was done. That way the category behaviors could throw that kind of message if need be, too, with your existing control flow.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
```diff
@@ -0,0 +1,188 @@
...
+                    "Verify that the label size matches the output size of the model."
+            );
+        }
+
+        // Find the predominant sentiment
+        Sentiment sentiment = null;
+        for (int i = 0; i < labels.size(); i++) {
+            SentimentType sentimentType = SentimentTypeAdapter.fromTensorFlow(labels.get(i));
+            float confidenceScore = output[0][i] * PERCENT;
+            if (sentiment == null || sentiment.getConfidence() < confidenceScore) {
```
jamesonwilliams, line 157: Code that is more algorithmic in nature like this is well suited for unit test coverage.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/TensorFlowPredictionsPlugin.java
raphkim: yea seems they miss some sometimes

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationDictionary.java
raphkim: I would prefer to leave it as is, since most Android developers should be more familiar with passing in an instance of `Context` than an `AssetManager`.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
raphkim: That would be unnecessary since the main idea is to block everything until after initialization. Technically the user won't be able to reach the state in question.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
raphkim: Seems like an overkill to me haha

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
raphkim: Looking back at the code, that would actually break the flow. It checks first to see if `loadingError` is null to escape early if it has attempted loading and failed. THEN it waits for load to finish.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
raphkim: I'll make a separate PR addressing both unit and integration tests for this module. For now, I will prioritize completing the AWS plugin implementation.

jamesonwilliams: (TLDR, "Okay, please write the tests soon!")

--------------

In the recent work on Auth, one of our developers decided to split up his work into different PRs for source and test. We could potentially use that same model for this Predictions, too. But, I just want to be sure there is a concrete plan to follow up with tests _next_.

It is a best practice is to build unit/component tests around the same time as the source code. The developer is most familiar with the source code at this time, and can catch the most bugs. This is also the most _frugal_ time to write tests, for the same reason (no context switching.)

Personally, I get _worried_ when I write too much source code without unit/component coverage. My memory happens to be slightly worse than the average person. Regardless, I _know_ that I made (at least a few!) mistakes, or held some false assumptions, forgot edge cases, etc. Software is _complex_, and it becomes unmanageable very very quickly without guardrails. At a certain point, the complexity becomes overwhelming and I'm no longer able to see how a code base functions. That's a very scary place to be - as an engineer, as a stakeholder in a software business. Unit/component tests are one of best tools I have personally found to help define functional baselines between a project's layered components.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/Loadable.java
```diff
@@ -44,33 +44,29 @@
...
-    Loadable<V, E> onLoaded(Consumer<V> onLoaded);
-
-    /**
-     * Sets the consumer of an error which will be triggered
-     * upon encountering an error while loading/unloading.
-     * @param onError the consumer of the thrown exception
-     * @return this loadable instance for chaining
-     */
-    Loadable<V, E> onError(Consumer<E> onError);
+    Loadable<V, E> onLoaded(Consumer<V> onLoaded, Consumer<E> onLoadError);
```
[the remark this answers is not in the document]
raphkim: E already extends `Exception`

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationDictionary.java
```diff
@@ -0,0 +1,247 @@
...
+ * <pre>
+ * &#60;PAD&#62; 0
+ * &#60;START&#62; 1
+ * &#60;UNKNOWN&#62; 2
+ * word 3
+ * word 4
+ * ...
+ * </pre>
+ */
+public class TextClassificationDictionary implements Loadable<Map<String, Integer>, PredictionsException> {
```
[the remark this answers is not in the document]
raphkim: oops good catch

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/TensorFlowPredictionsEscapeHatch.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    /**
+     * Return a map of pre-trained TensorFlow Lite interpreters
+     * used by the plugin.
+     * @return the map of {service key -> interpreter}
+     */
+    public Map<String, Interpreter> getInterpreters() {
+        return interpreters;
```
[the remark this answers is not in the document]
raphkim: will do

PR #337, opened 2020-03-31
Improve WhitespaceAround checks

jamesonwilliams:
Utility classes are meant to have private constructors, and those
constructors should be empty. There are many such constructors around
the codebase, but they generate checkstyle warnings, which are
suppressed each time one occurs. Instead, allow empty construtors as a
general rule, in the checkstyle config.

Continue to dis-allow empty lambda bodies, but make the rationale
explicit in the configuration file. Preferred style is a no-op
implementation class which expressed the intention, through naming.

Add explicit check for colon symbol in for-each style for loop.


configuration/checkstyle-rules.xml
```diff
@@ -88,7 +88,24 @@
...
+            <property name="allowEmptyConstructors" value="true" />
+            <!--
+                 false is already the default value for
+                 WhitespaceAround.  This documents that we have this
+                 rule, intentionally. Don't use an empty lambda.  Create
+                 a subclass which expresses this semantically:
+                 EmptyConsumer.create(), etc., instead of value -> {}.
+            -->
+            <property name="allowEmptyLambdas" value="false" />
+        </module>
```
raphkim, line 108: finally! :)

PR #339, opened 2020-04-01
[aws-datastore] Persist model sync time to enable delta sync

jamesonwilliams:
When sync for a model type compeltes, save the time.  Next time a sync
attempted, the last sync time is inspected. If this value is within the
delta sync window, perform a delta sync. Otherwise, perform a base sync.


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
```diff
@@ -18,53 +18,134 @@
...
-        } catch (JSONException exception) {
-            throw new DataStoreException(
+        final Iterator<String> jsonKeys = pluginJson.keys();
+        while (jsonKeys.hasNext()) {
+            final String keyString = jsonKeys.next();
+            final ConfigKey configKey;
+            try {
+                configKey = ConfigKey.fromString(keyString);
+            } catch (IllegalArgumentException noSuchConfigKey) {
+                throw new DataStoreException(
```
TrekSoft, line 63: Hmm, it seems like this will throw an exception if it sees any key in the config it doesn't support, not just if it's missing the required keys. I think it would be better not to do this since (a) It doesn't hurt us in any way if there's extra information so why throw an error? and (b) I could see a time in the future where in transitioning between versions of our code in a non-breaking way there ends up being extraneous information in people's old config files which shouldn't be a problem but this would make into a problem.

TrekSoft: It also makes performance worse because you have to loop through the keys vs just check if it has the ones you want (not significant but still, given the above reasons, it adds fuel to the fire of removing it)

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
```diff
@@ -18,53 +18,134 @@
...
+        return new AWSDataStorePluginConfiguration(syncMode, baseSyncIntervalMs);
     }
 
     /**
      * Gets the synchronization mode.
      * @return Synchronization mode
      */
     @NonNull
-    SyncMode getSyncMode() {
+    SyncMode syncMode() {
```
TrekSoft, line 96: Is there a reason you're dropping the get verb prefix? I debated this on my current Auth work and landed on keeping the get at the beginning of the method name since, in addition to being consistent with our pattern of starting method names with verbs, I also think it's a better experience for the developer if all getter methods start with get and thus are grouped together in autosuggest.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.UUID;
+
+/**
+ * Metadata about the last time that a model class was sync'd with AppSync backend.
+ * This metadata is persisted locally as a system model. This metadata is inspected
+ * whenever the Sync Engine starts up. The system consider the value of
+ * {@link LastSyncMetadata#getLastSyncTime()} to decide whether or not it should
+ * perform a "Base Sync" or a "Delta Sync".
+ */
+public final class LastSyncMetadata implements Model {
```
TrekSoft, line 35: There's probably a good reason for this I'm just missing here but was there a reason you decided to go with a new model for this metadata vs. supplementing the existing metadata object with a lastSyncTime field?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -46,25 +46,35 @@
...
-            @NonNull final AppSync appSync) {
+            @NonNull final AppSync appSync,
+            @NonNull final SyncProcessor.BaseSyncIntervalProvider baseSyncIntervalProvider) {
         Objects.requireNonNull(modelSchemaRegistry);
         Objects.requireNonNull(modelProvider);
         Objects.requireNonNull(appSync);
         Objects.requireNonNull(localStorageAdapter);
 
         RemoteModelMutations remoteModelMutations = new RemoteModelMutations(appSync, modelProvider);
-        RemoteModelState remoteModelState = new RemoteModelState(appSync, modelProvider);
```
TrekSoft, line 61: Thanks for getting rid of this - I'm a huge proponent of not overly separating things out since it leads to endless rabbit trails to find the code you're looking for so I fully support consolidating this functionality into the SyncProcessor.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -15,49 +15,70 @@
...
  */
 final class SyncProcessor {
-    private final RemoteModelState remoteModelState;
-    private final Merger merger;
     private final ModelProvider modelProvider;
     private final ModelSchemaRegistry modelSchemaRegistry;
+    private final SyncTimeRegistry syncTimeRegistry;
+    private final AppSync appSync;
+    private final Merger merger;
+    private final BaseSyncIntervalProvider syncIntervalProvider;
```
TrekSoft, line 59: What is the need for this vs just passing in the interval?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTime.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.amplifyframework.util.Time;
+
+/**
+ * Represents the time at which a type of model was last sync'd.
+ * Why isn't that just an @Nullable Long (capital 'L')?
+ * Rx doesn't allow nulls, so we have to encapsulate the Long value into this
+ * class. TODO: move this to become a static inner class of {@link LastSyncMetadata}?
+ */
+final class SyncTime {
```
TrekSoft, line 29: Ooof, the fact we have to have this class due to RX hurts my eyes - I like the idea of moving it to a static inner class.

TrekSoft: Also, do we still need Rx now that we're requiring Java 1.8 compatability sections be added to the code in the docs?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
jamesonwilliams: This one did take it too far. The Sync Processor and Remote Model State concepts had a lot of the same concerns. This became clear when I needed to lookup/save sync times. Basically, I was going to have to weave two Rx chains together. So I just flattened all the crap into the same Java document.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
jamesonwilliams: The `ModelMetadata` deals with metadata for a single _instance_ of a model.

The `LastSyncMetadata` deals with metadata for the model _type_, inclusive of any/all instances there-of.

However, in the biggest picture, I agree with what you're saying, and I do think we could probably simplify our metadata / system models a bit.

[Here are the ones we have right now, complete with (decent) doco](https://github.com/aws-amplify/amplify-android/blob/master/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/SystemModelsProviderFactory.java#L39-L51).

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
jamesonwilliams: Thanks for bringing this up.

Currently, we have inconsistent naming across the codebase for accessor methods. Some use the "`OkHttp` style" `foo()` (no `get`), and some use the style of prefixing by `getFoo()`.

Our builders mostly use the `.someValue(...)` configurator style, instead of the `.withSomeValue(...)`, but this is probably another area where we have some inconsistency.

Aligning the accessor methods to one naming style is going to be a large change in our code base. Let's discuss the preferred style amongst the team. For methods that simply returned a cached field value, I do prefer the version without `get...()`, personally.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
jamesonwilliams: Assuming I check all of the keys, the number of computations should be the same with or without the control flow.

I did intentionally code this to dis-allow unknown keys. The reasoning is that I don't want a user to have like a "syncMODE" and think it's okay. I want them to get a warning like "Hey bud, what's a syncMODE? The valid keys are: (acceptable keys including syncMode)."

I agree with the comments about transitioning keys in the future, but let's cross that bridge when we get to it. A different way of addressing it would be to have like:
```
// This is a 100% contrived, hypothetical example of us supporting keys during a graceful deprecation.
enum ConfigKey {
    /**
     * @deprecated Use {@link #SYNCHRONIZATION_MODE} instead.
     */
    @Deprecated
    SYNC_MODE("syncMode"),

    /**
     * How to synchronize data with the cloud?
     */
    SYNCHRONIZATION_MODE("syncrhonizationMode")
}
```
Then somewhere:
```
while (keyIterator.hasNext()) {
    ConfigKey configKey;
    try {
        configKey = ConfigKey.fromString(keyIterator.next());
    } catch (IllegalArgumentException noKnownConfigKey) {
        // Tell user what are the acceptable keys ....
    }
    switch (configKey) {
        case SYNC_MODE /* deprecated, but allowed for now... */:
        case SYNCHRONIZATION_MODE /* transitioning to this... */:
            // handle sync mode value
            break;
    }
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePluginConfiguration.java
jamesonwilliams: Resolution: starting to move these things to include `get...()`, as per kotlin interop doc:

https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters

TrekSoft:
Due to the rebased commit there's no way for me to review the changes since my last PR 😞

PR #340, opened 2020-04-02
[aws-datastore] Re-provision if system models change

jamesonwilliams:
There was some exiting logic to blow away the SQL tables if the user's
model structure changes (as informed by the version in the model
provider).

However, no such logic existed for changes to the system models
provider.

To acommodate this, form a compound provider of both, and consider its
version, instead.


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -141,8 +140,7 @@ private SQLiteStorageAdapter(
             ModelProvider userModelsProvider,
             ModelProvider systemModelsProvider) {
         this.modelSchemaRegistry = modelSchemaRegistry;
-        this.userModelsProvider = userModelsProvider;
-        this.systemModelsProvider = systemModelsProvider;
+        this.modelsProvider = CompoundModelProvider.of(userModelsProvider, systemModelsProvider);
```
raphkim, line 143: This is nice! LGTM

PR #345, opened 2020-04-02
Fix language code -> enum conversion

raphkim:
*Issue #, if available:*

*Description of changes:*
I introduced a bug that fails to convert language code to language type enum. The issue is fixed in this PR by creating a static look-up map in the enum class.

### DIFF
--- a/core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
+++ b/core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
@@ -15,6 +15,9 @@
 
 package com.amplifyframework.predictions.models;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Enum of different types of language. These are the languages
  * that are recognized and supported by Amazon Translate Service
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
```diff
@@ -145,11 +158,10 @@
      * @return An enum value of matching language code
      */
     public static LanguageType from(String languageCode) {
```
jamesonwilliams, line 160: It's likely that this will only be called occasionally, for ~one language code. In which case, it might be better optimize the disk use, at the cost of (occasionally) doing some extra computations.

Storing the code map would be a good choice if you need to do a lot of lookups all the time. But, I don't think that's the case here.

For that reason, the other enums have the `for (EnumType value : values()) ` loops that do matching and return a match if found.

raphkim: Ah good point. I'll do that.

core/src/main/java/com/amplifyframework/predictions/models/LanguageType.java
```diff
@@ -145,11 +145,42 @@
...
-        try {
-            return LanguageType.valueOf(languageCode);
-        } catch (IllegalArgumentException noMatchError) {
-            return UNKNOWN;
+        // Micro-optimization... for fun :)
```
jamesonwilliams, line 148: Haha :-D Alright, cool.

Also, I did not realize Spanish was #2, above English!

Quizas debo practicarlo con mas frequencia!

PR #350, opened 2020-04-05
[aws-datastore] Re-enable AWSDataStorePluginInstrumentationTest

jamesonwilliams:
This test had been flakey for two reasons.

Firstly, due to polluted state on the backend.  Sometimes, when the
DataStore starts up in this test, and it tries to sync records, the
records don't make sense.

That makes the test fail during arrangement, way before we actually
exercise any of the business logic of the test.

A solution to this problem is to clear all records from the backend,
before allowing the Sync Engine to start up on the client.

Secondly, the matching logic on the accumulated hub events was loose.
The logic has been tightened up, so that we are looking for more
specific events arrive (by model ID!).

### DIFF
--- a/aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
+++ b/aws-api/src/androidTest/java/com/amplifyframework/api/aws/CodeGenerationInstrumentationTest.java
@@ -66,7 +66,7 @@ public final class CodeGenerationInstrumentationTest {
     @BeforeClass
     public static void onceBeforeTests() throws AmplifyException {
         AmplifyTestConfigurator.configureIfNotConfigured();
-        api = SynchronousApi.singleton();
+        api = SynchronousApi.delegatingToAmplify();
     }
 
     /**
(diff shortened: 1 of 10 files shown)


jamesonwilliams:
> The code itself looks good, but it doesn't seem to pass locally on my machine and the test hangs afterwards.
> 
> ```
> com.amplifyframework.datastore.AWSDataStorePluginInstrumentedTest > null[Pixel_2_API_29(AVD) - 10] FAILED 
>         java.lang.RuntimeException: Neither result nor error consumers accepted a value within 10000ms.
>         at com.amplifyframework.testutils.Await.await(Await.java:156)
> 
> com.amplifyframework.datastore.AppSyncClientInstrumentationTest > null[Pixel_2_API_29(AVD) - 10] SKIPPED 
> <============-> 99% EXECUTING [30m 58s]
> > IDLE
> > :aws-datastore:connectedDebugAndroidTest
> > IDLE
> > IDLE
> ```

On further attempts, I am _still_ getting this sometimes, too. It's ~33-50% of the time. So, not very good. Will uptick my testing of the test, and update this PR when more reliably resolved.

jamesonwilliams:

PR #353, opened 2020-04-05
[aws-analytics-pinpoint] Introduce builder for AnalyticsEvent

jpignata:
**TODO**:

- [x] API Bar Raiser approval
- [ ] Code review approvals
- [x] End to end testing

---

Previously, when using the Amazon Pinpoint plugin, a developer had to instantiate and pass a PinpointProperties object to the constructor of AnalyticsEvent. This change exposes a fluent builder that configures and returns a complete AnalyticsEvent. This moves Analytics closer to the design document and the iOS implementation.

The new syntax for creating an AnalyticsEvent is:

```java
AnalyticsEvent event = AnalyticsEvent.builder()
     .name("EventName")
     .addProperty("StringProperty", "Pancakes")
     .addProperty("BooleanProperty", true)
     .addProperty("DoubleProperty", 3.14)
     .addProperty("IntegerProperty", 42)
     .build();
```

This change also cleans up some of the related interfaces and behavior, namely:

- Rename AnalyticsEvent (the interface) to AnalyticsEventBehavior and BasicAnalyticsEvent (the implementation) to AnalyticsEvent
- Remove the concept of PinpointProperties and instead hoist Property-related classes into the Category itself; prefix all of these classes with "Analytics"
- Expand Properties to support Boolean and Integer as well as String and Double by adding BooleanProperty and IntegerProperty and handling their conversion in the Pinpoint plugin; add a generic add method to allow a developer to add any type that implements AnalyticsPropertyBehavior
- Change the signature of AnalyticsCategoryBehavior#unregisterGlobalProperties to be accept a variable number of Strings instead of a Set
- Stop guarding configuration methods like registering and unregistering properties with an enabled check
- Add iterator, size, and get methods to AnalyticsProperties to encapsulate the underlying Map


undefobj:
Since this is a change to the public API we should do an API review before releasing.

core/src/main/java/com/amplifyframework/analytics/BasicAnalyticsEvent.java
```diff
@@ -15,31 +15,148 @@
...
+    }
+
+    /**
+     * Builder is used to create and configure an instance of {@link BasicAnalyticsEvent}. Its
+     * methods return the Builder instance to allow for fluent method chaining. This Builder reuses
+     * {@link Properties.Builder} to construct the properties to store in the event.
+     *
+     * @see Properties
+     */
+    public static final class Builder {
```
jamesonwilliams, line 85: All of the methods in this builder can also be marked as `@NonNull`, I believe. (A human wouldn't benefit from the cue, on this file. But, a lint-ing program / Kotlin type interpreter, could.)

For the values that we pass into the builder - should we allow them to be null, or no? If no, we have a convention of using `Objects.requireNonNull(...)` to "fail fast" as soon as we see a bad user-provided value, at its first location of use.

core/src/main/java/com/amplifyframework/analytics/BooleanProperty.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Factory method to instantiate a {@link BooleanProperty} from a {@link Boolean} value.
+     *
+     * @param value a boolean value
+     * @return an instance of {@link BooleanProperty}
+     */
+    public static BooleanProperty from(@NonNull Boolean value) {
+        return new BooleanProperty(value);
```
jamesonwilliams, line 47: Potentially, `BooleanProperty(Objects.requireNonNull(value))`  (Same for all of the arguments to various `FooProperty` factories.)

core/src/main/java/com/amplifyframework/analytics/Properties.java
```diff
@@ -15,34 +15,152 @@
...
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 /**
- * Wrapper for {@link Map} to restrict the types that may be passed in.
+ * Properties provides key/value pairs to represent attributes of an {@link AnalyticsEvent}. While
+ * different analytics services will have support for different datatypes, Properties provides a
+ * baseline of support for String, Boolean, Double, and Integer. If one of these aren't available in
+ * a given service, the plugin is expected to cast it to something supported. (e.g., convert Boolean
```
jamesonwilliams, line 28: The docs you've added in this PR are first-class, @jpignata!

core/src/main/java/com/amplifyframework/analytics/Properties.java
```diff
@@ -15,34 +15,152 @@
...
+        return properties.entrySet().iterator();
+    }
+
+    /**
+     * size returns the number of properties.
+     *
+     * @return The number of properties
+     */
+    @NonNull
+    public Integer size() {
```
jamesonwilliams, line 72: Would return primitive `int`, which can't be null, anyway.

core/src/main/java/com/amplifyframework/analytics/Properties.java
```diff
@@ -15,34 +15,152 @@
...
+     * get returns a specific {@link Property} by its name.
+     *
+     * @param propertyName The name of the property to retrieve
+     * @return The corresponding {@link Property}
      */
-    @SuppressWarnings("VisibilityModifier")
-    protected final Map<String, Property<?>> properties;
+    @NonNull
+    public Property<?> get(@NonNull String propertyName) {
+        return properties.get(propertyName);
```
jamesonwilliams, line 84: If I pass a non-existing property name, this would currently return `null`, but the flag says `@NonNull`. I think we need some logic like:
```
Property<?> property = properties.get(propertyName);
if (property == null) {
   // there was no match ... throw an exception?
}
return property;
```

core/src/main/java/com/amplifyframework/analytics/Properties.java
```diff
@@ -15,34 +15,152 @@
...
 
     /**
-     * Returns the underlying property map.
-     * @return underlying properties map
+     * Builder is used to create and configure an instance of {@link Properties}. Its
+     * methods return the Builder instance to allow for fluent method chaining.
      */
-    public final Map<String, Property<?>> get() {
-        return Immutable.of(properties);
+    public static final class Builder {
```
jamesonwilliams, line 101: Hm, I forget. Since this is part of the `core` spec: Do we need an `<T, P extends Property<T>> Builder add(String, P property)` method on this builder, as well? That way the user has an extension point, if they're using a property other than the ones our plugin does? (That is, another plugin might only work with `GizmoProperty` values.)

core/src/main/java/com/amplifyframework/analytics/PropertyNames.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * PropertyNames holds the names of {@link Property} objects used in configuration calls.
+ *
+ * <pre>
+ *     PropertyNames propertyNames = PropertyNames.of("Height", "HairColor", "StarSign");
+ * </pre>
+ */
+public final class PropertyNames implements Iterable<String> {
+    private final HashSet<String> propertyNames;
```
jamesonwilliams, line 31: The convention in the rest of the project is to declare the field using its type of least specificity (Liskov Principle); here, this would be just `Set<String>`.

core/src/main/java/com/amplifyframework/analytics/PropertyNames.java
```diff
@@ -0,0 +1,85 @@
...
+        return propertyNames.iterator();
+    }
+
+    /**
+     * size returns the number of names contained in PropertyNames.
+     *
+     * @return The number of names contained
+     */
+    @NonNull
+    public Integer size() {
```
jamesonwilliams, line 78: (Suggest to use primitive `int` for size methods.)

core/src/test/java/com/amplifyframework/analytics/BasicAnalyticsEventTest.java
```diff
@@ -0,0 +1,51 @@
...
+package com.amplifyframework.analytics;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class BasicAnalyticsEventTest {
+
+    @Test
+    @SuppressWarnings("MagicNumber")
```
jamesonwilliams, line 25: Not part of this PR -- but just mentioning while I'm here --

I've been meaning to do a PR that adds `MagicNumber` to the `checkstyle-suppressions.xml` when found under `src/test` or `src/androidTest` (these trees tend to have lots of abitrarily arranged data values.)

core/src/test/java/com/amplifyframework/analytics/PropertyNamesTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public void of() {
+        // Arrange: Use the static factory method to create a PropertyNames instance
+        Set<String> names = new HashSet<>(Arrays.asList("Height", "Weight", "StarSign"));
+
+        PropertyNames propertyNames;
+        propertyNames = PropertyNames.of(names.toArray(new String[names.size()]));
+
+        // Assert: Iterate through the items and remove each name from the set, then check all items
+        // were removed after iteration
+        assertEquals(3, (long) propertyNames.size());
```
jamesonwilliams, line 40: (By the time you get this returning a primitive, I think the case will go away, and the `assertEquals(int, int)` would be automatically selected, instead of `assertEquals(long, long)`, as it is now (or `assertEquals(Object, Object)`, as it might have been without the primitive casting.)

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -145,52 +166,66 @@ public void testAutoFlush() throws AnalyticsException {
...
-        assertEquals(2, analyticsClient.getAllEvents().size());
-        assertTrue(eventAttributes.has("Property"));
-        assertTrue(eventAttributes.has("GlobalProperty"));
-        assertFalse(event2Attributes.has("Property"));
+        // Global properties are attached to all events
+        assertTrue(event1Attributes.has("GlobalProperty"));
         assertTrue(event2Attributes.has("GlobalProperty"));
+
+        // Local properties are only attached is passed explicitly as in event
+        assertTrue(event1Attributes.has("Property"));
```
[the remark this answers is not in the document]
jpignata: :+1:

core/src/main/java/com/amplifyframework/analytics/BasicAnalyticsEvent.java
jpignata: Ahh, yes, you are correct. Thanks! I'll get the hang of this eventually. Adding the fail-fast incantation where appropriate.

core/src/main/java/com/amplifyframework/analytics/Properties.java
jpignata: Thank you for noting this. I thought about this this morning a while and wasn't sure of the best way to present an extension point. Will grab you tomorrow to chat about it.

core/src/test/java/com/amplifyframework/analytics/BasicAnalyticsEventTest.java
jpignata: In the spirit of the annotations slightly irking me, I've tackled that in a commit in this pull. If you'd rather, I can split it out.

core/src/main/java/com/amplifyframework/analytics/Properties.java
jpignata: I suppose we could also remove the @NonNull and embrace the null-ness as this method is mainly used for test purposes. I gave it a shot by raising `NoSuchElementException`.

core/src/test/java/com/amplifyframework/analytics/PropertiesTest.java
```diff
@@ -0,0 +1,60 @@
...
+        final String timeZone = ((StringProperty) properties.get("TimeZone")).getValue();
+        final Boolean inGeoFence = ((BooleanProperty) properties.get("inGeoFence")).getValue();
+        final Double loginDuration = ((DoubleProperty) properties.get("LoginDuration")).getValue();
+        final Integer profileCount = ((IntegerProperty) properties.get("ProfileCount")).getValue();
+
+        assertEquals(4, properties.size());
+        assertEquals("EDT", timeZone);
+        assertEquals(false, inGeoFence);
+        assertEquals(new Double(98.36), loginDuration);
+        assertEquals(2, (long) profileCount);
```
jamesonwilliams, line 49:
```
assertEquals(4, properties.size());
assertEquals("EDT", timeZone);
assertEquals(Boolean.FALSE, inGeoFence);
assertEquals(Double.valueOf(98.36), loginDuration);
assertEquals(Integer.valueOf(2), profileCount);
```

core/src/main/java/com/amplifyframework/analytics/Properties.java
```diff
@@ -15,34 +15,181 @@
...
+ * <pre>
+ *     Properties properties = Properties.builder()
+ *          .add("PostType", "UserImage")
+ *          .add("LikedUserID", 78219)
+ *          .add("FirstLike", true)
+ *          .build();
+ * </pre>
  */
-public class Properties {
+public final class Properties implements Iterable<Map.Entry<String, Property<?>>> {
```
lawmicha, line 41: `AnalyticsProperties`? and each `Property` -> `AnalyticsProperty`?
- would we benefit having a less genericly named "Property" class and builder, in case "Property" is used else where in the future?
- by having this, developer experience would be:
```
Amplify.Analytics.registerGlobalProperties(
    AnalyticsProperties.builder()
        .add("Key1", "Value1")
        .add("Key2", "Value2")
        .build()
);
```

core/src/test/java/com/amplifyframework/analytics/AnalyticsPropertiesTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    public void getRaisesNoSuchElementExceptionWhenPropertyNotFound() {
+        AnalyticsProperties properties = AnalyticsProperties.builder().build();
+
+        assertThrows(NoSuchElementException.class, () -> {
+            properties.get("key-that-doesnt-exist");
+        });
+    }
+
+    class FooProperty implements AnalyticsPropertyBehavior<Date> {
```
jamesonwilliams, line 74: nit: could be marked as a `static final class`

PR #354, opened 2020-04-07
[aws-api] Run tests against ApiCategory, not Amplify

jamesonwilliams:
The Amplify singleton is problematic to test, since it hangs onto static
references of its once-configured categories.

Instead, test fresh instances of the ApiCategory, instead of the Ampilfy
facade.


aws-api/src/androidTest/java/com/amplifyframework/api/aws/TestApiCategory.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raphkim, line 2: 2020

jamesonwilliams: Good eye - fixed

PR #355, opened 2020-04-07
[aws-storage-s3] Test against StorageCategory not Amplify

jamesonwilliams:
Amplify is a process-wide singleton, that can be configured only once.
It maintains state until the process dies. This is true of its reference
to the Amplify.Storage category (and its plugins) as well.

This is a brittle way to test the functionality of storage. Better is to
create fresh instances of the StorageCategory, and run assertions
against them.


aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestStorageCategory.java
```diff
@@ -0,0 +1,56 @@
...
+final class TestStorageCategory {
+    private TestStorageCategory() {}
+
+    /**
+     * Creates an instance of {@link StorageCategory} using the provided configuration resource.
+     * @param resourceId Android resource ID for a configuration file
+     * @param context Android Context
+     * @return A StorageCategory instance using the provided configuration
+     */
+    static StorageCategory create(@RawRes int resourceId, @NonNull Context context) {
```
raphkim, line 42: nit: Not sure if it's a convention that is documented somewhere, but I think `context` is always the first argument if it is required. It already bothers me that our `configure` methods break this pattern haha.

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousStorage.java
```diff
@@ -38,25 +40,33 @@
  * code, where we want to make a series of sequential assertions after
  * performing various operations.
  */
-@SuppressWarnings("unused")
 public final class SynchronousStorage {
-    // 5 seconds seemed to be insufficient to reliably cover both initial auth calls + storage network ops
-    private static final long STORAGE_OPERATION_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10);
+    private static final long STORAGE_OPERATION_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(5);
```
raphkim, line 44: Hmm, I guess having smaller default value is fine since `SynchronousStorage` does allow timeout override for especially long operations (e.g. large file upload)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/UserCredentials.java
```diff
@@ -0,0 +1,172 @@
...
+
+final class UserCredentials implements Iterable<UserCredentials.Credential> {
+    private static final String CREDENTIALS_RESOURCE_NAME = "credentials";
+    private final List<Credential> credentials;
+
+    private UserCredentials(List<Credential> credentials) {
+        this.credentials = credentials;
+    }
+
+    static UserCredentials create(@NonNull IdentityIdSource identityIdSource, @NonNull Context context) {
```
raphkim, line 45: nit: context as first parameter?

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/UserCredentials.java
```diff
@@ -0,0 +1,172 @@
...
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+final class UserCredentials implements Iterable<UserCredentials.Credential> {
```
raphkim, line 37: this is clean, and likely reusable for other categories like API :)

As far as I know, we lack tests for testing schemas that contain `@auth`

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/TestStorageCategory.java
jamesonwilliams: Hmm, true. I've never seen that written anywhere, but it "feels" true enough... I'll update all locations in the PR.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/UserCredentials.java
jamesonwilliams: To this point -- I'll move this over into `testutils`, instead of `aws-storage-s3`. That way it _can_ be re-used by other modules' tests.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/UserCredentials.java
jamesonwilliams: done

PR #356, opened 2020-04-07
[aws-datastore] Test against DataStoreCategory not Amplify

jamesonwilliams:


testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousDataStore.java
```diff
@@ -33,20 +34,21 @@
...
+     * @param asyncDelegate This performs actual DataStore operations
+     * @return A synchronizing wrapper around the provided {@link DataStoreCategoryBehavior}
      */
     @NonNull
-    public static synchronized SynchronousDataStore singleton() {
-        if (SynchronousDataStore.singleton == null) {
-            SynchronousDataStore.singleton = new SynchronousDataStore();
-        }
-        return SynchronousDataStore.singleton;
+    public static SynchronousDataStore delegatingTo(@NonNull DataStoreCategoryBehavior asyncDelegate) {
```
rjuliano, line 49: is this supposed to change to `delegateTo`?

jamesonwilliams: Thanks for pointing this out! I have been inconsistent so far between the two in the various categories. Right now in `master`, `SynchronousApi` and `SynchronousStorage` use `delegatingTo`, so I guess I'll try to stick with that here, as well.

PR #363, opened 2020-04-07
Add unit tests for offline predictions

raphkim:
*Issue #, if available:*

*Description of changes:*
- Add unit tests for `fetchSentiment()` and `tokenizeInputText()`
- Remove `final` keyword from loaders to make them mockable

### DIFF
--- a/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationModel.java
+++ b/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationModel.java
@@ -37,7 +37,7 @@ import java.nio.channels.FileChannel;
  * Loads the pre-trained text classification model into
  * a TensorFlow Lite interpreter instance.
  */
-public final class TextClassificationModel implements Loadable<Interpreter, PredictionsException> {
+public class TextClassificationModel implements Loadable<Interpreter, PredictionsException> {
 
     private static final String MODEL_PATH = "text_classification.tflite";
 
--- a/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
+++ b/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
@@ -18,6 +18,7 @@ package com.amplifyframework.predictions.tensorflow.service;
 import android.content.Context;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import androidx.annotation.WorkerThread;
 
 import com.amplifyframework.core.Consumer;
@@ -50,6 +51,7 @@ final class TensorFlowTextClassificationService {
     private final TextClassificationModel interpreter;
     private final TextClassificationDictionary dictionary;
     private final TextClassificationLabels labels;
+
     private final List<Loadable<?, PredictionsException>> assets;
     private final CountDownLatch loaded;
 
@@ -62,13 +64,23 @@ final class TensorFlowTextClassificationService {
      * @param context the Android context
      */
     TensorFlowTextClassificationService(@NonNull Context context) {
-        this.interpreter = new TextClassificationModel(context);
-        this.dictionary = new TextClassificationDictionary(context);
-        this.labels = new TextClassificationLabels(context);
+        this(new TextClassificationModel(context),
+                new TextClassificationDictionary(context),
+                new TextClassificationLabels(context));
+    }
+
+    @VisibleForTesting
+    TensorFlowTextClassificationService(
+            TextClassificationModel interpreter,
+            TextClassificationDictionary dictionary,
+            TextClassificationLabels labels
+    ) {
+        this.interpreter = interpreter;
+        this.dictionary = dictionary;
+        this.labels = labels;
 
         this.assets = Arrays.asList(interpreter, dictionary, labels);
         this.loaded = new CountDownLatch(assets.size());
-
         for (Loadable<?, PredictionsException> asset : assets) {
             asset.onLoaded(
                 onLoad -> this.loaded.countDown(),
@@ -87,6 +99,10 @@ final class TensorFlowTextClassificationService {
         return SERVICE_KEY;
     }
 
+    /**
+     * Load all of the required assets. No-op for each asset
+     * that is already loaded.
+     */
     @WorkerThread
     synchronized void loadIfNotLoaded() {
         for (Loadable<?, PredictionsException> asset : assets) {
@@ -132,7 +148,8 @@ final class TensorFlowTextClassificationService {
         }
     }
 
-    private Sentiment fetchSentiment(String text) throws PredictionsException {
+    @VisibleForTesting
+    Sentiment fetchSentiment(String text) throws PredictionsException {
         float[][] input;
         float[][] output;
 
--- a/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
+++ b/aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
@@ -36,6 +36,7 @@ import org.tensorflow.lite.Interpreter;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 
 /**
@@ -61,14 +62,12 @@ final class TensorFlowTextClassificationService {
      * Constructs an instance of service to perform text
      * sentiment interpretation using TensorFlow Lite
      * interpreter.
-     * @param context the Android context
+     * @param interpreter the TensorFlow Lite interpreter with
+     *                    loaded model
+     * @param dictionary the dictionary of words and respective
+     *                   tokens
+     * @param labels the list of labels for a feature
      */
-    TensorFlowTextClassificationService(@NonNull Context context) {
-        this(new TextClassificationModel(context),
-                new TextClassificationDictionary(context),
-                new TextClassificationLabels(context));
-    }
-
     @VisibleForTesting
     TensorFlowTextClassificationService(
             TextClassificationModel interpreter,
@@ -87,6 +86,20 @@ final class TensorFlowTextClassificationService {
                 error -> this.loadingError = error
             );
         }
+    }
+
+    /**
+     * Constructs an instance of text classifier service by
+     * loading the assets from provided Android context.
+     * @param context the Android context
+     * @return an instance of text classification service
+     */
+    static TensorFlowTextClassificationService fromContext(@NonNull Context context) {
+        Objects.requireNonNull(context);
+        TextClassificationModel model = new TextClassificationModel(context);
+        TextClassificationDictionary dictionary = new TextClassificationDictionary(context);
+        TextClassificationLabels labels = new TextClassificationLabels(context);
+        return new TensorFlowTextClassificationService(model, dictionary, labels);
     }
 
     /**
--- a/aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
+++ b/aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
@@ -22,8 +22,6 @@ import com.amplifyframework.predictions.PredictionsException;
 
 import org.junit.Before;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
 
 import java.io.FileInputStream;
 import java.io.InputStream;
@@ -43,7 +41,6 @@ import static org.mockito.Mockito.when;
  * Test that input conversion to TensorFlow Lite tokens
  * works as intended.
  */
-@RunWith(JUnit4.class)
 public final class InputTokenizerTest {
     private static final long LOAD_TIMEOUT_MS = 100;
 
--- a/aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
+++ b/aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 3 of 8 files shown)


aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
```diff
@@ -62,13 +64,23 @@
      * @param context the Android context
      */
     TensorFlowTextClassificationService(@NonNull Context context) {
```
jamesonwilliams, line 66: One of the reason I user factories all the time is that they allow you to write code above the equivalent of `this(...)`. e.g.,

```
static TensorFlowTextClassificationService fromContext(@NonNull Context context) {
    Objects.requireNonNull(context);
    TextClassificationModel model = new TextClassificationModel(context);
    TextClassificationDictionary dictionary = new TextClassificationDictionary(context);
    TextClassifcationLabels labels = new TextClassificationLabels(context);
    return new TensorFlowTextClassificationService(model, dictionary, labels);
}
```

Overall, you just don't have to fight the syntax as hard, with this approach ^^ as compared the constructor-way.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
```diff
@@ -132,7 +148,8 @@ void classify(
         }
     }
 
-    private Sentiment fetchSentiment(String text) throws PredictionsException {
```
jamesonwilliams, line 135: If you already have package visibility for `classify(...)`, can you just test through it? 

The only problem I can see might be that latch, which would block stuff.

Ah, haha. Okay. Remember that crazy `Loadable<List<Loadable<?, PredictionsException>>, PredictionsException>`? If you use that, and then accept it as a dependency in the constructor, you can "fake" whether or not this service is loaded.

Assuming you can push `loaded.await();` into the implementation of that big Loadable, that'll help you when you run `classify` from test. (In source, you provide an impl that blocks. In test, you provide a `mock(Loadable.class)` that does whatever you want.

I guess the actual latch await call would change to:
```
if (!compoundLoadable.isLoaded()) {
    compountLoadable.load();
}
```
Or something like that.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
```diff
@@ -0,0 +1,114 @@
...
+    public void testInputTextTokenizer() throws Exception {
+        final CountDownLatch loaded = new CountDownLatch(1);
+        final AtomicReference<Map<String, Integer>> tokens = new AtomicReference<>();
+        final AtomicReference<PredictionsException> error = new AtomicReference<>();
+
+        final String inputText = "Where is the bathroom?";
+        final InputStream stream = new FileInputStream("src/test/resources/word-tokens.txt");
+
+        // Make mock context return mock asset manager
+        when(mockContext.getAssets()).thenReturn(mockAssets);
```
jamesonwilliams, line 76: Since both of these are in `@Before` scope, I would move this `when` arrangement up right underneath them, there, too.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
```diff
@@ -0,0 +1,114 @@
...
+                .onLoaded(
+                    onLoad -> {
+                        loaded.countDown();
+                        tokens.set(onLoad);
+                    },
+                    onError -> {
+                        loaded.countDown();
+                        error.set(onError);
+                    }
+                );
```
jamesonwilliams, line 92: You can use the `Await` utility for this:

```
Map<String, Integer> tokens = Await.<WhateverResultType, PredictionsException>result(
    (onResult, onError) -> dictionary.onLoaded(onResult::accept, onError::accept)
);
```

This way, you don't have to mess around with a `CountDownLatch`,  yourself.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
```diff
@@ -0,0 +1,114 @@
...
+        dictionary.load();
+        loaded.await(LOAD_TIMEOUT_MS, TimeUnit.MICROSECONDS);
```
jamesonwilliams, line 94: (This will go away with `Await`, too. But ___nit:___ `MILLISECONDS`! Not `MICROSECONDS`.)

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        // Tokenize input
+        float[][] input = dictionary.tokenizeInputText(inputText);
+        float[][] expected = new float[1][input[0].length];
+        expected[0][0] = 1; // <START>
+        expected[0][1] = 2; // Where (<UNKNOWN>)
+        expected[0][2] = 9; // is
+        expected[0][3] = 4; // the
+        expected[0][4] = 2; // bathroom (<UNKNOWN>)
+        // Followed by 0's (<PAD>)
```
jamesonwilliams, line 110:
```
int[] values = { 1, 2, 9, 4, 2};
for (int index = 0; index < : values.size(); index++) {
    expected[0][index] = values[index];
}
```

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/service/TextClassificationTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private TensorFlowTextClassificationService service;
+
+    /**
+     * Set up mock behavior before each test.
+     */
+    @Before
+    public void setUp() {
+        // Mock asset loaders
+        mockInterpreter = mock(TextClassificationModel.class);
+        mockDictionary = mock(TextClassificationDictionary.class);
```
jamesonwilliams, line 59: Instead of mocking the dictionary, can your provide `text_classification_vocab.txt`, and use real one?

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationModel.java
```diff
@@ -37,7 +37,7 @@
  * Loads the pre-trained text classification model into
  * a TensorFlow Lite interpreter instance.
  */
-public final class TextClassificationModel implements Loadable<Interpreter, PredictionsException> {
+public class TextClassificationModel implements Loadable<Interpreter, PredictionsException> {
```
jamesonwilliams, line 40: It might make sense to have a 
```
public interface TextClassificationModel<T> extends Loadable<T, PredictionsException> {
    ...
}
```
with a:
```
public final class TensorFlowTextInterpreter implements TextClassificationModel<Interpreter> {
    ...
}
```

Then, in your test you can `mock(TestClassificationModel.class)`, but in your source, you can keep `TensorflowTextInperpreter`.

aws-predictions-tensorflow/src/test/resources/word-tokens.txt
```diff
@@ -0,0 +1,10 @@
+<PAD> 0
```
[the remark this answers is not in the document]
raphkim: Hmm I guess I could, but this was how the asset in the official sample tensorflow lite app was formatted. It wouldn't be a huge deal, since we can internalize how we want them to be formatted, but I feel like it could create confusion to customers who are directly copying the assets from the official tensorflow sample.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
raphkim: I envision adding more tests in this suite that tests different types of assets to load (for different predictions verbs), so I structured it like this intentionally.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
raphkim: Ah I was wrestling with that for a bit, but I think I got it now!

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
raphkim: oh wow i guess I shouldn't have relied on autocomplete LOL

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
raphkim: I was considering that, but I think the way I currently have it written makes it easier to document what each number means

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/service/TextClassificationTest.java
raphkim: I was trying to stay away from relying on androidx and robolectric if possible. I think this is sufficient for now.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationModel.java
raphkim: interfaces are _almost_ always great, but I think we shouldn't overuse it either. `TextClassificationModel` would be a bit _too_ specific of a concept to be appropriate as an interface, and we are not really getting concrete advantage out of it besides allowing its instance to be `final`.

aws-predictions-tensorflow/src/test/resources/word-tokens.txt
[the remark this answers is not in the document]
jamesonwilliams: Oh yea, okay! I agree. Let's stick with the TensorFlow sample.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
jamesonwilliams: Haha yea, it's a compound statement with a lot of syntax...

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/asset/InputTokenizerTest.java
jamesonwilliams: This exact thing happens to me all the time. It's probably something IntelliSense could improve.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/service/TextClassificationTest.java
jamesonwilliams: There are pros and cons -- the less you mock, the more "real" your test is, and the more coverage you get. Plus, you'll be able to keep `final` on your class, to keep it "closed."

The only con is really just that Robolectric is slow.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowTextClassificationService.java
raphkim: This test is to directly test the behavior of `fetchSentiment`. How about if I write more tests for blocking behavior in the future when I improve on the current latch design?

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/asset/TextClassificationModel.java
jamesonwilliams: Yea, actually, agreed. There aren't going be multiple text classification services -- it's the TensorFlow plugin. The only alternate would be for test purposes. So I do agree, that's overkill.

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/service/TextClassificationTest.java
jamesonwilliams: One more note -- 

I used to write (and I have seen a lot of others write) tests like this:

```
// Assuming a simple actor like ...
class ObjectUnderTest {
    private final Dependency dependency;
    ObjectUnderTest(Dependency dependency) {
        this.dependency = dependency;
    }
    Value lookupValue() {
        return dependency.getValue();
     }
}
```

```
// Here, we mock out all dependencies of the unit, and write a test
// that basically just re-codifies or implementation logic into Mockito syntax.
public final class AnemicUnitTest {
    private Dependency dependency;
    @Before
    public void setup() {
        this.dependency = mock(Dependency.class);
        this.objectUnderTest = new ObjectUnderTest(dependency);
    }

    @Test
    public void lookupValueCallsDependency() {
        when(dependency.getValue()).thenReturn(Value.of(2));
        assertEquals(Value.of(2), doSomething.lookupValue());
    }
}
```

This is an "anemic" unit test, where all dependencies are mocked. We've verified that the unit of code (the ObjectUnderTest) calls our Dependency. But so what? The complexity is very low.

Now, let's imagine `Dependency` is something like TensorFlow, and `getValue()` triggers _thousands_ of lines of code to be executed. If we can avoid using the mock, we have a bonafide component test on our hands, and a good indication that the `ObjectUnderTest` itself will work in a real-world application.

```
// Here, we test *both* ObjectUnderTest, and the direct dependencies it needs to its job
public final class ComponentTest {
     private RealDependency dependency;
     private ObjectUnderTest objectUnderTest;

     @Before
      public void setup() {
          this.dependency = new RealDependency();
          this.objectUnderTest = new ObjectUnderTest(dependency);
      }

      @Test
      public void lookupValueReturnsExpectedValue() {
          // Value is not arranged, but derived form real world business conditions
          dependency.setColor("#441155"); // Some logic mutates the dependency state
          assertTrue(Value.of(44), objectUnderTest.getValue());
      }
}
```

aws-predictions-tensorflow/src/test/java/com/amplifyframework/predictions/tensorflow/service/TextClassificationTest.java
```diff
@@ -0,0 +1,117 @@
...
+                "unknown2"
+        );
+        // Mock random confidence score
+        final float confidenceScore = new Random().nextFloat();
+
+        // Make mock interpreter set confidence score for
+        // "positive" label into pre-determined random value
+        doAnswer(invocation -> {
+            float[][] output = invocation.getArgument(1, float[][].class);
+            output[0][labels.indexOf("positive")] = confidenceScore;
```
jamesonwilliams, line 98: Huh, I like this indexing.

PR #364, opened 2020-04-07
Run tests against PredictionsCategory, not Amplify

raphkim:
*Issue #, if available:*

*Description of changes:*
- Delivering on the promised follow-up in #346 to re-factor Predictions instrumentation test
- Using more robust sample text for testing


aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsInterpretTest.java
```diff
@@ -44,16 +51,28 @@
...
-        TestConfiguration.configureIfNotConfigured();
-        predictions = SynchronousPredictions.delegatingToAmplify();
+        SynchronousMobileClient.instance().initialize();
+    }
+
+    /**
+     * Configure Predictions category before each test.
+     */
+    @Before
+    public void setUp() {
```
jamesonwilliams, line 69: It should be okay to re-init the mobile client in `@Before`, right? The benefit is, it gets all this setup stuff back into just few lines inside of one method body. Also, a guaranteed baseline before each `@Test` (for the win!)

Singletons and static fields are sort of anti-pattern; I think we should aim to reduce our use of `@BeforeClass` (we use it more than most (quality) code bases I have seen, right now.)

raphkim: Yea, they would do the same thing, but I think leaving it as is is good to remind us that `AWSMobileClient` is a singleton.

PR #365, opened 2020-04-08
Improve feature comparison in tests

raphkim:
*Issue #, if available:*

*Description of changes:*
- `FeatureAssert` utility class helps compare a feature or features with the values they are assigned while ignoring other properties like confidence score.
- Flatten `TargetText` into direct properties of `TextFeature` to reduce nesting
  (i.e. `keyPhrase.getTargetText()` is easier to write than `keyPhrase.getTarget().getText()`)


aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/FeatureAssert.java
```diff
@@ -0,0 +1,62 @@
...
+final class FeatureAssert {
+    private FeatureAssert() {}
+
+    /**
+     * Assert that a given value equals the feature's assigned value.
+     * @param expectedValue the value to expect in a feature
+     * @param actual the actual feature
+     * @param <T> the type of expected value
+     */
+    static <T> void assertMatches(T expectedValue, Feature<T> actual) {
```
jamesonwilliams, line 39: Maybe this could be named _`assertFeatureValue(T expectedValue, Feature<T> actualFeature)`_?

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/FeatureAssert.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Assert that a list of features all match the expected values.
+     * @param expected the list of feature values to expect
+     * @param actual the list of actual features
+     * @param <F> the data type of the feature
+     * @param <T> the data type of the value of the feature
+     */
+    static <F extends Feature<T>, T> void assertMatches(Collection<T> expected, Collection<F> actual) {
```
jamesonwilliams, line 52: Maybe, `assertValuesForFeatures(...)`, or `assertFeatureValues(...)`? 

Hm, I wonder if you can do
```
static <F extends Feature<T>, T> void assertFeatureValues(T... values, F... features);
```
? I don't think you can have two varargs like that ...

core/src/main/java/com/amplifyframework/predictions/models/TextFeature.java
```diff
@@ -28,20 +28,41 @@
  */
 @SuppressWarnings("unchecked")
 public abstract class TextFeature<T> extends Feature<T> {
```
jamesonwilliams, line 30: We'd talked about calling this like `TestPortion` or `TextSelection` or `TextSection` or `TextSegment` or something ... now that it has the positioning info embedded, worth revisiting? Or no?

core/src/main/java/com/amplifyframework/predictions/models/TextFeature.java
```diff
@@ -28,20 +28,41 @@
...
     @NonNull
-    public final TargetText getTarget() {
-        return target;
+    public final String getTargetText() {
+        return targetText;
+    }
+
+    /**
+     * Gets the starting position of the target text
+     * with respect to the full input text.
```
jamesonwilliams, line 53: So the `targetText` starts at `startIndex` within a (>= size) input text?

raphkim: yep!

core/src/main/java/com/amplifyframework/predictions/models/TextFeature.java
raphkim: Hmm, are you talking about renaming the `TextFeature` class? It is still a feature, so I would like to keep that word in the name. I believe our old discussion was regarding the name of `TargetText` class, but that on longer exists.

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/FeatureAssert.java
raphkim: haha, I feel like there should be a verb to make grammatical sense (i.e. "matches"). I was actually considering `assertValueMatches` or `assertFeatureValueMatches`, but I thought that was a bit too long. Besides, this method is static, so calling `FeatureAssert` first would imply that we are asserting to see if features match.

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/FeatureAssert.java
raphkim: Yea the varargs would not work since it is required to be the last parameter only.

I was modeling this after `Assert#assertEquals(Object, Object)`, which works for pretty much anything (list, or singular struct) even with the same name. I wanted to maintain this simplicity in the assertion name.

raphkim:
> raphkim dismissed jamesonwilliams’s stale review via 31018d8 5 hours ago

But I didn't want to 😢

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsInterpretTest.java
```diff
@@ -168,17 +165,24 @@ public void testNegativeSentimentDetection() throws Exception {
      */
     @Test
     public void testKeyPhraseDetection() throws Exception {
+        final String sampleText = "My mama always said life was like a box of chocolates.";
```
jpignata, line 168: Seems before your time.

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsInterpretTest.java
```diff
@@ -168,17 +165,24 @@ public void testNegativeSentimentDetection() throws Exception {
...
-        // Assert detected key phrases are not empty
-        List<KeyPhrase> keyPhrases = result.getKeyPhrases();
-        assertNotNull(keyPhrases);
-        assertFalse(keyPhrases.isEmpty());
+        // Assert key phrase detection
+        List<KeyPhrase> actual = result.getKeyPhrases();
+        List<String> expected = Arrays.asList(
+                "My mama",
+                "life",
+                "a box",
```
jpignata, line 182: Idle thought: Does Comprehend provide a guarantee around what it considers a "key phrase?" For example, will we regret coupling our test to whatever Comprehend's models return for a given sentence. For example, in the future, could Comprehend omit articles from key phrases and just return the noun? Perhaps the answer is, "idk? We'd just update the test."

raphkim: That's my approach for now haha. We'll keep this the way is until the tests break -> update the tests. If it becomes a semi-frequent enough issue, we will revisit to see if we can have a "snapshot" of the comprehend model

PR #368, opened 2020-04-08
Sign In and Fetch Session

TrekSoft:
Adds support for normal sign in + fetchAuthSession per the latest design review.

Design review hit a couple more snags so it's still ongoing and will be writing the unit tests for  once that's finished before the AuthCategory branch is merged to master.


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -230,13 +230,9 @@ public void onResult(SignInResult result) {
...
 
-                if (result.getSignInState().equals(SignInState.DONE)) {
-                    state = AuthSignedInStatus.SIGNED_IN;
-                } else {
-                    state = AuthSignedInStatus.SIGNED_OUT;
-                }
-
-                onSuccess.accept(new AuthSignInResult(state, newDetails));
+                onSuccess.accept(
+                        new AuthSignInResult(AuthSignInState.getEnum(result.getSignInState().toString()), newDetails)
```
TrekSoft, line 234: Note that in the future, depending on the API naming meeting, I may need to write a map to convert Cognito sign in state values to match ours but for now, per the team's agreement, we're starting with the same values.

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
```diff
@@ -0,0 +1,114 @@
...
+        this.value = Objects.requireNonNull(value);
+    }
+
+    @NonNull
+    public String getValue() {
+        return value;
+    }
+
+    @NonNull
+    public static AuthSignInState getEnum(String value) {
```
jamesonwilliams, line 105: Naming for this method, I like any of these better:
1. `fromValue(String possibleValue)` (to match `getValue()`, which returns one.
2. `fromString(String possibleValue)`
3. `lookup(String possibleValue)`;
4. `from(String possibleValue)`

(`getEnum` is [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation), technically - and [folks say not to use that](https://stackoverflow.com/a/111972/12442906), these days.)

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
```diff
@@ -0,0 +1,114 @@
...
+        for (AuthSignInState v : values()) {
+            if (v.getValue().equalsIgnoreCase(value)) {
+                return v;
+            }
+        }
+
+        return AuthSignInState.UNKNOWN;
+    }
+}
```
jamesonwilliams, line 114: Since the names of your enum identifiers are the same as the contents of their bound `values`, the default `toString()` (which calls `name()`) works, as is.

However, should any enumerated value to added/changed here, this could break in a hard-to-trace way.

I recommend adding a `toString()` impl which emits the `getValue()`, instead of the enum _`name()`_ (the default behavior).

-------------

Actually, your enum values are currently the same as all of your enum identifiers. So you _could_ just skip the inner `value field, and use `name()` for everything. The `fromString(...)` would even still work, you'd just match against `name()`, instead of against `getValue()`. I guess you could keep `getValue()`, and just have that return `name()`, as well. Or, axe `getValue()` and call into the enum via the `name()`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
```diff
@@ -110,8 +120,12 @@ public boolean equals(Object obj) {
     public String toString() {
         return new StringBuilder()
```
jamesonwilliams, line 121: For me, Android Studio generates a Lint about these `StringBuilder ` in `toString()`; it prefers just `"foo" + "bar"` style, as gets IDE-generated by default for `toString()`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -426,6 +418,19 @@ public void onError(Exception exception) {
...
 
+        AWSMobileClient.getInstance().getUserAttributes(new Callback<Map<String, String>>() {
+            @Override
+            public void onResult(Map<String, String> result) {
+                sessionBuilder.userSub(result.get("sub"));
+                latch.countDown();
+            }
+
+            @Override
+            public void onError(Exception exception) {
```
jamesonwilliams, line 429: What happens to the content of this error? It looks ignored. How does it get back to the consumer, for auditing/reporting?

TrekSoft:
> What's the `userSub` business?

As part of the architecture discussions we decided that similar to including identity ID, it makes sense to include the userSub which is the UserPool equivalent of user id in this escape hatch session object as well.

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
TrekSoft: Yeah so I was debating this while writing it and was thinking the flexibility of having a different value than the ENUM name might be worthwhile but I guess that could also just make things more confusing so I'll go ahead and drop it.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
TrekSoft: Hmmm... that's weird since StringBuilder is more performant from what I understand. Maybe they don't think it's worth the extra dependency..? Anyway I'll update.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: Currently, we are ignoring it because the most common scenario for a failure is that the user attributes aren't available for the given user type but we still want to return the rest of the information for this session object which is available. However, it would be nice to also return errors for optional inspection so I'll be talking with JIthin and the Architecture team about this as an enhancement in a follow-up PR.

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
TrekSoft: I see we're using fromString elsewhere so I'll update to that

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
jamesonwilliams: I could go either way as well!

core/src/main/java/com/amplifyframework/auth/AuthSignInState.java
```diff
@@ -0,0 +1,101 @@
...
+    DONE,
+
+    /**
+     * Unknown sign-in state, potentially unsupported state.
+     */
+    UNKNOWN;
+
+    @NonNull
+    public static AuthSignInState fromString(String value) {
+        for (AuthSignInState v : values()) {
```
jamesonwilliams, line 93: Better to avoid single letter variable names -- `value`, instead? I think our checkstyle only checks for user-exposed fields like parameter names, though. This isn't a major issue, internal to the method, here.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
jamesonwilliams: Can you add this whole thing into an `/** Why we're ignoring this error ... */` comment?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
jamesonwilliams: I thought it was weird, too. To be fair, I didn't actually look deep enough into it, to tell if the IDE was being reasonable.

TrekSoft:
Will go ahead and incorporate the extra suggestions into my next PR

PR #372, opened 2020-04-10
Enforce a consistent annotation specification style

jamesonwilliams:


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/StorageItemChange.java
```diff
@@ -309,7 +309,7 @@ public String toString() {
      */
     @SuppressWarnings({"unused", "WeakerAccess"})
     @ModelConfig(pluralName = "Records")
-    @Index(fields = {"itemClass"}, name = "itemClassBasedIndex")
```
raphkim, line 312: while it doesn't REALLY matter since this is a system side model, does the CLI model codegen deal with this style appropriately?

jamesonwilliams: No, the code-gen'd stuff has numerous checkstyle issues. (But, the user doesn't apply our library checkstyle config, anyway, so its okay.)

PR #374, opened 2020-04-11
[testutils] Simplify count down logic in HubAccumulator

jamesonwilliams:

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
@@ -100,7 +100,7 @@ public final class AWSDataStorePluginInstrumentedTest {
     public void blogOwnerSavedIntoDataStoreIsThenQueriableInRemoteAppSyncApi() throws DataStoreException, ApiException {
         // Start listening for model publication events on the Hub.
         HubAccumulator outboundModelEventAccumulator =
-            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.PUBLISHED_TO_CLOUD);
+            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.PUBLISHED_TO_CLOUD, 1);
         outboundModelEventAccumulator.clear().start();
 
         // Save Charley Crockett, a guy who has a blog, into the DataStore.
@@ -110,7 +110,7 @@ public final class AWSDataStorePluginInstrumentedTest {
         dataStore.save(localCharley);
 
         // Wait for a Hub event telling us that our Charley model got published to the cloud.
-        outboundModelEventAccumulator.takeOne();
+        outboundModelEventAccumulator.takeAll();
         outboundModelEventAccumulator.stop().clear();
 
         // Try to get Charley from the backend.
@@ -149,7 +149,7 @@ public final class AWSDataStorePluginInstrumentedTest {
                     return false;
                 }
                 return modelWithMetadata.getSyncMetadata().getId().equals(expectedId);
-            });
+            }, 1);
         inboundModelEventAccumulator.clear().start();
 
         // Act: externally in the Cloud, someone creates a record for a blog owner,
@@ -165,7 +165,7 @@ public final class AWSDataStorePluginInstrumentedTest {
 
         // A hub event tells us that a model was created in the cloud;
         // this model was synced into our local store.
-        inboundModelEventAccumulator.takeOne();
+        inboundModelEventAccumulator.takeAll();
         inboundModelEventAccumulator.stop().clear().start();
 
         // Act: externally, the record in the Cloud is updated, to correct the entry's last name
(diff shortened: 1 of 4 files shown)


testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -139,28 +144,20 @@ public HubAccumulator clear() {
...
-    public List<HubEvent<?>> take(int desiredQuantity) {
-        // If we haven't yet received the desired quantity of events on the subscription,
-        // setup a latch to await the desired quantity, less the number of existing events.
-        // For example: I desire 5, I already have 3, I wait for 2 more.
-        if (events.size() < desiredQuantity) {
-            latch = new CountDownLatch(desiredQuantity - events.size());
-            Latch.await(latch);
-            latch = null;
-        }
+    public List<HubEvent<?>> takeAll() {
```
raphkim, line 152: is there any reason to specify "all" now that it's the only option you provide? I would suggest just `take` but that doesn't sound like its very descriptive of what it's doing. What would be a good verb to follow up on accumulating something? `drain`? Maybe `consume`?

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -139,28 +144,20 @@ public HubAccumulator clear() {
...
+        Latch.await(latch);
 
         // If we already had the right number of events,
         // or if our latch counted down as a result of receiving the remaining number of events,
         // return the requested number of events.
```
raphkim, line 157: is this comment still relevant?

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -139,28 +144,20 @@ public HubAccumulator clear() {
      * block until the rest show up. If there are enough, they will
      * be returned immediately. The returned items will be cleared from the
      * accumulator.
-     * @param desiredQuantity Number of items being awaited
      * @return A list of desiredQuantity many items
```
raphkim, line 147: `desiredQuantity` is no longer a parameter

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -139,28 +144,20 @@ public HubAccumulator clear() {
...
-            latch = null;
-        }
+    public List<HubEvent<?>> takeAll() {
+        Latch.await(latch);
 
         // If we already had the right number of events,
         // or if our latch counted down as a result of receiving the remaining number of events,
         // return the requested number of events.
-        List<HubEvent<?>> returning = new ArrayList<>(events.subList(0, desiredQuantity));
+        CopyOnWriteArrayList<HubEvent<?>> returning = new CopyOnWriteArrayList<>(events);
```
raphkim, line 158: Whoa any reason to do this? Will a user ever need to write anything on the returned list? Can't you just return an immutable list of hub events?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
```diff
@@ -181,7 +181,7 @@ public void blogOwnerCreatedAndUpdatedRemotelyIsFoundLocally() throws DataStoreE
 
         // Another HubEvent tells us that an update occurred in the Cloud;
         // the update was applied locally, to an existing record.
-        inboundModelEventAccumulator.take(1);
+        inboundModelEventAccumulator.takeAll();
```
raphkim, line 184: Isn't `HubAccumulator` now not-recyclable after this code change? Wouldn't calling `takeAll()` the second time break it?

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
jamesonwilliams: I like `drain`. I checked Merriam Webster for antonyms of "accumulate", but none of them felt appropriate here. The "all" was added to indicate the collection return type.

`drain` actually describes the behavior a little better though. This method does drain all items out of the accumulator.

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
jamesonwilliams: NO! Thanks, good catch.

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
jamesonwilliams: Thanks, will fix

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
jamesonwilliams: Good point!

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
jamesonwilliams: You're right; it is no longer recycle-able. I would expect that calling `takeAll()` the second time would return an empty list immediately.

jamesonwilliams:

PR #375, opened 2020-04-11
[testutils] Prevent deadlocks in Latch and Await

jamesonwilliams:
`Await.java`: if `Await.result(...)` is called from thread A, and its method
fires callbacks on that same thread A, they may never be invoked. This
is because the `CountDownLatch` is sleeping the thread A. To address this
situation, ensure that the `ResultErrorEmitter` function always run on a
different (new) thread.

`Latch.java`: `InterruptedException`was being gobbled. But, this needs to be
raised to crash the thread. Otherwise, the thread can hang.

### DIFF
--- a/testutils/src/main/java/com/amplifyframework/testutils/Await.java
+++ b/testutils/src/main/java/com/amplifyframework/testutils/Await.java
@@ -134,28 +134,37 @@ public final class Await {
             @NonNull ResultErrorEmitter<R, E> resultErrorEmitter,
             @NonNull AtomicReference<R> resultContainer,
             @NonNull AtomicReference<E> errorContainer) {
-
-        CountDownLatch latch = new CountDownLatch(1);
-
-        resultErrorEmitter.emitTo(
-            result -> {
-                resultContainer.set(result);
-                latch.countDown();
-            }, error -> {
-                errorContainer.set(error);
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<Throwable> unexpectedErrorContainer = new AtomicReference<>();
+        final Thread thread = new Thread(() -> {
+            try {
+                resultErrorEmitter.emitTo(
+                    result -> {
+                        resultContainer.set(result);
+                        latch.countDown();
+                    }, error -> {
+                        errorContainer.set(error);
+                        latch.countDown();
+                    }
+                );
+            } catch (Throwable unexpectedFailure) {
+                unexpectedErrorContainer.set(unexpectedFailure);
                 latch.countDown();
             }
-        );
+        });
+        thread.setDaemon(true);
+        thread.start();
+
+        Latch.await(latch, timeMs);
 
         try {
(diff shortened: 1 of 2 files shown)


testutils/src/main/java/com/amplifyframework/testutils/Await.java
```diff
@@ -134,28 +134,37 @@ private Await() {}
...
+                        latch.countDown();
+                    }
+                );
+            } catch (Throwable unexpectedFailure) {
+                unexpectedErrorContainer.set(unexpectedFailure);
                 latch.countDown();
             }
-        );
+        });
+        thread.setDaemon(true);
```
raphkim, line 155: is there any reason why you want these threads to persist? I feel like all threads should be stopped if the test suite that started it crashes due to a certain error that is unrelated to this thread.

testutils/src/main/java/com/amplifyframework/testutils/Await.java
```diff
@@ -134,28 +134,37 @@ private Await() {}
...
+                        latch.countDown();
+                    }, error -> {
+                        errorContainer.set(error);
+                        latch.countDown();
+                    }
+                );
+            } catch (Throwable unexpectedFailure) {
+                unexpectedErrorContainer.set(unexpectedFailure);
                 latch.countDown();
             }
```
raphkim, line 153: would anything ever get thrown directly to this thread?

testutils/src/main/java/com/amplifyframework/testutils/Await.java
jamesonwilliams: Ahhh!!! Good catch. I set out to make this say _`false`_, (`true` is the default anyway.) But, this does _not_ say `false`, does it.

testutils/src/main/java/com/amplifyframework/testutils/Await.java
jamesonwilliams: ___Not___ if it is working properly. Before, if something was broken, it would throw on the calling thread, here. But since I've moved the call onto a new thread, I don't want the exception to just kill the thread, while the caller is still alive waiting for a response. I want it to kill the _calling_ thread, too.

But yea - this is a safeguard. Software that is functioning as we intent it do would not trigger this condition. (But, this is after-all a _testutils_ component, so we can't expect things passed into this method will always be functioning correctly (they may throw instead of calling error consumer, accidentally.))

testutils/src/main/java/com/amplifyframework/testutils/Await.java
raphkim: ah okay. that sounds like a good reason.

testutils/src/main/java/com/amplifyframework/testutils/Await.java
raphkim: I guess it's always better safe than sorry. I just couldn't think of a scenario where anything would ever be thrown directly here now that you created a new well-contained thread.

PR #391, opened 2020-04-22
Use File instead of String input parameters

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Currently `downloadFile` and `uploadFile` expect a String with the local path for the source file and destination file respectively. To move us closer to iOS and to simplify the iOS, use a File instead as the input parameter.

### DIFF
--- a/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
+++ b/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
@@ -71,13 +71,13 @@ public final class AWSS3StorageDownloadTest {
     private static SynchronousStorage synchronousStorage;
 
     private File downloadFile;
-    private String downloadPath;
     private StorageDownloadFileOptions options;
     private Set<SubscriptionToken> subscriptions;
 
     /**
      * Initialize mobile client and configure the storage.
      * Upload the test files ahead of time.
+     *
      * @throws Exception if mobile client initialization fails
      */
     @BeforeClass
@@ -101,14 +101,12 @@ public final class AWSS3StorageDownloadTest {
         // Upload large test file
         largeFile = new RandomTempFile(LARGE_FILE_NAME, LARGE_FILE_SIZE);
         key = LARGE_FILE_NAME;
-        local = largeFile.getAbsolutePath();
-        synchronousStorage.uploadFile(key, local, uploadOptions, EXTENDED_TIMEOUT_MS);
+        synchronousStorage.uploadFile(key, largeFile, uploadOptions, EXTENDED_TIMEOUT_MS);
 
         // Upload small test file
         smallFile = new RandomTempFile(SMALL_FILE_NAME, SMALL_FILE_SIZE);
         key = SMALL_FILE_NAME;
-        local = smallFile.getAbsolutePath();
-        synchronousStorage.uploadFile(key, local, uploadOptions);
+        synchronousStorage.uploadFile(key, smallFile, uploadOptions);
     }
 
     /**
@@ -128,7 +126,6 @@ public final class AWSS3StorageDownloadTest {
 
         // Create a file to download to
         downloadFile = new RandomTempFile();
-        downloadPath = downloadFile.getPath();
     }
 
     /**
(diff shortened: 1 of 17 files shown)


aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
```diff
@@ -101,14 +101,12 @@ public static void setUpOnce() throws Exception {
         // Upload large test file
         largeFile = new RandomTempFile(LARGE_FILE_NAME, LARGE_FILE_SIZE);
         key = LARGE_FILE_NAME;
-        local = largeFile.getAbsolutePath();
```
jamesonwilliams, line 104: Can we kill the declaration on L96, now? Is `String local` still used?

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadAccessLevelTest.java
```diff
@@ -88,9 +88,8 @@ public void setUp() throws Exception {
         mobileClient.signOut();
 
         // Create a new file to upload
-        File fileToUpload = new RandomTempFile(UPLOAD_SIZE);
-        fileName = fileToUpload.getName();
-        filePath = fileToUpload.getPath();
+        uploadFile = new RandomTempFile(UPLOAD_SIZE);
+        fileName = uploadFile.getName();
```
jamesonwilliams, line 92: This existed before this PR -- but just noting it since I saw it:
`fileName` is a weird name for the first argument, in these various calls to `SynchronousStorage#uploadFile(String, File, StorageOptions)`. I think it should be named `remoteKey`, right?

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
```diff
@@ -144,11 +145,12 @@ public void testGenerateUrlGetsPresignedUrl() throws StorageException {
...
      * {@link StorageDownloadFileResult} with correct file path.
      * @throws StorageException when an error is encountered while downloading
+     * @throws IOException when a temporary file cannot be created
      */
     @Test
-    public void testDownloadToFileGetsFile() throws StorageException {
+    public void testDownloadToFileGetsFile() throws StorageException, IOException {
         final String fromRemoteKey = RandomString.string();
-        final String toLocalPath = RandomString.string();
+        final File toLocalFile = File.createTempFile("download", "data");
```
jamesonwilliams, line 153: We have a utility in the code base called `RandomTempFile`, I think that could be a good replacement to these `File.createTemptFile(..., ...)` calls:
1. It can throw some kind of a `E extends RuntimeException`, which would prevent the need for us to snake this `IOException` through, and onto every `@Test`;
2. Instead of passing `"data"` every time, we could let the `RandomTempFile` populate some random string data.

jpignata: The second parameter is actually a suffix, or file extension, so it's just creating an empty text file called `{prefix}.{suffix}`. `RandomTempFile` sounds better for upload tests, however for download tests we'd still like an empty file.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
jpignata: It isn't, I missed it. TIL: Analyze -> Run Inspection By Name... -> Unused Declarations!

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadAccessLevelTest.java
jpignata: Cool. I thought perhaps there was a reason for this difference I was missing. Renamed.

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/StorageComponentTest.java
jpignata: `RandomTempFile` also can throw an `IOException`, so instead I left the throws declaration on each `@Test` if it was the only one, if not I up-scoped to `Exception`. Let me know if you had something else in mind.

TrekSoft:
Jameson do we still need all the Rx stuff in here now that we're requiring users add the Java 8 language compatability segment to the build?

PR #392, opened 2020-04-22
Expose all GraphQL error properties

richardmcclellan:
*Description of changes:*
The [2018 GraphQL spec](https://spec.graphql.org/June2018/#sec-Errors) specifies that the error type has 4 fields: `message`, `location`, `path`, `extensions`).   The `extensions` field is designed to point to a Map, and can contain any custom error information, as desired by the implementor.   AppSync added some additional fields at the root (`errorType`, `errorInfo`, `data`) of the error object, which goes against the 2018 GraphQL specification (likely because AppSync implemented it prior to 2018, when `extensions` was not part of the spec).   This PR handles this by merged all non-expected fields into the `extensions` map.  If the same field exist in both places, the field inside of extensions take precedence, and the field at the root level is discarded.

After deserialization, the extensions data is stored as a Map<String, Object>.  I considered adding a type parameter, E, to represent the data inside of extensions, but I did not end up going with this approach because it would have needed to be added as a second type parameter to all of the APICategoryBehavior method response types, and that seemed overly complex.  I'm open to other suggestions for how to make this work though, there may be another approach I'm not thinking of.

I did create an `AppSyncExtensions` model object (currently only used by the unit test), which can be used by DataStore to convert from a Map to a strongly-typed object.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.aws;
+
+import com.amplifyframework.api.graphql.GraphQLResponse;
+import com.amplifyframework.api.graphql.error.GraphQLLocation;
+import com.amplifyframework.api.graphql.error.GraphQLPathSegment;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResponse.Error> {
+    private static final String MESSAGE_KEY = "message";
+    private static final String LOCATIONS_KEY = "locations";
+    private static final String PATH_KEY = "path";
+    private static final String EXTENSIONS_KEY = "extensions";
+
+    @Override
+    public GraphQLResponse.Error deserialize(JsonElement json,
+                 Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+        if (!(json.isJsonObject())) {
+            throw new JsonParseException("Expected a JSONObject but found a " +
+                    json.getClass().getName() + " while deserializing error");
+        }
+
+        String message = null;
+        List<GraphQLLocation> locations = null;
+        List<GraphQLPathSegment> path = null;
+        Map<String, Object> extensions = null;
+        JsonObject extensionsJson = new JsonObject();
+        JsonObject nonSpecifiedData = new JsonObject();
+
+        JsonObject error = json.getAsJsonObject();
+
+        for (String key : error.keySet()) {
+            JsonElement value = error.get(key);
+            if (value != null && !value.isJsonNull()) {
+                if (key.equals(MESSAGE_KEY)) {
+                    message = context.deserialize(value, String.class);
+                } else if (key.equals(LOCATIONS_KEY)) {
+                    Type locationsType = new TypeToken<List<GraphQLLocation>>() { }.getType();
+                    locations = context.deserialize(value, locationsType);
+                } else if (key.equals(PATH_KEY)) {
+                    path = getPath(value);
+                } else if (key.equals(EXTENSIONS_KEY)) {
+                    extensionsJson = value.getAsJsonObject();
+                } else {
+                    nonSpecifiedData.add(key, value);
+                }
+            }
+        }
+
+        // Merge nonSpecifiedData into extensions, but don't overwrite anything already there.
+        for (String key : nonSpecifiedData.keySet()) {
+            if (!extensionsJson.has(key)) {
+                extensionsJson.add(key, nonSpecifiedData.get(key));
+            }
+        }
+
+        // Deserialize the extensions JSON to a Map
+        if (extensionsJson.size() > 0) {
+            Type extensionType = new TypeToken<Map<String, Object>>() {
+            }.getType();
+            extensions = context.deserialize(extensionsJson, extensionType);
+        }
+
+        return new GraphQLResponse.Error(message, locations, path, extensions);
+    }
+
+    private List<GraphQLPathSegment> getPath(JsonElement pathElement) {
+        List<GraphQLPathSegment> path = new ArrayList<>();
+        if (!pathElement.isJsonArray()) {
+            throw new JsonParseException("Expected a JsonArray but found a " +
+                    pathElement.getClass().getName() + " while deserializing path");
+        }
+        for (JsonElement element : pathElement.getAsJsonArray()) {
+            JsonPrimitive primitive = (JsonPrimitive) element;
+            if (primitive.isNumber()) {
+                path.add(new GraphQLPathSegment(primitive.getAsInt()));
+            } else if (primitive.isString()) {
+                path.add(new GraphQLPathSegment(primitive.getAsString()));
+            } else {
+                throw new JsonParseException("Expected a String or int, but found a " +
+                        primitive.getClass().getSimpleName() + " while deserializing path segment");
+            }
+        }
+        return path;
+    }
+}
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
@@ -32,7 +32,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-public final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResponse.Error> {
+final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResponse.Error> {
     private static final String MESSAGE_KEY = "message";
     private static final String LOCATIONS_KEY = "locations";
     private static final String PATH_KEY = "path";
@@ -57,11 +57,11 @@ public final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResp
 
         for (String key : error.keySet()) {
             JsonElement value = error.get(key);
-            if (value != null && !value.isJsonNull()) {
+            if (value != null) {
                 if (key.equals(MESSAGE_KEY)) {
                     message = context.deserialize(value, String.class);
                 } else if (key.equals(LOCATIONS_KEY)) {
-                    Type locationsType = new TypeToken<List<GraphQLLocation>>() { }.getType();
+                    Type locationsType = new TypeToken<List<GraphQLLocation>>() {}.getType();
                     locations = context.deserialize(value, locationsType);
                 } else if (key.equals(PATH_KEY)) {
                     path = getPath(value);
@@ -82,8 +82,7 @@ public final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResp
 
         // Deserialize the extensions JSON to a Map
         if (extensionsJson.size() > 0) {
-            Type extensionType = new TypeToken<Map<String, Object>>() {
-            }.getType();
+            Type extensionType = new TypeToken<Map<String, Object>>() {}.getType();
             extensions = context.deserialize(extensionsJson, extensionType);
         }
 
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
@@ -17,9 +17,11 @@ package com.amplifyframework.api.aws;
 
 import com.amplifyframework.api.ApiException;
 import com.amplifyframework.api.graphql.GraphQLResponse;
-import com.amplifyframework.api.graphql.error.AppSyncExtensions;
+import com.amplifyframework.api.graphql.error.GraphQLLocation;
+import com.amplifyframework.api.graphql.error.GraphQLPathSegment;
 import com.amplifyframework.testutils.Resources;
 
+import com.google.gson.internal.LinkedTreeMap;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Before;
@@ -30,6 +32,7 @@ import org.robolectric.RobolectricTestRunner;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -123,9 +126,25 @@ public final class GsonGraphQLResponseFactoryTest {
         // Assert that we parsed the errors successfully.
         assertNotNull(response.getErrors());
 
-        for (GraphQLResponse.Error error : response.getErrors()) {
-            assertEquals(error.getMessage(), "failed");
+        final List<GraphQLResponse.Error> expectedErrors = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
+            String message = "failed";
+            List<GraphQLLocation> locations = Arrays.asList(
+                    new GraphQLLocation(5, 7));
+            List<GraphQLPathSegment> path = Arrays.asList(
+                    new GraphQLPathSegment("listTodos"),
+                    new GraphQLPathSegment("items"),
+                    new GraphQLPathSegment(i),
+                    new GraphQLPathSegment("name")
+            );
+            Map<String, Object> extensions = new LinkedTreeMap<>();
+            extensions.put("errorType", null);
+            extensions.put("errorInfo", null);
+            extensions.put("data", null);
+            expectedErrors.add(new GraphQLResponse.Error(message, locations, path, extensions));
         }
+
+        assertEquals(expectedErrors, response.getErrors());
     }
 
     /**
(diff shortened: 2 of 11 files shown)


core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
```diff
@@ -0,0 +1,42 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.graphql.error;
+
+/**
+ * Location mapping to a particular line and column in the request.
+ */
+public class GraphQLLocation {
```
raphkim, line 21: can it be final?

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
```diff
@@ -0,0 +1,72 @@
...
+     * Convenience method to get this element as an int.
+     *
+     * @return get this element as an int
+     * @throws IllegalStateException if the value contained is not an Integer.
+     */
+    public int getAsInt() {
+        if (isInteger()) {
+            return ((Integer) value).intValue();
+        }
+        throw new IllegalStateException("Not an int: " + value.getClass().getSimpleName());
```
raphkim, line 70: Maybe `ClassCastException` might be a better fit?

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.graphql.error;
+
+public class GraphQLPathSegment {
```
raphkim, line 18: `final`?

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +113,73 @@ public int hashCode() {
...
+
         /**
          * Constructs error response in accordance with GraphQL specs.
          * @param message error message
+         * @param locations list of locations describing the syntax element
+         * @param path The key path of the response field with error.
+         * @param extensions additional error map, reserved for implementors.
          */
-        public Error(@NonNull String message) {
+        public Error(@NonNull String message, List<GraphQLLocation> locations, List<GraphQLPathSegment> path,
```
raphkim, line 135: `@Nullable` annotation on the rest of the parameters would be helpful, since it is a public constructor

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +113,73 @@ public int hashCode() {
...
         }
 
+        /**
+         * Gets the list of locations where each location describes the beginning of an associated
+         * syntax element.
+         *
+         * @return locations
+         */
+        public List<GraphQLLocation> getLocations() {
+            return locations;
```
raphkim, line 159: Let's make the list properties immutable. Luckily, we have nice `Immutable` utility class to help with that :)

```
return Immutable.of(locations);
```

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +113,73 @@ public int hashCode() {
...
+
+        /**
+         * Gets the key path of the response field which experienced the error.  This allows clients
+         * to identify whether a null result is intentional or caused by a runtime error.  The
+         * values are either strings or 0-index integers.
+         *
+         * @return path
+         */
+        public List<GraphQLPathSegment> getPath() {
+            return path;
```
raphkim, line 170: ^

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +113,73 @@ public int hashCode() {
...
+        }
+
+        /**
+         * Returns additional error information of type E.  Reserved for GraphQL implementors to
+         * add details however they see fit.  No additional restrictions on its contents.
+         *
+         * @return extensions
+         */
+        public Map<String, Object> getExtensions() {
+            return extensions;
```
raphkim, line 180: ^

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
raphkim: Also might be helpful to mark the return type as `@Nullable`, along with the other properties.

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
raphkim, line 21: So I see this class doesn't have a real constructor that allows instantiation with specific line + column. This is functional for now since we use GSON to directly write these values. Still, we should provide a constructor or static factory method for other plugin implementers that may not rely on a deserializer.

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
richardmcclellan: Added

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
richardmcclellan: I modeled GraphQLPathSegment off of JsonElement, which has a similar set of methods.  They throw an `IllegalStateException` here, though I tend to agree with you that `ClassCastException` is more logical.  Fixed!

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
richardmcclellan: yep, added.

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
richardmcclellan: Added!

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
richardmcclellan: Added!

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
richardmcclellan: Added!

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
```diff
@@ -0,0 +1,75 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.graphql.error;
+
+import androidx.annotation.NonNull;
+
+public final class GraphQLPathSegment {
```
jamesonwilliams, line 20: Ah crumbs. When I did the checkstyle rules last year,  I think I had intended to include
[`MissingJavadocType`](https://checkstyle.sourceforge.io/config_javadoc.html#MissingJavadocType), so that we remembered to make Javadocs on public entities. I suspect we are missing some around the code base. Anyway, if a user will get this `GraphQLPathSegment` back in response, it would be nice to have doco for any `public` entity they might access. (Not expected to access constructors, but class-level doco may benefit the user?)

core/src/main/java/com/amplifyframework/api/graphql/error/AppSyncExtensions.java
```diff
@@ -0,0 +1,72 @@
...
+
+package com.amplifyframework.api.graphql.error;
+
+import androidx.annotation.Nullable;
+
+import com.amplifyframework.util.Immutable;
+
+import java.util.Map;
+
+public final class AppSyncExtensions {
```
jamesonwilliams, line 24: It'd be good to avoid putting this in the `core` module, if possible. Can we put it under `aws-api`, or `aws-datastore` modules, instead?

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +114,79 @@ public int hashCode() {
...
+         *
+         * @return path
+         */
+        @Nullable
+        public List<GraphQLPathSegment> getPath() {
+            return Immutable.of(path);
+        }
+
+        /**
+         * Returns additional error information of type E.  Reserved for GraphQL implementors to
```
jamesonwilliams, line 180: Whoops I think the JavaDoc is left from a prior pre-commit factoring.

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -110,24 +114,79 @@ public int hashCode() {
...
          * @param message error message
+         * @param locations list of locations describing the syntax element
+         * @param path The key path of the response field with error.
+         * @param extensions additional error map, reserved for implementors.
          */
-        public Error(@NonNull String message) {
+        public Error(@NonNull String message,
+                     @Nullable List<GraphQLLocation> locations,
+                     @Nullable List<GraphQLPathSegment> path,
+                     @Nullable Map<String, Object> extensions) {
```
jamesonwilliams, line 139: (TLDR Ignore this, no action needed.)

------

I just realized how we could do this 100% to spec. It ends up being way overkill, and I still do think we should stick what you have, here.

It would be possible for us to model the extensions as a native GraphQL entity. To do so, we would need Java models for all of the various [types defined in the GraphQL spec](https://graphql.org/learn/schema/).

iOS uses the type `JSONValue`, here. While GraphQL and JSON share some similarities, the type definitions are _not_ the same. As a simple example, even the scalar definitions are slightly different (ref: [GraphQL](https://graphql.org/learn/schema/#scalar-types), [JSON](https://tools.ietf.org/html/rfc7159#section-3).) The `String`-`Object` `Map` in Java is a _looser_ specification of typing than the GraphQL spec.

Part of the scope of work we have taken up with Amplify is to create a lightweight GraphQL client. The AppSync SDK leverages an [Apollo](https://github.com/apollographql/apollo-android) distribution for this purpose. While we building our GraphQL client, we have identified needed functionalities by breadth-first discovery. But in instances like this, we sometimes uncover considerable depth needed to fully model all of the parts of the spec.

I hadn't checked until now. But for what its worth, [Apollo _also does use_ a `String`-`Object` `Map` for this model, too](https://github.com/apollographql/apollo-android/blob/master/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Error.kt#L20).

aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
```diff
@@ -0,0 +1,112 @@
...
+        // Merge nonSpecifiedData into extensions, but don't overwrite anything already there.
+        for (String key : nonSpecifiedData.keySet()) {
+            if (!extensionsJson.has(key)) {
+                extensionsJson.add(key, nonSpecifiedData.get(key));
+            }
+        }
+
+        // Deserialize the extensions JSON to a Map
+        if (extensionsJson.size() > 0) {
+            Type extensionType = new TypeToken<Map<String, Object>>() {
```
jamesonwilliams, line 85: You can add a checkstyle suppression if you want to keep this `{}` together. Or, we could update the checkstyle rules to not enforce this rule for class bodies.

I think this is [`WhitespaceAround`](https://github.com/aws-amplify/amplify-android/blob/master/configuration/checkstyle-rules.xml#L91) right? Might be addressed by adding a property for [`allowEmptyTypes`](https://checkstyle.sourceforge.io/config_whitespace.html#WhitespaceAround_Properties) as `true`?

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
```diff
@@ -122,15 +123,59 @@ public void partialResponseRendersWithTodoDataAndErrors() throws ApiException {
         // Assert that we parsed the errors successfully.
         assertNotNull(response.getErrors());
 
-        final List<GraphQLResponse.Error> expectedErrors = Arrays.asList(
-            new GraphQLResponse.Error("failed"),
-            new GraphQLResponse.Error("failed"),
-            new GraphQLResponse.Error("failed")
-        );
+        for (GraphQLResponse.Error error : response.getErrors()) {
+            assertEquals(error.getMessage(), "failed");
```
jamesonwilliams, line 127: These are backwards, the meaning of the arguments are `assertEquals(expectedValue, actualValue)` -- `"failed"` is the one you expected to find, `error.getMessage()` is the actual value under test.

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
```diff
@@ -122,15 +123,59 @@ public void partialResponseRendersWithTodoDataAndErrors() throws ApiException {
...
+        final GraphQLResponse<ListTodosResult> response =
+                responseFactory.buildSingleItemResponse(partialResponseJson, ListTodosResult.class);
 
-        assertEquals(expectedErrors, actualErrors);
+        // Assert that the response contained things...
+        assertNotNull(response);
+        assertNotNull(response.getErrors());
+        assertEquals(response.getErrors().size(), 4);
+
+        // Assert that each error is has been parsed to the same output object
```
jamesonwilliams, line 158: Under the hood, [`assertEquals(..., ...)`](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/Assert.java#L99-L122) is basically just this:
```
if (!expect.equals(actual)) {
    throw new AssertionError(providedMessage);
}
```

So, so long as your objects have a sensible `equals()` implementation (they do!), you can make high-level assertions on objects using jUnit's `assertEquals(...)` methods.

Here, what I propose is to reword this into one bit assertion statement, like:
```
assertEquals(
    /* expected */ new GraphQLResponse(
        null,
        Arrays.asList(
            new GraphQLError("Confict resolver rejects ...", ,, ) // build these up, from the fields in your assertions
        )
    ),
    actualResponse /* built the same exact way you are now, in the Act portion. */
);
```

core/src/main/java/com/amplifyframework/api/graphql/error/AppSyncExtensions.java
richardmcclellan: I initially put it in `core` since [that's the current approach on iOS](https://github.com/aws-amplify/amplify-ios/pull/368/files#diff-2eab2075f7c39215e0895d8018aed92a) (though this PR is pending refactoring) and I understand we don't really want AppSync specific stuff in core.

AppSyncExtensions is really just a helper object with a couple potential consumers:
(1)  A unit test in `aws-api`
(2) An end user consuming the `aws-api` 
(3) `aws-datastore` could use it to parse the error to handle conflict resolution.

Any of the above three use cases has the ability to access extensions via the Map<String, Object> though.

If we put it in `aws-api`, then it would not be accessible by `aws-datastore`.  

I'm sort of inclined to move it to `aws-datastore` because that satisfies (3).  For (1), we can just move the unit test to datastore, and for (2), not worry about it now since the end user use case isn't totally understood, and they can just use the Map<String, Object>.

@jamesonwilliams I don't have much context on datastore yet, so you tell me, does it make sense to put it in `aws-datastore`?

jamesonwilliams: Hey @richardmcclellan, yea, I think `aws-datastore` should work fine, and be an improvement over `core`.

We have App Sync details in both `aws-api`, and `aws-datastore`, at the moment, with no _real_ rules about how or why.

When we get some time, I think it probably makes sense to create a new module called `appsync`. `aws-api` and `aws-datastore` would impelment the contracts in `core`, by means of leveraging a common `appsync` library module.

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
richardmcclellan: Originally I didn't do this because `GraphQLLocation` (which is only created via gson) didn't have a constructor, but I added that constructor while addressing @raphkim's feedback, essentially because a customer may actually want a constructor.  

I do like the suggestion generally because it is a good way to ensure everything is tested, even if new fields are added later, so I've gone ahead and made this change.  I did keep a couple intermediate asserts though.  When the test succeeds, they are redundant, but when the test fails, the intermediate ones are very useful to debug the problem.

I even caught and fixed a bug in after refactoring this test, which was that elements equal to JsonNull in the error object were getting dropped instead of getting merged into extensions.

Note: I did have to implement `equals` in the GraphQLLocation, GraphQLPathSegment, AppSyncExtensions classes to get this to work, but it's good that it's there now :)

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
richardmcclellan: Fixed!

aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
richardmcclellan: Thanks for the tip!  I updated `checkstyle-rules.xml`

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
richardmcclellan: Got it, I added Javadoc for this class and a few others as well.  If I get a chance later, I will follow up in a separate PR to add MissingJavadocType to checkstyle-rules.xml.

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
richardmcclellan: Yep, fixed!

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
richardmcclellan: Thanks for the thoughts.  I think this would be a small improvement over what I have now, but for now, I think Map<String, Object> works pretty well.  

It would be nice to actually parse extensions as a template type though.  I think this could actually be done by creating a new template type R (response) which wraps a <T extends Model> and an extension type, E.   This would require updating all methods in ApiBehavior though, and I'm not sure if we have the bandwidth to make that change right now.

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
```diff
@@ -122,15 +126,94 @@ public void partialResponseRendersWithTodoDataAndErrors() throws ApiException {
...
+
+        // Assert that each error has been parsed to the same output object.
+        for (int i = 0; i < response.getErrors().size(); i++) {
+            GraphQLResponse.Error actual = response.getErrors().get(i);
+            GraphQLResponse.Error expected = expectedResponse.getErrors().get(i);
+            assertEquals("Unexpected message on error[" + i + "]", expected.getMessage(), actual.getMessage());
+            assertEquals("Unexpected extensions on error[" + i + "]", expected.getExtensions(), actual.getExtensions());
+            assertEquals("Unexpected path on error[" + i + "]", expected.getPath(), actual.getPath());
+            assertEquals("Unexpected locations on error[" + i + "]", expected.getLocations(), actual.getLocations());
+            assertEquals("Unexpected error[" + i + "]", expected, actual);
```
jamesonwilliams, line 211: Assuming that `GraphQLResponse.Error`'s `equals()` implementation considers all of `message`, `extensions`, `path`, `locations`, you should _only_ need L211, and not L207-210.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * For conflict unhandled errors, returns a map containing the same fields as the model type.
+     *
+     * @return data
+     */
+    @Nullable
+    public Map<String, Object> getData() {
+        return Immutable.of(data);
+    }
+
```
jamesonwilliams, line 82: Can you provide a `toString()` implementation, as well?

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
```diff
@@ -139,12 +199,19 @@ public boolean equals(Object thatObject) {
 
             Error error = (Error) thatObject;
 
-            return message.equals(error.message);
+            return ObjectsCompat.equals(message, error.message) &&
+                   ObjectsCompat.equals(path, error.path) &&
+                   ObjectsCompat.equals(extensions, error.extensions) &&
+                   ObjectsCompat.equals(locations, error.locations);
         }
```
jamesonwilliams, line 207: Need to update `GraphQLResponse.Error`'s `toString()` as well, I think. Right now it just only prints the `message` , but not any of your new fields.

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.graphql.error;
```
jamesonwilliams, line 16: Since `GraphQLResponse` is at `com.amplifyframework.api.graphql`, and since `Error lives in it, that means Error is com.amplifyframework.api.graphql`. All of these inner components should live in the same package as well.

In other words, I think this package should be `com.amplifyframework.api.graphql`.

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Returns column number corresponding to position in the request where an error occurred.
+     *
+     * @return column
+     */
+    public int getColumn() {
+        return column;
+    }
+
```
jamesonwilliams, line 47: `toString()` ?

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
```diff
@@ -0,0 +1,110 @@
...
+
+        if (isString() && segment.isString()) {
+            return ObjectsCompat.equals(getAsString(), segment.getAsString());
+        }
+        if (isInteger() && segment.isInteger()) {
+            return ObjectsCompat.equals(getAsInt(), segment.getAsInt());
+        }
+        return false;
+    }
+
```
jamesonwilliams, line 105: `toString()` impl?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncExtensionsTest.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.appsync;
+
+import com.google.gson.internal.LinkedTreeMap;
```
jamesonwilliams, line 18: The `AppSyncExtensions` utility and its test are "above" the abstraction of Gson, at this point. So, we should not import any Gson components as that's a leaking abstraction. If you need a map which preserves insertion order, consider a `LinkedHashMap`, or if you need one that preserves a natural ordering of keys, look at a (Java Collections) `TreeMap`.

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
richardmcclellan: Makes sense, I will move this as well as GraphQLPathSegment!

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLPathSegment.java
richardmcclellan: Will add

core/src/main/java/com/amplifyframework/api/graphql/error/GraphQLLocation.java
richardmcclellan: Will add.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncExtensionsTest.java
richardmcclellan: Good point!  No need to preserve order here, so I'll just use a `HashMap`.  

I also just realized that while `GsonErrorDeserializer` is serializing `extensions` as a `HashMap`, all `JsonObject`s _inside_ of `extensions` are serialized as Gson `LinkedTreeMap`s, since that is the default when using gson.  I will update this to make sure `HashMap`s are used.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
richardmcclellan: will do!

aws-api/src/test/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactoryTest.java
richardmcclellan: Agreed.  I had left these here because it makes it easier to debug when this test fails.  That said, adding toString() implementations to each class should solve that problem, so I'll go ahead and remove them!

core/src/main/java/com/amplifyframework/api/graphql/GraphQLResponse.java
richardmcclellan: Will do!

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncExtensionsTest.java
jamesonwilliams: Oh! Good catch. Getting into the fine points here :-D

PR #393, opened 2020-04-22
Adds getCurrentUser and changePassword

TrekSoft:
- Adds getCurrentUser and changePassword
- Fixes exception handling in the configure method
- Updates signInWithSocial to have proper result object
- Updates result objects to be extendable since plugins may very well have extra information they want to include in their result objects.


core/src/main/java/com/amplifyframework/auth/result/AuthSocialSignInResult.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Wraps the result of a sign in operation.
+     * @param isSignedIn true if the user is now completely signed in, false otherwise
+     */
+    public AuthSocialSignInResult(boolean isSignedIn) {
+        this.isSignedIn = isSignedIn;
+    }
+
+    /**
+     * True if the user is now completely signed in, false otherwise.
```
jamesonwilliams, line 32: In the JavaDoc, can you qualify what "completely" means, here? Does it mean successful completion of any sign-in challenges, and (something about that validation data object?)

core/src/main/java/com/amplifyframework/auth/result/AuthSocialSignInResult.java
```diff
@@ -0,0 +1,76 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.result;
+
+import androidx.core.util.ObjectsCompat;
+
+public class AuthSocialSignInResult {
```
jamesonwilliams, line 20: When would someone extend this? Can you suggest a use case in a class-level JavaDoc? Or, alternately, add `final` to close the class?

How come we `AuthSocialSignInResult` needs to be distinct from just an `AuthSignInResult`?

core/src/main/java/com/amplifyframework/auth/result/AuthSignUpResult.java
```diff
@@ -19,7 +19,7 @@
 
 import com.amplifyframework.auth.AuthCodeDeliveryDetails;
 
-public final class AuthSignUpResult {
+public class AuthSignUpResult {
```
jamesonwilliams, line 22: Nothing in this PR actually extends it, though, right? Why open it?

core/src/main/java/com/amplifyframework/auth/result/AuthSignInResult.java
```diff
@@ -20,7 +20,7 @@
 import com.amplifyframework.auth.AuthCodeDeliveryDetails;
 import com.amplifyframework.auth.AuthSignInState;
 
-public final class AuthSignInResult {
+public class AuthSignInResult {
```
jamesonwilliams, line 23: Why open?

core/src/main/java/com/amplifyframework/auth/AuthUser.java
```diff
@@ -52,6 +55,24 @@ public String getUsername() {
...
     }
 
+    /**
+     * Interface for changing the password of an existing user.
+     * @param oldPassword The user's existing password
+     * @param newPassword The new password desired on the user account
+     * @param onSuccess Success callback
+     * @param onError Error callback
+     */
+    public abstract void changePassword(
```
jamesonwilliams, line 65: If `AuthUser` is a valid working user, and then `changePassword(...)` is called and fails, then `AuthUser` is no longer a valid working user. This might have significance to the auth system, and to the app that integrates it. How can other components come to know that that `AuthUser` no longer has a valid password? Only by attempting to use the instance?

This also makes it look like there's one auth, and it mutates. Is that the case? If so, should you make this a singleton? If so, you could end up with an access pattern like `Amplify.Auth.User.changePassword(...)` which is certainly very _readable_, anyway.

Instead of making an `abstract` class, and allowing an implementer to add a variety of customizations onto this class, you could accomodate the `changePassword(...)` by accepting a `ChangePasswordStrategy` into the constructor of the `final` `AuthUser`. e.g.,

```
@FunctionalInterface interface ChangePasswordStrategy {
    void changePassword(String oldPass, String newPass, Action onSuccess, Consumer<AuthException> onFailure);
}
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthUser.java
```diff
@@ -0,0 +1,72 @@
...
+import com.amplifyframework.auth.AuthUser;
+import com.amplifyframework.core.Action;
+
+import com.amazonaws.mobile.client.AWSMobileClient;
+import com.amazonaws.mobile.client.Callback;
+
+import java.util.Objects;
+
+public final class AWSCognitoAuthUser extends AuthUser {
+    private AWSMobileClient client;
```
jamesonwilliams, line 31: `final`

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -69,35 +76,64 @@ public String getPluginKey() {
     public void configure(
             @NonNull JSONObject pluginConfiguration,
             @NonNull Context context
-    ) throws AmplifyException {
+    ) throws AuthException {
         final CountDownLatch latch = new CountDownLatch(1);
+        final AsyncResult<Exception> asyncException = new AsyncResult<>();
 
         AWSMobileClient.getInstance().initialize(
```
jamesonwilliams, line 83: We had agreed to put this in `configure(...)` since we read the code and it looked like it was only doing configuration.

This current PR is adding initialization logic into the callback, though. We should split that out into the `initialize()` hook.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -69,35 +76,64 @@ public String getPluginKey() {
...
                 @Override
                 public void onError(Exception error) {
-                    throw new IllegalStateException("Failed to instantiate AWSMobileClient", error);
+                    asyncException.set(error);
+                    latch.countDown();
                 }
             }
         );
 
         try {
```
jamesonwilliams, line 115: The three exceptions being created from within this try are all very similar. Think of some ways you can create one error, and throw it if any of the scenarios arrise.

core/src/main/java/com/amplifyframework/auth/result/AuthSignUpResult.java
TrekSoft: By design, we want to enable plugin developers to put any plugin specific return values in an extension of this class.

core/src/main/java/com/amplifyframework/auth/result/AuthSignInResult.java
TrekSoft: Same as above

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -69,35 +76,64 @@ public String getPluginKey() {
...
+        final AsyncResult<Exception> asyncException = new AsyncResult<>();
 
         AWSMobileClient.getInstance().initialize(
             context,
             new AWSConfiguration(pluginConfiguration),
             new Callback<UserStateDetails>() {
                 @Override
                 public void onResult(UserStateDetails result) {
-                    latch.countDown();
+                    if (result.getUserState().equals(UserState.SIGNED_IN)) {
```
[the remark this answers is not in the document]
TrekSoft: Ah, good call out - fixing.

core/src/main/java/com/amplifyframework/util/AsyncResult.java
```diff
@@ -0,0 +1,55 @@
...
+package com.amplifyframework.util;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+/**
+ * A simple class which starts empty and will indicate if it's ever been set with a value and if so, what that value is.
+ * @param <T> The type of result stored in this object.
+ */
+public class AsyncResult<T> {
```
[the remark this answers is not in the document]
TrekSoft: Ugh - thanks - I figured there had to be a default class for doing this but just couldn't find it. Will update.

core/src/main/java/com/amplifyframework/auth/result/AuthSignUpResult.java
TrekSoft: My understanding is that, in general, Objects in the core library for a category should be non-final because, by nature, they are simply capturing what must be common across all plugins but we also expect there to be plugin specific values which we don't want be put up at the core level but rather extended at the plugin implementation level.

core/src/main/java/com/amplifyframework/auth/result/AuthSocialSignInResult.java
TrekSoft: While the core value of isSignedIn is common to all sign in results, the nextStep and possible additional data for the results of the different sign in method can be completely different - AWSMobileClient already has a different result object for the federatedSignIn and showSignIn methods than normal sign in and there's reason to believe other plugins would as well.

core/src/main/java/com/amplifyframework/auth/AuthUser.java
TrekSoft: Let's talk about this on Chime - I'm not fully understanding the concerns you raise.

core/src/main/java/com/amplifyframework/auth/result/AuthSocialSignInResult.java
TrekSoft: Actually in thinking about it, I decided to remove the completely term since I could imagine a scenario where a user is signed in but there are still steps suggested to be taken so isSignedIn here should just be simply whether the user is actually authenticated. Also updated the comment to the technical term of authenticated to be precise.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: I'll catch up with you on this on a brief call

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: Also need clarification on this so I'll ask about it on our call

core/src/main/java/com/amplifyframework/auth/result/AuthSocialSignInResult.java
TrekSoft: Possibly implement a common interface for isSignedIn

core/src/main/java/com/amplifyframework/auth/result/AuthSignUpResult.java
jamesonwilliams: [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle (of [SOLID](https://en.wikipedia.org/wiki/SOLID)) is:
> software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification

In practice, there are two principle ways to achieve extension points: inheritance, and composition. Software texts like Design Patterns tell us to [prefer composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance).

Even so, there are [benefits](https://en.wikipedia.org/wiki/Composition_over_inheritance#Benefits) and [drawbacks](https://en.wikipedia.org/wiki/Composition_over_inheritance#Drawbacks) to using composition over inheritance, and both can be valuable tools.

_Personally_, the issues I have with inheritance are mostly a few:
1. Base classes grow over time, to include muddled concerns; `AbstractFooBase` (what is it?)
2. With a closed class implementation, you can encapsulate all of the policy and assumptions you hold about the function of the software. When you open a class, there is no guarantee that the derived class will meet all of these assumptions.

An interface has problem (2). An open non-abstract class is susceptible to (1). An abstract class may be used to enforce policy while offering means for extension/composition, and so contains partial mitigation to both (1) and (2).

Ref:
- [Delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern)

PR #371, opened 2020-04-10
[aws-datstore][DRAFT] Prototype of user-provided configs

jamesonwilliams:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #395, opened 2020-04-23
[datastore] Update dataStore configuration for consistency with iOS 

rjuliano:
This PR builds on #371. I can also push my commit there if that's preferred, but didn't what to hijack Jameson's PR :)

Summary of my changes

## [core] Category and AmplifyConfiguration
Modified `forCategoryType` to return an empty `CategoryConfiguration` object instead of throwing an exception if it can't find the configuration key in the JSON file. The implication here is that it passes the burden onto the plugins to have sane defaults. I don't think that's an unreasonable thing but definitely open for discussion. 

One option I considered was to have an abstract method added to `CategoryConfiguration` called `defaults()` for example.

## [datastore] DataStoreConfiguration
- Added syncInterval, syncPageSize and syncMaxRecords to the `ConfigKeys` enum. Naming lines up with [iOS](https://github.com/aws-amplify/amplify-ios/pull/369)
- Added two overloads of the `builder()` method. One that takes in a `JSONObject` and another that allows developers to pass in the application context from which the `amplifyconfiguration.json` can be read. This allows a `Builder` instance to be returned with settings from the JSON file while still allowing developers to specify the error and conflict handler if desired.
- Account for the additional conifg keys.

## [datastore] AWSDataStorePlugin
- Created an overload of the `forModels` method which takes in an instance of `DataStoreConfiguration`

## [datastore] DataStoreConflictHandlerResult
- Renamed the items in the enum based on discussions with the iOS team.
    - APPLY_REMOTE: discarded local model and apply the remote changes to the local store
    - RETRY_LOCAL: retried to save the local model to the remote store. Need to add implementation for this handler still.
    - RETRY: conflict was handled by passing in a new version of the model to the remote store which will eventually sync with the local store. This would have to be returned from a custom handler provided by the developer.

Please let me know if I'm off on these names.

## [datastore] DiscardingConflictHandler -> ApplyRemoteConflictHandler
Renamed the class to align with the enum names mentioned above

## DX
From a developer experience standpoint, this is what the initialization code would look like

Java
```java
ModelProvider modelProvider = AmplifyModelProvider.getInstance();
DataStoreConfiguration dataStoreConfig = DataStoreConfiguration
    .builder(getApplicationContext(), AWSDataStorePlugin.PLUGIN_CONFIG_KEY) //creates a builder that reads from the JSON file
    .dataStoreConflictHandler(ApplyRemoteConflictHandler.instance(DefaultDataStoreErrorHandler.instance())) //or passes in the a custom handler
    .dataStoreErrorHandler(DefaultDataStoreErrorHandler.instance()) //or passes in a custom handler
    .build();
Amplify.addPlugin(new AWSApiPlugin());
Amplify.addPlugin(AWSDataStorePlugin.forModels(modelProvider, dataStoreConfig));
```
Kotlin
```kotlin
val modelProvider: ModelProvider = AmplifyModelProvider.getInstance()
val dataStoreConfig = DataStoreConfiguration
    .builder(applicationContext, AWSDataStorePlugin.PLUGIN_CONFIG_KEY) //creates a builder that reads from the JSON file
    .dataStoreConflictHandler(ApplyRemoteConflictHandler.instance(DefaultDataStoreErrorHandler.instance())) //or passes in the a custom handler
    .dataStoreErrorHandler(DefaultDataStoreErrorHandler.instance()) //or passes in a custom handler
    .build()
Amplify.addPlugin(AWSApiPlugin())
Amplify.addPlugin(AWSDataStorePlugin.forModels(modelProvider, dataStoreConfig));
```


jpignata:
Thanks for the details. Perhaps answered in the code, but: would I be required to call `dataStoreConflictHandler` and  `dataStoreErrorHandler` during initialization or does this assume the defaults? Would I even need to pass in `DataStoreConfiguration` to the plugin factory?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -206,4 +288,41 @@ public DataStoreConfiguration build() {
...
         }
     }
+
+    enum ConfigKey {
+        /**
+         * At most one base sync will be performed within this interval of time.
+         * The interval is expressed in milliseconds.
+         */
+        SYNC_INTERVAL("syncInterval"),
+        SYNC_PAGE_SIZE("syncPageSize"),
```
jpignata, line 298: Should we write descriptions for these as well?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
```diff
@@ -25,17 +25,17 @@
...
     /**
      * Conflict is handled by keeping the local (client-side) changes, discarding whatever
      * was on the server.
      */
-    KEEP,
+    RETRY_LOCAL,
 
     /**
-     * Conflict is handled by return a new `Model` instance that should used
+     * Conflict is handled by return a new `Model` instance that should be used
```
jpignata, line 37: returning?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
```diff
@@ -25,17 +25,17 @@
      * Conflict is handled by discarding the local (client-side) changes, preferring whatever
      * was on the server.
      */
-    DISCARD,
+    APPLY_REMOTE,
 
     /**
      * Conflict is handled by keeping the local (client-side) changes, discarding whatever
      * was on the server.
```
jpignata, line 32: Thinking about this description and its precision. The conflict will be handled by retrying the local change which attempts to persist it again on the backend. Is it a given that the end result "discarding whatever was on the server" will occur, or could it result in another conflict (say, from another client)?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
jpignata: I'm having a hard time understanding `RETRY` versus `RETRY_LOCAL`.  What version of attributes will the new model reflect?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -98,27 +102,44 @@ private AWSDataStorePlugin(
      * the supplied {@link ModelProvider}. If remote synchronization is enabled, it will be
      * performed through {@link Amplify#API}.
      * @param modelProvider Provider of models to be usable by plugin
+     * @param dataStoreConfiguration Instance of the data store configuration object
```
jpignata, line 105: Nit: Stylized as DataStore

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -58,6 +70,76 @@ public static Builder builder() {
...
+    public static Builder builder(@NonNull Context context, @NonNull String pluginConfigKey) throws DataStoreException {
+        Objects.requireNonNull(context);
+        Objects.requireNonNull(pluginConfigKey);
+        JSONObject pluginConfig = null;
+        try {
+            pluginConfig = Objects.requireNonNull(AmplifyConfiguration.fromConfigFile(context)
+                .forCategoryType(CategoryType.DATASTORE))
+                .getPluginConfig(pluginConfigKey);
+        } catch (AmplifyException | NullPointerException exception) {
+            Log.w(TAG, "Unable to read DataStore configuration from file.", exception);
```
jpignata, line 90: I'm a little unclear on our approach to logging. Should this use a unified logger, a la:

```java
private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore);
```

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -83,7 +83,7 @@ public final synchronized void configure(
                 for (P plugin : getPlugins()) {
                     String pluginKey = plugin.getPluginKey();
                     JSONObject pluginConfig = configuration.getPluginConfig(pluginKey);
-                    plugin.configure(pluginConfig, context);
+                    plugin.configure(pluginConfig != null ? pluginConfig : new JSONObject(), context);
```
jpignata, line 86: Is this how we handle this on other platforms? I'd imagine we want users to explicitly configure their use of any plugin via the CLI, so I'm not sure this is ideal. What problem are you tackling here?

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -169,13 +169,14 @@ private static JSONObject readInputJson(Context context, int resourceId) throws
...
         final CategoryConfiguration categoryConfiguration =
             categoryConfigurations.get(categoryType.getConfigurationKey());
-        if (categoryConfiguration == null) {
-            throw new AmplifyException(
-                "Unknown/bad category type: " + categoryType,
-                "Be sure to use one of the supported Categories in your current version of Amplify"
-            );
-        }
-        return categoryConfiguration;
+        return categoryConfiguration != null ? categoryConfiguration :
```
jpignata, line 172: Will this return a dummy `CategoryConfiguration` for any key I pass? For example, if I pass a key for category `pretend`, I'll get back a Null-y object for it?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -53,6 +53,10 @@
  * An AWS implementation of the {@link DataStorePlugin}.
  */
 public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
+    /**
+     * Configuration key name of the plugin in the amplify config file.
```
jpignata, line 57: Nit: Amplify

core/src/main/java/com/amplifyframework/core/category/Category.java
jpignata: Idle thought: In general I'm anti-ternary for readability - do we have an opinion in our own team style guide?

rjuliano:
> Thanks for the details. Perhaps answered in the code, but: would I be required to call `dataStoreConflictHandler` and `dataStoreErrorHandler` during initialization or does this assume the defaults? Would I even need to pass in `DataStoreConfiguration` to the plugin factory?

You don't need to. You can just call `DataStoreConfiguration.builder().build()` and it will assume default values

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -58,6 +70,76 @@ public static Builder builder() {
...
+            } catch (IllegalArgumentException noSuchConfigKey) {
+                throw new DataStoreException(
+                        "Saw unexpected config key: " + keyString,
+                        "Make sure your amplifyconfiguration.json is valid."
+                );
+            }
+
+            try {
+                switch (configKey) {
+                    case SYNC_INTERVAL:
```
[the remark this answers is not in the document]
rjuliano: iOS appears to be using milliseconds while JS uses minutes. Probably something we need to align on. @drochetti @wooj2 @palpatim @jamesonwilliams 

I don't have a super strong opinion one way or another, but milliseconds does leave the door open for some potential chattiness.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
rjuliano: That was my miss. I'll add that. Checkstyle didn't flag because the interface is not public, but it definitely warrants comments.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
rjuliano: RETRY_LOCAL will try to update the remote data store with the remote model. While RETRY would be returned from a custom conflict handler that would allow the developer to retry with a model that may be a mix of the remote and local (if it makes sense for their use case). RETRY_LOCAL is kind of a shortcut for RETRY(localModel)

There was some discussion on RETRY when we met with the iOS team, but I think we'll need to have some solid documentation behind this one.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
rjuliano: I'm going to update the wording here. Will also remove the discarding part because the retry may actually result in another conflict

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: done

core/src/main/java/com/amplifyframework/core/category/Category.java
rjuliano: The goal here is to prevent the issue we're having when the dataStore (or any other category) is not in the `amplifyconfiguration.json` file but the developer adds it as a plugin. Definitely more than one way to skin the cat here...let me give it some more thought

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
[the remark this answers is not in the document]
drochetti: on Swift we follow the platform standard of representing the time intervals with `Double`. However, there's an idiomatic way of doing that, so our actual configuration code looks like this:

```swift
.custom(
    syncTimeInterval: .minutes(10)
)
```

Therefore the code itself is explicit about the time unit being used. Behind the scenes the Double represents the time in **seconds**. But since it's a Double, milliseconds can be achieved by using decimal points (or through `.milliseconds(500)`). At least the intent is explicit.

https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift#L76

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
rjuliano: You can only pass one of the members of the `CategoryType` enum. It will return an empty CategoryConfiguration instance of the type you requested if one was no loaded via the config file. 

This is also related to handling plugins that get added by the developer without corresponding entries in the config file, which seems to be the case with DataStore for the most part since the CLI does not add that to the config file. Perhaps this is something we can handle in a separate PR?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: done

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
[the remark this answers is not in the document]
jpignata: Would it be reasonable for Android to use minutes?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -206,4 +291,48 @@ public DataStoreConfiguration build() {
...
+        public String toString() {
+            return key;
+        }
+
+        /**
+         * Lookup a config key by its string value.
+         * @param anything Any string found in the key position of the plugin config JSON
+         * @return An enumerate config key
+         */
+        @SuppressWarnings("unused")
```
jamesonwilliams, line 328: I think this is from an earlier phase of this code's evolution, where this method was not being used, yet. But, it does get used now, up on L113. So we could remote this suppression.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandlerResult.java
```diff
@@ -22,20 +22,20 @@
...
+     * The conflict handled by retrying the local change which attempts to persist
+     * it again on the backend.
      */
-    KEEP,
+    RETRY_LOCAL,
 
     /**
-     * Conflict is handled by return a new `Model` instance that should used
+     * Conflict handled by returning a new `Model` instance that should be used
      * instead of the local and remote changes.
```
jamesonwilliams, line 38: The description here is not as good as what you wrote in the PR description. To that point, your PR description was actually better than most of the existing JavaDoc in the code. I wonder if you can find places to copy your PR description documentation into the JavaDoc? Please feel more than free to remove/overwrite anything that I had.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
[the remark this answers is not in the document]
rjuliano: We could modify the `DataStoreConfiguration` constructor and rename the `syncIntervalMs` to `syncInterval` and the do the following in the constructor body:

```
this.syncInterval = TimeUnit.MINUTES.toMillis(syncInterval);
```

Thoughts?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -98,27 +102,44 @@ private AWSDataStorePlugin(
      * the supplied {@link ModelProvider}. If remote synchronization is enabled, it will be
      * performed through {@link Amplify#API}.
      * @param modelProvider Provider of models to be usable by plugin
+     * @param dataStoreConfiguration Instance of the data store configuration object
      * @return An {@link AWSDataStorePlugin} which warehouses the provided models
      */
     @NonNull
     @SuppressWarnings("WeakerAccess")
-    public static AWSDataStorePlugin forModels(@NonNull ModelProvider modelProvider) {
+    public static AWSDataStorePlugin forModels(
```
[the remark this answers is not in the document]
rjuliano: That's a good point. Looking at the pattern across the other plugins, it seems like we have `Builder` objects for configuration objects and constructor initialization for plugins. Probably a change worth making unless there are objections.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -169,13 +169,14 @@ private static JSONObject readInputJson(Context context, int resourceId) throws
...
+            new CategoryConfiguration() {
```
jamesonwilliams, line 173: If I have a config file like:
```
{
    "foo": ...
}
```
I believe the current behavior is that I'll get a non-null config for `FooCategory`, but I'd get a `null` config for `BarCategory`, right? I do actually thing that's a correct use of optionality in Java. The JSON document does not contain/specify any config for `"bar"`.

Is there anyway we can preserve that attribute of the system?

If not, as a mechnical note to simplify the current code block, I'd suggest pulling this anonymous `CategoryConfiguration` out into a subclass. You could make like a :
```
final class EmptyCategoryConfiguration extends CategoryConfiguration {
    private final CategoryType categoryType;

    private EmptyCategoryConfiguration(CategoryType categoryType) {
        this.categoryType = categoryType;
    }

    static EmptyCategoryConfiguration forCategory(CategoryType categoryType) {
        Objects.requireNonNull(categoryType);
        return new EmptyCategoryConfiguration(categoryType);
    }
    
    @NonNull
    @Override
    public CategoryType getCategoryType() {
        return categoryType;
    }
}
```
```
EmptyCategoryConfiguration.forCategory(CategoryType.DATASTORE); // Java is verbose isn't it, gee.
```
Hm. Double check the source and test code for current extensions of `CategoryConfiguration` though - I feel like something like this might already be there? (I don't remember.) It probably has a different name, if it exists. You could "promote" it from test to source code and fixup the name, if its around.

core/src/main/java/com/amplifyframework/core/category/Category.java
jamesonwilliams: I think upgrading to this non-null object ~here is probably a good idea. Jotted down my thinking --

-------
Let's suppose we have a config file like:
```
{
    "fooCategory": {
        "plugins": {
            "fooCategoryPlugin": {
            }
        }
     },
     "barCategory": {
         "plugins": {
             "barCategoryPlugin": {
                "key1": "val1"
             }
         }
     },
     "bazCategory": {
         "plugins": {
         }
     }
}
```
Potentially, the specification of these three categories imply three distinct things:
1. For _foo_, it looks to me like "Hey, use this plugin with default values."
2. For _bar_, it looks like "Hey, use this plugin, and use my values."
3. For _baz_, it looks like: "Huh? What's a baz plugin? we're not expecting to use that kind of a plugin."

If that's so, then we might have three corresponding ways to represent this in Java:
1. `new JSONObject()` with no contents;
2. `new JSONObject().put("key1", "val2")` representing the chosen keys;
3. `null`.

The discrepancy arrises could situations where the config file enumerates config for plugins _p1...pn_, but the user, at runtime, enumerates `π1... πn` via `addPlugin(...)`.

For each plugin, there six code paths to consider, while `configure(...)`ing a plugin at run-time:
1. File has populated config, user doesn't add plugin.
2. File has empty config, user doesn't add plugin.
3. File doesn't have config, user doesn't add plugin.
4. File has populated config, user adds plugin.
5. File has empty config, user adds plugin.
6. File doesn't have config, user adds plugin.

1-3 are probably a non-issue - we can just ignore the unused plugin blocks.
4 - We should probably validate the config block, then use the values
5 - We should whatever default plugin values we can, throw only if config file missing some required value
6 - This is the interesting one, I guess.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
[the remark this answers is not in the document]
drochetti: @rjuliano that was our path on iOS. We respected the platform default, of expressing seconds as a `Double` and make sure we have an API to clearly express the duration unit (without the `Ms` suffix).

My 2 cents is that maybe a wrapper around `TimeUnit` could be a good way to express the API that would work well in both Java and Kotlin (and very similar to the one in Swift):

```java
import static com.amplifyframework.util.TimeUnitUtils.minutes; // Utils for a lack of better word now

Long syncInterval = minutes(5); // delegates to TimeUnit.MINUTES.toMillis(5);
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -206,4 +291,48 @@ public DataStoreConfiguration build() {
             );
         }
     }
+
+    enum ConfigKey {
+        /**
+         * At most one base sync will be performed within this interval of time.
+         * The interval is expressed in milliseconds.
+         */
+        SYNC_INTERVAL("syncInterval"),
```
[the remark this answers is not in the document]
drochetti: I posted my 2 cents there https://github.com/aws-amplify/amplify-android/pull/395#discussion_r414137596

I completely agree we need to be crystal clear on how time duration and intervals are expressed. I think name is one way and I don't disagree with it being a good option.

However, I think having an expressive API like `syncInterval = minutes(5);` could be even better. Intent is clear, customer doesn't even need to know if that's backed up by milliseconds, all they know is "it's 5 minutes".

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -53,6 +53,10 @@
  * An AWS implementation of the {@link DataStorePlugin}.
  */
 public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
+    /**
+     * Configuration key name of the plugin in the Amplify config file.
+     */
+    public static final String PLUGIN_CONFIG_KEY = "awsDataStorePlugin";
```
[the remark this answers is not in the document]
drochetti: I thought these constants could also be used when referencing the plugin directly (for escape hatch use cases).

```java
Amplify.DataStore.getPlugin(AWSDataStorePlugin.PLUGIN_CONFIG_KEY);
```

Which makes me wonder now, shouldn't we get plugins by their class `Amplify.DataStore.getPlugin(AWSDataStorePlugin.class)`? Do we (need to) support two instances of the same plugin implementation?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -58,6 +71,78 @@ public static Builder builder() {
...
+     * Begin building a new instance of {@link DataStoreConfiguration} by reading dataStore
+     * settings from the config file.
+     * @param context application context which will be used to retrieve amplicationconfiguration.json
+     * @param pluginConfigKey the desired dataStore plugin config key (i.e. awsDataStorePlugin)
+     * @return A new builder instance
+     * @throws DataStoreException exception thrown if there's an unexpected configuration key or
+     * an invalid configuration value
+     */
+    @NonNull
+    public static Builder builder(@NonNull Context context, @NonNull String pluginConfigKey) throws DataStoreException {
```
drochetti, line 84: is `Context` require at this point? I think of the configuration object should be a simple pojo. It won't execute any configuration logic, but just carry their values. So the context should be used when the Plugin's `configure(JSONObject pluginConfiguration, @NonNull Context context)` is invoked. Your pojo is available to execute any configuration workflow you need at that moment of the plugin lifecycle.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -58,6 +71,78 @@ public static Builder builder() {
...
+        Objects.requireNonNull(pluginConfigKey);
+        JSONObject pluginConfig = null;
+        try {
+            pluginConfig = Objects.requireNonNull(AmplifyConfiguration.fromConfigFile(context)
+                .forCategoryType(CategoryType.DATASTORE))
+                .getPluginConfig(pluginConfigKey);
+        } catch (AmplifyException | NullPointerException exception) {
+            LOG.warn("Unable to read DataStore configuration from file.", exception);
+        }
+        return pluginConfig != null ? builder(pluginConfig) : builder();
```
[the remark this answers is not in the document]
drochetti: I posted a comment above (https://github.com/aws-amplify/amplify-android/pull/395/files#r414157825) also related to this. I believe the configuration logic should be executed by the plugin's `configure(configuration, context)` method, that way the sequence is enforced by `Amplify.configure(context)`. The `DataStoreConfiguration` should just be a pojo holding the config values.

rjuliano:
Closing this in favor of #401

PR #396, opened 2020-04-23
[aws-api] OkHttp to support all claimed API levels

jamesonwilliams:
As of OkHttp 3.13, support was dropped for API levels below 21. To
maintain support of these older versions of Android, a few changes are
needed, as discussed in OkHttp's blog.

Refer: https://developer.squareup.com/blog/okhttp-3-13-requires-android-5/
Refer: https://medium.com/tech-quizlet/working-with-tls-1-2-on-android-4-4-and-lower-f4f5205629a

### DIFF
--- a/aws-api/build.gradle
+++ b/aws-api/build.gradle
@@ -21,6 +21,8 @@ dependencies {
     api project(':core')
 
     implementation dependency.androidx.appcompat
+    implementation dependency.google.playservices.base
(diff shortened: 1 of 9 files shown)


jamesonwilliams:
There's some good work here, but this PR is going stale and is not actively being worked on. Will re-open if that changes. For now, Amplify Android is 21+ plus. Adults only! No, anyone is welcome. But you should target Android 5.0 as your `minSdk`.

PR #397, opened 2020-04-24
[aws-datastore] AWSDataStorePlugin via Constructor

jamesonwilliams:
Disposition 1. All of the other plugins currently used in Amplify can be
created by constructor. The `AWSDataStorePlugin.forModels(...)` factory
breaks the customer's expectations of how plugins are instantiated.

Disposition 2. It is possible to dynamically locate the code-generated
model provider, internal to the plugin, and use it to instantiate the
plugin. To simplify the happy path of on-boarding with the DataStore, a
no-arguments constructor is provided.

The plugin may now be added into the system by the same means as the
other plugins:

    Amplify.addPlugin(new AWSDataStorePlugin());


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -86,33 +86,50 @@ private AWSDataStorePlugin(
...
 
     /**
-     * Creates an {@link AWSDataStorePlugin} which can warehouse the model types provided by
+     * Constructs an {@link AWSDataStorePlugin} which can warehouse the model types provided by
+     * your application's code-generated model provider. This model provider is expected to have the
+     * fully qualified class name of com.amplifyframework.datastore.generated.model.AmplifyModelProvider.
+     *
+     * This constructor will attempt to find that provide by means of reflection. If you have changed
+     * the path to your {@link ModelProvider} instance, and do not use the default path generated by
+     * the Amplify Code Generator, then this will be break. Likewise, if you need to provide a custom
```
rjuliano, line 95: Minor: 

```suggestion
     * the Amplify Code Generator, then this will break. Likewise, if you need to provide a custom
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -86,33 +86,50 @@ private AWSDataStorePlugin(
...
+     * {@link ModelProvider}, this will not work. In both cases, you should prefer one of the other
+     * overloads such as {@link #AWSDataStorePlugin(ModelProvider)}.
+     *
+     * If remote synchronization is enabled, it will be performed via {@link Amplify#API}.
```
rjuliano, line 99: When I started looking at DataStore, this was one of the points of confusion for me. How do I tell if remote sync is enabled? It doesn't appear to be a setting in the category or plugin configuration, so I'm asserting that it's simply by virtue of an API being present. If that's the case, maybe we should say something like:

> If the API plugin is present and configured, then remote synchronization will be performed.

Or something along those lines

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -86,33 +86,50 @@ private AWSDataStorePlugin(
...
+     *
+     * @throws DataStoreException If it is not possible to access the code-generated model provider
+     */
+    @SuppressWarnings({"unused", "WeakerAccess"})
+    public AWSDataStorePlugin() throws DataStoreException {
+        this(ModelProviderLocator.locate(), Amplify.API);
```
drochetti, line 105: oh wow, this is fancy! I like it.

I wasn't expecting this to happen anytime soon, you're a ninja @jamesonwilliams

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -86,33 +86,50 @@ private AWSDataStorePlugin(
...
-     * @return An {@link AWSDataStorePlugin} which warehouses the provided model types
      */
-    @NonNull
-    public static AWSDataStorePlugin create(@NonNull ModelProvider modelProvider, @NonNull GraphQlBehavior api) {
-        Objects.requireNonNull(modelProvider);
-        Objects.requireNonNull(api);
-        ModelSchemaRegistry modelSchemaRegistry = ModelSchemaRegistry.instance();
-        return new AWSDataStorePlugin(modelSchemaRegistry, modelProvider, api);
+    @SuppressWarnings("WeakerAccess")
+    public AWSDataStorePlugin(@NonNull ModelProvider modelProvider, @NonNull GraphQlBehavior api) {
```
drochetti, line 127: on iOS we're treating this constructor as internal, only for dependency injection purposes. Customers cannot pass their own API implementation because of how strict our sync engine is when it comes to how API should behave.

jamesonwilliams: This is a good point. I'll look to remove the `public` access.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: Good point, done!

PR #399, opened 2020-04-25
Implement online entity detection

raphkim:
*Issue #, if available:*

*Description of changes:*
* Add configuration for `identifyEntities`
* Implement entities detection + entity match detection

### DIFF
--- a/aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyEntitiesTest.java
+++ b/aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyEntitiesTest.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.aws;
+
+import android.content.Context;
+import android.graphics.Bitmap;
+
+import com.amplifyframework.predictions.PredictionsCategory;
+import com.amplifyframework.predictions.aws.test.R;
+import com.amplifyframework.predictions.models.EmotionType;
+import com.amplifyframework.predictions.models.EntityDetails;
+import com.amplifyframework.predictions.models.GenderBinaryType;
+import com.amplifyframework.predictions.models.IdentifyActionType;
+import com.amplifyframework.predictions.result.IdentifyEntitiesResult;
+import com.amplifyframework.testutils.Assets;
+import com.amplifyframework.testutils.FeatureAssert;
+import com.amplifyframework.testutils.sync.SynchronousMobileClient;
+import com.amplifyframework.testutils.sync.SynchronousPredictions;
+import com.amplifyframework.util.CollectionUtils;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Collections;
+
+import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests that Predictions identify delivers a non-null
+ * entity detection result for valid input.
+ */
+public final class AWSPredictionsIdentifyEntitiesTest {
+
+    private static final IdentifyActionType TYPE = IdentifyActionType.DETECT_ENTITIES;
+
+    private SynchronousPredictions predictions;
+
+    /**
+     * Configure Predictions category before each test.
+     * @throws Exception if mobile client initialization fails
+     */
+    @Before
+    public void setUp() throws Exception {
+        Context context = getApplicationContext();
+
+        // Set up Auth
+        SynchronousMobileClient.instance().initialize();
+
+        // Delegate to Predictions category
+        PredictionsCategory asyncDelegate =
+                TestPredictionsCategory.create(context, R.raw.amplifyconfiguration);
+        predictions = SynchronousPredictions.delegatingTo(asyncDelegate);
+    }
+
+    /**
+     * Assert general entity detection works.
+     * @throws Exception if prediction fails
+     */
+    @Test
+    @SuppressWarnings("MagicNumber") // Jeff Bezos' current age
+    public void testIdentifyEntities() throws Exception {
+        final Bitmap image = Assets.readAsBitmap("jeff_bezos.jpg");
+
+        // Identify the entity inside given image and assert non-null result.
+        IdentifyEntitiesResult result = (IdentifyEntitiesResult) predictions.identify(TYPE, image);
+        assertNotNull(result);
+
+        // Assert that at least one entity is detected
+        assertFalse(CollectionUtils.isNullOrEmpty(result.getEntities()));
+        EntityDetails entity = result.getEntities().get(0);
+
+        // Assert features from detected entity
+        FeatureAssert.assertMatches(GenderBinaryType.MALE, entity.getGender());
+        FeatureAssert.assertMatches(EmotionType.HAPPY, Collections.max(entity.getEmotions()));
+        assertNotNull(entity.getAgeRange());
+        assertTrue(entity.getAgeRange().contains(56));
+    }
+}
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/NetworkPolicy.java
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/NetworkPolicy.java
@@ -23,12 +23,6 @@ import androidx.annotation.NonNull;
  */
 public enum NetworkPolicy {
     /**
-     * Only use cloud resources to perform online Predictions
-     * operation.
-     */
-    ONLINE("online"),
-
-    /**
      * Only use local resources to perform offline Predictions
      * operation.
      */
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
@@ -29,6 +29,7 @@ import com.amazonaws.services.rekognition.model.BoundingBox;
 import com.amazonaws.services.rekognition.model.FaceDetail;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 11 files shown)


aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
```diff
@@ -114,6 +133,29 @@ void recognizeCelebrities(
...
+            } else {
+                int maxEntities = config.getMaxEntities();
+                String collectionId = config.getCollectionId();
+                List<EntityMatch> matches = detectEntityMatches(image, maxEntities, collectionId);
+                onSuccess.accept(IdentifyEntityMatchesResult.fromEntityMatches(matches));
+            }
+        } catch (PredictionsException exception) {
+            onError.accept(exception);
+        }
+
```
jamesonwilliams, line 156: nit: extra line

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/IdentifyEntitiesConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public boolean isGeneralEntityDetection() {
+        return generalEntityDetection;
+    }
+
+    /**
+     * Returns true if configured to allow celebrity detection.
+     * @return true if configured to allow celebrity detection
+     */
+    public boolean celebrityDetectionEnabled() {
```
jamesonwilliams, line 124: Naming: `isCelebritryDection...()` (with `is[A-Z]` that Kotlin likes.)

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/IdentifyEntitiesConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+        NetworkPolicy networkPolicy = NetworkPolicy.fromKey(networkPolicyString);
+
+        // Optional fields
+        int maxEntities;
+        boolean generalEntity;
+        String collectionId;
+        String maxEntitiesString;
+        try {
+            maxEntitiesString = identifyEntitiesJson.getString("maxEntities");
+            collectionId = identifyEntitiesJson.getString("collectionId");
```
jamesonwilliams, line 81: So if this throws, you'lll end up in the `catch` block, and then `maxEntities` will be `0`, and `generalEntity` will be `true`. Is that the control flow you were going for, here? Seems like there's a lot of branches here, depending on which lines throws, and what the defaults will get set to.

Actually, I guess if `maxEntitiesString` line throws, then the outcome here is even weirder, with `maxEntitiesString` as `null` and `collectionId` as `""`.

`generalEntity` is very hard to trace. It starts as `false` (default for a boolean.) It _might_ get set to true/false on L83. But. But if there is an error before L83, it gets reset to `true` in the `catch`.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
```diff
@@ -108,4 +123,50 @@ public static Pose fromRekognitionPose(@NonNull com.amazonaws.services.rekogniti
...
     }
+
+    /**
+     * Gets all the binary features from AWS Rekognition's face
+     * details and compiles them into a single list.
+     * @param face the Rekognition face detail object
+     * @return the list of Amplify {@link BinaryFeature}
+     */
+    public static List<BinaryFeature> fromFaceDetail(FaceDetail face) {
+        List<BinaryFeature> features = new ArrayList<>();
```
jamesonwilliams, line 134: Just use:
```
Arrays.astList(
    new Foo<>("asdf"),
    new Foo<>("asdfeee"),
    new Foo<>("sdfgsdfgr")
)
```

Almost as good as [Java 9 `List.of()`](https://docs.oracle.com/javase/9/docs/api/java/util/List.html#of-E...-), or [Kotlin `listOf()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/list-of.html) :-D

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/NetworkPolicy.java
```diff
@@ -22,12 +22,6 @@
  * perform a Predictions operation.
  */
 public enum NetworkPolicy {
-    /**
-     * Only use cloud resources to perform online Predictions
-     * operation.
-     */
-    ONLINE("online"),
```
jamesonwilliams, line 29: Hm. How come?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/IdentifyEntitiesConfiguration.java
raphkim: https://docs.amplify.aws/lib/predictions/identify-entity?platform=ios

> The value of maxEntities should be a number greater than 0 or less than 51 (50 is the max number of entities Rekognition can detect from a collection). If both collectionId and maxEntities do not have valid values in the amplifyconfiguration.json file, then this call will just detect entities in general with facial features, landmarks, etc.

Very weird flow indeed, but I think what I currently have kind of describes this sufficiently?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
raphkim: haha i was thinking about it, and the only reason I decided against it was to reduce indentation depth. Will fix

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/IdentifyEntitiesConfiguration.java
raphkim: will do

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
raphkim: removed

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/NetworkPolicy.java
raphkim: We still give them the implicit option of doing so (i.e. choose "auto" and just not add any offline models). iOS currently doesn't support "online" only mode and there was apparently already a discussion that decided against giving the customer an option to explicitly specify online-only mode.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/IdentifyEntitiesConfiguration.java
raphkim: also `generalEntity` (which is pretty badly named I guess. Should have been something like `isGeneralEntityDetectionEnabled`, but that was too long haha) should kind of be defaulted to `true`. However, the IDE saw that its value is specified in both try and catch clauses, so it didn't like being given n initial value.

The main idea is, `isGeneralEntityDetection` should be true, and `collectionId` and `maxEntities` should not be referenced if it is. Therefore, they are given arbitrary values for the time being (inside `catch`). `isGeneralEntityDetection` will be true ONLY if both `maxEntities` and `collectionId` are specified in the configuration, and the `maxEntities` value is valid (0 < x < 51).

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
```diff
@@ -226,6 +267,99 @@ void recognizeCelebrities(
         return celebrities;
     }
 
+    private List<EntityDetails> detectEntities(Image image) throws PredictionsException {
+        DetectFacesRequest request = new DetectFacesRequest()
+                .withImage(image)
+                .withAttributes("ALL");
```
jamesonwilliams, line 273: Does Rekognition store `"ALL"` as a labeled constant anywhere, that you could use instead?

jamesonwilliams, line 273: Does Rekognition store `"ALL"` as a labeled constant anywhere, that you could use instead?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
```diff
@@ -108,4 +124,50 @@ public static Pose fromRekognitionPose(@NonNull com.amazonaws.services.rekogniti
...
+    /**
+     * Gets all the binary features from AWS Rekognition's face
+     * details and compiles them into a single list.
+     * @param face the Rekognition face detail object
+     * @return the list of Amplify {@link BinaryFeature}
+     */
+    public static List<BinaryFeature> fromFaceDetail(FaceDetail face) {
+        return Arrays.asList(
+                BinaryFeature.builder()
+                        .type("Beard")
```
jamesonwilliams, line 137: Does Rekognition have labeled constants somewhere for like `"Beard"` ,  etc., that you could use instead of hard-coding the string?

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyEntitiesTest.java
```diff
@@ -0,0 +1,94 @@
...
+        IdentifyEntitiesResult result = (IdentifyEntitiesResult) predictions.identify(TYPE, image);
+        assertNotNull(result);
+
+        // Assert that at least one entity is detected
+        assertFalse(CollectionUtils.isNullOrEmpty(result.getEntities()));
+        EntityDetails entity = result.getEntities().get(0);
+
+        // Assert features from detected entity
+        FeatureAssert.assertMatches(GenderBinaryType.MALE, entity.getGender());
+        FeatureAssert.assertMatches(EmotionType.HAPPY, Collections.max(entity.getEmotions()));
```
jamesonwilliams, line 90: Wasn't there a `getPredominantSentiment()` or something like that? Should we provide an API for this, `getPredominantEmotion()` or something?

The thing about test code is that it tends to look like what a customer will do. If the customer just wants to know "Is Jeff Bezos happy?" (Insert joke here...)

(... back from joke) what about like `FeatureAssert.assertMatches(EmotionType.HAPPY, entity.getPredominantEmotion())`? // Returns the one with most confidence?

raphkim: it would be convenient in a way... but that would unfortunately break the platform parity. 

To be completely fair, iOS's implementation of `interpret` provides BOTH predominant sentiment AND the map of all other sentiments (and their confidence scores). I'll leave it as for now, but it would certainly be a nice addition in the future on both platforms.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
raphkim: they are individual features with strongly typed classes in Rekognition, but iOS folks decided to make them all into generic binary attributes (binary features). So `FaceDetail#getBeard()` returns a `Beard` typed object, which makes it hard to obtain constants. Perhaps I can do something like `Beard.class.getSimpleName()`, but that seems somewhat pointless.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
raphkim: good catch. Indeed there was!

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyResultTransformers.java
raphkim: ^The above solution would also cause SLIGHT disparity with iOS since the binary feature for glasses is hardcoded as "EyeGlasses", but the Rekognition class is named "Eyeglasses"

raphkim: Hmm, perhaps so. These are specific types of binary features that OUR implementation happens to detect, so that wouldn't belong in the core. However, I can see it being useful inside the actual plugin module.

I'll make a new PR to address this since it is big enough of a change to add to an already approved PR.

PR #400, opened 2020-04-25
Expose all GraphQL error properties (part 2)

richardmcclellan:
Addresses the last bits of feedback on https://github.com/aws-amplify/amplify-android/pull/392, specifically:
 - Moves GraphQLLocation, GraphQLPathSegment into same package as GraphQLResponse
 - Implements custom deserialization of GraphQL error extension into HashMap to eliminate gson's LinkedTreeMap leaking to public interface.
 - Adds `toString` implementations to error models

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
@@ -29,8 +29,8 @@ import com.google.gson.reflect.TypeToken;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 @SuppressWarnings("unchecked")
 final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResponse.Error> {
@@ -50,7 +50,7 @@ final class GsonErrorDeserializer implements JsonDeserializer<GraphQLResponse.Er
         String message = null;
         List<GraphQLLocation> locations = null;
         List<GraphQLPathSegment> path = null;
-        HashMap<String, Object> extensions = null;
+        Map<String, Object> extensions = null;
         JsonObject extensionsJson = new JsonObject();
         JsonObject nonSpecifiedData = new JsonObject();
 
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
@@ -15,9 +15,9 @@
 
 package com.amplifyframework.api.aws;
 
+import com.amplifyframework.api.graphql.GraphQLLocation;
+import com.amplifyframework.api.graphql.GraphQLPathSegment;
 import com.amplifyframework.api.graphql.GraphQLResponse;
-import com.amplifyframework.api.graphql.error.GraphQLLocation;
-import com.amplifyframework.api.graphql.error.GraphQLPathSegment;
 
 import com.google.gson.JsonDeserializationContext;
 import com.google.gson.JsonDeserializer;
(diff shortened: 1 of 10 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
```diff
@@ -49,7 +50,7 @@
         String message = null;
         List<GraphQLLocation> locations = null;
         List<GraphQLPathSegment> path = null;
-        Map<String, Object> extensions = null;
+        HashMap<String, Object> extensions = null;
```
jamesonwilliams, line 53: Ordinarily the declaration is the bast `Map<KeyType, ValType>`, unless you are using some exotic property of a map that requires specification of which map type you need. For example, if you need ordering properties, you'd both instantiate and declared `LinkedHashMap`, since the type _is_ important, you're no longer dealing in "any old map." But here, I think you're just dealing with any old map.

Most common pattern is like:
```
static Map<Foo, Bar> createOne() {
    Map<Foo, Bar> fooBar = new HashMap<>(); // This is the only place the special type is mentioned
    fooBar.put("ham", "sandwich");
    return fooBar;
}
```

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
```diff
@@ -0,0 +1,75 @@
...
+        throw new UnsupportedOperationException("No instances allowed.");
+    }
+
+    static HashMap<String, Object> toMap(JsonObject object) {
+        HashMap<String, Object> map = new HashMap<>();
+        for (String key : object.keySet()) {
+            JsonElement element = object.get(key);
+            map.put(key, toObject(element));
+        }
+        return map;
```
jamesonwilliams, line 39: `return Immutable.of(map)`?

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    static List<Object> toList(JsonArray array) {
+        List<Object> list = new ArrayList<>();
+        for (int i = 0; i < array.size(); i++) {
+            JsonElement element = array.get(i);
+            list.add(toObject(element));
+        }
+        return list;
```
jamesonwilliams, line 48: `return Immutable.of(list)`?

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private static Object toObject(JsonElement element) {
+        if (element != null) {
```
jamesonwilliams, line 52: There is the [`element.isJsonNull()`](https://javadoc.io/doc/com.google.code.gson/gson/2.6.2/com/google/gson/JsonElement.html#isJsonNull--) -- not sure what the deal is with that... do we need to treat this case, or does the final `return null` handle it appropriately?

aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
richardmcclellan: Agreed!  Not sure why I specified `HashMap` instead of `Map`.  I will change to just `Map`.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
richardmcclellan: Will do

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
richardmcclellan: Will do

aws-api/src/main/java/com/amplifyframework/api/aws/GsonUtil.java
richardmcclellan: The final `return null` will handle `element.isJsonNull()`.

PR #401, opened 2020-04-25
[aws-datstore] Implementation of user-provided DataStore configs

rjuliano:
Sorry guys, had to end up creating a new PR (this is a continuation of #395 which I botched on a merge)
*Description of changes:*

## Highlights of this edition as compared to #395 
- Move the configuration responsibility back into the the configure method of the DataStore plugin. As a result, I had to add a bit of synchronization code in the constructor of the DataStore plugin class to wait for the configuration to be initialized before creating an instance of the Orchestrator.
- Change syncInterval to syncInterval in minutes
- Updated javadoc descriptions for the DataStoreConflictResult enum.
- Re-structure the code that returns an empty CategoryConfiguration if one can't be found in the Amplify config file.


aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+
+        /**
+         * Sets the number of items requested in each page of sync results.
+         * @param syncPageSize Number of items requested per page in sync operation
+         * @return Current builder
+         */
+        @NonNull
+        public Builder syncPageSize(@IntRange(from = 0) int syncPageSize) {
+            //Only set this value if the incoming value is not equal to the default
+            if (syncPageSize != DEFAULT_SYNC_PAGE_SIZE) {
```
jamesonwilliams, line 276: Shouldn't need this `if`. In the case that `syncPageSize == DEFAULT_SYNC_PAGE_SIZE`, you'd just set the value to the same thing again. Which wouldn't change the system behavior.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+
+        /**
+         * Sets the maximum number of records, from the server, to process from a sync operation.
+         * @param syncMaxRecords Max number of records client will consumer from server at one time
+         * @return Current builder instance
+         */
+        @NonNull
+        public Builder syncMaxRecords(@IntRange(from = 0) int syncMaxRecords) {
+            //Only set this value if the incoming value is not equal to the default
+            if (syncMaxRecords != DEFAULT_SYNC_MAX_RECORDS) {
```
jamesonwilliams, line 262: Same, shouldn't need the `if`.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+
+        /**
+         * Sets the duration of time after which delta syncs will not be preferred over base syncs.
+         * @param syncIntervalMs The amount of time that must elapse for delta syncs to not be considered
+         * @return Current builder instance
+         */
+        @NonNull
+        public Builder syncIntervalMs(@IntRange(from = 0) long syncIntervalMs) {
+            //Only set this value if the incoming value is not equal to the default
+            if (syncIntervalMs != DEFAULT_SYNC_INTERVAL_MINUTES) {
```
jamesonwilliams, line 248: Uh oh! This builder didn't get updated. This still says `ms`, but it should say:

```
@NonNull
public Builder syncIntervalInMinutes(@IntRange(from = 0) long syncIntervalInMinutes) {
    this.syncIntervalMs = TimeUnit.MINUTES.toMilliseconds(syncIntervalInMunutes);
    return Builder.this;
}
```

No `if` needed.

As per comments further down -- this is a ~good place to do the unit conversion into `ms`, I think.

(Human beings like minutes, but computers like milliseconds.)

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+    /**
+     * Builds instances of {@link AWSDataStorePlugin} by providing a variety of
+     * configuration methods.
+     */
+    public static final class Builder {
+        private DataStoreErrorHandler dataStoreErrorHandler;
+        private DataStoreConflictHandler dataStoreConflictHandler;
+        private Long syncIntervalMs;
+        private Integer syncMaxRecords;
+        private Integer syncPageSize;
```
jamesonwilliams, line 206: I don't think you ever use the fact that these are objects. (They have non-null default values, and you overwrite the m via primitive types.)

So, I think it should be both safe and preferable to store these as their primitive counterparsts,`long` and `int` .

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+            } catch (IllegalArgumentException noSuchConfigKey) {
+                throw new DataStoreException(
+                        "Saw unexpected config key: " + keyString,
+                        "Make sure your amplifyconfiguration.json is valid."
+                );
+            }
+            try {
+                switch (configKey) {
+                    case SYNC_INTERVAL:
+                        builder.syncIntervalMs(pluginJson.getLong(ConfigKey.SYNC_INTERVAL.toString()));
```
jamesonwilliams, line 118: Since this one is user-exposed, I think we want this work in minutes, here. _Inside_ the builder, convert to ms.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+     */
+    @NonNull
+    public static Builder builder(@NonNull JSONObject pluginJson,
+                                  @NonNull DataStoreConfiguration userProvidedConfiguration) throws DataStoreException {
+        return builder(pluginJson)
+            .syncPageSize(userProvidedConfiguration.getSyncPageSize())
+            .syncMaxRecords(userProvidedConfiguration.getSyncMaxRecords())
+            .syncIntervalMs(userProvidedConfiguration.getSyncIntervalInMinutes())
+            .dataStoreErrorHandler(userProvidedConfiguration.getDataStoreErrorHandler())
+            .dataStoreConflictHandler(userProvidedConfiguration.getDataStoreConflictHandler());
```
jamesonwilliams, line 90: Hm, how does this  work? Doesn't the user config always clobber the file config?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+import org.json.JSONObject;
+
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A user-provided configuration for the DataStore.
+ */
+@SuppressWarnings("unused")
```
jamesonwilliams, line 35: You can probably strip this off now that it's being referenced at least from test

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,340 @@
...
+public final class DataStoreConfiguration {
+    static final String PLUGIN_CONFIG_KEY = "awsDataStorePlugin";
+    static final long DEFAULT_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
+    static final int DEFAULT_SYNC_MAX_RECORDS = 1_000;
+    static final int DEFAULT_SYNC_PAGE_SIZE = 1_000;
```
jamesonwilliams, line 40: Are these default values the same on iOS and JavaScript? I don't remember, but I may have just made these up on the fly.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -148,13 +163,24 @@ public String getPluginKey() {
...
-            this.pluginConfiguration =
-                AWSDataStorePluginConfiguration.fromJson(pluginConfiguration);
+            DataStoreConfiguration dataStoreConfiguration;
+            if (userProvidedConfiguration == null) {
+                dataStoreConfiguration = DataStoreConfiguration
+                    .builder(pluginConfiguration)
+                    .build();
+            } else {
+                dataStoreConfiguration = DataStoreConfiguration
+                    .builder(pluginConfiguration, userProvidedConfiguration)
```
jamesonwilliams, line 177: Can we push this `null` check down into the `builder(...)` so that we don't have to process it in this block here in the plugin? That is,
```
dataStoreConfiguration = DataStoreConfiguration.builder(configFromDisk, null).build();
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -62,28 +65,39 @@
...
-        this.orchestrator = new Orchestrator(
-            modelProvider,
-            modelSchemaRegistry,
-            sqliteStorageAdapter,
-            AppSyncClient.via(api),
-            () -> pluginConfiguration.getBaseSyncIntervalMs()
+        dataStoreConfigurationProvider = AsyncSubject.create();
+
+        dataStoreConfigurationProvider.subscribe(initializedConfiguration -> {
+            orchestrator = new Orchestrator(
```
jamesonwilliams, line 89: I guess there are two ways of looking at this:

1. The current view:
> The orchestrator is an optional component. It's presence is determined by the content of the configuration. If the configuration does not permit remote sync, then don't create an orchestrator.
2. The original view:
> The orchestrator is a data store component, it is constructed at the same time as the plugin. Later, if the configuration determines that sync should be enabled, the orchestrator is started.

Realistically, the only concrete concern is whether or not `orchestrator` is null at the time you call `start`. So long as configuration is definitely done by the time `initialize()` rolls around, you should be good.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -66,6 +67,7 @@ public void setup() {
      * @throws JSONException Not expected; on failure to arrange configuration object
      * @throws DataStoreException Not expected; on failure to configure of initialize plugin
      */
+    @Ignore("TODO: need a mechanism to re-enable this scenario........")
```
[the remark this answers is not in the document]
jamesonwilliams: Agreed. Okay, sounds great.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * found in the remote AppSync system.
+ */
+public interface DataStoreConflictHandler {
+    /**
+     * Resolves a DataStore conflict.
+     * @param conflictData Data about the conflict.
+     * @param onResult A callback that should be invoked when the conflict handling is complete.
+     * @param <T> The type of model for which a conflict was observed
+     */
+    <T extends Model> void resolveConflict(
```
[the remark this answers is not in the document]
jamesonwilliams: Hmm. Maybe @drochetti thought about this one. If so / if not, maybe we should pow-wow with him for 15 minutes later today to discuss this specific topic?

We can also do a variant of #2, where-in we try/catch the conflict handler, and then dispatch any exception to the datastore handler. The user would continue to provide the same handler that they do now. And we would enforce the policy of falling back to the error handler, in an internal final class that the user couldn't get at.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
jamesonwilliams: Okay cool I understand the intent now! Thanks

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
jamesonwilliams: > non-default values provided in userProvidedConfiguration to override whatever was in the JSON

I think this means that you'll still overwrite config file values, if they aren't the defaults. No?

I like the second mechanism you propose, better, I think.

Basically:
1. Create a config from the file json. This will have 0...all values as null;
2. Create a config from user. This will have 0...all values as null;
3. Merge the two configs according to the rule:
```
if (userProvided.getSomeConfigValue() != null) {
    use(userProvided.getSomeConfigValue());
} else {
    use(configFileProvided.getSomeConfigValue());
```
4. As a final step, evaluate the merged config. If any fields are still `null`, then use a default value.

That is kind of a lot of machinery isn't it (eye-roll) sigh.

rjuliano: > I think this means that you'll still overwrite config file values, if they aren't the defaults. No?

Correct. The assumption here is that any non-default values passed via the `userProvidedConfiguration` will clobber anything that was set via the file. That being said, if we want to respect any non-default settings set via the config file, I can add an extra check in the setters:

```
public Builder syncMaxRecords(@IntRange(from = 0) int syncMaxRecords) {
    //Only set this value if the current value is default AND the incoming value is NOT default
    if (this.syncMaxRecords == DEFAULT_SYNC_MAX_RECORDS && 
        syncMaxRecords != DEFAULT_SYNC_MAX_RECORDS) {
        this.syncMaxRecords = syncMaxRecords;
    }
    return Builder.this;
}
```

I think in the end, it comes down to deciding which is more authoritative: the config file or the java object passed in via the constructor.

The drawback of making them nullable is that you need to have that extra step to ensure defaults are set. When I started going that route, there was no clear place for that. I couldn't do it on the build method of the builder, because I actually need the values to remain null. So I would need something (probably in the plugin's configure method) that would set those defaults.

Another thing I just thought of if we want to go the null route, is that I could modify the getters to return their respective defaults if the underlying field is null. I may give that a try.

rjuliano: That actually worked, so I updated the code to make the fields null (with the exception of the handler) to start with and then I only write to them if the underlying field is null. The getters, in turn, return a default value if the underlying field is null

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandler.java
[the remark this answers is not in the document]
rjuliano: Agreed. I think once we actually wire the call to the conflict handler, we do exactly what you described. Wrap it in a try catch and use the configured error handler

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
rjuliano: My latest commit makes the scalar values nullable and returns defaults from the getters if the underlying field is null

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -121,14 +122,13 @@ Completable hydrate() {
      * @param lastSyncTime The time of a last successful sync.
      * @return The input, or {@link SyncTime#never()}, if the last sync time is "too old."
      */
-    private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreException {
-        if (!lastSyncTime.exists()) {
+    private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) {
+        if (!lastSyncTime.exists() || dataStoreConfiguration == null) {
```
jamesonwilliams, line 126: I'd expect that `dataStoreConfiguration` would be guaranteed non-null, here. That is, if there's no config file, and the user supplies no config, then you still have some system defaults you can use. Right? So, _those_ could go into a non-null `dataStoreConfiguration`. And that would have a different effect here.

The current logic says basically "never do a delta sync."

But that's not what we want -- we want a delta sync if the last sync time is within some san default (1 day? Whatever it is?)

In other words, I'd prefer to have an `Objects.requireNonNull(...)` and `@NonNull` directive on the `SyncProcessor` constructor, and then just never have to worry about this branch, in this spot.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+
+        /**
+         * Sets the maximum number of records, from the server, to process from a sync operation.
+         * @param syncMaxRecords Max number of records client will consumer from server at one time
+         * @return Current builder instance
+         */
+        @NonNull
+        public Builder syncMaxRecords(@IntRange(from = 0) Integer syncMaxRecords) {
+            //Only set this value if the incoming value is null
+            if (this.syncMaxRecords == null) {
```
jamesonwilliams, line 271: This is still a little spooky. If do:
```
builder.syncMaxRecords(1000)
    .syncMaxRecords(500)
    .build();
```
The value is 1000 even though I'd probably expect it to be 500...

(Same for others with this pattern.)

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+     * Gets the number of items that should be requested in page, from AppSync, during
+     * a sync operation.
+     * @return Desired size of a page of results from an AppSync sync response
+     */
+    @IntRange(from = 0)
+    public Integer getSyncPageSize() {
+        return this.syncPageSize == null ? DEFAULT_SYNC_PAGE_SIZE : this.syncPageSize;
```
jamesonwilliams, line 206: Can you put this construction stuff in the `build()` so that the `DataStoreConfiguration` doesn't have any logic, it just returned a single, final, stored instance value?

(Same for others with this pattern.)

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A user-provided configuration for the DataStore.
```
jamesonwilliams, line 33: You actually use it to model a few things right, it is no longer just user-provided?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
rjuliano: Yeah...i agree, it feels wonky. I’ll review this tomorrow just to make sure. The struggle here is trying to come up with a mechanism that we can use to determine which values have precedence over the other. I would imagine we want to create the object with defaults, then read from file, then use anything passed in by the user. 

I’ll take another look at it tomorrow.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: Yeah. I added this here to prevent a potential NPE, but by the time it gets here, we’re sure to have at least default values.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
rjuliano: Changes made in the [latest commit](https://github.com/aws-amplify/amplify-android/pull/401/commits/6c3c4c5522dcd21c055d511c03afb970e64bede5)

- Removed the logic from the getters and modified the builder's build method to ensure defaults when instructed to do so.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: Changes made in the [latest commit](https://github.com/aws-amplify/amplify-android/pull/401/commits/6c3c4c5522dcd21c055d511c03afb970e64bede5)

I created a provider interface that I'm now passing from the the plugin down to the `Orchestrator` and `SyncProcessors`. By the time these classes invoke the provider method to get the configuration, the plugin will have already setup the configuration in the `configure` method.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
rjuliano: Changes made in the [latest commit](https://github.com/aws-amplify/amplify-android/pull/401/commits/6c3c4c5522dcd21c055d511c03afb970e64bede5)

With the latest commit, the following rules should apply:
* Read settings from config file (if any)
* Read settings from user-provided POJO (if any). Apply any non-null values
* Fill in any missing fields with default values if instructed to do so (More on that below)

## Changes to `Builder`
* Moved most of the logic out of the `DataStoreConfiguration` methods
* Added a constructor that takes an instance of `DataStoreConfiguration` as the user-provided settings in addition to the JSON values
* Added a field to the `Builder` called `ensureDefaults`. This field is only set to true if the new constructor is called.
* Modified `build()` method to populate defaults if  `ensureDefaults` is true

## Changes to `DataStoreConfiguration`
* Marked two builder methods(`builder(JSONObject)`, and `builder(JSONObject, DataStoreConfiguration)`) as package private so they should only be callable internally. This leaves the parameter-less `builder()` method as the only option when user's want to create an instance of `DataStoreConfiguration` to pass in overrides. When created via this method, the builder's `build()` method will leave any scalar fields not explicitly set by the user as null.
* Simplified getters and added a new getter for `getSyncIntervalInMinutes`

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -62,28 +64,36 @@
...
-            @NonNull GraphQlBehavior api) {
+            @NonNull ModelSchemaRegistry modelSchemaRegistry,
+            @NonNull GraphQlBehavior api,
+            @Nullable DataStoreConfiguration userProvidedConfiguration) {
         this.sqliteStorageAdapter = SQLiteStorageAdapter.forModels(modelSchemaRegistry, modelProvider);
         this.storageItemChangeConverter = new GsonStorageItemChangeConverter();
         this.categoryInitializationsPending = new CountDownLatch(1);
-        this.orchestrator = new Orchestrator(
+
+        orchestrator = new Orchestrator(
```
jamesonwilliams, line 89:
```suggestion
        this.orchestrator = new Orchestrator(
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -62,28 +64,36 @@
 
     // A component which synchronizes data state between the
     // local storage adapter, and a remote API
-    private final Orchestrator orchestrator;
+    private Orchestrator orchestrator;
```
jamesonwilliams, line 67:
```suggestion
    private final Orchestrator orchestrator;
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -148,12 +159,13 @@ public String getPluginKey() {
...
     @Override
     public void configure(
-            @Nullable JSONObject pluginConfiguration,
+            @NonNull JSONObject pluginConfiguration,
             @NonNull Context context
     ) throws DataStoreException {
         try {
-            this.pluginConfiguration =
-                AWSDataStorePluginConfiguration.fromJson(pluginConfiguration);
+            this.pluginConfiguration = DataStoreConfiguration
```
jamesonwilliams, line 166:
```suggestion
            // Applies user-provided configs on-top-of any values from the file.
            this.pluginConfiguration = DataStoreConfiguration
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,387 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore;
+
+import androidx.annotation.IntRange;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
```
jamesonwilliams, line 20:
```suggestion
import androidx.annotation.Nullable;
import androidx.annotation.VisibleForTesting;
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * Configuration options for {@link AWSDataStorePlugin}.
+ */
+public final class DataStoreConfiguration {
+    static final String PLUGIN_CONFIG_KEY = "awsDataStorePlugin";
+    static final long DEFAULT_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
+    static final long DEFAULT_SYNC_INTERVAL_MS = TimeUnit.MINUTES.toMillis(DEFAULT_SYNC_INTERVAL_MINUTES);
+    static final int DEFAULT_SYNC_MAX_RECORDS = 10_000;
+    static final int DEFAULT_SYNC_PAGE_SIZE = 1_000;
```
jamesonwilliams, line 40:
```suggestion
    @VisibleForTesting
    static final long DEFAULT_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
    @VisibleForTesting
    static final long DEFAULT_SYNC_INTERVAL_MS = TimeUnit.MINUTES.toMillis(DEFAULT_SYNC_INTERVAL_MINUTES);
    @VisibleForTesting 
    static final int DEFAULT_SYNC_MAX_RECORDS = 10_000;
    @VisibleForTesting 
    static final int DEFAULT_SYNC_PAGE_SIZE = 1_000;
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,387 @@
...
+                return;
+            }
+            dataStoreErrorHandler = userProvidedConfiguration.getDataStoreErrorHandler();
+            dataStoreConflictHandler = userProvidedConfiguration.getDataStoreConflictHandler();
+            syncIntervalInMinutes = getValueOrDefault(
+                userProvidedConfiguration.getSyncIntervalInMinutes(),
+                syncIntervalInMinutes);
+            syncMaxRecords = getValueOrDefault(userProvidedConfiguration.getSyncMaxRecords(), syncMaxRecords);
+            syncPageSize = getValueOrDefault(userProvidedConfiguration.getSyncPageSize(), syncPageSize);
+
```
jamesonwilliams, line 307:
```suggestion
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -0,0 +1,387 @@
...
+    /**
+     * Creates an {@link DataStoreConfiguration} which uses all default values.
+     * @return A default {@link DataStoreConfiguration}
+     * @throws DataStoreException exception thrown if there's an unexpected configuration key or
+     * an invalid configuration value
+     */
+    @NonNull
+    public static DataStoreConfiguration defaults() throws DataStoreException {
+        return builder()
```
jamesonwilliams, line 112: Should this also include our opinionated error and conflict handlers?

rjuliano: Those get set in the default constructor of the builder since there's no ambiguity with those two fields. That being said, it doesn't hurt to be explicit about it and setting it here.

rjuliano:
> Nice, this looks ready to ship to me! Thanks for sticking with it and investing in a high bar @rjuliano. We can iterate on the "how to do local-only" now, but to me it makes more sense to commit this work and build now. What do you think?

Agreed. Thanks for spending the time reviewing this. I made the last couple tweaks and just committed and will be merging it soon

PR #409, opened 2020-04-29
Implement online text detection 

raphkim:
*Issue #, if available:*

*Description of changes:*
Implement text detection using rekognition + textract

[Amazon Textract documentation](https://docs.aws.amazon.com/textract/latest/dg/how-it-works-document-layout.html) for document analysis


aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSTextractService.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Predictions service for performing text translation.
+ */
+final class AWSTextractService {
+    private final AmazonTextractClient textract;
+    private final AWSPredictionsPluginConfiguration pluginConfiguration;
```
jamesonwilliams, line 61: Looks like this is [dead-store](https://en.wikipedia.org/wiki/Dead_store)?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
```diff
@@ -226,6 +246,60 @@ void recognizeCelebrities(
...
+        try {
+            result = rekognition.detectText(request);
+        } catch (AmazonClientException serviceException) {
+            throw new PredictionsException(
+                    "Amazon Rekognition encountered an error while detecting text.",
+                    serviceException, "See attached service exception for more details."
+            );
+        }
+
+        StringBuilder fullTextBuilder = new StringBuilder();
```
jamesonwilliams, line 264: I like the pattern you used elsewhere where split out the "convert to result type" step into a helper.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
```diff
@@ -226,6 +246,60 @@ void recognizeCelebrities(
...
+        List<String> rawLineText = new ArrayList<>();
+        List<IdentifiedText> words = new ArrayList<>();
+        List<IdentifiedText> lines = new ArrayList<>();
+        for (TextDetection detection : result.getTextDetections()) {
+            String detectedText = detection.getDetectedText();
+            Geometry geometry = detection.getGeometry();
```
jamesonwilliams, line 270: Minimal savings from caching these. I think you're running out of width on lines 271 and 272 because `IdentifyEntitiesResultTransformers` is so long. Maybe we can rename that to something shorter?

Also you're starting at +12 from class indentation, method indentation, and `for` indentation.

Maybe the body of the `for` loop can be externalized to a helper? Among other things, that'd set you back to +8 indentation at least.

(Point in short, by the time you actually get to call `fromPoints`, it's pretty far to the right side of the screen.)

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPredictionsService.java
```diff
@@ -100,6 +105,36 @@ public void recognizeCelebrities(
...
     }
 
+    /**
+     * Delegate to {@link AWSRekognitionService} to detect plain text.
+     * @param image the Rekognition input image
+     * @param onSuccess triggered upon successful result
+     * @param onError triggered upon encountering error
+     */
+    public void detectPlainText(
+            @NonNull Image image,
```
jamesonwilliams, line 115: This method is weird in that it takes an AWS  service type as an input, but then renders an Amplify type as an output.

Or in other words, right now,`AWSPredictionsService` is "leaking" some abstractions.

Let's consider the duties of the `AWSPredictionsService`, vs the `AWSRecognitionService`.

The lower level ones like `AWSRecognitionService` are handling the business of creating service-modeled requests, and decoding service-modeled responses. But, they are still requiring some types like `Image`, `Document`, which are contained in the requests.

Theoretically, you could "patch" the "leak" even here. Your `AWSRekognitionService` could accept only Amplify types, and return only Amplify types.

Then, your `AWSPredictionsService` is just a bundle of the various service clients.

But certainly, _at minimum_ you should encapsulate the AWS services underneath the `AWSPredictionsService`.

I do believe that `AWSPredictionsService` is an architecturally desirable component. If you look at some other plugin modules, they `WhateverPlugin extends Plugin` are chokepoints, and get huge. And its because the developer(s) encoded business rules directly into the plugin classes, instead of encapsulating the behaviors into a self contained service component, as you have here. The `AWSPredictionsPlugin` is just a platform-binding adapter, once you have this service class.

Anyway, you're leaking the AWS stuff up through the service, into the plugin. The plugin would need to know what an `Image` is, and that's not so desirable.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyTextResultTransformers.java
```diff
@@ -0,0 +1,232 @@
...
+            return null;
+        }
+        return new RectF(
+                box.getLeft(),
+                box.getTop(),
+                box.getLeft() + box.getWidth(),
+                box.getTop() + box.getHeight()
+        );
+    }
+
```
jamesonwilliams, line 72: nit: empty line

core/src/main/java/com/amplifyframework/predictions/models/Selection.java
```diff
@@ -86,4 +62,65 @@ public Polygon getPolygon() {
     public boolean isSelected() {
         return selected;
     }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static final class Builder {
```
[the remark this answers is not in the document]
raphkim: Apparently yes! AWS provides both shapes so that polygon can define a finer shape WITHIN the provided bounding box.

testutils/src/main/java/com/amplifyframework/testutils/FeatureAssert.java
```diff
@@ -58,4 +58,22 @@ private FeatureAssert() {}
...
+     */
+    public static <F extends Feature<T>, T> void assertContains(T expectedValue, Collection<F> actual) {
+        Assert.assertNotNull(expectedValue);
+        Assert.assertNotNull(actual);
+        for (F feature : actual) {
+            if (expectedValue.equals(feature.getValue())) {
+                return;
+            }
+        }
+        Assert.fail("Collection does not contain expected feature element.");
```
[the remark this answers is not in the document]
raphkim: I like that more. thank you

core/src/main/java/com/amplifyframework/predictions/models/IdentifiedText.java
```diff
@@ -49,13 +49,23 @@ public String getText() {
     }
 
     /**
-     * Gets the page value.
-     * @return the page
+     * Gets the page of the document in which this text was
```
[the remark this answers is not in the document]
raphkim: Further elaborated. The sneaking suspicion is that this is actually an unnecessary property that was defined simply because Textract provides this value. It doesn't seem to be applicable to other potential types of implementation

core/src/main/java/com/amplifyframework/predictions/models/Selection.java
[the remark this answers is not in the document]
jamesonwilliams: Haha ah gee, alright.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSTextractService.java
```diff
@@ -0,0 +1,232 @@
...
+            throw new PredictionsException(
+                    "AWS Textract encountered an error while analyzing document.",
+                    serviceException, "See attached service exception for more details."
+            );
+        }
+
+        return processTextractBlocks(result.getBlocks());
+    }
+
+    private IdentifyDocumentTextResult processTextractBlocks(List<Block> blocks) {
```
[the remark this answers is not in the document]
raphkim: Fixed!

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyTextTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+
+import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Tests that Predictions identify delivers a non-null
+ * text detection result for valid input.
```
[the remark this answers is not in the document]
raphkim: Hmm, I would consider `IdentifyActionType#DETECT_TEXT` an *invalid* input, so it would still be a true statement. It just isn't a very descriptive of a javadoc haha

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyTextTest.java
```diff
@@ -0,0 +1,156 @@
...
+        IdentifyDocumentTextResult result =
+                (IdentifyDocumentTextResult) predictions.identify(TextFormatType.FORM, image);
+
+        // Assert that four key-values are detected.
+        List<BoundedKeyValue> keyValues = result.getKeyValues();
+        assertFalse(CollectionUtils.isNullOrEmpty(keyValues));
+        assertEquals(4, keyValues.size());
+
+        // Assert that key-value pairs have correct values.
+        FeatureAssert.assertContains(new Pair<>("Name:", "Jane Doe"), keyValues);
```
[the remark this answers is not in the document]
raphkim: > `Pair.create("Name: ", "Jane Doe")` skips the ugle `new` and `<>` .

This is a game changer

> These results are split by tab? Or space? Why not ':'? Does the user have a way to split on ':'?

This is the output of AWS Textract, so I have no control over this :/

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyEntitiesResultTransformers.java
```diff
@@ -56,15 +61,40 @@ public static RectF fromBoundingBox(@NonNull BoundingBox box) {
...
     }
 
+    /**
+     * Converts geometric polygon from Amazon Rekognition into
+     * Amplify-compatible polygon object.
+     * @param polygon the polygon object by Amazon Rekognition
+     * @return the polygon object representing vertices
+     */
+    @Nullable
+    public static Polygon fromPoints(@Nullable List<Point> polygon) {
```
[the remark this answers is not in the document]
raphkim: Platform parity with iOS! They treat `Rect` and `Polygon` separately, likely because the AWS backend does as well.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyTextResultTransformers.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Converts a given Amazon Textract block into Amplify-compatible
+     * key-value pair feature. Returns null if not a valid table.
+     * @param block Textract text block
+     * @param blockMap map of Textract blocks by their IDs
+     * @return Amplify Table instance
+     */
+    @Nullable
+    public static BoundedKeyValue processKeyValue(@NonNull Block block, @NonNull Map<String, Block> blockMap) {
```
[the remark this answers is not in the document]
raphkim: Hmm, these transformers are internal, so naming isn't all that important. But I agree that my naming has been somewhat inconsistent (mostly because I was doing `from<Something>`, but Textract is now requiring a bunch of conversion from `Block`, causing naming conflicts). 

Will start using `fetch<Something>` again, like I do in other services.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSTextractService.java
raphkim: It is indeed (for the time being)! I will add these in when I go over configuration for predictions after being finished, but both labels detection and text detection are missing configuration.

For now, I'm only preemptively adding the ones that are required for the plugin to work.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
raphkim: `AWSTextractService#processTextractBlocks()` was split into its own method mainly because the same logic was required for both `detectDocumentText()` and `analyzeDocument()` methods. Generally for other services, I included the conversion directly inside the method calling the service APIs.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSRekognitionService.java
raphkim: > Maybe the body of the `for` loop can be externalized to a helper? Among other things, that'd set you back to +8 indentation at least.

It would still be +12 no?
```
for () {
    foo();
    bar();
    baz();
}
``` 
has the same indentation as
```
for () {
    doFooBarBaz();
}
```
(although definitely cleaner)

Meanwhile, I also moved the `IdentifiedText` builder into utility class for simplicity inside service implementation

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPredictionsService.java
raphkim: Reminder @raphkim to refactor this

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyTextResultTransformers.java
raphkim: fixed

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/adapter/IdentifyTextResultTransformers.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Converts a given Amazon Textract block into Amplify-compatible
+     * table object. Returns null if not a valid table.
+     * @param block Textract text block
+     * @param blockMap map of Textract blocks by their IDs
+     * @return Amplify Table instance
+     */
+    @Nullable
+    public static Table processTable(@NonNull Block block, @NonNull Map<String, Block> blockMap) {
```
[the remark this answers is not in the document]
raphkim: Renaming to `fetchTable`

`process` was originally borrowed from iOS utility class

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyTextTest.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    @Test
+    public void testIdentifyPlainText() throws Exception {
+        final Bitmap image = Assets.readAsBitmap("sample-table.png");
+
+        // Identify the text inside given image and assert non-null result.
+        IdentifyTextResult result =
+                (IdentifyTextResult) predictions.identify(TextFormatType.PLAIN, image);
+        assertNotNull(result);
+        assertNotNull(result.getFullText());
```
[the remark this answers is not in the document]
raphkim: Reminder to address this @raphkim

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPredictionsService.java
jamesonwilliams: ^ Or just the same `Bitmap` you got in the input, right? That was you don't have to do much mapping between the plugin and the service. Can you map from `Bitmap` (Android) to `Image` (AWS service model) inside this service class?

raphkim: I like leaving the `Bitmap` -> `ByteBuffer` conversion work inside `AWSImageIdentifyRequest` object, where it is actually converting the Amplify input into an acceptable service parameter. So I'll just pass the `ByteBuffer` directly to the services.

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsIdentifyTextTest.java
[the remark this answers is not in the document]
raphkim: It is literally a piece of document that contains an arbitrary table and some text haha. I just happen to be using it for both tables detection and plain text detection for testing. I will definitely add expected text to compare against in the assets folder.

testutils/src/main/java/com/amplifyframework/testutils/Assets.java
```diff
@@ -51,8 +51,11 @@ public static String readAsString(final String name) throws RuntimeException {
 
             String line = bufferedReader.readLine();
             while (line != null) {
-                stringBuilder.append(line).append("\n");
+                stringBuilder.append(line);
                 line = bufferedReader.readLine();
+                if (line != null) {
+                    stringBuilder.append("/n");
```
jamesonwilliams, line 57: I think you meant to keep this as `\n`, not `/n`.

raphkim: Ahhh, you are absolutely right! That's an embarrassing mistake. Thanks for catching it

PR #412, opened 2020-04-29
[aws-api] Fix JsonParseException when parsing null path in Error object

richardmcclellan:
*Description of changes:* Fixes a bug where a `JsonParseException` is thrown if `path` on an error object is null.  `path` is not required, so no exception should be thrown.

### DIFF
--- a/aws-api/src/test/resources/error-null-properties.json
+++ b/aws-api/src/test/resources/error-null-properties.json
@@ -1,0 +1,15 @@
+{
+  "data": {
+    "listTodos": null
+  },
+  "errors": [
+    {
+      "path": null,
+      "data": null,
+      "errorType": null,
+      "errorInfo": null,
+      "locations": null,
+      "message": "the message"
+    }
+  ]
+}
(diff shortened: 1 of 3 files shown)


aws-api/src/test/resources/error-null-properties.json
```diff
@@ -0,0 +1,15 @@
...
+  "data": {
+    "listTodos": null
+  },
+  "errors": [
+    {
+      "path": null,
+      "data": null,
+      "errorType": null,
+      "errorInfo": null,
+      "locations": null,
```
jamesonwilliams, line 11: Out of curiosity, you did observe this in a real world AppSync response, in the field? Should we file a bug and ask them to not send null-valued fields?

richardmcclellan: I did observe "path" : null in a response I got a few days ago.  Unfortunately I can't remember now exactly what that error was.  

I suppose it would be better to just not include null-valued fields, though I don't think the [GraphQL spec](https://spec.graphql.org/June2018/#sec-Errors) is explicit one way or the other.  Regardless though, the client should ignore null-valued fields, not throw an exception, which is what was happening.

PR #416, opened 2020-04-30
[aws-datastore] Implement DataStore clear feature

rjuliano:
*Issue #, if available:*

*Description of changes:*

# DataStore clear
Adding clear method to the DataStore plugin.

The intent of the functionality of the clear method is to blow away the entire underlying data store and reset it as needed. The current implementation of the DataStore plugin uses SQLite for the implementation of the LocalStorageAdapter interface.

## Summary of changes

* Added `clear` method to the `DataStoreCategory`, `DataStoreCategoryBehavior`, and `LocalStorageAdapter`
* Implement `clear` for `AWSDataStorePlugin`. It stops the orchestrator, invokes the clear method of the storage adapter, ~~and restarts the orchestrator~~.  The orchestrator is not restarted right away; instead, it is restarted upon the next interaction with the DataStore category.
* Implement `clear` for `SQLiteStorageAdapter`. 
  * The implementation closes the adapter and the current handle to the DB connection. Then it deletes the database file using the context and attempts to re-initialize the adapter.
   * Keep a reference to the `Context` object received in the `configure` method so we can use it to get access to the database file for clearing.

* Updated some of the test classes that are used to create a synchronous interface to the storage adapter.


jamesonwilliams:
I think if you `git rebase --pull origin master` or `git merge origin/master`, the integ tests should be ~passing now. @raphkim [\@Ignore'd the broken REST API tests, here](https://github.com/aws-amplify/amplify-android/pull/411/files).

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
```diff
@@ -285,4 +286,14 @@ public void terminate() throws DataStoreException {
             asyncDelegate.observe(emitter::onNext, emitter::onError, emitter::onComplete)
         );
     }
+
+    /**
+     * Invokes the clear method of the underlying adapter and
+     * either completes or throws an exception.
+     */
+    public void clear() {
+        Completable.fromObservable(emitter -> {
```
jamesonwilliams, line 295: Ah yea, good call. The `Await.result(...)` thing could basically be replaced with a `Single` and a `blockingGet()`. At the time I wrote `Await.result(...)` we hadn't started using `Rx` from the tests across different modules. Here in the DataStore tests, its doubly okay, though, since the actual _source code_ uses Rx.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesonwilliams, line 2: 2020

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+
+import io.reactivex.observers.TestObserver;
+
+import static com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.DATABASE_NAME;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class SQLiteStorageAdapterClearTest {
```
jamesonwilliams, line 42: `final`

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @BeforeClass
+    public static void enableStrictMode() {
+        StrictMode.enable();
+    }
+
+    @Before
+    public void setup() {
+        TestStorageAdapter.cleanup();
+        context = (ApplicationProvider.getApplicationContext());
```
jamesonwilliams, line 55: Drop the outer `(`,`)`

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+        adapter = TestStorageAdapter.create(AmplifyModelProvider.getInstance());
+        observer = TestObserver.create();
+        adapter.observe().subscribe(observer);
```
jamesonwilliams, line 58: There's a short-hand for this, [`Observable#test()`](http://reactivex.io/RxJava/javadoc/rx/Observable.html#test--):

```
TestObserver<StorageItemChange.Record> observer = adapter.observe().test();
```

It only saves a line there, but it doesn't need to be disposed, which is neat. And it's _slightly_ more expressive of the intention.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
wooj2, line 2: nit: date change from. 2019 to 2020

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -390,6 +390,22 @@ public void observe(
         onObservationFailure.accept(new DataStoreException("Not implemented yet, buster!", "Check back later!"));
     }
 
+    /**
+     * Stops all synchronization processes and deletes the
+     * SQLite database from device's local storage.
+     * @param onComplete Invoked if the call is successful.
+     * @param onError Invoked if an exception occurs.
```
wooj2, line 397: Seems like this should be changed to: "Invoked if not successful" ?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+            .build();
+
+        adapter.save(dummy);
+        List<BlogOwner> query = adapter.query(BlogOwner.class);
+        assertEquals(1, query.size());
+        //Check that exactly one event was received by the observer.
+        observer.assertValueCount(1);
+
+        adapter.clear();
+        assertFalse(context.getDatabasePath(DATABASE_NAME).exists());
```
jamesonwilliams, line 85: Nice this reads really well

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+        //check that it's not subscribed anymore.
+        observer.assertNotSubscribed();
+        //Assert the observer was terminated during the file removal.
+        observer.assertTerminated();
+    }
```
jamesonwilliams, line 90: What happens if you do `adapter.query(BlogOwner.class)` at this point?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
```diff
@@ -186,4 +186,14 @@ Cancelable observe(
      * @throws DataStoreException if something goes wrong during terminate
      */
     void terminate() throws DataStoreException;
+
+    /**
+     * Each implementation of this adapter interface will have its own
+     * interpretation of what clear means. At a high level, the intent is to destroy
```
wooj2, line 192: 1.  Remove "At a high level"
2.  In iOS, we are not making any guarantees on in-memory objects...  We still have some objects still in memory (that have read in configuration) so that customers  do not need to call configure() on amplify again if they want to use Datastore.
3.  This comment seems to be duplicated from the "*behavior.java" class... I'm not familiar with your code base, but my preference would be to only have it at the interface level.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertTrue(context.getDatabasePath(DATABASE_NAME).exists());
+        //check that it's subscribed
+        observer.assertSubscribed();
+
+        adapter.clear();
+        //check that it's not subscribed anymore.
+        observer.assertNotSubscribed();
+        assertFalse(context.getDatabasePath(DATABASE_NAME).exists());
+
+        adapter.clear();
```
jamesonwilliams, line 109: Ah, this is the actual `// Act` part of the test, right? That `clear()` can be called multiple times without raising hell? I missed it until my second read - can you highlight this via a comment or something?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -390,6 +390,22 @@ public void observe(
         onObservationFailure.accept(new DataStoreException("Not implemented yet, buster!", "Check back later!"));
     }
 
+    /**
+     * Stops all synchronization processes and deletes the
+     * SQLite database from device's local storage.
```
jamesonwilliams, line 395: Maybe a note about how/when this stuff is able to resume, afterwards? What if I have some changes in the outbound mutations queue? Will they be sent to AppSync, still?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.List;
+
+import io.reactivex.observers.TestObserver;
+
+import static com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.DATABASE_NAME;
```
wooj2, line 37: For unit tests, I think it is a better test not to share constants between test code and production code.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
```diff
@@ -186,4 +186,14 @@ Cancelable observe(
      * @throws DataStoreException if something goes wrong during terminate
      */
     void terminate() throws DataStoreException;
+
+    /**
+     * Each implementation of this adapter interface will have its own
```
jamesonwilliams, line 191: 👍 Good docs.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals(1, query.size());
+        //Check that exactly one event was received by the observer.
+        observer.assertValueCount(1);
+
+        adapter.clear();
+        assertFalse(context.getDatabasePath(DATABASE_NAME).exists());
+        //check that it's not subscribed anymore.
+        observer.assertNotSubscribed();
+        //Assert the observer was terminated during the file removal.
+        observer.assertTerminated();
```
wooj2, line 89: is there a way to assert that the success handler gets called?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
wooj2, line 109: is there a way to make sure that the error handler gets called?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -579,6 +579,25 @@ public synchronized void terminate() throws DataStoreException {
...
+    @Override
+    public void clear(@NonNull Action onComplete,
+                      @NonNull Consumer<DataStoreException> onError) {
+        try {
+            terminate();
+            sqliteStorageHelper.deleteDatabaseFromDisk();
+            onComplete.call();
+        } catch (Exception exception) {
+            DataStoreException dataStoreException = new DataStoreException(
+                "Error while trying to delete database from local storage.",
```
jamesonwilliams, line 594: Maybe, for more clarity, `"Error while trying to clear data from the local DataStore storage"` ?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -579,6 +579,25 @@ public synchronized void terminate() throws DataStoreException {
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void clear(@NonNull Action onComplete,
+                      @NonNull Consumer<DataStoreException> onError) {
+        try {
+            terminate();
+            sqliteStorageHelper.deleteDatabaseFromDisk();
+            onComplete.call();
+        } catch (Exception exception) {
```
jamesonwilliams, line 592: Let's try to make this exception handling more narrow. Which of these calls are actually expected to throw, and what types of exceptions will they throw?

Can you do:
```
try {
    justTheThingsThatThrow();
} catch (OnePossibleCaseException | AnotherCaseException failureToClear) {
    // wrap in DataStoreException and re-throw
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageHelper.java
```diff
@@ -172,6 +175,13 @@ public synchronized void update(
         }
     }
 
+    /**
+     * Deletes the SQLite database file from the device's local storage.
+     */
+    public void deleteDatabaseFromDisk() {
+        context.deleteDatabase(getDatabaseName());
```
jamesonwilliams, line 182: Should you call `super.close()` first, to be safe?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
jamesonwilliams: ^ I do agree. I made this change in my own tests probably 1-2 years ago.

It's a tradeoff of maintaining two copies of a constant, versus keeping the constant as an implementation detail (`private` field access, etc.) 

If the source code version changes, a self-maintained version in test would be able to catch it. But, if you just use the constant, you could have like:

```
@VisibleForTesting static final String GOOD_VALUE = "oopsActuallyBadPsych!";
```

If test still had a hardcoded `"goodValue"` somewhere, it'd catch it.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: Prior to my latest commits, it would just blow up. I've added code to more gracefully reset the local storage adapter after the file is removed.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: Fixing this.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Yeah...that makes more sense. Updating it.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
rjuliano: I've updated this. We have a little bit of duplication because we have an interface for the plugin as well as an interface for the storage adapter.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: Updating this one. I had made the change locally then stashed it and ended up getting lost.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: I modified the test quite a bit in the latest couple of commits. Added a few comments on the assertions.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageHelper.java
rjuliano: I actually removed this code. I'm calling the helper's close method from the adapter where I can have better control to restart things

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: We created a [wrapper class](https://github.com/aws-amplify/amplify-android/blob/782eb601676c1e94b85718c4a66028784e90a802/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java#L296) for that we use for testing that basically blocks until either the success handler gets called or bubbles up an exception. So if the call does not throw, we know the success handler was called.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: See comment above. The adapter is wrapped by a class that blocks the call until it either succeeds or throws.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Updated the comment and made a note to come up with a test to verify this.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
rjuliano: Updated message.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
rjuliano: This method has been changes. Take a look at the updated logic and let me know.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+        List<BlogOwner> blogOwners = adapter.query(BlogOwner.class, BlogOwner.NAME.eq(name));
+        assertNotNull(blogOwners);
+        assertEquals(0, blogOwners.size());
+    }
+
+    private void assertDbFileExists() {
+        assertTrue(context.getDatabasePath(DATABASE_NAME).exists());
+    }
+
+    private static final class TestFileObserver extends FileObserver {
```
jamesonwilliams, line 166: Ah, cool beans. I've never seen or used the [`FileObserver`](https://developer.android.com/reference/android/os/FileObserver), before! Good stuff.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+     * @throws IOException if it can't ready file creation time
+     * @throws InterruptedException if something happens when sleeping for 1 second
+     */
+    @Test
+    public void clearDeletesAndRecreatesDatabase() throws DataStoreException, IOException, InterruptedException {
+        assertDbFileExists();
+        assertEquals(0, fileObserver.createFileEventCount);
+        assertEquals(0, fileObserver.deleteFileEventCount);
+        BlogOwner blogger1 = createBlogger("Dummy Blogger Sr.");
+        BlogOwner blogger2 = createBlogger("Dummy Blogger Jr.");
```
jamesonwilliams, line 102: :-D

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+        //Save a record and check if it's there
+        adapter.save(blogger1);
+        assertRecordIsInDb(blogger1.getName());
+        //Verify observer is still alive
+        assertFalse(subscriberDisposableRef.get().isDisposed());
+        assertObserverReceivedRecord(blogger1);
+        //The ensures files are created at least 1 second apart
+        Thread.sleep(1000);
```
jamesonwilliams, line 110: But how come? Is it necessary?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+    @Test
+    public void clearDeletesAndRecreatesDatabase() throws DataStoreException, IOException, InterruptedException {
+        assertDbFileExists();
+        assertEquals(0, fileObserver.createFileEventCount);
+        assertEquals(0, fileObserver.deleteFileEventCount);
+        BlogOwner blogger1 = createBlogger("Dummy Blogger Sr.");
+        BlogOwner blogger2 = createBlogger("Dummy Blogger Jr.");
+        //Save a record and check if it's there
+        adapter.save(blogger1);
+        assertRecordIsInDb(blogger1.getName());
```
jamesonwilliams, line 105: In theory you could achieve this with:
```
assertEquals(blogger1, adapter.query(BlogOwner.class, blogger1.getId()).get(0));
```
But that gets kind of long, and loses the DAMP you have.

Maybe:
```
assertRecordIsInDB(BlogOwner.class, blogger1.getId());
```
To use the same type of arguments as the storage interface usually do, though?
```
private <T extends Model> void assertRecordIsInDB(T item) {
    List<T> results = adapter.query(item.getClass(), item.getId());
    assertEquals(1, results.size());
    assertEquals(item, results.get(0));
}
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+        //Check the new record is in the database
+        //and the old record is not.
+        assertRecordIsInDb(blogger2.getName());
+        assertRecordIsNotInDb(blogger1.getName());
+        assertObserverReceivedRecord(blogger2);
+        //Terminate the adapter
+        adapter.terminate();
+        //Verify observer was disposed.
+        assertTrue(subscriberDisposableRef.get().isDisposed());
+
```
jamesonwilliams, line 131: nit: an extra empty

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,210 @@
...
+    private BlogOwner createBlogger(String name) throws DataStoreException {
+        return BlogOwner.builder()
+            .name(name)
+            .build();
+    }
+
+    private void assertObserverReceivedRecord(BlogOwner blogger) {
+        for (StorageItemChange<? extends Model> owner : observer.values()) {
+            if (owner.item() instanceof BlogOwner &&
```
jamesonwilliams, line 142: There is also an `item.itemClass()`, you could do like `BlogOwner.class.isAssignableFrom(item.itemClass())` or something like that. (Or, `BlogOwner.class.equals(item.itemClass())`.)

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: It's not needed anymore. Before I started using the `FileObserver`, I was checking the created date of the file. Had to move away from that because creation date attribute is not available in older versions of Android.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: I'm going with `BlogOwner.class.isAssignableFrom(item.itemClass())`.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
rjuliano: Yeah...I initially used a method that would return me the created date of the file as an instance of `FileTime`, but it appears creation date was not available as an attribute in older versions of Android.

rjuliano:
> Looks pretty good to me. Have a question about the re-init part, inline. Love the main test case you wrote out!

I made a change to the logic that controls the re-initialization of the synchronization processes. It's no longer restarting right away. Instead, it restarts on the next interaction with the DataStore.

I believe I addressed most of the comments, but let me know if I missed anything. In the process of changing the way we restart the sync process, i realized there were some gaps and no real test to verify. 

The most significant change of this batch of changes is in `AWSDataStorePluginTest`. I added a test to verify the restart behavior.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -568,6 +572,35 @@ public synchronized void terminate() throws DataStoreException {
...
+     */
+    @Override
+    public synchronized void clear(@NonNull Action onComplete,
+                                   @NonNull Consumer<DataStoreException> onError) {
+        sqliteStorageHelper.close();
+        databaseConnectionHandle.close();
+
+        try {
+            context.deleteDatabase(DATABASE_NAME);
+        } catch (Exception exception) {
```
jamesonwilliams, line 586: Can we tighted this up at all? `SQLSpecialException | SQLOtherSpecialException sqlError)` ?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -95,22 +98,38 @@ public Orchestrator(
...
-        return Completable.fromAction(() -> {
-            storageObserver.startObservingStorageChanges();
-            subscriptionProcessor.startSubscriptions();
-            syncProcessor.hydrate().blockingAwait();
-            mutationProcessor.startDrainingMutationOutbox();
-            subscriptionProcessor.startDrainingMutationBuffer();
+    public synchronized Completable start() {
+        return Completable.fromSingle(single -> {
+            try {
+                if (!storageObserver.isObserving()) {
```
jamesonwilliams, line 104: Let's align the naming of all of these to their corresponding `start<whatever>` methods. Here, this would be come:
```
storageObserver.isObservingStorageChanges()
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -95,22 +98,38 @@ public Orchestrator(
...
+                    LOG.debug("Starting local storage observer.");
+                    storageObserver.startObservingStorageChanges();
+                }
+                if (!subscriptionProcessor.isObserving()) {
```
jamesonwilliams, line 108: And this one,
```
subscriptionsProcessor.isSubcribed()
```
Or
```
subscriptionsProcessor.hasSubscriptions()
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -95,22 +98,38 @@ public Orchestrator(
...
+                    LOG.debug("Starting subscription processor.");
+                    subscriptionProcessor.startSubscriptions();
+                }
+                syncProcessor.hydrate().blockingAwait();
+                if (!mutationProcessor.isObserving()) {
```
jamesonwilliams, line 113:
```
mutationProcessor.isDrainingMutationOutbox()
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -95,22 +98,38 @@ public Orchestrator(
...
+                    LOG.debug("Starting mutation processor.");
+                    mutationProcessor.startDrainingMutationOutbox();
+                }
+                subscriptionProcessor.startDrainingMutationBuffer();
```
jamesonwilliams, line 117: How come this one doesn't need an `if` guard like the others?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -95,22 +98,38 @@ public Orchestrator(
...
     @NonNull
-    public Completable start() {
-        return Completable.fromAction(() -> {
-            storageObserver.startObservingStorageChanges();
-            subscriptionProcessor.startSubscriptions();
-            syncProcessor.hydrate().blockingAwait();
-            mutationProcessor.startDrainingMutationOutbox();
-            subscriptionProcessor.startDrainingMutationBuffer();
+    public synchronized Completable start() {
+        return Completable.fromSingle(single -> {
```
jamesonwilliams, line 102: Would this be better as just:
```
Completable.create(emitter -> {
    try {
        tryToSucceed();
        emitter.onComplete();
    } catch (WhateverException whatever) {
       emitter.onError(whatever);
    } 
});
```

`fromCallable` or `fromAction` or `forRunnable` ... one of these factories must do the exception handling for you, I think, too?

like,
```
Completable.fromAction(() -> {
}).doOnError(thrownInsideOfAction -> {
    // I need to look up if it works like this ...
});
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -84,6 +84,15 @@ void startDrainingMutationOutbox() {
...
     }
 
+    /**
+     * Checks if the mutation processor is listening
+     * for events.
+     * @return True if there are listeners. False otherwise.
+     */
+    boolean isObserving() {
+        return disposable.size() > 0;
+    }
```
jamesonwilliams, line 94:
```suggestion
    /**
     * Checks if the mutation processor is actively observing the mutation outbox.
     * @return True if the mutation processor is subscribed the mutation outbox.
     */
    boolean isDrainingMutationOutbox() {
        return !disposable.isEmpty();
    }
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -123,6 +125,18 @@ void startSubscriptions() {
...
+            // so it can properly dispose of resources if necessary. For the AWS API plugin,
+            // this means means closing the underlying network connection.
+            synchronized (cancelable) {
+                Cancelable subscriptionOperation = cancelable.get();
+                if (subscriptionOperation != null) {
+                    LOG.info("Terminating subscription operation.");
+                    subscriptionOperation.cancel();
+                }
+            }
+        })
```
jamesonwilliams, line 139: L108 was _supposed_ to do this.

```
            emitter.setCancellable(cancelable::get);
```

However, I think I found elsewhere, that `setCancellable(...)` doesn't do what I thought it was supposed to do. There is also `emitter.setDispoable(Disposable...)`, though. I have a factory method elsewhere in the code base that builds a `Disposable` from our own `Cancelable` type. There's a note about "pulling it out into a utility class". Maybe its time.....

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -149,6 +163,15 @@ void startDrainingMutationBuffer() {
...
 
+    /**
+     * Checks if the subscription processor is listening
+     * for events.
+     * @return true if there are listeners. False otherwise.
+     */
+    boolean isObserving() {
+        return disposable.size() > 0;
+    }
+
```
jamesonwilliams, line 174: This one is tricker than in the other case, because the `SubscriptionProcessor` has an intermediate state, where it is buffering content, but is not processing it. So, we should decided which state we mean. From your current implementation, it would match two states:
- subscribed, not processing
- subscribed, processing under way

```suggestion
    /**
     * Checks if the subscription processor is currently buffering events from the remote AppSync instance.
     * @return true if the subscription processor has active subscriptions to AppSync
     */
    boolean isSubscribed() {
        return !disposable.isEmpty();
    }

```
The other meaning would also be easy to implement:
```
/**
 * Checks if the subscription processor is actively processing events found on its AppSync subscriptions.
 * @return true if the subscription processor is actively processing subscription events
 */
public isDrainingEventBuffer() {
    return disposable.size() == 2;
}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -74,16 +84,17 @@
     @Before
     public void setup() {
         this.context = getApplicationContext();
-        modelProvider = spy(SimpleModelProvider.builder()
-            .version(RandomString.string())
-            .addModel(Person.class)
-            .build());
+        modelProvider = spy(AmplifyCliGeneratedModelProvider.singletonInstance());
+        subscriptionCancelledCounter = new AtomicInteger(0);
+        subscriptionStartedCounter = new AtomicInteger(0);
```
jamesonwilliams, line 89: 0 is already the default for [`AtomicInteger()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html#AtomicInteger--)

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -158,29 +168,117 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
...
+            Completable.fromSingle(single -> { // Verify the record has been saved
+                awsDataStorePlugin.query(Person.class, results -> {
+                    Person actualPerson = results.next();
+                    assertNotNull(actualPerson);
+                    assertFalse(results.hasNext()); //we should only have one result.
+                    assertEquals(person2, actualPerson);
+                    single.onSuccess(true);
+                }, single::onError);
+            })
+        ).blockingGet(OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
jamesonwilliams, line 238: Does this construct work? It looks like you're making `assertFoo` statements on a different thread, which the jUnit test runner thread would not capture.

 If you put a `fail()` inside one these things, does the test runner thread catch it?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -158,29 +168,117 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
...
+
+        assertRemoteSubscriptionsStarted();
+
+        Person person1 = createPerson("Test", "Dummy I");
+        Person person2 = createPerson("Test", "Dummy II");
+        Completable.fromSingle(single -> { // Save a record to local store
+            awsDataStorePlugin.save(person1, itemSaved -> {
+                assertNotNull(itemSaved.item().getId());
+                assertEquals(person1.getLastName(), itemSaved.item().getLastName());
+                single.onSuccess(true);
```
jamesonwilliams, line 199: Since you ignore the result of the single, can't you just use `Completable.create`, `Completable.fromAction`, `Completable.fromRunnable`, etc?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -158,29 +168,117 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
...
+    private void assertRemoteSubscriptionsCancelled() {
+        // Check that we've had active subscriptions
+        assertTrue(subscriptionStartedCounter.get() > 0);
+        // And the number of started and cancelled are the same
+        assertEquals(subscriptionStartedCounter.get(), subscriptionCancelledCounter.get());
+    }
 
-        assertTrue(syncProcessorInvoked);
+    private void assertRemoteSubscriptionsStarted() {
+        // For each model, there should be 3 subscriptions setup.
```
jamesonwilliams, line 252: You could phrase this as,
```
final int diffTypesOfSubscriptions  = SubscriptionType.values().length;
```
and then get out of hte business of writing `3`, while also clarifying its meaning.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -194,16 +292,26 @@ private static ApiCategory mockApiCategoryWithGraphQlApi() throws AmplifyExcepti
...
-            onResponse.accept(new GraphQLResponse<>("{}", Collections.emptyList()));
+            // Calling onResponse with an invalid input generates an error in MutationOutbox.hasPendingMutation.
+            // Disabling it for now since it does not affect the assertions we need for this test.
+            // onResponse.accept(new GraphQLResponse<>("{}", Collections.emptyList()));
             return null;
         }).when(mockApiPlugin).mutate(any(GraphQLRequest.class), any(Consumer.class), any(Consumer.class));
 
         // Make believe that subscriptions return response immediately
         doAnswer(invocation -> {
-            int indexOfStartConsumer = 2;
```
jamesonwilliams, line 203: Oh dang, looks like this wasn't doin' much before huh lol

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -194,16 +292,26 @@ private static ApiCategory mockApiCategoryWithGraphQlApi() throws AmplifyExcepti
...
-            Consumer<String> onResponse = invocation.getArgument(indexOfStartConsumer);
-            onResponse.accept(RandomString.string());
-            return null;
+            int indexOfStartConsumer = 1;
+            Consumer<String> onStart = invocation.getArgument(indexOfStartConsumer);
+            GraphQLOperation<?> mockOperation = mock(GraphQLOperation.class);
+            doAnswer(opAnswer -> {
+                this.subscriptionCancelledCounter.incrementAndGet();
+                return null;
+            }).when(mockOperation).cancel();
```
jamesonwilliams, line 309: Oh fuck, a `doAnswer` inside of a `doAnswer`, Head just exploded.

The situation with the `Cancelable` return type and `String` "on started" token is such a nightmare, isn't it? On a brighter day, maybe we could cleanup that API signature so we don't have to get so clever about managing/mocking/testing it.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -568,6 +573,40 @@ public synchronized void terminate() throws DataStoreException {
...
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public synchronized void clear(@NonNull Action onComplete,
+                                   @NonNull Consumer<DataStoreException> onError) {
+        try {
+            LOG.debug("Shutting down thread pool for the storage adapter.");
+            threadPool.awaitTermination(THREAD_POOL_TERMINATE_TIMEOUT, TimeUnit.MILLISECONDS);
```
jpignata, line 584: Don't we need to explicitly shut down the threadpool?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -94,32 +97,88 @@ public Orchestrator(
...
-            .andThen(Completable.fromAction(() -> {
-                storageObserver.startObservingStorageChanges();
-                subscriptionProcessor.startSubscriptions();
-                syncProcessor.hydrate().blockingAwait();
-                mutationProcessor.startDrainingMutationOutbox();
-                subscriptionProcessor.startDrainingMutationBuffer();
-                LOG.info("Cloud synchronization is now fully active.");
-            }));
+        // Only start if it's stopped.
+        if (OrchestratorStatus.STOPPED.equals(status.get())) {
```
jpignata, line 116: I really like the use of finite state machines in the iOS codebase.. we should nick some of that.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
jpignata: Agree! These are good docs!

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -568,6 +573,40 @@ public synchronized void terminate() throws DataStoreException {
...
+            LOG.debug("Storage adapter thread pool shutdown.");
+        } catch (InterruptedException exception) {
+            LOG.warn("Storage adapter thread pool was interrupted during shutdown.", exception);
```
jpignata, line 587: Was it interrupted or just didn't stop within some timeframe?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -568,6 +573,40 @@ public synchronized void terminate() throws DataStoreException {
...
+                "See attached exception for details.");
+            onError.accept(dataStoreException);
+        }
+        LOG.debug("DataStore cleared. Re-initializing storage adapter.");
+
+        //Re-initialize the adapter.
+        initialize(context, schemaList -> {
+            onComplete.call();
+        }, exception -> {
+                onError.accept(new DataStoreException("Error occurred whilte trying to re-initialize " +
```
jpignata, line 604: maybe move the string to a new line instead of concatenating?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -94,32 +97,88 @@ public Orchestrator(
...
 
     /**
      * Stop all model synchronization.
      */
     public void stop() {
-        LOG.info("Intentionally stopping cloud synchronization, now.");
-        storageObserver.stopObservingStorageChanges();
-        subscriptionProcessor.stopAllSubscriptionActivity();
-        mutationProcessor.stopDrainingMutationOutbox();
+        if (OrchestratorStatus.STARTED.equals(status.get())) {
```
jpignata, line 150: Can you use the convenience method you created with this body above?

aws-datastore/src/main/java/com/amplifyframework/datastore/AsyncUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.reactivex.disposables.Disposable;
+
+/**
+ * Utility class for functions used throughout DataStore.
+ */
+public final class AsyncUtils {
+    private AsyncUtils() {
+
```
jpignata, line 32: Nit: extra line?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -185,12 +189,16 @@ public void configure(
...
-    public void initialize(@NonNull Context context) {
-        Completable completable = initializeStorageAdapter(context);
-        if (!api.getPlugins().isEmpty()) {
-            completable = completable.andThen(orchestrator.start());
+    public void initialize(@NonNull Context context) throws AmplifyException {
+        Throwable initError = initializeStorageAdapter(context)
+            .andThen(initializeOrchestrator())
+            .blockingGet(PLUGIN_INIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        if (initError != null) {
+            throw new AmplifyException("Failed to initialize the local storage " +
```
jpignata, line 197: Nit: maybe newline instead of concat?

jpignata: Why is this more likely to happen after clear?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,217 @@
...
+        TestFileObserver(@NonNull String path) {
+            super(path, FileObserver.CREATE | FileObserver.DELETE | FileObserver.DELETE_SELF);
+
+            this.deleteFileEventCount = 0;
+            this.createFileEventCount = 0;
+        }
+
+        /**
+         * The event handler, which must be implemented by subclasses.
+         *
```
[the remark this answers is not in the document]
jamesonwilliams: I believe the best practice is _not_ to provide Javadocs for `@Override` methods, since it will be mostly duplicative of the content of the interface/base class being overriden.

The [`JavadocMethod`](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocMethod) checkstyle has this a default setting:

> Javadoc is not required on a method that is tagged with the @Override annotation.

In a few place in this code base, we have the `@inheritDoc` Javadoc directive on `@Override`n methods.

A guy on [StackOverflow](https://stackoverflow.com/a/3062697/695787) goes on:

> I studied this problem during my PhD and found that in general folks will never be aware of the extra information in the overriding version if they are invoking through a supertype.

Holy Cow. Talk about "Diving Deep!"

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: @jpignata I don't think its more likely. I have this code in one of my PRs, too, though. Probably, we could carve this out into its own targeted PR. (Not saying we should, at this point, just that it would be possible.)

@rjuliano and I have been talking offline about reducing/removing all of our dangerous `blockPermanently()` type calls, which can hose threads. We only have a few such calls in source code, but there are quite a few in the tests, right now. We should start whittling away at them by providing some sane timeouts, so that nothing is wedged with permanence.

aws-datastore/src/main/java/com/amplifyframework/datastore/AsyncUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.reactivex.disposables.Disposable;
+
+/**
+ * Utility class for functions used throughout DataStore.
+ */
+public final class AsyncUtils {
```
jamesonwilliams, line 30: I've been working hard for us to avoid having any `FooUtils` classes (by name).

Would it make sense to call this class `Cancelables`? (with one `l`, like our `Cancelable`, and unlike the Java concurrency `Cancellable`.)

It would read like:

```
Disposable disposables = Cancelables.asDisposable(cancelable);
```

The advantage here is that `Cancelables` is only ever going to have a few things in it, and they'll all relate to that data type.

`AsyncUtils` could end up being abused in the future, and have all kinds of executors, schedulers, etc., in it., over time.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -568,6 +573,43 @@ public synchronized void terminate() throws DataStoreException {
...
+     */
+    @Override
+    public synchronized void clear(@NonNull Action onComplete,
+                                   @NonNull Consumer<DataStoreException> onError) {
+        try {
+            LOG.debug("Shutting down thread pool for the storage adapter.");
+            threadPool.shutdown();
+            if (!threadPool.awaitTermination(THREAD_POOL_TERMINATE_TIMEOUT, TimeUnit.MILLISECONDS)) {
+                threadPool.shutdownNow();
+            }
```
jamesonwilliams, line 587: This solved a bug, right? This must have been very tricky to find. Nice work.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
jamesonwilliams, line 116: To help with the indentation / nesting, can we invert this?

```
if (!OrchestratorStatus.STOPPED.equals(status.get())) {
    return initializationCompletable;
}
// Save 4 cols, here ...
```

You can save another eight more cols with the `andThen(` on L120, too, by:
```
initializationCompletable = mutationOutbox.load().andThen(Completable.fromAction(() -> {
    // 8 left of where it was 
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -94,32 +97,88 @@ public Orchestrator(
...
+        }
+
+    }
+
+    /**
+     * Represents possible status of the orchestrator.
+     */
+    enum OrchestratorStatus {
+        /**
+         * The orchestrator is stopping.
```
jamesonwilliams, line 167: True 'nuff, but when it is stopping, what does that mean for the system -- request to `start()` will be denied? What are the valid input and output states, from this step?

Same Q for all

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -114,15 +113,19 @@ void startSubscriptions() {
...
                 emitter::onError,
                 emitter::onComplete
-            ));
+            );
+            // When the observable is disposed, we need to call cancel() on the subscription
+            // so it can properly dispose of resources if necessary. For the AWS API plugin,
+            // this means means closing the underlying network connection.
+            emitter.setDisposable(
+                AsyncUtils.asDisposable(subscriptionCancelable)
+            );
```
jamesonwilliams, line 128: I think this probably solves a bug I've been seeing lately. Thanks!

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -21,49 +21,60 @@
...
-import io.reactivex.Completable;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.mockito.ArgumentMatchers.argThat;
```
jamesonwilliams, line 59: Can you use the one on the `org.mockito.Mockito` facade, instead? `org.mockito.Mockito.argThat`?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -21,49 +21,60 @@
...
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 @RunWith(RobolectricTestRunner.class)
 public final class AWSDataStorePluginTest {
-    private static final long OPERATION_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(1);
+    private static final String MOCK_API_PLUGIN_NAME = "MockApiPlugin";
+    private static final int ASSERTION_TIMEOUT_MS = 1000;
```
jamesonwilliams, line 72: ` = 1_000 /* ms */`, or ` = TimeUnit.SECONDS.toMillis(1);`

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -127,60 +137,144 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
...
+
+        synchronousDataStore.clear();
+        assertRemoteSubscriptionsCancelled();
+
+        // Save person 2
+        synchronousDataStore.save(person2);
+        Person result2 = synchronousDataStore.get(Person.class, person2.getId());
+        assertEquals(person2, result2);
+
+        verify(mockApiCategory, timeout(ASSERTION_TIMEOUT_MS).atLeastOnce())
```
jamesonwilliams, line 232: Oooh! I've never see this before. Nice find: `timeout(value).atLeastOnce()`

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
rjuliano: Done.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: Heck yeah...that's much better.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: I went through and updated the comments for each of those to be a bit more descriptive.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
rjuliano: Yeah...it's a pesky one too.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -225,4 +225,14 @@ void observe(
             @NonNull Consumer<DataStoreException> onObservationFailure,
             @NonNull Action onObservationCompleted
     );
+
+    /**
+     * Every class that implements this behavior will have its own
```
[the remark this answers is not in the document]
rjuliano: Updated the docs.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterClearTest.java
```diff
@@ -0,0 +1,217 @@
...
+        private int deleteFileEventCount;
+        /**
+         * Equivalent to FileObserver(file, FileObserver.ALL_EVENTS).
+         *
+         * @param path Directory to watch
+         */
+
+        @SuppressWarnings("deprecation")
+        TestFileObserver(@NonNull String path) {
+            super(path, FileObserver.CREATE | FileObserver.DELETE | FileObserver.DELETE_SELF);
```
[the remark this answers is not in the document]
rjuliano: I wasn't sure what Android's `deleteDatabase` method would do, so I figured I'd look for either one. It deletes both the database and the journal file as well.

PR #417, opened 2020-04-30
Auth Sign In/Sign Up Options Updates

TrekSoft:
Ignore the messy commit message history - this is due to this being forked off another work in progress branch while that was pending code review. The three changes made here are:

- Modifying user attributes to be a hard type vs. a generic Map<String, String> with predefined values

- Updating signIn/signUp options to match the design review

- Moving Cognito specific signIn/signUp options down to the plugin level


jamesonwilliams:
> Ignore the messy commit message history - this is due to this being forked off another work in progress branch while that was pending code review. The three changes made here are:

Potentially, try (key point is the `--squash`):
```
git checkout --track origin/AuthCategory -B AuthCategory
git merge --squash AuthChores # or whatever the local branch was
git checkout -B branch-to-push-for-pr
git push origin --set-upstream branch-to-push-for-pr
```

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -80,47 +75,73 @@ public boolean equals(Object obj) {
...
                 "userAttributes=" + userAttributes +
-                ", validationData=" + validationData +
                 '}';
     }
 
-    public static final class Builder {
-        private Map<String, String> userAttributes;
-        private Map<String, String> validationData;
+    public abstract static class Builder<T extends Builder<T>> {
+        private List<AuthUserAttribute> userAttributes;
```
jamesonwilliams, line 94: the list reference can be `final`, since you only do `addAll` on it, not re-assign the reference, beyond the constrcutor.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -18,60 +18,55 @@
...
-     */
-    @NonNull
-    public Map<String, String> getValidationData() {
-        return validationData;
-    }
-
     @NonNull
-    public static Builder builder() {
-        return new Builder();
+    public static Builder<?> builder() {
```
jamesonwilliams, line 51: `Builder<? extends AuthSignUpOptions>` would be an easy improvement, if you can't achieve like a `Builder<T>`, or `B` (is a` Builder<SubClassType>`, etc.)

Sync up with @raphkim he has known these struggles in his predictions builders as well

core/src/main/java/com/amplifyframework/auth/options/AuthSignInOptions.java
```diff
@@ -16,83 +16,21 @@
...
-        @NonNull
-        public Builder validationData(@NonNull Map<String, String> validationData) {
-            Objects.requireNonNull(validationData);
-            this.validationData.clear();
-            this.validationData.putAll(validationData);
-            return this;
-        }
 
+// Note: This is currently empty but exists here to support common sign in options in the future.
+public abstract class AuthSignInOptions {
```
jamesonwilliams, line 21: An abstract class that has no implementation and no policy is really just like an interface. Can this, and its inner-Builder be interfaces, instead?

Interfaces are always better than abstract classes, since Java doesn't support multiple inheritence of classes, but you can implement as many interfaces as you want.

The only exception is when you _have_ to use a class, because you want to enforce policy. And in that category:
1. Use a final class with composition. Don't need to worry about multiple inheritence - much less any inheritence, if the class is closed.
2. Use an abstract class that acts like an interface, but constrains what the extensible methods are able to achieve, through hard-coded policy.

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
```diff
@@ -0,0 +1,225 @@
...
+    public static AuthUserAttributeKey custom(String attributeKey) {
+        return new AuthUserAttributeKey(attributeKey);
+    }
+
+    /**
+     * Returns the String key for the attribute.
+     * @return the String key for the attribute
+     */
+    @NonNull
+    public String getKeyString() {
```
jamesonwilliams, line 196: `getAttributeKey()` , to match the field? Or, change the field name to match the method name?

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
```diff
@@ -0,0 +1,225 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth;
+
+import androidx.annotation.NonNull;
+import androidx.core.util.ObjectsCompat;
+
+public final class AuthUserAttributeKey {
```
jamesonwilliams, line 21: Hm. There is a lot of "opinion" in this class, in core. How many of these attributes come from Cognito?

core/src/main/java/com/amplifyframework/auth/AuthUserAttribute.java
```diff
@@ -0,0 +1,78 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth;
+
+import androidx.core.util.ObjectsCompat;
+
+public final class AuthUserAttribute {
```
jamesonwilliams, line 20: Yea, maybe. But, it looks a lot like an `androidx.core.util.Pair`. 

That might be so, but I do agree it is nice to own the entire surface of our own API, without relying on third-party types (AndroidX.)

You could, though, make `AuthUserAttribute` basically just a wrapper around a `Pair delegate` from `androidx.core.util.Pair`.

It's also so lightweight of a class that basically nothing I've said here matters lol :-D haha

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -86,7 +86,7 @@ void resendSignUpCode(
     void signIn(
             @Nullable String username,
             @Nullable String password,
-            @NonNull AuthSignInOptions options,
+            @Nullable AuthSignInOptions options,
```
jamesonwilliams, line 89: Instead of doing this, It would be better to provide an overloaded method that does not require any `options`.

If your options as a user are to call:
```
void signIn(@Nullable first, @Nullable last, @Nullable options)
```
as:
```
signIn(null, null, null, ok -> {}, notOk -> {})
```
Wouldn't you rather have just done:
```
signIn(ok -> {}, notOk -> {})
```

The null params end up snaking down through the code base and creating a lot of branch complexity once everything has a `if (thing == null)`  around it.

So, the way I would handle this is to have two copies of sign-in, one that requires some non-null params, and one that does not take in those params, at all.

Internal to the implementation of the second method, choose "sane defaults." `AuthSignInOptions.defaults()` or something, that you can proxy into the other method, which _does_ need the anon-null param.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/options/AWSCognitoAuthSignUpOptions.java
```diff
@@ -0,0 +1,117 @@
...
+                getUserAttributes(),
+                getValidationData()
+        );
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        } else if (obj == null || getClass() != obj.getClass()) {
```
jamesonwilliams, line 72: If I have empty Cognito sign-up options, and empty Third party sign-up options, are the options the same?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -148,11 +152,24 @@ public void signUp(
...
+        Map<String, String> validationData = new HashMap<>();
+
+        if (options.getUserAttributes() != null) {
+            for (AuthUserAttribute attribute : options.getUserAttributes()) {
+                userAttributes.put(attribute.getKey().getKeyString(), attribute.getValue());
+            }
+        }
+
+        if (options instanceof AWSCognitoAuthSignUpOptions) {
+            validationData = ((AWSCognitoAuthSignUpOptions) options).getValidationData();
```
jamesonwilliams, line 165: `.putAll()`?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -253,11 +270,17 @@ public void onError(Exception error) {
...
-        @NonNull AuthSignInOptions options,
+        @Nullable AuthSignInOptions options,
         @NonNull final Consumer<AuthSignInResult> onSuccess,
         @NonNull final Consumer<AuthException> onException
     ) {
-        AWSMobileClient.getInstance().signIn(username, password, null, new Callback<SignInResult>() {
+        Map<String, String> metadata = null;
+
+        if (options != null && options instanceof AWSCognitoAuthSignInOptions) {
+            metadata = ((AWSCognitoAuthSignInOptions) options).getMetadata();
```
jamesonwilliams, line 280: `putAll()`, into a non-`null` collection?

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -80,47 +75,73 @@ public boolean equals(Object obj) {
...
         } else {
             AuthSignUpOptions authSignUpOptions = (AuthSignUpOptions) obj;
-            return ObjectsCompat.equals(getValidationData(), authSignUpOptions.getValidationData()) &&
-                    ObjectsCompat.equals(getUserAttributes(), authSignUpOptions.getUserAttributes());
+            return ObjectsCompat.equals(getUserAttributes(), authSignUpOptions.getUserAttributes());
         }
     }
 
+    /**
+     * When overriding, be sure to include the parent properties in the output string.
```
jamesonwilliams, line 83: --> `be sure to include userAttributes in the ...`

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -18,60 +18,55 @@
...
     public int hashCode() {
         return ObjectsCompat.hash(
-                getUserAttributes(),
-                getValidationData()
+                getUserAttributes()
         );
     }
 
+    /**
+     * When overriding, be sure to include the parent properties in the comparison.
```
jamesonwilliams, line 67: In your base classes, you could probably do like:
```
if (!super.equals(obj))
```
But, you'll have to remove the class-type checks. You probably want to relax them to `(obj instanceof BaseType)` instead of "obj _is_ basetype"?

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
raphkim: ahaha i'm the master of making unnecessarily complicated generic type builders now

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
```diff
@@ -80,47 +75,73 @@ public boolean equals(Object obj) {
...
         }
+    }
 
-        @NonNull
-        public AuthSignUpOptions build() {
-            return new AuthSignUpOptions(Immutable.of(userAttributes), Immutable.of(validationData));
+    public static final class CoreBuilder extends Builder<CoreBuilder> {
+
+        @Override
+        public CoreBuilder getThis() {
```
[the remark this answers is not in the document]
TrekSoft: Yeah I'd done a bunch of research online because Builder extension is actually quite a difficult problem and this seemed to be the most widely accepted answer to the problem.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: Gotcha - good point

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: Ok yeah Builder<T> didn't work but that sounds like it would work and doubly enforce what already is enforced from the abstract Builder class definition.

core/src/main/java/com/amplifyframework/auth/options/AuthSignInOptions.java
TrekSoft: I debated this in my head and chose to go this route because this is a direct mirror of AuthSignUpOptions and very much expect to have common options go in here just like in it's counterpart AuthSignUpOptions so want to have it implemented in such a way that there is minimal code impact to the rest of the library when we do that.

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
TrekSoft: This was another thing I debated quite a bit. The reason for choosing this name is the following - if the user was wanting to get the string value of the attribute key, it would look like: myAttributeVariable.getKey().getAttributeKey() which seems very duplicative and confusing. 

I thought in this case it would be clearer what that method does naming it this way so the user would be doing myAttributeVariable.getKey().getKeyString()

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
TrekSoft: All of them - however, none of these are specific to the Cognito implementation so I think it's fair to start with this as the initial common list.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
TrekSoft: So I actually do have an overloaded method that does not require any options. I made this Nullable so the method that didn't take options could call this with null options since the options aren't the kind which take default values - they either exist or they don't. But if you feel strongly about this, I can have it pass a blank options object over to the other method instead and change this back to NonNull.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/options/AWSCognitoAuthSignUpOptions.java
TrekSoft: Hmm, I guessss so... will update

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: validationData in SignUpOptions is immutable so I don't think this is necessary

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: Same as above

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: ok

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: Alright

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/options/AWSCognitoAuthSignUpOptions.java
TrekSoft: In attempting to implement this here, I don't think this is a helpful thing to check and opens a whole new can of worms where I don't know what all other classes implement the base class so I can't check whether they have options defined which I don't have defined and thus are not equal. 

I'm going to stick with the stricter (and I think more traditional) definition that for two objects to be equal, they have to at least be the same class.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: If the child class has properties set which are not present in the base class, the two are not equal so combined with my explanatory comment above, I don't think this is the right path to go.

core/src/main/java/com/amplifyframework/auth/options/AuthSignUpOptions.java
TrekSoft: Hmm, so yeah this doesn't work because when you do ? extends AuthSignUpOptions.Builder it throws a raw types error because the type of T is itself a generic. I don't think it's a problem not having this since, as the online resource I got this from pointed out, the type of T is enforced in the builder class definition.

PR #418, opened 2020-05-01
fix datastore javadoc checkstyle

drochetti:
small PR to fix a checkstyle violation in master.

I'm not sure my docs are 100% accurate but my main goal is to get a green build. Follow up with improvements will be more than welcome. :)

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfigurationProvider.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfigurationProvider.java
@@ -16,7 +16,7 @@
 package com.amplifyframework.datastore;
 
 /**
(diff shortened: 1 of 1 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfigurationProvider.java
```diff
@@ -15,6 +15,18 @@
 
 package com.amplifyframework.datastore;
 
+/**
+ * Implementations of this interface should provide access to a <code>DataStoreConfiguration</code>
```
raphkim, line 19: instead of `<code>DataStoreConfiguration</code>`, there is `{@link DataStoreConfiguration}` :)

PR #424, opened 2020-05-05
Get integ tests passing again

jamesonwilliams:
This single PR is composed of three git commits, which are only related by a common goal of re-instating _passing tests_.

1. A recent change to refactor date/time handling changed the serialization strategy for `java.util.Date`. However, the integration tests rely on a set of existing Model classes, in `testmodels`, that require the old strategy. To re-enable the AWS API Plugin instrumentation tests, this PR re-instates the original behavior of the _`Date`_ type. To accomodate the new improvements that had been made, a new _`AWSTimestamp`_ type it added. It will behave how the _new_ use of `Date` _would_ have, and the Amplify CLI should code-gen using this new type in place of `Date`. 

2. The method count in the single dex file for the AWS API Plugin instrumentation tests now exceeds the limit of 64k methods. To address this, multi-dex support is added (it is needed only for devices < API 21, which we _do_ target, as our `minSdk` is `16`.)

3. Ignore a flaky storage around pause/resume.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.aws;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Custom list deserializer since some lists come back not as an array of the items but as an object which contains
+ * an items property with the list of items and a nextToken property for pagination purposes.
+ */
+final class GsonListDeserializer implements JsonDeserializer<List<Object>> {
+    private static final String APP_SYNC_ITEMS_KEY = "items";
+
+    @Override
+    @SuppressWarnings("unchecked") // Cast to Class<Object> for templateClassType
+    public List<Object> deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+            throws JsonParseException {
+
+        // Because this is a list and typeOfT is ParameterizedType we can be sure this is a safe cast.
+        final Class<Object> templateClassType;
+        if (typeOfT instanceof ParameterizedType) {
+            templateClassType = (Class<Object>) ((ParameterizedType) typeOfT).getActualTypeArguments()[0];
+        } else {
+            throw new JsonParseException("Expected a parameterized type during list deserialization.");
+        }
+
+        // If the json we got is not really a List and this List has a generics type...
+        if (json.isJsonObject()) {
+            JsonObject jsonObject = json.getAsJsonObject();
+
+            // ...and it is in the format we expect from AppSync for a list of objects in a relationship
+            if (jsonObject.has(APP_SYNC_ITEMS_KEY) && jsonObject.get(APP_SYNC_ITEMS_KEY).isJsonArray()) {
+                List<Object> response = new ArrayList<>();
+                JsonArray items = jsonObject.get(APP_SYNC_ITEMS_KEY).getAsJsonArray();
+
+                if (items.size() == 0) {
+                    return null;
+                } else {
+                    for (JsonElement item : items) {
+                        response.add(context.deserialize(item, templateClassType));
+                    }
+
+                    return response;
+                }
+            } else {
+                throw new JsonParseException(
+                    "Got JSON from an API call which was supposed to go with a List " +
+                    "but is in the form of an object rather than an array. It also is not in the standard " +
+                    "format of having an items property with the actual array of data so we do not know how " +
+                    "to deserialize it."
+                );
+            }
+        } else if (json.isJsonArray()) {
+            final List<Object> items = new ArrayList<>();
+            for (JsonElement item : json.getAsJsonArray()) {
+                items.add(context.deserialize(item, templateClassType));
+            }
+            return items;
+        }
+
+        throw new JsonParseException(
+            "Got a JSON value that was not an object or a list. " +
+            "Refusing to deserialize into a Java list."
+        );
+    }
+}
--- a/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
+++ b/aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
@@ -37,6 +37,7 @@ import com.amazonaws.mobileconnectors.s3.transferutility.TransferState;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.File;
@@ -195,6 +196,10 @@ public final class AWSS3StorageUploadTest {
      *         completed successfully before timeout
      */
     @Test
+    @Ignore(
+        "This test is not passing reliably. It is not currently known if this " +
+        "constitutes a defect in source or test code."
+    )
     @SuppressWarnings("unchecked")
     public void testUploadFileIsResumable() throws Exception {
         final CountDownLatch completed = new CountDownLatch(1);
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,7 @@ ext {
             annotation: 'androidx.annotation:annotation:1.1.0',
             appcompat: 'androidx.appcompat:appcompat:1.1.0',
             core: 'androidx.core:core:1.2.0',
+            multidex: 'androidx.multidex:multidex:2.0.1',
             test: [
                 core: 'androidx.test:core:1.2.0',
                 runner: 'androidx.test:runner:1.2.0',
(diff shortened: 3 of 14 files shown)


core/src/main/java/com/amplifyframework/core/model/AWSTimestamp.java
```diff
@@ -0,0 +1,60 @@
...
+
+package com.amplifyframework.core.model;
+
+/**
+ * The AWSTimestamp scalar type represents the number of seconds that have elapsed
+ * since 1970-01-01T00:00Z. Timestamps are serialized and deserialized as numbers.
+ * Negative values are also accepted and these represent the number of seconds
+ * til 1970-01-01T00:00Z.
+ */
+public final class AWSTimestamp {
```
raphkim, line 24: I think it would be helpful to provide additional constructors like 
```
public AWSTimestamp(Date date);
public AWSTimestamp(String text);
```

I think we should try to strongly associate `Date` with our new formats since it is what the customers are likely more familiar with. Our other AWS temporal datatypes are already coupled with `Date` if I remember correctly.

build.gradle
```diff
@@ -98,6 +99,7 @@ private void configureAndroidLibrary(Project project) {
         compileSdkVersion rootProject.ext.compileSdkVersion
 
         defaultConfig {
+            multiDexEnabled true
```
raphkim, line 102: wooo

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
```diff
@@ -195,6 +196,7 @@ public void testUploadFileIsCancelable() throws Exception {
      *         completed successfully before timeout
      */
     @Test
+    @Ignore("This test is failing reliably.")
```
raphkim, line 199: hmm is this an oxymoron? lol

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raphkim, line 2: 2020

testmodels/src/main/java/com/amplifyframework/testmodels/meeting/Meeting.java
```diff
@@ -31,7 +31,7 @@
     private final @ModelField(targetType="AWSDate") AWSDate date;
     private final @ModelField(targetType="AWSDateTime") AWSDateTime dateTime;
     private final @ModelField(targetType="AWSTime") AWSTime time;
-    private final @ModelField(targetType="AWSTimestamp") Date timestamp;
+    private final @ModelField(targetType="AWSTimestamp") AWSTimestamp timestamp;
```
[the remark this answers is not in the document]
jamesonwilliams: It would, yea.

core/src/main/java/com/amplifyframework/core/model/AWSTimestamp.java
jamesonwilliams: Good idea, will do

build.gradle
jamesonwilliams: Eh, ~sorta, haha :-D

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
jamesonwilliams: Haha yea. Actually, sorry, this `@Ignore` message is not setting a good example. It should have a bit more. Will update.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
jamesonwilliams: This file was actually from 2019. It was removed only for a few days.

raphkim: makes sense! Go on ahead then :)

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageUploadTest.java
raphkim: haha I thought it was a funny message because "prone-to-failure" vs "reliability".
I think the reason for `@Ignore` is sufficient with this already, so need to fix

build.gradle
raphkim: At one point this error got brought up early on in the discussion (when discussing extensively overloading our API methods for convenience), but we dismissed it because 60k is A LOT. At the time, we didn't think we would be adding a direct dependency on a third-party framework...

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
```diff
@@ -0,0 +1,80 @@
...
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Custom list deserializer since some lists come back not as an array of the items but as an object which contains
+ * an items property with the list of items and a nextToken property for pagination purposes.
+ */
+final class GsonListDeserializer implements JsonDeserializer<List<Object>> {
```
richardmcclellan, line 36: Strangely I could have sworn this deserializer was not actually used, and instead the logic for deserializing the "items" array was handled [in GsonGraphQLResponseFactory](https://github.com/aws-amplify/amplify-android/blob/4d85b459af2539c6d6b0586228fd7ed6cc320986/aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java#L141-L142). 

I'm guessing there's a use case that I'm not familiar with yet - I'll take a closer look at the integration tests today to try and understand this further.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Gson singleGson;
+    private final Gson listGson;
+
+    GsonListDeserializer() {
+        this.singleGson = GsonFactory.create();
+        this.listGson = GsonFactory.create(Collections.singletonMap(List.class, this));
```
richardmcclellan, line 42: Injecting these into the constructor is a pretty good approach, but I think a better is to use the `JsonDeserializationContext` that is already passed in.

Instead of:

```
listGson.fromJson(item, clazz)
```

use:

```
context.deserialize(item, clazz)
```

And instead of:

```
singleGson.fromJson(json, typeOfT)
```

use (you have to iterate through the array manually rather than just deserializing `json` itself to avoid endless recursion):

```
List<Object> items = new ArrayList();
for(JsonElement item : json.getAsJsonArray()) {
    items.add(context.deserialize(item, clazz);
}
return items;
```

Two main benefits to this approach:
 - No dependencies at all
 - You are guaranteed to be using the same Gson instance that was provided to deserialize the root node.

build.gradle
jamesonwilliams: Yea, we'd be busy coding to get to 65,536 ourselves. The vast majority of these come from our dependencies.

core/src/main/java/com/amplifyframework/core/model/AWSTimestamp.java
jamesonwilliams: I ended up with:
```
// "now"
AWSTimestamp timestamp = new AWSTimestamp();

// Any date
AWSTimestamp timestamp = new AWSTimestamp(new Date());

// An amount of time since the epoch
AWSTimestamp timestamp = new AWSTimestamp(5_000L, TimeUnit.MILLISECONDS);
```

testmodels/src/main/java/com/amplifyframework/testmodels/meeting/Meeting.java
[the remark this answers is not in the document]
richardmcclellan: Is this for compat with existing integration tests, or are you thinking this should be the ultimate design?

On one hand, my thought was that `AWSTimestamp` should map directly to `java.util.Date` because:
 - The data itself can be represented completely by `java.util.Date`
 - Simpler developer experience to have a way to have a `java.util.Date` object on the model itself, since it eliminates the need to call a `getDate()` conversion method or something before they can actually use it.


On the other hand, having our own `AWSTimestamp` is more congruent with `AWSTime`, `AWSDateTime`, `AWSDate`.

I guess what I'm saying is I don't feel that strongly one way or the other.   What were your thoughts?

build.gradle
richardmcclellan: Do you think this was due to the addition of the threeten library?  That would be unfortunate :/

raphkim: I think after we agreed to add rxJava as a core dependency, we already set a precedent. From that point on, going over the limit was kind of inevitable, so I wouldn't blame it on any particular library

jamesonwilliams: John had run this over the weekend - https://github.com/KeepSafe/dexcount-gradle-plugin

That probably added a few, but it definitely wasn't one of the big offenders. OkHttp and RxJava alone are almost half of it.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
jamesonwilliams: I was surprised by this, too. I was pretty sure we have some unit tests that contain listed items. Not sure where the discrepancy is coming from!

testmodels/src/main/java/com/amplifyframework/testmodels/meeting/Meeting.java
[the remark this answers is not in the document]
jamesonwilliams: To document our offline conversation, we agreed on these as the right trade-offs.

We should confirm with @drochetti, but believe that the developer has no path to create a `java.util.Date` in a non-model type. So, if our models _also_ don't create `java.util.Date`, they would not be part of the system.

This does allow us to re-visit what to do with `Date` in the future, too, I guess.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
jamesonwilliams: Awesome!! Thanks. (We implemented this together on call. Pushing update now.)

aws-api/src/main/java/com/amplifyframework/api/aws/GsonListDeserializer.java
richardmcclellan: Ok, I figured this out. 

[`buildSingleArrayResponse`](https://github.com/aws-amplify/amplify-android/blob/be8560576ed6d22da73b5b7239e942401309c7be/aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java#L95-L116) automatically skips the "items" level before parsing it with Gson.

[`buildSingleItemResponse`](https://github.com/aws-amplify/amplify-android/blob/be8560576ed6d22da73b5b7239e942401309c7be/aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java#L57-L93) does not, so it relies on `GsonListDeserializer`.

I guess that means we have no unit tests for `buildSingleItemResponse`, so definitely something to add at some point.

PR #430, opened 2020-05-05
[core] Ignore HubInstrumentedTest#multiplePublications

jamesonwilliams:
This test is flakey, and frequently fails. Further effort will be
required to determine a root cause. Meanwhile, we would like to bulid a
consistent baseline of passing tests, from which we can build.

### DIFF
--- a/core/src/androidTest/java/com/amplifyframework/hub/HubInstrumentedTest.java
+++ b/core/src/androidTest/java/com/amplifyframework/hub/HubInstrumentedTest.java
@@ -25,6 +25,7 @@ import com.amplifyframework.logging.Logger;
 
 import com.amazonaws.amplify.core.test.R;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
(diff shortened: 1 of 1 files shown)


core/src/androidTest/java/com/amplifyframework/hub/HubInstrumentedTest.java
```diff
@@ -140,6 +141,7 @@ public void noSubscriptionReceivedAfterUnsubscribe() throws InterruptedException
      * @throws InterruptedException when waiting for CountDownLatch to
      *                              meet the desired condition is interrupted.
      */
+    @Ignore("This test doest not pass reliably and warrants further investigration.")
```
jpignata, line 144: Small typo!

jamesonwilliams: Fixed, thanks!

PR #432, opened 2020-05-06
Implement online text to speech conversion

raphkim:
*Issue #, if available:*

*Description of changes:*
* No integration tests yet (how do I test against audio data output?)
* Tested in an sample app using [`MediaPlayer`](https://developer.android.com/reference/android/media/MediaPlayer) to play the outputted MP3 audio data


core/src/main/java/com/amplifyframework/predictions/result/TextToSpeechResult.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.amplifyframework.core.async.Result;
+
+import java.io.InputStream;
+import java.util.Objects;
+
+/**
+ * The result of the call to synthesize speech from text.
+ */
+public final class TextToSpeechResult implements Result {
```
richardmcclellan, line 28: (unrelated to this PR)

Thoughts on renaming `Result` to `Response` (and all subclasses) or something else?   I don't really like `Result` because it might cause some customer confusion with [Kotlin's Result class](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/).

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
```diff
@@ -0,0 +1,116 @@
...
+import com.amazonaws.services.polly.model.SynthesizeSpeechResult;
+import com.amazonaws.services.polly.model.TextType;
+
+import java.io.InputStream;
+
+/**
+ * Predictions service for performing text to speech conversion.
+ */
+final class AWSPollyService {
+    private static final int MP3_SAMPLE_RATE = 24_000;
```
richardmcclellan, line 45: Woah, TIL about [underscore literals](https://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html) :)

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
```diff
@@ -0,0 +1,116 @@
...
+            onSuccess.accept(TextToSpeechResult.fromAudioData(data));
+        } catch (PredictionsException exception) {
+            onError.accept(exception);
+        }
+    }
+
+    private InputStream synthesizeSpeech(String text, VoiceType voiceType) throws PredictionsException {
+        final String languageCode;
+        final String voiceId;
+        if (VoiceType.UNKNOWN.equals(voiceType)) {
```
richardmcclellan, line 79: Could you represent an unknown VoiceType with `null` instead of a `UNKNOWN` type?

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsTextToSpeechTest.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    @Test
+    public void testTextToSpeechReturnsNonNullResult() throws PredictionsException {
+        // Interpret english text and assert non-null result
+        final String englishText = Assets.readAsString("sample-text-en.txt");
+        TextToSpeechResult result = predictions.convertTextToSpeech(englishText);
+
+        // Assert non-null audio data
+        InputStream audioData = result.getAudioData();
+        assertNotNull(audioData);
```
jamesonwilliams, line 73: When I was in Alexa, we had some tests that used AWS services to do TTS and ASR. This was a ~fair external verification, since they were ~separate services. Here, we don't have such an external verification mechanism.

However, we do have mechanisms to:
1. Convert text to speech;
2. Convert speech to text.

you can have a test that does the "does speech to text work?" by providing an input wav, and expecting some valid text back.

For the "convert text to speech," you can assume the "speech to text" part works. Then do:
1. Arrange some text.
2. Act: convert it to speech.
3. Run speech to text (assuming that this component works), and get text back.
4. Assert that the returned text is ~close to the arranged text.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
```diff
@@ -0,0 +1,116 @@
...
+import com.amazonaws.services.polly.model.SynthesizeSpeechRequest;
+import com.amazonaws.services.polly.model.SynthesizeSpeechResult;
+import com.amazonaws.services.polly.model.TextType;
+
+import java.io.InputStream;
+
+/**
+ * Predictions service for performing text to speech conversion.
+ */
+final class AWSPollyService {
```
richardmcclellan, line 44: I think this should be "Amazon Polly", not "AWS Polly".   And not just for Polly, for the same change should probably be made for all the other Predictions Services as well.  Maybe something to address in a separate PR.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/SpeechGeneratorConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+import org.json.JSONObject;
+
+/**
+ * Configures the behavior for speech generation.
+ */
+public final class SpeechGeneratorConfiguration {
+    private static final String CONFIG_NAME = "speechGenerator";
+
+    private final String voice;
+    private final String language;
```
jamesonwilliams, line 33: If you have enums for these things, why not just use them everywhere?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/request/AWSPollyRequest.java
```diff
@@ -0,0 +1,61 @@
...
+import androidx.annotation.NonNull;
+
+import com.amplifyframework.predictions.aws.models.VoiceType;
+
+import java.util.Objects;
+
+/**
+ * Simple request instance for speech synthesis operation.
+ */
+public final class AWSPollyRequest {
```
jamesonwilliams, line 27: Don't forget to do equals, hashCode, toString on all of your models, request and response types.

core/src/main/java/com/amplifyframework/predictions/options/TextToSpeechOptions.java
```diff
@@ -0,0 +1,93 @@
...
+        this.voiceType = builder.getVoiceType();
+    }
+
+    /**
+     * Gets the custom voice type if specified.
+     * Null otherwise.
+     * @return the custom voice type
+     */
+    @Nullable
+    public String getVoiceType() {
```
jamesonwilliams, line 39: This is `String` since `VoiceType` is in the plugin, right? Any other options? (I agree, this is better than including the current `VoiceType` into core!)

core/src/main/java/com/amplifyframework/predictions/result/TextToSpeechResult.java
jamesonwilliams: Re: convos with Tim recently, I'm not sure our `interface Result {}` marker interface is providing us a lot of value. It seems like mostly an architectural vestige at this point. Originally, it might have turned into something like the Swift/Kotlin construct. But we don't really use it for anything important. Could even argue that its harmful, since it _looks_ important, but isn't actually.

My vote is just to `git rm` the thing.

https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/core/async/Result.java#L25

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
jamesonwilliams: Our internal tech writers' page on this topic does call it "Amazon Polly". Actually, all of the services that are part of the Predictions plugins are _Amazon_ whatevers.

jamesonwilliams: Verified against our internal tech writers' page -- all of the services in this category are branded as _Amazon_ whatevers.

core/src/main/java/com/amplifyframework/predictions/result/TextToSpeechResult.java
raphkim: ^ I agree... Result doesn't really serve any purpose.

Meanwhile, calling the actual classes `<purpose>Result` is in alignment with what iOS has, so I don't think I will be changing that to `Response`. Especially if we are removing `Result` class, then there shouldn't be namespace conflict with Kotlin anymore right?

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsTextToSpeechTest.java
raphkim: I was thinking about this... but it turns out we won't be able to support speech to text for GA... Android SDK doesn't have transcribe-streaming implemented

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/configuration/SpeechGeneratorConfiguration.java
raphkim: Enum is scoped to the plugin, since these names only make sense in the context of AWS Polly, and "voice" is a configurable value in core category options for text-to-speech...

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/request/AWSPollyRequest.java
raphkim: After this PR + dummy transcribe PR, I will go through every model to add them!

aws-predictions/src/androidTest/java/com/amplifyframework/predictions/aws/AWSPredictionsTextToSpeechTest.java
jamesonwilliams: Ah okay! Makes sense

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
raphkim: I'd prefer to leave all instances where a `VoiceType` is expected to be `@NonNull` as to minimize null-checking

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
raphkim: shoot! I'll have to go through the code to fix those ugh

core/src/main/java/com/amplifyframework/predictions/result/TextToSpeechResult.java
jamesonwilliams: https://github.com/aws-amplify/amplify-android/pull/440

core/src/main/java/com/amplifyframework/predictions/options/TextToSpeechOptions.java
raphkim: I figured it out! `VoiceType` is now a core predictions interface with `String getVoice();` method, and `AWSVoiceType` implements that voice type inside the plugin!

core/src/main/java/com/amplifyframework/predictions/models/VoiceType.java
```diff
@@ -0,0 +1,32 @@
...
+
+package com.amplifyframework.predictions.models;
+
+import androidx.annotation.NonNull;
+
+/**
+ * Contains details about a specific voice to be
+ * used for synthesizing speech.
+ */
+public interface VoiceType {
```
jamesonwilliams, line 24: Should this have both a `getVoiceName()` and `getLanguageCode()`, to mirror the values that are in the `AWSVoiceType`?

raphkim: iOS's core [`VoiceType`](https://github.com/aws-amplify/amplify-ios/blob/master/Amplify/Categories/Predictions/Models/VoiceType.swift) doesn't contain anything about language. Strangely enough, even the [plugin's extension](https://github.com/aws-amplify/amplify-ios/blob/cac730cc218edd7718c5bb4640975545fb6d0d2a/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/VoiceType%2BAWSExtension.swift) doesn't seem to contain language code (outside of the constant name that is), which means that they won't be able to support bilingual TTS feature.

raphkim: Since it wasn't clear whether the intention was to completely remove language from the equation or if they wanted to keep core `VoiceType` as generic as possible, I decided to keep the core `VoiceType` interface match up with iOS and only introduce the language code in the plugin specific voice type.

This allows us to support bilingual voices by separating `AWSVoiceType#HINDI_ADITI` and `AWSVoiceType#INDIAN_ENGLISH_ADITI`

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
raphkim: (to be done in a separate PR as to not introduce non-polly stuff in this PR)

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/AWSVoiceType.java
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Types of voice to synthesize speech with.
+ *
+ * @see <a href=https://docs.aws.amazon.com/polly/latest/dg/voicelist.html>Voices in Amazon Polly</a>
+ */
+@SuppressWarnings("JavadocVariable")
+public enum AWSVoiceType implements VoiceType {
+    ARABIC_ZEINA("Zeina", "arb"),
+    CHINESE_ZHIYU("Zhiyu", "cmn-CN"),
```
richardmcclellan, line 31:
```suggestion
    MANDARIN_CHINESE_ZHIYU("Zhiyu", "cmn-CN"),
```

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            // Override configuration defaults if explicitly specified in the options
+            languageCode = voiceType.getLanguageCode();
+            voiceId = voiceType.getName();
+        }
+
+        SynthesizeSpeechRequest request = new SynthesizeSpeechRequest()
+                 .withText(text)
+                 .withTextType(TextType.Text)
+                 .withLanguageCode(languageCode)
```
richardmcclellan, line 93: Apparently [languageCode is an optional parameter](https://github.com/aws-amplify/aws-sdk-android/blob/09ccb6110232c8103dcb5037f34840f1c2240261/aws-android-sdk-polly/src/main/java/com/amazonaws/services/polly/model/SynthesizeSpeechRequest.java#L47-L68), and if you don't include it, multilingual voices will fall back to a default language.  I guess this is how iOS is able to not include this parameter at all, and still have a mostly working implementation.  

I like that we now have multilingual support for Android.  I'd suggest reaching out to iOS and suggesting that they add similar support.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/AWSVoiceType.java
raphkim: good suggestion! Makes sense since we are following {VARIATION}_{LANGUAGE}_{NAME} scheme

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSPollyService.java
raphkim: yep! I also noticed this shortly after. It should definitely be supported by both platforms ideally.

PR #433, opened 2020-05-06
Refactors methods to match design and adds unit tests

TrekSoft:
Refactors existing methods to match final interface design: https://quip-amazon.com/V550Arr1aKJN/Amplify-Auth-Category-Interface

Also adds unit tests to ensure that this wrapper implementation calls the proper AWSMobileClient methods with the expected parameters and parses its results into the proper Amplify result objects.


aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Test that the current implementation of Auth as a wrapper of AWSMobileClient calls the correct
+ * AWSMobileClient methods with the correct parameters when the different Auth methods are called.
+ */
+@RunWith(RobolectricTestRunner.class)
+@SuppressWarnings("unchecked")
```
jamesonwilliams, line 74: Can you add a `// Where do they come from?` explanation after it?

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+    private final String destination = "e***@email.com";
+    private final String deliveryMedium = "sms";
+    private final String attributeName = "email";
+    private final String username = "username";
+    private final String password = "password123";
+    private final String newPassword = "newPassword123";
+    private final String attributeKey = AuthUserAttributeKey.email().getKeyString();
+    private final String attributeVal = "email@email.com";
+    private final String confirmationCode = "confirm";
+    private final Map<String, String> metadata = Collections.singletonMap("aCustomKey", "aCustomVal");
```
jamesonwilliams, line 88: At least `private static final String SCREAMING_SNAKE_VAR` -- but, do you have some data models that you can put these in, so that there aren't just a heap of `String`? Like maybe an `new SomeCognitoType("username", "email@foo.tld");`, etc?

Lastly, the non-final fields should appear _after_ these constants (L76, L77.)

In unit tests, I prefer _not_ to store constants at the class level. This is because, often, the values are only used in individual tests. There are still a lot of good reasons to label your constants, though.

```
@Test(expected = FrickinMurdersItException.class)
public void testCookieMonsterWillPower() {
    final String temptation = "Cookie";
    CookieMonster monster = new CookieMonster();
    monster.temp(temptation);
    // Well, what did we expect, sigh.
}
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+    @Before
+    public void setup() throws AmplifyException {
+        mobileClient = mock(AWSMobileClient.class);
+        authPlugin = new AWSCognitoAuthPlugin(mobileClient, userId);
```
jamesonwilliams, line 93: Were you able to do:
```
AuthCategory auth = new AuthCategory();
auth.addPlugin(new AWSCognitoAuthPlugin());
CategoryConfiguration config = // blah blach
auth.configure(config, getApplicationContext());
auth.initialize(getApplicationContext());
```
And then test against the category? You'd get more coverage that way.

Testing a thin veneer of invocation verifications on just the plugin is going to be a little brittle.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+    /**
+     * Tests that the configure method of the Auth wrapper of AWSMobileClient (AMC) calls AMC.initialize() with the
+     * passed in context, a new AWSConfiguration object containing the passed in JSONObject, and a callback object
+     * which causes configure to complete successfully in the onResult case.
+     * @throws AmplifyException an exception wrapping the exception returned in onError of AMC.initialize()
+     * @throws JSONException has to be declared as part of creating a test JSON object
+     */
+    @Test
+    public void testConfigure() throws AmplifyException, JSONException {
+        UserStateDetails userStateDetails = new UserStateDetails(UserState.SIGNED_OUT, null);
```
jamesonwilliams, line 105: Can you provide a version of this constructor which does not require the `null` value?

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+        Context context = getApplicationContext();
+        JSONObject testConfig = new JSONObject().put("TestKey", "TestVal");
+
+        doAnswer(invocation -> {
+            assertEquals(context, invocation.getArgument(0));
+            assertEquals(testConfig.toString(), invocation.getArgument(1).toString());
```
jamesonwilliams, line 111: You shouldn't need to check this. This is guaranteed, based on the matching you do in your `.when(...)`.

The way this construct usually fails is if `.when` does not match, when you expect it to. In this case, the `doAnswer` part never gets called. So the problem isn't what happens inside of the lambda, rather that it's tricky to know if the lambda even got called (without checking for _that_.)

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+            Callback<UserStateDetails> callback = invocation.getArgument(2);
+            callback.onResult(userStateDetails);
+            return null;
+        }).when(mobileClient).initialize(any(Context.class), any(AWSConfiguration.class), any(Callback.class));
+
+        authPlugin.configure(testConfig, context);
+        verify(mobileClient, times(1))
```
jamesonwilliams, line 119: `times(1)` is the default, I believe. So you can just say `verify(foo).gotFooed()`. The `times(int)` comes into play when the value is _not_ `1`.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+                .initialize(any(Context.class), any(AWSConfiguration.class), any(Callback.class));
+    }
+
+    /**
+     * Tests that if AWSMobileClient returns an error callback during initialization, the Auth configure method
```
jamesonwilliams, line 124: > If {@link AWSMobileClient} emits an error during initialization, the {@link Auth#configure(...)} method should throw that same exception on its calling thread.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+            null
+        );
+
+        doAnswer(invocation -> {
+            assertEquals(username, invocation.getArgument(0));
+            assertEquals(password, invocation.getArgument(1));
+            Map<String, String> attributeMap = invocation.getArgument(2);
+            assertTrue(attributeMap.containsKey(attributeKey));
+            assertEquals(attributeVal, attributeMap.get(attributeKey));
+            assertEquals(metadata, invocation.getArgument(3));
```
jamesonwilliams, line 162: These are basically just testing your test. The `doAnswer` for faking a behavior on a a component. You'd want to make assertions on the final state, after your test. But here, it's sort of intermediate, before you do `callback.onResult(amcResult)`.

Outside of this `doAnswer().when()` constrcut, later, I'd expect some like `assertTrue(hasSomePropertyOfInterest(amcResult))`.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+        authPlugin.signUp(
+            username,
+            password,
+            AWSCognitoAuthSignUpOptions.builder()
+            .userAttributes(
+                Collections.singletonList(new AuthUserAttribute(AuthUserAttributeKey.email(), attributeVal))
+            )
+            .validationData(metadata)
+            .build(),
```
jamesonwilliams, line 178: Indentation here is -4 from where it ought to be, I believe. It's not clearly indented from the `username`, `password,` etc.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+            result -> validateSignUpResult(result, AuthSignUpStep.CONFIRM_SIGN_UP_STEP),
+            error -> fail());
```
jamesonwilliams, line 180: If this callback occurs on a different thread, `fail()` may not do what you think it will. `fail()` needs to be called on the _test runner_ thread, to halt the test. If a background thread crashes, but the `@Test` has already moved on, jUnit won't catch it.

TLDR: use something like `Await.result` to pull the exception back into your calling thread, and then do something with it, from that thread. `Completable.fromAction(() -> {}).blockingAwait()` is another way to achieve the same.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+        verify(mobileClient, times(1))
+                .signUp(any(String.class), any(String.class), any(Map.class), any(Map.class), any(Callback.class));
```
jamesonwilliams, line 183: You can do these `any` in a way that respects the type system:

```
Mockito.<Map<Key, Value>>any()
```
Instead of:
```
Mockito.any(/* rawtypes */ Map.class)
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+        authPlugin.signIn(
+            username,
+            password,
+            AWSCognitoAuthSignInOptions.builder().metadata(metadata).build(),
+            result -> validateSignInResult(
+                result,
+                false,
+                AuthSignInStep.CONFIRM_SIGN_IN_WITH_SMS_MFA_CODE
+            ),
+            error -> fail());
```
jamesonwilliams, line 294: To my earlier point about needing to make these calls synchronous -- this is why our other components all have a corresponding test utility called like `SynchronousWhatever`. It allows us to write easy jUnit tests that perform actions in a serial way. I would suggest you make like a `SynchronousAuth` utility, which wraps the `AuthCategoryBehavior`. Then, you could either do `SynchronousAuth.delegatingTo(authCategory)`, or `SynchronousAuth.delegatingTo(authPlugin)`, since both are examples of the `AuthCategoryBehavior`.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+     * Validate the sign up result is what was expected.
+     * @param result The received result
+     * @param targetStep The correct value for the next step (the only part of the response that varies in these tests)
+     */
+    private void validateSignUpResult(AuthSignUpResult result, AuthSignUpStep targetStep) {
+        AuthNextSignUpStep nextStep = result.getNextStep();
+        validateCodeDeliveryDetails(nextStep.getCodeDeliveryDetails());
+        assertTrue(result.isSignUpComplete());
+        assertEquals(targetStep, nextStep.getSignUpStep());
+
```
jamesonwilliams, line 443: nit empty line

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+        AWSMobileClient client = authPlugin.getEscapeHatch();
+        assertEquals(mobileClient, client);
+    }
+
+    /**
+     * Validate the sign up result is what was expected.
+     * @param result The received result
+     * @param targetStep The correct value for the next step (the only part of the response that varies in these tests)
+     */
+    private void validateSignUpResult(AuthSignUpResult result, AuthSignUpStep targetStep) {
```
jamesonwilliams, line 438: Instead of this, you can usually do like:
```
org.junit.Assert.assertEquals(highLevelExpectedObj, highLevelActualObj)
```

This works if the high level object has a sane `equals()` implementation.

Seems likely that `AWSMobileClient` (and its associated classes) may not.

If that's the case, the next best you can do is like:
```
private static void assertEquals(HighLevelType expected, HighLevelType actual) {
    // Assert field-by-field equality
}
```

aws-auth-cognito/build.gradle
```diff
@@ -22,4 +22,18 @@ dependencies {
     implementation dependency.androidx.appcompat
     implementation dependency.aws.mobileclient
     implementation dependency.aws.cognitoauth
+
+    testImplementation project(path: ':testutils')
+    testImplementation dependency.junit
+    testImplementation dependency.mockito
```
jamesonwilliams, line 28: I was surprised that you don't need to include a special second dependency for the `final` value mocking stuff. Isn't there like an `mockito-inline`, or something? Maybe that was an older version....

aws-auth-cognito/build.gradle
```diff
@@ -22,4 +22,18 @@ dependencies {
...
+    testImplementation (dependency.robolectric) {
+        // https://github.com/robolectric/robolectric/issues/5245
+        exclude group: 'com.google.auto.service', module: 'auto-service'
+    }
+    testImplementation dependency.androidx.test.core
+
+    androidTestImplementation project(path: ':testutils')
+    androidTestImplementation dependency.androidx.annotation
+    androidTestImplementation dependency.androidx.test.runner
+    androidTestImplementation dependency.androidx.test.junit
```
jamesonwilliams, line 38: You don't have an instrumentation tests, so you don't need to include anything that begins with `androidTest*`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -452,16 +414,23 @@ public void onError(Exception exception) {
...
             @NonNull Consumer<AuthException> onException
     ) {
-        AWSMobileClient.getInstance().forgotPassword(username, new Callback<ForgotPasswordResult>() {
+        awsMobileClient.forgotPassword(username, new Callback<ForgotPasswordResult>() {
             @Override
             public void onResult(ForgotPasswordResult result) {
                 if (result.getState().equals(ForgotPasswordState.CONFIRMATION_CODE)) {
-                    onSuccess.accept(convertCodeDeliveryDetails(result.getParameters()));
+                    onSuccess.accept(new AuthResetPasswordResult(
+                            false,
```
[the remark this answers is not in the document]
TrekSoft: The thing about this is that we do want the value being stored and accessed by the customer to be a boolean vs. an enum and a customer would never need to construct this result object so in this case, I think the extra complexity this would introduce isn't justified.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -452,16 +414,23 @@ public void onError(Exception exception) {
...
+                            new AuthNextResetPasswordStep(
+                                    AuthResetPasswordStep.CONFIRM_RESET_PASSWORD_WITH_CODE,
+                                    null,
```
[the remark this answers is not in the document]
TrekSoft: Ok

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -517,14 +486,34 @@ public void onError(Exception exception) {
...
+            public void onResult(Void result) {
+                onSuccess.call();
+            }
+
+            @Override
+            public void onError(Exception error) {
+                onError.accept(new AuthException(
+                        "Failed to change password",
+                        error,
+                        "See attached exception for more details"
```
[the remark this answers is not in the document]
TrekSoft: Hmm, I think we still want to force developers to think about whether there is a helpful explanation they can give even with a cause - there are many cases where there's going to be an underlying exception where a helpful suggestion is still needed.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/SignInStateConverter.java
```diff
@@ -0,0 +1,56 @@
...
+        convertSignInStateInit.put(SignInState.SMS_MFA, AuthSignInStep.CONFIRM_SIGN_IN_WITH_SMS_MFA_CODE);
+        convertSignInStateInit.put(SignInState.CUSTOM_CHALLENGE, AuthSignInStep.CONFIRM_SIGN_IN_WITH_CUSTOM_CHALLENGE);
+        convertSignInStateInit.put(SignInState.NEW_PASSWORD_REQUIRED, AuthSignInStep.CONFIRM_SIGN_IN_WITH_NEW_PASSWORD);
+        convertSignInStateInit.put(SignInState.DONE, AuthSignInStep.DONE);
+        CONVERT_SIGN_IN_STATE = Collections.unmodifiableMap(convertSignInStateInit);
+    }
+
+    public static AuthSignInStep getAuthSignInStep(SignInState fromState) throws AuthException {
+        if (CONVERT_SIGN_IN_STATE.containsKey(fromState)) {
+            return CONVERT_SIGN_IN_STATE.get(fromState);
```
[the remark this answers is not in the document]
TrekSoft: Yup yup, makes sense

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/SignInStateConverter.java
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+            throw new AuthException("Unsupported sign in state",
+                "We currently do not support the " + fromState + " state from AWSMobileClient. " +
+                    "If this error is reached, please report it as a bug");
```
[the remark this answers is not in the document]
TrekSoft: Yep it's one of the things I have on our platform wide discussion topics list since this is something we should align on across Amplify Android.

core/src/main/java/com/amplifyframework/auth/result/step/AuthSignUpStep.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.result.step;
+
+public enum AuthSignUpStep {
```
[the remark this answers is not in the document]
TrekSoft: We debated this quite a bit in the naming meeting and landed on this because this is only used to indicate what the next step in a flow is, not what state the plugin is in. When someone calls a method which could start a flow, they will receive a boolean indicating if the operation is successful and a nextStep variable indicating if there is a next step or if it's done (in which case nextStep is not null but DONE).

core/src/main/java/com/amplifyframework/auth/result/step/AuthNextSignInStep.java
```diff
@@ -0,0 +1,121 @@
...
+import androidx.core.util.ObjectsCompat;
+
+import com.amplifyframework.auth.AuthCodeDeliveryDetails;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+public final class AuthNextSignInStep {
+    private final AuthSignInStep signInStep;
```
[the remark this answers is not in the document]
TrekSoft: Yeah this is what the naming meetings landed on - will add a comment.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -176,12 +163,26 @@ void forgotPassword(
...
+     * @param oldPassword The user's existing password
+     * @param newPassword The new password desired on the user account
+     * @param onSuccess Success callback
+     * @param onError Error callback
+     */
+    void updatePassword(
+            @NonNull String oldPassword,
+            @NonNull String newPassword,
+            @Nullable Action onSuccess,
+            @Nullable Consumer<AuthException> onError
```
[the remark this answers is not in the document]
TrekSoft: Yes - this was actually something I wanted to ask you about. Will try to catch up with you on this today.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -132,19 +132,6 @@ void signInWithUI(
...
     void handleSignInWithUIResponse(Intent intent);
 
-    /**
-     * Allows a user to sign in to your app using their credentials from another provider (e.g. Facebook, Google, etc).
-     * @param provider The provider you are signing in with (e.g. Facebook, Google, etc.)
-     * @param token Token retrieved from the social provider's authentication code.
-     * @param onSuccess Success callback
-     * @param onError Error callback
-     */
-    void signInWithSocial(
```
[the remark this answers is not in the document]
TrekSoft: Yeah it was decided to scratch this for P0 since they don't want a method which doesn't create a user in the user pool.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,459 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import java.util.Map;
+
+import edu.emory.mathcs.backport.java.util.Collections;
```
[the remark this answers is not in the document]
TrekSoft: Riiiip - I don't know how that happened. Thanks for the callout.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Sure

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Metadata is an open ended map in a number of our methods which we decided to keep open because it's meant for sending custom data. Also, I put these at the class level because these same values are used throughout the various methods so figured it would be better to have them defined once rather than redefining the same dummy values in each test.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Oh yeah, I was going to update later to try that - let me see.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: This is an old AWSMobileClient object

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Soo... the actual issue here is that we're passing in a JSON Object (testConfig) and that then gets wrapped in an AWSConfiguration object (invocation.getArgument(1)). The AWSConfiguration object returns the JSON object in its toString method so I'm using that to make sure it's wrapping the original JSON object passed in. I should probably add another validation here though to ensure that invocation.getArgument(1) is indeed an AWSConfiguration object.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Aha, good to know

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Ok

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: No so what's happening here is I'm testing that AWSMobileClient is being called with the right parameters. The whole point of these tests, per my comments on them, is to ensure that the parameters the user is passing in are then being properly routed to the AWSMobileClient method we're wrapping and that the result which comes back from that method is properly wrapped / converted into the right Amplify result object.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Ok

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Ok, I may need to sync up with you on this

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Ok

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Will sync up with you

aws-auth-cognito/build.gradle
TrekSoft: It's a property under resources/mockito-extensions

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Yeah they aren't compatible and in this case I'm using default dummy values so while I could do it that second way, it'd just be less efficient code wise.

aws-auth-cognito/build.gradle
TrekSoft: Gotcha - thanks

aws-auth-cognito/build.gradle
```diff
@@ -22,4 +22,18 @@ dependencies {
...
+    testImplementation dependency.junit
+    testImplementation dependency.mockito
+    testImplementation (dependency.robolectric) {
+        // https://github.com/robolectric/robolectric/issues/5245
+        exclude group: 'com.google.auto.service', module: 'auto-service'
+    }
+    testImplementation dependency.androidx.test.core
+
+    androidTestImplementation project(path: ':testutils')
+    androidTestImplementation dependency.androidx.annotation
```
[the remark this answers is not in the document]
TrekSoft: Kk

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: It was actually from the generics issue which you commented on below so I was able to just remove it altogether after addressing that.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Turns out that with Java 8 the type is redundant and the best practice is to just use any() which infers the proper type

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,458 @@
...
+            return null;
+        }).when(mobileClient).signUp(any(), any(), any(), any(), any());
+
+        AuthSignUpResult result = synchronousAuth.signUp(
+                USERNAME,
+                PASSWORD,
+                AWSCognitoAuthSignUpOptions.builder()
+                    .userAttributes(
+                        Collections.singletonList(new AuthUserAttribute(AuthUserAttributeKey.email(), ATTRIBUTE_VAL))
+                    )
```
jamesonwilliams, line 205: Feels like a lot of boilerplate for the user to add a dang attribute, no?

Can we achieve soemthing more like:

```
AWSCognitoAuthSignUpOptions.builder()
    .userAttribute(AuthUserAttributeKey.email(), ATTRIBUTE_VAL)
    .build();
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,458 @@
...
+                    .validationData(METADATA)
+                    .build()
+        );
+
+        validateSignUpResult(result, AuthSignUpStep.CONFIRM_SIGN_UP_STEP);
+        verify(mobileClient).signUp(any(), any(), any(), any(), any());
```
jamesonwilliams, line 211: You can probably skip the `doAnswer` stuff, and then just use `eq` to check what is passed in.

Most of these tests should just be like:
```
// Arrange some options foo, bar, baz
Foo foo = new Foo();
Bar bar = mock(Bar.class);
Baz baz = getABaz();

// Act by calling Auth API
auth.signUp(foo, bar, baz, whatever);

// veriify that the mobile client was called in a certain way...
verfiy(mobileClient).signUp(eq(foo), eq(baz));
```

Or, in some cases you might not be able to use`eq`, directly. In which case you can do:
```
auth.signUo("foo", "bar", "baz");

ArgumentCaptor<TypeYouCantMake> whatDidItGiveMe = ArgumentCaptor.forClass(TypeYouCantMake.class);
verify(mobileClient).signUp(whatDidItGiveMe.capture(), eq(bar), eq(baz));

Foo foo = whatDidItGiveMe.value().getSomeContainedField();
assertEquals(expectedFoo, foo); // Ah!!  the thing it gave us _contained_ an expected foo.
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,458 @@
...
+     * @throws JSONException has to be declared as part of creating a test JSON object
+     */
+    @Test
+    public void testConfigure() throws AmplifyException, JSONException {
+        UserStateDetails userStateDetails = new UserStateDetails(UserState.SIGNED_OUT, null);
+        Context context = getApplicationContext();
+        JSONObject pluginConfig = new JSONObject().put("TestKey", "TestVal");
+        JSONObject json = new JSONObject().put("plugins",
+                new JSONObject().put(
+                    new AWSCognitoAuthPlugin().getPluginKey(),
```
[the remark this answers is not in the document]
TrekSoft: Fair enough - definitely felt dirty doing this

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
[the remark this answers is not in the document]
TrekSoft: Per the design meeting, the value the customer is interacting with should be a boolean (in this case isPasswordReset) so given that plugin developers are constructing this as a result object, I think it's more straightforward for them to construct it with the raw values they intend the customer to receive rather than having to check the implementation of isPasswordReset() to see that while it's constructed with an ENUM, it ends up being converted to a boolean result.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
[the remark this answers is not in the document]
TrekSoft: ok

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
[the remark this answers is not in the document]
TrekSoft: I agree - per my other comment, this is a framework wide discussion we should have about the proper way to do this on Android.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Yeah it is - the options object stores a list of attributes but I guess I could write some logic in the builder to handle taking a single attribute. Would just have to handle the case properly where someone uses both like `builder().userAttribute(...).userAttributes(...)`

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -0,0 +1,458 @@
...
+                    pluginConfig
+                )
+        );
+        AuthCategoryConfiguration authConfig = new AuthCategoryConfiguration();
+        authConfig.populateFromJSON(json);
+
+        doAnswer(invocation -> {
+            assertEquals(context, invocation.getArgument(0));
+            assertEquals(pluginConfig.toString(), invocation.getArgument(1).toString());
+            assertTrue(invocation.getArgument(1) instanceof AWSConfiguration);
```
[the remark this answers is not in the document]
TrekSoft: Ok

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Ok

TrekSoft:
I debated the issue of separating the chores updating methods to match the design doc from the unit tests but in the end decided to just keep them together since I thought the end result wouldn't be that big. At this point probably just worth rolling with it.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
TrekSoft: Haha so this was pretty neat - I'd made a slight error when refactoring this and my unit test caught it. :D

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -58,18 +64,33 @@
...
+ */
 public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     private static final String AWS_COGNITO_AUTH_PLUGIN_KEY = "awsCognitoAuthPlugin";
     private static final long SECONDS_BEFORE_TIMEOUT = 10;
     private static final String COGNITO_USER_ID_ATTRIBUTE = "sub";
     private static final String MOBILE_CLIENT_TOKEN_KEY = "token";
     private String userId;
+    private AWSMobileClient awsMobileClient;
+
+    public AWSCognitoAuthPlugin() {
```
jamesonwilliams, line 85: Blech, we're gonna need Javadocs on this, now that we have the new Checkstyle rule.

TrekSoft: Done

PR #435, opened 2020-05-06
[aws-datastore] Move model concerns into model sub-package

jamesonwilliams:
The various implementations of `ModelProvider` (and the
`ModelProviderLocator`) are concerns of the data modeling aspect of the
DataStore.

These components are moved into the newly created
`com.amplifyframework.datastore.model` subpackage, which deals with such
concerns.


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/ModelProviderFactory.java
```diff
@@ -1,47 +0,0 @@
...
-package com.amplifyframework.datastore;
-
-import com.amplifyframework.core.model.Model;
-import com.amplifyframework.core.model.ModelProvider;
-
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Set;
-
-final class ModelProviderFactory {
```
jamesonwilliams, line 25: There is a very similar thing in the _source_ code, now, called a `SimpleModelProvider`. All code now uses that, instead.

aws-datastore/src/main/java/com/amplifyframework/datastore/model/ModelProviderLocator.java
```diff
@@ -33,16 +34,21 @@
  * There is no compile-time guarantee that this class actually exists, or that it has that method,
  * or that we can use that method. But in the "happy path," this simplifies the amount of complexity
  * needed to begin using the DataStore. This utility class is used from
- * {@link AWSDataStorePlugin#AWSDataStorePlugin()}.
+ * {@link com.amplifyframework.datastore.AWSDataStorePlugin#AWSDataStorePlugin()}.
```
raphkim, line 37: I think you can just import it to avoid typing out the full name

aws-datastore/src/main/java/com/amplifyframework/datastore/model/SimpleModelProvider.java
```diff
@@ -146,22 +146,24 @@ public String toString() {
     /**
      * Configures and builds instances of SimpleModelProvider.
      */
-    static final class Builder {
+    @SuppressWarnings("WeakerAccess")
+    public static final class Builder {
```
raphkim, line 150: Won't the public methods require javadoc?

aws-datastore/src/main/java/com/amplifyframework/datastore/model/ModelProviderLocator.java
jamesonwilliams: I had it like that originally, however, then `import com.amplifyframework.datastore.AWSDataStorePlugin` appears at the top, and makes it look like a source-code dependency. The relationship is actually the _opposite_ in reality, the plugin uses the locator, not vice-versa. To avoid ambiguity or misconception, I left it this way.

aws-datastore/src/main/java/com/amplifyframework/datastore/model/SimpleModelProvider.java
jamesonwilliams: Should our checkstyle work this way? Yes, I think so.

Does it actually? Unfortunately not. When I built the rule-set last year, I forgot to add [`MissingJavadocMethod`](https://checkstyle.sourceforge.io/config_javadoc.html#MissingJavadocMethod).

Later, we added [`MissingJavadocType`](https://github.com/aws-amplify/amplify-android/blob/master/configuration/checkstyle-rules.xml#L245), but we still don't have the method-level presence check.

We _do_ have checks that say like "_If_ it is there, it had dang-nabbit better be well formed as heck!".

As a result, I think we probably have a number of places in the code that are missing Javadocs. Anyway, I will add them for _these_, since we observed the issue in code review.

As a separate exercise, we should go back and add the `MissingJavadocMethod` for `protected` and `public`s, so that we can catch it before code review, in the future!

jamesonwilliams: Ehh, actually update -- I'll just add them all when I add `MissingJavadocType`. Cause that likely will require adding several more, anyway.

PR #436, opened 2020-05-07
[core] Remove a superfluous utility class

jamesonwilliams:
ArrayUtils contains one method which is a glorified call to
java.util.Arrays.copyOf(...). Instead, just call that method, itself.


jamesonwilliams:
> Hmmm I thought it might be useful elsewhere too haha

I was going through and trying to figure out how to rename/remove things with `Utils` in the name. If it had just been called like `ArrayCopy` or something, I probably would have left it lol :-D

PR #437, opened 2020-05-07
Adds some miscellaneous, missing method documentation.

jamesonwilliams:
In order to add a Checkstyle for MissingJavadocMethod, we need to first
have Javadoc on our public and protected methods.

This commit begins chiseling away at that problem.


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
```diff
@@ -365,34 +364,17 @@ public String toString() {
...
          */
-        UPDATE(MutationType.UPDATE),
+        UPDATE,
 
         /**
          * The removal of a previously-created model.
          */
-        DELETE(MutationType.DELETE);
-
-        private final MutationType apiMutationType;
```
jamesonwilliams, line 380: All of this stuff was not used. And, it required more Javadocs. So, I just got rid of it, instead.

TrekSoft: Yay for simplifcation!

core/src/test/java/com/amplifyframework/core/model/scalar/AWSTimeTest.java
```diff
@@ -62,9 +73,14 @@ public void convertsToAndFromDate() {
         Date date = cal.getTime();
         AWSTime awsTime = new AWSTime(date);
         assertEquals(date, awsTime.toDate());
-        assertThrows(IllegalStateException.class, () -> awsTime.getOffsetTotalSeconds());
+        assertThrows(IllegalStateException.class, awsTime::getOffsetTotalSeconds);
```
TrekSoft, line 76: Hmm, interesting syntax. Will remember.

rxbindings/src/test/java/com/amplifyframework/rx/RxHubBindingTest.java
```diff
@@ -29,18 +29,24 @@
...
     @Before
     public void setup() {
         this.rxHub = new RxHubBinding();
     }
 
+    /**
+     * A subscriber should receive a value published on the Hub.
+     */
     @Test
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
TrekSoft, line 49: Yeah this is better down at the method level

PR #438, opened 2020-05-07
[DataStore] query API changes with pagination support

drochetti:
## Query API changes and Pagination

This PR adds pagination support to the `DataStore.query()`. However, when adding the pagination support I quickly realized that things could get messy real quick. Allow me to explain... =)

Currently, the query API supports a model type and an optional predicate:

```java
// simplified representation of current API
query(@NonNull Class<M>, @Nullable QueryPredicate)
```

### Overload-load

This means we have two overloads, another one that just accepts the model type. By adding a third optional parameter to handle pagination we would increase the number of overloads from 2 to 4. However it doesn't stop there. We intend to add support to sorting and projections/aggregations (i.e. count, sum, max, avg, etc). You see where I'm coming from...

The number of overloads would not only be hard to maintain, but also impact the DX since the IDE won't be able to guess which parameter type you intend to use next very accurately.

### Builder DSL

We used builders before to overcome this issues and I think it's a good fit here too. However, not only the pattern itself, but how we use it as a expressive DSL, something that could match what Swift and Kotlin can achieve with named arguments.

#### Current Swift API

For extra context with a goal to acompli

```swift
// API signature
func query<M: Model>(_ modelType: M.Type,
                     where predicate: QueryPredicateFactory?,
                     paginate paginationInput: QueryPaginationInput?,
                     completion: DataStoreCallback<[M]>)

// Usage

Amplify.DataStore.query(Todo.self,
                        where: Todo.keys.title.beginsWith("Amplify"),
                        paginate: .page(2, limit: 20)) { result in
    // handle result
}
```

#### DSL Flavor 1: static imports ("functional" style)

*This loses. Refer to Flavor 2 for chosen design.*

```java
import static com.amplifyframework.core.model.query.QueryOptions.byId;
import static com.amplifyframework.core.model.query.QueryOptions.where;
import static com.amplifyframework.core.model.query.QueryPaginationInput.firstPage;
import static com.amplifyframework.core.model.query.QueryPaginationInput.page;


Amplify.DataStore.query(
    Todo.class,
    where(
        Todo.TITLE.beginsWith("Amplify")
    )
    .paginate(page(2).withLimit(20)),
    success -> {},
    failure -> {}
);

// query by id
Amplify.DataStore.query(Todo.class, byId(""),
    success -> {},
    failure -> {}
);

// pagination helpers (exists on iOS)
Amplify.DataStore.query(
    Todo.class,
    where(Todo.DONE.eq(false)) // or all()
    .paginate(firstPage()),
    success -> {},
    failure -> {}
);
```

#### DSL Flavor 2: classes with static builders

**This wins!**

```java
import com.amplifyframework.core.model.query.Where;
import com.amplifyframework.core.model.query.Page;

Amplify.DataStore.query(
    Todo.class,
    Where.matches(
        Todo.TITLE.beginsWith("Amplify")
    )
    .paginate(
        Page.startingAt(2).withLimit(20)
    ),
    success -> {},
    failure -> {}
);

// query by id
Amplify.DataStore.query(Todo.class, Where.id(""),
    success -> {},
    failure -> {}
);

// pagination helpers (exists on iOS)
Amplify.DataStore.query(
    Todo.class,
    Where
        .matchesAll()
        .paginate(Page.firstPage()),
    success -> {},
    failure -> {}
);
```


core/src/main/java/com/amplifyframework/core/model/query/QueryPaginationInput.java
```diff
@@ -0,0 +1,96 @@
...
+     * The default page size.
+     */
+    public static final Integer DEFAULT_LIMIT = 100;
+
+    private final Integer page;
+    private final Integer limit;
+
+    private QueryPaginationInput(@NonNull Integer page, @NonNull Integer limit) {
+        this.page = page;
+        this.limit = limit;
```
jamesonwilliams, line 37: If it is not intended for these values to ever be null, I think you could use `int` instead of `Integer`, everywhere in the class.

core/src/main/java/com/amplifyframework/core/model/query/QueryPaginationInput.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * A simple data structure that holds pagination information that can be applied queries.
+ */
+public class QueryPaginationInput {
+
+    /**
+     * The default page size.
+     */
+    public static final Integer DEFAULT_LIMIT = 100;
```
jamesonwilliams, line 30: Can we use the term "page size", instead of limit? "limit" is ambiguous. It would either be a max results, or a number of results per page. I guess here it means limit for each page.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -367,16 +370,16 @@ public synchronized void initialize(
     @Override
     public <T extends Model> void query(
             @NonNull Class<T> itemClass,
-            @Nullable QueryPredicate predicate,
+            @NonNull QueryOptions options,
             @NonNull Consumer<Iterator<T>> onSuccess,
             @NonNull Consumer<DataStoreException> onError) {
         Objects.requireNonNull(itemClass);
-        // Objects.requireNonNull(predicate); It is not required!
```
jamesonwilliams, line 374: Can you uncomment it, and assert the non-null state, now that you have `@NonNull`? Also thank you for making this API improvement. The null-ness of this param has always bothered me.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -237,10 +252,57 @@ public void queryWithMaliciousPredicates() throws DataStoreException {
...
     }
+
+    @Test
+    public void queryWithPaginationWithCustomValues() throws DataStoreException {
+        final int pageSize = 10;
+        createBlogOwnerRecords(pageSize * 2);
+
+        List<BlogOwner> result = adapter.query(
+            BlogOwner.class,
+            all().paginated(page(0).withLimit(pageSize))
```
jamesonwilliams, line 269: I like this style, but I'd change the method names for the pagination. To me, this would be more clear:
```
Where.allMatch().paginated(startAtPage(0).withPageSize(pageSize))
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterModelConverterTest.java
```diff
@@ -70,7 +71,7 @@ public void saveModelWithAllTypesThenQuery() throws DataStoreException {
         final Todo todo = createTestTodoModel();
         this.adapter.save(todo);
 
-        final List<Todo> result = this.adapter.query(Todo.class, Todo.ID.eq(todo.getId()));
+        final List<Todo> result = this.adapter.query(Todo.class, byId(todo.getId()));
```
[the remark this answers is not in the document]
drochetti: I like the concept of the `Where` as a class to avoid the static imports (even though I'm torn about the reasons why the Java community perceives static imports a something generally bad).

I updated the readme with more details and will push an update to this PR introducing this way to build the DSL.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
drochetti: > Nice work @drochetti! Overall looks great. Some ideas mostly about naming, inline.

Thanks for the early review @jamesonwilliams. I'll update the PR with your suggestions.

drochetti:
> Nice work @drochetti! Overall looks great. Some ideas mostly about naming, inline.

Thanks for the early review @jamesonwilliams. I'll update the PR with your suggestions.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
drochetti: I like your suggestion better than the current implementation. My only concern is that JS went GA with the `page`/`limit` [pagination API](https://docs.amplify.aws/lib/datastore/data-access/q/platform/js#pagination). I'm not strongly opposed to change it into something more expressive, but I'd say we should align iOS and Android in that case.

drochetti:
> Thanks for the great PR documentation! IMO if we can get the instrumentation tests back passing, this would be ready to go in.

Oops, what am I doing? :D

> I think I like DSL flavor 2 the best.
> 
> This flavor actually gives the user the _option_ to use static imports, if their style guide allows it...

agree! looks very clean, pretty close to the iOS version and with or without the static imports is provides a nice DX. The current PR is already implemented flavor 2.

> Re the pagination method names: I can see the pros of staying consistent with the other platforms. That might be our best option, to maintain familiarity.

Yeah, I like that too... therefore I kept `withLimit()` so we're aligned with the API released in JS:

```java
Page.startingAt(0).withLimit(20)
```

I'll convert the iOS version to:

```swift
// from
.page(0, limit: 20)

// to
.page(startingAt: 0, withLimit: 20)
```

configuration/checkstyle-rules.xml
```diff
@@ -184,7 +184,10 @@
             how the code is composed, with only a quick visual inspection.
         -->
         <module name="AvoidStarImport" />
-        <module name="AvoidStaticImport" />
+        <module name="AvoidStaticImport">
+            <property name="excludes" value="com.amplifyframework.core.model.query.QueryOptions.*" />
+            <property name="excludes" value="com.amplifyframework.core.model.query.QueryPaginationInput.*" />
+        </module>
```
jamesonwilliams, line 190: This doesn't actually happen under the source tree, though, does it? (I couldn't spot it.) If not -- I'd say, remove these. The `src/test` has `checkstyle-suppressions.xml` rules which allow static imports in test code. (Mockito, for example, is designed for use with static imports.)

drochetti: That's correct, I forgot to remove them, thanks for catching it!

PR #439, opened 2020-05-07
[aws-datastore] PendingMutations are now Comparable according to their creation times

jamesonwilliams:
Changes the ID type of the `PendingMutation`. The `PendingMutation` now
has its own UUID. A mechnism still exists to obtain the ID of the
associated model:

    pendingMutation.getMutationId();
    pendingMutation.getMutatedItem().getId();

The `PendingMuation.PersistentRecord` now has *two* IDs, a "mutation
ID", and a "decoded model ID":

    record.getId();
    record.getDecodedModelId();

Previously, the getId() on the record would return the same ID as
`pendingMutation.getMutatedItem().getId()`.

After this change, the mutation ID is uniquely for the PendingMutation
itself. This ID is also of a different type. It is a time-based UUIDv1.
This type of UUID was chosen as it may be sorted according to its
timestamp. And as a consequence, `PendingMutation`s themselves may now
be sorted by their UUIDv1s. `PendingMutation.PersistentRecord`s use this
as their primary `getId()` value.

The old ID of `PendingMutation.PersistentRecord` is now available by
`getDecodedModelIs()`. This is the ID of the _associated model_ -- the
thing that had undergone some mutation, to begin with. As before, this
UUID is still a v4 UUID, maintained for its technical superiority over
the time-based V1. This value is stored on the record, so that we can
still query for records based on _model ID_, which is what is generally
used around the DataStore system.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
@@ -19,7 +19,6 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.util.ObjectsCompat;
 
-import com.amplifyframework.api.graphql.MutationType;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.annotations.Index;
 import com.amplifyframework.core.model.annotations.ModelConfig;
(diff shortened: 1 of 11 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
```diff
@@ -38,34 +37,55 @@
...
+     * Creates a {@link PendingMutation}, using a provided mutation ID.
+     * @param mutationId A globally-unique, *time-based* ID for this mutation event;
+     *                   the ID is used for temporal ordering of mutations
      * @param mutatedItem The item that undergone a mutation
      * @param classOfMutatedItem The class of the item that has undergone mutation
      * @param mutationType Type of mutation
      * @param <T> The type of the item that has undergone mutation
      * @return A {@link PendingMutation}
      */
     static <T extends Model> PendingMutation<T> instance(
```
rjuliano, line 63: Other than for testing purposes, do we want to leave this `instance` method callable from any class within this package? 

This is very low risk, but my only thought is that we want to ensure these time-based UUIDs are generated on the local device only right?

jamesonwilliams: This version is actually used by the Gson deserializer, when we convert back from a (peristable) `PersistentRecord`. This is how we "recover" the stored timestamp, back into a `PendingMutation`.

PR #440, opened 2020-05-08
[core] Remove some sparsely or unused constructs

jamesonwilliams:
1. `AmplifyOperationRequest`: It wasn't used anywhere.
3. `AsyncEvent`: It wasn't used anywhere.
2. `AsyncOperation`: `AmplifyOperation` did extend this, but for no apparent
   purpose, in practice. `AmplifyOperation` can include a `start()`,
   itself.
4. `EventListener`: It was being used cosmetically by Hub, but the
   implementation was a shell implementation that was not being called
   anywhere. Both were removed.
5. `Options`: It was a marker interface, with no explicit type references.
   Subclasses may be assigned by their direct type, now, instead.
6. `Result`: It was a marker interface, with no explicit type references.
   Subclasses may be assigned by their direct type, now, instead.


core/src/main/java/com/amplifyframework/hub/HubCategory.java
```diff
@@ -72,33 +70,6 @@ public CategoryType getCategoryType() {
...
-        HubChannel channel = HubChannel.forCategoryType(operation.getCategoryType());
-        HubSubscriber transformingListener = event -> {
-            // TODO: check for casting of Object to E and
-            // see if it can be prevented.
-            // eventListener.onEvent(event.getData());
-        };
-
-        return subscribe(channel, event -> true, transformingListener);
-    }
-
```
raphkim, line 101: Ah I see you are removing this also. Nevermind about my comment about operation ID then. 

Would it be a non-trivial task to fix this before GA?

jamesonwilliams: I haven't thought much about what we would need to do here. I think it will take a bit of study to see what iOS is doing, and rebuild from that. We should talk to Tim, and revisit this Hub integration, probably after GA.

core/src/main/java/com/amplifyframework/core/async/AmplifyOperation.java
```diff
@@ -59,18 +49,9 @@
...
         this.categoryType = categoryType;
-        this.operationId = UUID.randomUUID();
         this.request = request;
     }
 
-    /**
-     * Gets the ID of the operation.
-     * @return Operation unique ID
-     */
-    public final UUID getOperationId() {
```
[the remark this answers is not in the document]
jamesonwilliams: I think the operation-integration hub events might be cool, but we're not very close to having them, I don't think. We should come back and look at this integration after GA.

PR #442, opened 2020-05-08
[core] Break apart utils classes

jamesonwilliams:
`CollectionUtils` is now called `Empty`.

`StringUtils` is broken into `Quotes`, and `Casing`.

Tests coverage is added for all utilities mentioned.


core/src/main/java/com/amplifyframework/util/Quotes.java
```diff
@@ -0,0 +1,54 @@
...
+        return "'" + original + "'";
+    }
+
+    /**
+     * Returns original string wrapped with double quotes.
+     * @param original Original string to modify
+     * @return Original string wrapped with double quotes.
+     *         If original string is null or empty, it just returns the original.
+     */
+    @SuppressWarnings("WeakerAccess")
```
jpignata, line 46: Maybe obvious but curious - why is this suppression needed here and not on `wrapInSingle`?

core/src/main/java/com/amplifyframework/util/Quotes.java
```diff
@@ -0,0 +1,54 @@
...
+    @Nullable
+    public static String wrapInDouble(@Nullable String original) {
```
jpignata, line 48: Seems like this is unused.

jamesonwilliams: I guess it is not used in source code. It has test coverage though, so it doesn't required the `@SuppressWarnings("unused")`. Android Studio will count this as being used, since it is used in test, I guess.

core/src/main/java/com/amplifyframework/util/Quotes.java
jamesonwilliams: `wrapInSingle` gets used from `com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java` -- a different package, so it has to be `public`. `wrapInDouble` is only used from the same package (in a unit test, right now), so it _could_ be package-only access. What Android studio doesn't know is that the utility class has been designed for people to call it from different packages, in the future.

PR #446, opened 2020-05-12
[aws-datastore] Test reliability in TimeBasedUuidTest

jamesonwilliams:
The mechanism to obtain a unique, randomly-ordered list is subtly
flawed. It does produce a random list, but that list may not be unique
from the input.

As a consequence, the call to `assertNotEquals` would occasionally cause
the test to fail. This was observed on a development machine, already.

Instead, keep re-ordering the list in random order, until it is
different.

### DIFF
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/TimeBasedUuidTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/TimeBasedUuidTest.java
@@ -26,7 +26,6 @@ import java.util.Random;
 import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


richardmcclellan:
Wow, this was already observed on a development machine?  I believe the chance of that happening is 10! (3628800), right?

jpignata:
I'm assuming the sort with a random integer was not guaranteeing an even distribution. Any reason not to use `Collections.shuffle()` whose docs claims it does?

jamesonwilliams:
> I'm assuming the sort with a random integer was not guaranteeing an even distribution. Any reason not to use `Collections.shuffle()` whose docs claims it does?

I didn't know about it! That's all. Thanks!

PR #447, opened 2020-05-12
[DataStore] fix type conversion for enums and dates

drochetti:
**Notes:**
    
- reuse the type converter logic on predicate values
- fix enum conversion
- fix date types conversion


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelFieldTypeConverter.java
```diff
@@ -106,6 +152,11 @@ public Object convertValueFromSource(
                 case MODEL:
                     return convertModelAssociationToTarget(cursor, field);
                 case ENUM:
+                    @SuppressWarnings("rawtypes")
+                    final Class<? extends Enum> enumClass = field.getType().asSubclass(Enum.class);
+                    @SuppressWarnings({"rawtypes", "unchecked"})
+                    final Enum enumValue = Enum.valueOf(enumClass, valueAsString);
+                    return enumValue;
```
raphkim, line 159: `@SuppressWarnings({"rawtypes", "unchecked"})` can just go directly on the method itself to make it cleaner :)

Actually, rawtypes should be avoidable by doing this:
```
final Class<Enum<?>> enumClass = field.getType().asSubclass(Enum.class);
final Enum<?> enumValue = Enum.valueOf(enumClass, valueAsString);
```

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
```diff
@@ -50,14 +53,19 @@
...
     /**
-     * Represents the java.util.Date data type.
+     * Represents the Date data type.
      */
-    DATE(java.util.Date.class.getSimpleName()),
+    DATE(AWSDate.class.getSimpleName()),
 
     /**
-     * Represents the java.sql.Time data type.
+     * Represents the Date data type.
```
raphkim, line 61: maybe "date-time" instead of just "date"?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelFieldTypeConverter.java
raphkim: Actually, I'm not sure what the best way to represent `Class<T extends Enum<T>>` would be with a `?`, since `Class<? extends Enum<?>>` doesn't sufficiently enforce the two types to be the same... hmm

raphkim: Okay, my suggested solution doesn't work, and it seems that the only way to avoid `rawtypes` warnings is to do the following:
```
final Class<?> clazz = field.getType();
final Enum<?> enumValue = Enum.valueOf(clazz.asSubclass(Enum.class), "FOO");
```

Unchecked type warning is unfortunately very difficult to avoid, so you will need the annotation anyways :/

jamesonwilliams: Meh. Yea, I wasn't able to get rid of one final `unchecked` warning.
```
    @Test
    public void testEnumConversion() {
        assertEquals(Foo.BAR, valueOf(Foo.class, Foo.BAR.name()));
    }

    private static <E extends Enum<E>> Object valueOf(Class<?> clazz, String fieldName) {
        @SuppressWarnings("unchecked")
        Class<E> enumClazz = (Class<E>) clazz.asSubclass(Enum.class);
        return Enum.valueOf(enumClazz, fieldName);
    }

    enum Foo {
        BAR,
    }
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelFieldTypeConverter.java
```diff
@@ -49,7 +49,7 @@
  * to <code>Model</code> properties and from <code>Model</code> properties to values that are
  * valid in a <code>SQLiteStatement</code>.
  */
-final class SQLiteModelFieldTypeConverter implements ModelFieldTypeConverter<Cursor, Model> {
+public final class SQLiteModelFieldTypeConverter implements ModelFieldTypeConverter<Cursor, Model> {
```
richardmcclellan, line 52: Does this need to be public?

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
```diff
@@ -50,14 +53,19 @@
     STRING(String.class.getSimpleName()),
 
     /**
-     * Represents the java.util.Date data type.
+     * Represents the Date data type.
```
richardmcclellan, line 56: As written, a customer might get confused thinking `Date` is `java.util.Date`.    This should be `AWSDate` (or whatever the model class is renamed, if we are renaming it).  

If we do rename `AWSDate`, I don't like just dropping the `AWS` prefix, since `Date` would be confused with `java.util.Date`.  Maybe `AmplifyDate`?  Not sure about that either though.

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
```diff
@@ -17,6 +17,9 @@
 
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.core.model.AWSDate;
+import com.amplifyframework.core.model.AWSDateTime;
+import com.amplifyframework.core.model.AWSTime;
```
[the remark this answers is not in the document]
drochetti: yeah `AWSTimestamp` is confusing to me, because the implementation handles it like a `Long` (like it should imo), so I don't fully understand why AWSTimestamp exists...

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/TypeConverter.java
```diff
@@ -81,6 +83,26 @@ static JavaFieldType getJavaFieldType(@NonNull ModelField field) {
         }
     }
 
+    /**
+     * Gets the {@link JavaFieldType} from the value class.
+     * @param value The value to guess the type from.
+     * @return the {@link JavaFieldType} from the value class.
+     */
+    public static JavaFieldType getJavaFieldTypeFromValue(@NonNull Object value) {
```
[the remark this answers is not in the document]
drochetti: I'm not sure how this would be a benefit, since the value we have at the time we call this is of type `Object`. We're reading it using reflection from a customer provided type, so there's no way to know the actual type beforehand in a way the `T` would be relevant.

Thoughts?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelFieldTypeConverter.java
drochetti: unless I start to move stuff around, yes... There are two packages right now and some `public` members because those two packages share code. Cleaning this up would require a deeper refactor.

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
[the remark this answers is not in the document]
richardmcclellan: which "implementation" are you referring to?  

The codegen currently generates a model with a `Long`, but we've asked them to change this to `AWSTimestamp`.

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
drochetti: We have a big debate in iOS too since this is consumer-facing API. I'm honestly against this verbosity of adding AWS or Amplify as prefixes (or any other prefix tbh) not only to the verbosity itself, but why is it relevant that is an **Amplify** Date? From a consumer perspective that makes me wonder why that's a special Date/DateTime.

I really wanted customer to be able to ready is and understand what the intent is: it's a `Date`, a `DateTime` and `Time`.

But I'm starting to accept that I will lose this battle.

testmodels/src/main/java/com/amplifyframework/testmodels/todo/Todo.java
```diff
@@ -73,15 +74,15 @@ public TodoStatus getStatus() {
         return status;
     }
 
-    public Date getCreatedAt() {
+    public AWSDateTime getCreatedAt() {
         return createdAt;
     }
 
     public Long getLastUpdated() {
```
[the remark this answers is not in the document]
drochetti: I still don't agree with `AWSTimestamp` existence :D

I'm willing to change it once I'm convinced we need `AWSTimestamp`. Why not a `Long` instead?

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
richardmcclellan: Just throwing this out there, one other option for naming would be:

AWSDate > com.amplifyframework.core.model.OffsetDate
AWSDateTime > com.amplifyframework.core.model.OffsetDateTime
AWSTime > com.amplifyframework.core.model.OffsetTime
AWSTimestamp > java.util.Date

These are basically the same names that are used by java.time, but we've sort of reimplemented them ourselves, since java.time isn't available.

@jamesonwilliams any thoughts?

core/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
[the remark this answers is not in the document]
drochetti: The SQLite adapter handles as a `Long`.

Maybe there's a reason for using `AWSTimestamp` instead, if that's the preferred way I can change this. But it's worth noting that both iOS and JS handle timestamp as a number. The alternative for a more "controlled type" in those platforms is `AWSDateTime`.

PR #450, opened 2020-05-12
[aws-datastore] Clarify AWSDataStorePlugin dependency on API Category

jamesonwilliams:
The `AWSDataStorePlugin` has a dependency on a `GraphQlBehavior` in one
place, as the portal through which to perform synchronization. In
another place, `Amplify.API` is referenced directly, to determine if there
are plugins available in the API category.

For the purposes of test, consolidate both uses to an `ApiCategory`
instance, which satisfies both purposes. This dependency is wired
through to the constructor, for dependency injection.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreCategoryConfigurator.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreCategoryConfigurator.java
@@ -20,7 +20,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.RawRes;
 
 import com.amplifyframework.AmplifyException;
-import com.amplifyframework.api.GraphQlBehavior;
+import com.amplifyframework.api.ApiCategory;
 import com.amplifyframework.core.AmplifyConfiguration;
 import com.amplifyframework.core.InitializationStatus;
 import com.amplifyframework.core.category.CategoryConfiguration;
@@ -36,7 +36,7 @@ final class DataStoreCategoryConfigurator {
     private Context context;
     @RawRes private Integer resourceId;
     private ModelProvider modelProvider;
-    private GraphQlBehavior api;
+    private ApiCategory api;
     private boolean clearRequested;
 
     private DataStoreCategoryConfigurator() {}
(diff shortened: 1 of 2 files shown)


jamesonwilliams:
> Naive question I'd ask in person if people met in person anymore: is this a step in the wrong direction? Shouldn't we reduce these kinds of cross-category dependencies and move to something where core itself injects the details needed by other categories?

`ApiCategory` implements `ApiCategoryBehavior`, which extends `GraphQlBehavior`. So, this change does _decrease_ the level of generality by two inheritance levels.

But, keep in mind that `ApiCategory` _is `core`_. Our extension point is to do `apiCategory.addPlugin(new HeresWhereAbstractionEnds())`.

---------

Alternately, it would be possible to go full-on dependency inversion here. That might look as follows. `AWSDataStorePlugin` would declare its wants and needs in an inner-interface, and someone would have to pass an instance of it, upon constructing the plugin:

```
public final class AWSDataStorePlugin {
    // I don't even want to know how you do it! I just need something that satisfies
    // my dependency contract. How is your business, plugin client.
    private final AWSDataStorePlugin.ApiDependency;

    public AWSDataStorePlugin(ApiDependency fullyInvertedDependency) {
        this.fullyInvertedDependency = fullyInvertedDependency;
    }
    ...
    public interface ApiDependency {
        boolean hasApiPlugins();
        void mutate(T item, Consumer<T> onSuccess, Consumer<DataStoreException> onFailure);
    }
    ...
```

In practice, this dependency inversion interface starts to look quite a bit like `ApiCategory`. Since `ApiCategory` is already generic and in core, it lives

jpignata:
Thanks! I follow.

jamesonwilliams:
> I like it. As a matter of fact, as I am currently wrangling with testing DataStore and trying to mock its interactions with the API category, I questioned whether it would make sense to make this change. I think this will go a long way in helping with testing those interactions.

Yea! It's _pretty close_ to just using a mock.
```
// category is closed to extension, so add a mock plugin as a delegate ...
ApiCategory category = new ApiCategory();
ApiPlugin<?> delegate = mock(ApiPlugin.class);
category.addPlugin(delegate);
when(delegate.whatever()).thenReturn(cookedValue);
```

PR #452, opened 2020-05-12
Amplify Tools Plugin Source

nikhname:
*Issue #, if available:*

*Description of changes:*
Adding the amplify-tools plugin to amplify-android repo.
Changes may need to be made to publishing and structure but using this PR to provide the source code.
Plugin is currently not included in the settings.gradle of amplify-android project

### DIFF
--- a/amplify-tools-gradle-plugin/build.gradle
+++ b/amplify-tools-gradle-plugin/build.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
--- a/amplify-tools-gradle-plugin/build.gradle
+++ b/amplify-tools-gradle-plugin/build.gradle
@@ -1,24 +1,27 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
 apply plugin: 'groovy'
 apply plugin: 'idea'
-apply plugin: 'maven-publish'
 
 apply plugin: 'com.android.library'
 
-sourceSets.main.groovy.srcDirs = ["src/main/java", "src/main/groovy"]
+sourceSets.main.groovy.srcDirs = ["src/main/groovy"]
 
 dependencies {
-    implementation 'androidx.appcompat:appcompat:1.1.0'
-}
-
-publishing {
-    publications {
-        pluginPublication(MavenPublication) {
-            from components.java
-            groupId 'com.amplifyframework'
-            artifactId 'amplify-tools-gradle-plugin'
-            version VERSION_NAME + "-SNAPSHOT"
-        }
-    }
+    implementation dependency.androidx.appcompat
 }
 
 sourceCompatibility = "1.7"
--- a/amplify-tools-gradle-plugin/build.gradle
+++ b/amplify-tools-gradle-plugin/build.gradle
@@ -23,6 +23,3 @@ sourceSets.main.groovy.srcDirs = ["src/main/groovy"]
 dependencies {
     implementation dependency.androidx.appcompat
 }
-
-sourceCompatibility = "1.7"
-targetCompatibility = "1.7"
--- a/amplify-tools-gradle-plugin/build.gradle
+++ b/amplify-tools-gradle-plugin/build.gradle
@@ -1,0 +1,25 @@
+apply plugin: 'groovy'
+apply plugin: 'idea'
+apply plugin: 'maven-publish'
+
+apply plugin: 'com.android.library'
+
+sourceSets.main.groovy.srcDirs = ["src/main/java", "src/main/groovy"]
+
+dependencies {
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+}
+
+publishing {
+    publications {
+        pluginPublication(MavenPublication) {
+            from components.java
+            groupId 'com.amplifyframework'
+            artifactId 'amplify-tools-gradle-plugin'
+            version VERSION_NAME + "-SNAPSHOT"
+        }
+    }
+}
+
+sourceCompatibility = "1.7"
+targetCompatibility = "1.7"
--- a/amplify-tools-gradle-plugin/gradle.properties
+++ b/amplify-tools-gradle-plugin/gradle.properties
@@ -1,0 +1,4 @@
+POM_ARTIFACT_ID=amplify-tools-gradle-plugin
+POM_NAME=Amplify Tools for Android - Gradle Plugin
+POM_DESCRIPTION=Gradle plugin for the amplify flow with android
+POM_PACKAGING=jar
--- a/amplify-tools-gradle-plugin/gradle.properties
+++ b/amplify-tools-gradle-plugin/gradle.properties
@@ -1,4 +1,4 @@
 POM_ARTIFACT_ID=amplify-tools-gradle-plugin
 POM_NAME=Amplify Tools for Android - Gradle Plugin
 POM_DESCRIPTION=A Gradle plugin that installs the Amplify CLI and enables the user to execute Amplify capabilities like modelgen and push within the IDE.
-POM_PACKAGING=jar
+POM_PACKAGING=aar
--- a/amplify-tools-gradle-plugin/gradle.properties
+++ b/amplify-tools-gradle-plugin/gradle.properties
@@ -1,4 +1,4 @@
 POM_ARTIFACT_ID=amplify-tools-gradle-plugin
 POM_NAME=Amplify Tools for Android - Gradle Plugin
-POM_DESCRIPTION=Gradle plugin for the amplify flow with android
+POM_DESCRIPTION=A Gradle plugin that installs the Amplify CLI and enables the user to execute Amplify capabilities like modelgen and push within the IDE.
 POM_PACKAGING=jar
--- a/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
+++ b/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
@@ -1,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.amplifyframework.tools" />
--- a/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
+++ b/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
    Licensed under the Apache License, Version 2.0 (the "License").
    You may not use this file except in compliance with the License.
--- a/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
+++ b/amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
@@ -1,2 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+   Licensed under the Apache License, Version 2.0 (the "License").
+   You may not use this file except in compliance with the License.
+   A copy of the License is located at
+
+    http://aws.amazon.com/apache2.0
+
+   or in the "license" file accompanying this file. This file is distributed
+   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+   express or implied. See the License for the specific language governing
+   permissions and limitations under the License.
+-->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.amplifyframework.tools" />
--- a/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
+++ b/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
@@ -1,0 +1,219 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+class AmplifyTools implements Plugin<Project> {
+    void apply(Project project) {
+        def doesNodeExist = true
+        def doesGradleConfigExist
+
+        def profile = 'default'
+        def accessKeyId = null
+        def secretAccessKey = null
+        def region = null
+        def envName = 'amplify'
+        def syncEnabled = 'true'
+
+        project.task('verifyNode') {
+            try {
+                project.exec {
+                    commandLine 'which', 'node'
+                    standardOutput = new ByteArrayOutputStream()
+                }
+            } catch (e) {
+                doesNodeExist = false
+                println("Node is not installed. Visit https://nodejs.org/en/download/ to install it")
+            }
+        }
+
+        project.task('createAmplifyApp') {
+            doesGradleConfigExist = project.file('amplify-gradle-config.json').isFile()
+            if (doesNodeExist && !doesGradleConfigExist) {
+                project.exec {
+                    commandLine 'npx', 'amplify-app', '--platform', 'android'
+                }
+            }
+        }
+
+        project.task('getConfig') {
+            def inputConfigFile = project.file('amplify-gradle-config.json')
+            if (inputConfigFile.isFile()) {
+                def configText = inputConfigFile.text
+                def jsonSlurper = new groovy.json.JsonSlurper()
+                def configJson = jsonSlurper.parseText(configText)
+                profile = configJson.profile
+                accessKeyId = configJson.accessKeyId
+                secretAccessKey = configJson.secretAccessKeyId
+                region = configJson.region
+                envName = configJson.envName
+                syncEnabled = configJson.syncEnabled
+            }
+        }
+
+        project.task('datastoreSync') {
+            def transformConfFile = project.file('amplify/backend/api/amplifyDatasource/transform.conf.json')
+            if (project.file('amplify/backend/api').exists()) {
+                new File('amplify/backend/api').eachFileRecurse(groovy.io.FileType.FILES) {
+                    if (it.name.endsWith('transform.conf.json')) {
+                        transformConfFile = project.file(it)
+                    }
+                }
+            }
+            if (transformConfFile.isFile()) {
+                def tranformConfText = transformConfFile.text
+                def jsonSlurper = new groovy.json.JsonSlurper()
+                def transformConfJson = jsonSlurper.parseText(tranformConfText)
+
+                def resolverConfigMap = ['ResolverConfig':
+                                                 ['project':
+                                                          ['ConflictHandler'  : 'AUTOMERGE',
+                                                           'ConflictDetection': 'VERSION']
+                                                 ]
+                ]
+                if (!syncEnabled) {
+                    transformConfJson.remove('ResolverConfig')
+                } else if (!transformConfJson.ResolverConfig) {
+                    transformConfJson << resolverConfigMap
+                }
+                def transformConfJsonStr = groovy.json.JsonOutput.toJson(transformConfJson)
+                def transformConfJsonStrPretty = groovy.json.JsonOutput.prettyPrint(transformConfJsonStr)
+                transformConfFile.write(transformConfJsonStrPretty)
+            }
+        }
+
+        project.task('modelgen') {
+            doLast {
+                project.exec {
+                    commandLine 'amplify', 'codegen', 'model'
+                }
+            }
+        }
+
+        project.task('amplifyPush') {
+            def AWSCLOUDFORMATIONCONFIG
+            if (!accessKeyId || !secretAccessKey || !region) {
+                AWSCLOUDFORMATIONCONFIG = """{\
+\"configLevel\":\"project\",\
+\"useProfile\":true,\
+\"profileName\":\"$profile\"\
+}"""
+            } else {
+                AWSCLOUDFORMATIONCONFIG = """{\
+\"configLevel\":\"project\",\
+\"useProfile\":true,\
+\"profileName\":\"$profile\",\
+\"accessKeyId\":\"$accessKeyId\",\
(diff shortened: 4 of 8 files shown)


amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+                    standardOutput = new ByteArrayOutputStream()
+                }
+            } catch (e) {
+                doesNodeExist = false
+                println("Node is not installed. Visit https://nodejs.org/en/download/ to install it")
+            }
+        }
+
+        project.task('createAmplifyApp') {
+            doesGradleConfigExist = project.file('amplify-gradle-config.json').isFile()
```
palpatim, line 29: How is this file created and managed? Also, by storing it in the project directory, we increase the risk of a customer accidentally committing their AWS account credentials to source control.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+import org.gradle.api.Project
+
+class AmplifyTools implements Plugin<Project> {
+    void apply(Project project) {
+        def doesNodeExist = true
+        def doesGradleConfigExist
+
+        def profile = 'default'
+        def accessKeyId = null
+        def secretAccessKey = null
```
palpatim, line 11: We need to support session-based temporary credentials as well, not just static and profile

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+
+            doLast {
+                def doesLocalEnvExist = project.file('./amplify/.config/local-env-info.json').exists()
+                if (doesLocalEnvExist) {
+                    project.exec {
+                        commandLine 'amplify', 'push', '--yes'
+                    }
+                } else {
+                    project.exec {
+                        commandLine 'amplify', 'init', '--amplify', AMPLIFY, '--providers', PROVIDERS, '--yes'
```
palpatim, line 133: The `PROVIDERS` variable has been configured with credentials above, and now we're passing those on the command line, where they will be visible to anyone who can view the command line. Can this command read `providers` from a file instead?

amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,25 @@
+apply plugin: 'groovy'
```
jamesonwilliams, line 1: Please add the copyright header that other `.gradle` files in this repo are using

amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+apply plugin: 'idea'
+apply plugin: 'maven-publish'
+
+apply plugin: 'com.android.library'
+
+sourceSets.main.groovy.srcDirs = ["src/main/java", "src/main/groovy"]
```
jamesonwilliams, line 7: I think `sourceSets.main.groovy.srcDirs` should only be `["src/main/groovy"]`. If you have groovy sources under `src/main/java`, can you move them to the groovy dir?

amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+
+dependencies {
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+}
+
+publishing {
```
jamesonwilliams, line 13: This is integrated into the root `build.gradle`, so doesn't need to be present in this module-level gradle.

amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,25 @@
+apply plugin: 'groovy'
+apply plugin: 'idea'
+apply plugin: 'maven-publish'
+
+apply plugin: 'com.android.library'
+
+sourceSets.main.groovy.srcDirs = ["src/main/java", "src/main/groovy"]
+
+dependencies {
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
jamesonwilliams, line 10: We define all of our dependencies in the project `build.gradle`, and then reference them as `implementation dependency.androidx.appcompat`. This ensures that we have a mutually consistent set of dependencies across the project modules.

amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.amplifyframework.tools" />
```
jamesonwilliams, line 2: Please add the copy-right header that other xml files in the project are using.

----------------

If this module uses an `AndroidManifest.xml`, and does `import com.android.library`, it probably means that you can package your plugin as an `aar`. This would be good since our current `configuration/publishing.gradle` has been tested for publishing `aar`. Also, since all of the other modules are. This would mean you'd need to update the `POM_PACKACING` to `aar` as well of course.

------------


`tools` is pretty generic. We also have a "Plugin" concept in our client library. So, we need to be careful about the name we choose here. How about `com.amplifyframework.gradle`, or `com.amplifyframework.gradleplugin`

amplify-tools-gradle-plugin/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=amplify-tools-gradle-plugin
+POM_NAME=Amplify Tools for Android - Gradle Plugin
+POM_DESCRIPTION=Gradle plugin for the amplify flow with android
```
jamesonwilliams, line 3: This doesn't really say anything new. Can you change this to like:
```
A Gradle plugin that enables the user to <achieve outcomes, here>.
```
Also, please capitalize `Amplify` and `Android` which are proper nouns

amplify-tools-gradle-plugin/src/main/resources/META-INF/gradle-plugins/com.amplifyframework.amplifytools.properties
```diff
@@ -0,0 +1 @@
+implementation-class=AmplifyTools
```
jamesonwilliams, line 1: This file name, `com.amplifyframework.amplifytools.properties `-- the import path needs to be consistent with the rest of the PR (for example, what you use in the manifest.)

amplify-tools-gradle-plugin/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">amplifytools</string>
```
jamesonwilliams, line 2: Do you use this anywhere? If not, just delete `src/main/res` entirely from the project.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+
+                //Save File
+                def writer = new FileWriter('./.idea/workspace.xml')
+
+                //Pretty print XML
+                groovy.xml.XmlUtil.serialize(xml, writer)
+            }
+        }
+
+        project.createAmplifyApp.dependsOn('verifyNode')
```
jamesonwilliams, line 213: These are easier to keep track of if you put them after the task definitions. Like:
```
        project.task('createAmplifyApp') {
            // blah blah
        }
        project.createAmplifyApp.dependsOn('verifyNode')
```

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+            } else {
+                AWSCLOUDFORMATIONCONFIG = """{\
+\"configLevel\":\"project\",\
+\"useProfile\":true,\
+\"profileName\":\"$profile\",\
+\"accessKeyId\":\"$accessKeyId\",\
+\"secretAccessKey\":\"$secretAccessKey\",\
+\"region\":\"$region\"\
+}"""
+            }
```
jamesonwilliams, line 108: This is not a very maintainable way to build JSON. I don't know how to do this in Groovy, but in Java, you could something like:
```
new JSONObject()
    .put("key", new JSONObject()
        .put("innerKey", "value")
        .put("innerKey2", new JSONArray(Arrays.asList("val1", "val2")))
   );
```
Or, actually, here's a page dedicated to the topic: https://groovy-lang.org/json.html

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+                        transformConfFile = project.file(it)
+                    }
+                }
+            }
+            if (transformConfFile.isFile()) {
+                def tranformConfText = transformConfFile.text
+                def jsonSlurper = new groovy.json.JsonSlurper()
+                def transformConfJson = jsonSlurper.parseText(tranformConfText)
+
+                def resolverConfigMap = ['ResolverConfig':
```
jamesonwilliams, line 66: Formatting here is weird, how about just:
```
def resolverConfigMap = [
    'ResolverConfig': [
        'project': [
            'ConflictHandler'  : 'AUTOMERGE',
            'ConflictDetection': 'VERSION'
        ]
    ]
]
```

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+                    commandLine 'npx', 'amplify-app', '--platform', 'android'
+                }
+            }
+        }
+
+        project.task('getConfig') {
+            def inputConfigFile = project.file('amplify-gradle-config.json')
+            if (inputConfigFile.isFile()) {
+                def configText = inputConfigFile.text
+                def jsonSlurper = new groovy.json.JsonSlurper()
```
jamesonwilliams, line 41: Can you import this, so you can just use `JsonSlurper()` without the full qualification?

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+        def envName = 'amplify'
+        def syncEnabled = 'true'
+
+        project.task('verifyNode') {
+            try {
+                project.exec {
+                    commandLine 'which', 'node'
+                    standardOutput = new ByteArrayOutputStream()
+                }
+            } catch (e) {
```
jamesonwilliams, line 22: Single letter variables are very difficult to find with search tools. Can you call this `error` at least, if not `commandLineFailure`.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: This file is created through amplify-app - https://github.com/aws-amplify/amplify-cli/blob/b38deb59223417c67d41838c7af21fe4184babd0/packages/amplify-app/src/index.js#L351

The file is added to the gitignore so it would have to be manually removed from there to be commited

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: This command line execution is not piped to STDOUT or any other output so it's not visible on the commandline. But if its better practice to use a file we can make this change

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
palpatim: Adding to .gitignore is a brittle solution; it's possible to override that with a `git add --force <path that accidentally includes this file even though you didn't actually intend to add it>` and then you're back to dealing with committed credentials. It sounds like this is something we need to discuss with the CLI team.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
palpatim: Simply executing the command in a shell makes it visible to utilities like `ps`, as well as potentially other process accounting, logging, and auditing systems. Let's use a file. We've been using a protected home directory: `$HOME/.aws-amplify` with `700` permissions as the root repository for these well-known paths. I'd suggest something like `$HOME/.aws-amplify/amplify-android` as a reasonable containing directory for this file, but I'll let @jamesonwilliams have the final say.

amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
jpignata: iOS has a similar plugin for XCode, so we should maintain some naming consistency for our own sanity. @undefobj

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
drochetti: +1 `JsonSlurper` would be the groovy way of doing it.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
drochetti: Related to this, when testing the tools in iOS yesterday I realized that we rely on the top-level `.gitignore` to omit some of the files inside the `amplify/` folder. It seems to me like that `amplify/` folder should have self-contained `.gitignore` rules that don't depend on the top-level file having the right config so we don't delegate that responsibility to customers.

amplify-tools-gradle-plugin/src/main/AndroidManifest.xml
drochetti: > iOS has a similar plugin for XCode, so we should maintain some naming consistency for our own sanity. @undefobj

yes please! :)

jamesonwilliams: Okay yes, I take back the name thing. :-D

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: I think this warrants further discussion but the change can't be made here. Could we open up an issue regarding ```.gitignore``` on the CLI repo so that we can continue the conversation there?

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: @undefobj do you have thoughts on using session-based temporary credentials with this plugin

undefobj: The way to do this is with an IAM role which uses a profile. We basically need to validate that what the CLI supports is supported here as well: https://docs.amplify.aws/cli/usage/iam-roles-mfa

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,219 @@
...
+        def accessKeyId = null
+        def secretAccessKey = null
+        def region = null
+        def envName = 'amplify'
+        def syncEnabled = 'true'
+
+        project.task('verifyNode') {
+            try {
+                project.exec {
+                    commandLine 'which', 'node'
```
jpignata, line 19: Seems like we need to make a simple change here to support Window. We can probably detect Windows using https://docs.gradle.org/current/javadoc/org/gradle/nativeplatform/platform/OperatingSystem.html and run a different command.

nikhname: It may also be that ```which``` (not a valid command in cmd) is causing the problem. Testing this out using ```node -v``` in windows

jpignata:
Perhaps a naive question - I understand that gradle files are written using Kotlin or Groovy. Does that mean we must choose one and use it throughout? Said another way, can't we use Kotlin or Java for the meat of the IDE integrations instead of groovy, which seems to be a language with which we aren't quite as familiar.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
jpignata: Nice. Did this work? Do we have a Windows compatible IDE integration now? 😀

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -0,0 +1,226 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import groovy.json.JsonSlurper
+import groovy.json.JsonOutput
+
+class AmplifyTools implements Plugin<Project> {
+    void apply(Project project) {
+        def doesNodeExist = true
```
jpignata, line 8: Is this boolean necessary? I'd think you should have any task that relies on node (and the CLI for the matter), just run `verifyNode` first and go boom if either is unavailable.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
jpignata: Another problem here is that we're relying on a profile called `default` which may or may not exist.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
jpignata: Yes. We should read from a file. Anything on the command line is available system wide, as @palpatim mentioned. I'm a little fuzzy and want to know more about why we need to transit credentials anywhere as they are available to the CLI already.

nikhname: These credentials are not available to the CLI because they're being pulled from the user editing amplify-gradle-config.json file, which is the CLI-less flow we worked on for this plugin. From my understanding, piping the PROVIDERS to the command line is no less secure than having the customer manually enter their secret key and access key to the command line when prompted by the CLI, which is an accepted flow that we already use.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: Thats only the placeholder profile name. If the customer is not using a profile named default (or secret key and access key) they should specify a different profile name in amplify-gradle-config.json and the plugin will pick up that profile

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
nikhname: Tested it out after a lot of struggles with windows, and got the build working! We need to use the OS detection to switch between npx.cmd and npx as well as amplify.cmd and amplify commands, and use node -v instead of which

nikhname:
> Perhaps a naive question - I understand that gradle files are written using Kotlin or Groovy. Does that mean we must choose one and use it throughout? Said another way, can't we use Kotlin or Java for the meat of the IDE integrations instead of groovy, which seems to be a language with which we aren't quite as familiar.

From my understanding Groovy is better at parsing and injecting XML which is valuable for the Android Studio IDE work that the plugin currently does and may do in the future. If Groovy is a blocker for development it still be possible to switch over to Java or Kotlin but I'd recommend having further discussion and a separate task for that after this source is merged.

jpignata:
> I'd recommend having further discussion and a separate task for that after this source is merged.

Yep, that's fine, we can handle it separately.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
jpignata: NICE!

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
jpignata: Okay, but we'll need to document that. Thanks.

amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/amplifytools.groovy
```diff
@@ -0,0 +1,226 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import groovy.json.JsonSlurper
+import groovy.json.JsonOutput
+
+class amplifytools implements Plugin<Project> {
+    void apply(Project project) {
+        def doesNodeExist = true
```
sammartinez, line 8: Because I keep getting some emails related to this PR, what is the difference between this line and line 16

nikhname: Hey Sam, line 16 has to do with toggling datastoreSync. If you mean the verifyNode task, that task sets ```doesNodeExist``` to false if node isn't found so that other tasks requiring node don't get executed (using the variable since a task cannot execute another task in gradle). An alternative which John mentioned is just erroring out of the build in the verifyNode task since the other tasks depend on it. Currently trying that out

jpignata:
Superseded by #505.

PR #453, opened 2020-05-13
Add new pagedQuery methods to GraphQLBehavior

richardmcclellan:
This PR replaces https://github.com/aws-amplify/amplify-android/pull/434

_Goal:_

- Expose pagination related properties so that datastore can query API and obtain all results.

_Description of changes:_

- Adds new `pagedQuery` method to GraphQLBehavior for obtaining the next set of results.
- Adds `Page` model, which encapsulates a GraphQLResponse, and knows if there is a nextPage (`hasNextPage()`), and how to get the next page (`queryNextPage(success, failure)`)

_To do before merging (but after team consensus on this approach):_

- Review this design and get consensus with iOS / JS
- Add integration test
- We don't actually need to add a new `pagedQuery` method to GraphQLBehavior, but it was easier for me to do it this way just to prove that this works.    Now that it does, we should just update the existing `query` methods which return a GraphQLResponse<Iterable<T>> to return a `Page<T>` instead.
 - Replace `Iterable` with `List`
 - Add Javadoc, address check-style issues.

_Customer Experience_

```
    private void listTodos() {
        try {
            int limit = 3;
            GraphQLRequest<Todo> request = AppSyncGraphQLRequestFactory.buildQuery(Todo.class, null, limit, null);
            Amplify.API.pagedQuery(request, this::handlePage, this::handleError);
        } catch(ApiException apiFailure) {
            handleError(apiFailure);
        }
    }

    private void handlePage(Page<Todo> page) {
        GraphQLResponse<Iterable<Todo>> response = page.getResponse();
        if (response.hasData()) {
            for (Todo todo : response.getData()) {
                Log.d(TAG, "todo: " + todo.getName());
            }
        }
        if (response.hasErrors()) {
            Log.d(TAG, "error: " + error);
        }

        if(page.hasNextPage()) {
            page.queryNextPage(this::handlePage, this::handleError);
        }
    }

    private void handleError(ApiException exception) {
        Log.e(TAG, exception.getMessage(), exception);
    }
```

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -17,6 +17,7 @@ package com.amplifyframework.api.aws;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
@@ -29,6 +30,7 @@ import com.amplifyframework.api.graphql.GraphQLOperation;
 import com.amplifyframework.api.graphql.GraphQLRequest;
 import com.amplifyframework.api.graphql.GraphQLResponse;
 import com.amplifyframework.api.graphql.MutationType;
+import com.amplifyframework.api.graphql.Page;
 import com.amplifyframework.api.graphql.SubscriptionType;
 import com.amplifyframework.api.rest.HttpMethod;
 import com.amplifyframework.api.rest.RestOperation;
@@ -212,6 +214,22 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
 
     @Nullable
     @Override
+    public <T> GraphQLOperation<T> pagedQuery(
+            @NonNull GraphQLRequest<T> graphQLRequest,
+            @NonNull Consumer<Page<T>> onResponse,
+            @NonNull Consumer<ApiException> onFailure) {
+        final String apiName;
+        try {
+            apiName = getSelectedApiName(EndpointType.GRAPHQL);
+        } catch (ApiException exception) {
+            onFailure.accept(exception);
+            return null;
+        }
+        return pagedQuery(apiName, graphQLRequest, onResponse, onFailure);
+    }
+
+    @Nullable
+    @Override
     public <T extends Model> GraphQLOperation<T> query(
             @NonNull String apiName,
             @NonNull Class<T> modelClass,
@@ -268,6 +286,24 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
         try {
             final GraphQLOperation<T> operation =
                 buildMultiResponseOperation(apiName, graphQLRequest, onResponse, onFailure);
+            operation.start();
+            return operation;
+        } catch (ApiException exception) {
+            onFailure.accept(exception);
+            return null;
+        }
+    }
+
+    @Nullable
+    @Override
+    public <T> GraphQLOperation<T> pagedQuery(
+            @NonNull String apiName,
+            @NonNull GraphQLRequest<T> graphQLRequest,
+            @NonNull Consumer<Page<T>> onResponse,
+            @NonNull Consumer<ApiException> onFailure) {
+        try {
+            final GraphQLOperation<T> operation =
+                    buildPageOperation(apiName, graphQLRequest, onResponse, onFailure);
             operation.start();
             return operation;
         } catch (ApiException exception) {
@@ -787,9 +823,9 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
         final ClientDetails clientDetails = apiDetails.get(apiName);
         if (clientDetails == null) {
             throw new ApiException(
-                "No client information for API named " + apiName,
-                "Check your amplify configuration to make sure there " +
-                "is a correctly configured section for " + apiName
+                    "No client information for API named " + apiName,
+                    "Check your amplify configuration to make sure there " +
+                            "is a correctly configured section for " + apiName
             );
         }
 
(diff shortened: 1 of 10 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -785,6 +821,32 @@ private String selectApiName(Set<String> apiClients) throws ApiException {
...
+                .client(clientDetails.okHttpClient())
+                .request(graphQLRequest)
+                .responseFactory(gqlResponseFactory)
+                .onResponse(onResponse)
+                .onFailure(onFailure)
+                .build();
+    }
+
+
+    private <T> PagedResultOperation<T> buildPageOperation(
```
jamesonwilliams, line 843: If it is only used once, does it make sense to inline it?

aws-api/src/main/java/com/amplifyframework/api/aws/PagedResultOperation.java
```diff
@@ -0,0 +1,209 @@
...
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+/**
+ * An operation to enqueue a GraphQL request to OkHttp client,
+ * with the goal of obtaining a list of responses. For example,
+ * this is used for a LIST query vs. a GET query or most mutations.
+ * @param <T> Casted type of GraphQL result data
+ */
+public final class PagedResultOperation<T> extends GraphQLOperation<T> {
```
jamesonwilliams, line 50: As prior tech debt, we forked the single and multiple response operations. So, we had two things, that are mostly the same. Now, we'll have three. Seems like a good time to pause and try to figure out a way to have less code duplication between the three.

aws-api/src/main/java/com/amplifyframework/api/aws/PagedResultOperation.java
```diff
@@ -0,0 +1,209 @@
...
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-api");
+    private static final String CONTENT_TYPE = "application/json";
```
jamesonwilliams, line 52: I think OkHttp has a constant for this, under `MediaType`, somewhere?

core/src/main/java/com/amplifyframework/api/GraphQlBehavior.java
```diff
@@ -139,6 +140,29 @@
...
 
+    /**
+     * Perform a GraphQL query against a configured GraphQL endpoint.
+     * This operation is asynchronous and may be canceled by calling
+     * cancel on the returned operation. The response will be provided
+     * to the `onResponse` callback.  If there is data present
+     * in the response, it will be cast as the requested class type.
+     * Requires that only one API is configured in your
+     * `amplifyconfiguration.json`. Otherwise, emits an ApiException to
+     * the provided `onFailure` callback.
```
jamesonwilliams, line 151: These new docs should start off with some strongly indicative language "PAges! How does a page work? You can use a page ... "

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.amplifyframework.api.ApiException;
+import com.amplifyframework.api.graphql.GraphQLOperation;
+import com.amplifyframework.api.graphql.GraphQLRequest;
+import com.amplifyframework.api.graphql.GraphQLResponse;
+import com.amplifyframework.api.graphql.Page;
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.core.Consumer;
+
+public class AppSyncPage<T> extends Page<T> {
```
jamesonwilliams, line 13: `final class`?

core/src/main/java/com/amplifyframework/api/graphql/Page.java
```diff
@@ -0,0 +1,23 @@
...
+    public abstract boolean hasNextPage();
+
+    public abstract GraphQLOperation<T> queryNextPage(@NonNull Consumer<Page<T>> onResponse,
+                                                      @NonNull Consumer<ApiException> onFailure);
+
+    protected Page(GraphQLResponse<Iterable<T>> response) {
+        this.response = response;
+    }
+
+    public GraphQLResponse<Iterable<T>> getResponse() {
```
jamesonwilliams, line 20: `final` to close the getter. Can this be just `protected`?

aws-api/src/main/java/com/amplifyframework/api/aws/PagedResultOperation.java
richardmcclellan: Agreed.  Adding a 3rd thing was the easiest way for me to get something going, but with a few more changes, I can pretty easily just replace the multi response operation with the paged operation, so we are left with 2.  I'll try to do some investigation as well to see if we can reduce it even further.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
richardmcclellan: Addressed in https://github.com/aws-amplify/amplify-android/pull/488

core/src/main/java/com/amplifyframework/api/graphql/Page.java
richardmcclellan: > `final` to close the getter. 

Addressed in https://github.com/aws-amplify/amplify-android/pull/488

> Can this be just `protected`?

No, this is designed to be exposed - this is the only way a customer would be able to access the `GraphQLResponse`, in order to access `data` and `errors`.

aws-api/src/main/java/com/amplifyframework/api/aws/PagedResultOperation.java
richardmcclellan: Hmm, I couldn't find a constant for it.  The [example in the OkHttp docs](https://square.github.io/okhttp/) seems to do this the same way, with a hardcoded String.  Further down in this file, it looks like this String is converted to a MediaType (`MediaType.get(CONTENT_TYPE)`), and then included on the request.

richardmcclellan:
Closing in favor of https://github.com/aws-amplify/amplify-android/pull/488

Issue #451, opened 2020-05-12
Possible race condition with publishSubject 


michalisAsp:
Facing issues with master branch latest version with DataStore sync enabled.
Didn't have a chance to debug yet and identify the exact issue, but I came across this.
https://github.com/aws-amplify/amplify-android/blob/72fd6f1a0f51123d7df4dfe61bbbaf77cfa8aaa0/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java#L102

`PublishSubject` is not thread safe. Not sure what are your plans on refactoring here, but a temporary "fix" is to make it serialized.
http://reactivex.io/RxJava/javadoc/io/reactivex/subjects/Subject.html#toSerialized--

jamesonwilliams:
@michalisAsp Good catch! Appreciate the close eye, here. I will put up a PR to address this briefly and would appreciate your review.

PR #454, opened 2020-05-13
[aws-datastore] thread safe rx publish subjects

jamesonwilliams:
The Rx `PublishSubject` is not thread-safe. It is being used around the
code base in locations where `PublishSubject#on<Event>(...)` methods may
be invoked from different threads.

One strategy to resolve this is to use a serialized `PublishSubject`,
via the `toSerialized()` operator.

Refer: https://stackoverflow.com/a/43008419/695787
Resolves: https://github.com/aws-amplify/amplify-android/issues/451


jamesonwilliams:
@michalisAsp

PR #457, opened 2020-05-13
Update Auth Session to Final Design

TrekSoft:
Update Auth Session to match final design - specifically adding error handling cases to categorize why different tokens failed to be retrieved.

**Note**: Because the branch was created off the old AuthCategory side branch, GitHub is showing all the commits in the commit history - this is inaccurate - only 3a39483 and  e1b352d are new.


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
```diff
@@ -16,104 +16,81 @@
...
 
     /**
-     * AWS Credentials from Identity Pools.
-     * @return AWS Credentials from Identity Pools.
+     * The credentials which come from Identity Pool.
+     * @return the credentials which come from Identity Pool
      */
-    @Nullable
-    public AWSCredentials getAWSCredentials() {
+    @NonNull
```
raphkim, line 64: I like that they are all `@NonNull` now :)

core/src/main/java/com/amplifyframework/auth/result/AuthSessionResult.java
```diff
@@ -0,0 +1,133 @@
...
+                getError(),
+                getType()
+        );
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        } else if (obj == null || !(obj instanceof AuthSessionResult<?>)) {
```
raphkim, line 99: `obj == null` is redundant and it can be removed

core/src/main/java/com/amplifyframework/auth/result/AuthSessionResult.java
```diff
@@ -0,0 +1,133 @@
...
+         * Fetch value operation was successful.
+         * You can retrieve the result with {@link #getValue()}.
+         */
+        SUCCESS,
+        /**
+         * Fetch value operation failed.
+         * You can retrieve the exception explaining what went wrong with {@link #getError()}.
+         */
+        FAILURE
+    }
```
raphkim, line 132: is it always either-or situation? Is it possible to have partial result, where both result & non-blocking error are returned? 

Also, can't this be represented with a simple boolean `isSuccessful` or something? Otherwise, maybe we can rename to `Status` or `State` instead? `AuthSessionResult#getType()` seems very vague, and gives off feeling that authentication is categorized into multiple types of operation.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -51,4 +51,144 @@ public AuthException(
     ) {
         super(message, recoverySuggestion);
     }
+
+    /**
+     * Auth exception caused by the user being signed out.
+     */
+    public static class SignedOutException extends AuthException {
```
raphkim, line 58: Ooh are we bringing back category sub-exceptions? I liked the design, but all of our other categories don't have them at the moment right?

core/src/main/java/com/amplifyframework/AmplifyException.java
```diff
@@ -78,4 +79,48 @@ public AmplifyException(
...
+        }
+    }
+
+    /**
+     * If any child classes override this method, they should be sure to include the recoverySuggestion, message, and
+     * cause in their returned string.
+     */
+    @Override
+    public String toString() {
+        return "AuthException {" +
```
raphkim, line 120: `AuthException` -> `AmplifyException`

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -376,29 +369,25 @@ public void fetchAuthSession(
...
+                        case SIGNED_OUT_USER_POOLS_TOKENS_INVALID:
                             onSuccess.accept(
-                                AWSCognitoAuthSession.builder().isSignedIn(false).build()
+                                    new AWSCognitoAuthSession(
+                                            true,
+                                            AuthSessionResult.failure(new AuthException.SessionExpiredException()),
+                                            AuthSessionResult.failure(new AuthException.SessionExpiredException()),
+                                            AuthSessionResult.failure(new AuthException.SessionExpiredException()),
+                                            AuthSessionResult.failure(new AuthException.SessionExpiredException())
+                                    )
```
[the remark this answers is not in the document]
TrekSoft: Alright, sure.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.java
```diff
@@ -16,104 +16,81 @@
...
+    private final AuthSessionResult<AWSCredentials> awsCredentials;
+    private final AuthSessionResult<String> userSub;
+    private final AuthSessionResult<AWSCognitoUserPoolTokens> userPoolTokens;
 
-    private AWSCognitoAuthSession(
+    /**
+     * Cognito extension of AuthSession containing AWS Cognito specific tokens.
+     * @param isSignedIn Are you currently in a signed in state (an AuthN indicator to be technical)
+     * @param awsCredentials The credentials which come from Identity Pool
+     * @param userSub The id which comes from User Pools
```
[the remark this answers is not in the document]
TrekSoft: This is the user pools user id - which is really the only ID for a user which is a legit ID in Cognito.

core/src/main/java/com/amplifyframework/auth/result/AuthSessionResult.java
TrekSoft: This was the auto generated code by the IDE so planning to leave it as is.

core/src/main/java/com/amplifyframework/auth/result/AuthSessionResult.java
TrekSoft: Yeah it is always either or - basically it's designed the way it is to mirror the Swift result object: https://www.swiftbysundell.com/articles/the-power-of-result-types-in-swift/

Also mirrors the design Jameson did for an object like this on Android with a getType method.

core/src/main/java/com/amplifyframework/auth/AuthException.java
TrekSoft: Yeah we need to discuss this as a team for adding to the other categories but for now, added this because it was super necessary for Auth.

core/src/main/java/com/amplifyframework/AmplifyException.java
TrekSoft: Ah, nice catch - thanks.

core/src/main/java/com/amplifyframework/auth/result/AuthSessionResult.java
raphkim: I guess using a specific enum to represent the state over boolean is fine, and it's not unprecedented as you mention. But I think we should still strive for a clearer name especially for customer-exposed data structures like this. Also, `Status` is what I'm leaning towards as the name, similarly to [this](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/core/InitializationStatus.java).

TrekSoft: Discussed offline that the idea of type in the context of request makes sense since there are static success/failure type results vs. state which reflects it could change and that this was the naming decision Jameson went with in mocking out the class: https://gist.github.com/jamesonwilliams/b15b8888c2e759d575c59fb1567ea32f

PR #459, opened 2020-05-13
[aws-datastore] Fix unhandled exceptions when no API is available

rjuliano:
## Overview
This PR addresses an issue where the developer has added both the DataStore and API plugins to Amplify but has not yet pushed the backend infrastructure (or an error happens while communicating with AppSync)

With this change, even if there are errors on the remote sync process startup, the app will not crash and local store will still be available. 

### RemoteModelMutations
Modified the observe function to return a merged obervable that contains one child observer per model type/subscription type combination. If any of these children fail, we log the error as a warning and complete. 

### SyncProcessor
Updated the hydrate function to handle errors by logging and forcing a the onComplete event.

Ideally, with both of these, the next step would be to implement a retry strategy. First step is to prevent the app from crashing if an unhandled exception occurs on startup.


aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -52,7 +61,6 @@
 @RunWith(RobolectricTestRunner.class)
 public final class AWSDataStorePluginTest {
```
jamesonwilliams, line 62: I think we should start splitting tests out into other classes. The ones I have in here right now are related to model synchronization. Probably `AWSDataStorePluginTest` was a bad name to begin with. All in all, the plugin does a lot of different stuff. I think we are ultimately funneling all of our high level tests into this class and its going to end up having a lot of unrelated stuff and being a bit of a maintenance issue.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -79,8 +87,10 @@ public void setup() {
     @Test
     public void configureAndInitializeInLocalMode() throws AmplifyException {
         //Configure DataStore with an empty config (All defaults)
-        awsDataStorePlugin.configure(new JSONObject(), context);
-        awsDataStorePlugin.initialize(context);
+        ApiCategory emptyApiCategory = spy(ApiCategory.class);
+        AWSDataStorePlugin standAloneDataStorePlugin = new AWSDataStorePlugin(modelProvider, emptyApiCategory);
```
jamesonwilliams, line 91: You may find utility in the `DataStoreCategoryConfigurator` . Er, maybe thats in `androidTest` though isn't it...

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -93,18 +103,57 @@ public void configureAndInitializeInLocalMode() throws AmplifyException {
...
+                    Person actualPerson = results.next();
+                    assertNotNull(actualPerson);
+                    assertFalse(results.hasNext()); // We should only have one result.
+                    assertEquals(person1, actualPerson);
+                    single.onSuccess(true);
+                }, single::onError);
+            })
+        ).doOnError(error -> {
+            fail(error.getMessage());
+        }).blockingAwait();
```
jamesonwilliams, line 154: I think you could do `blockingGet()`, save the return value, and then do `if (returnedValue != null) throw returnedValue;`

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -83,6 +89,34 @@ private AWSDataStorePlugin(
             @NonNull ModelSchemaRegistry modelSchemaRegistry,
             @NonNull ApiCategory api,
             @Nullable DataStoreConfiguration userProvidedConfiguration) {
+        // See https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling
+        // Without this, the host app crashes on an unhandled exception.
+        RxJavaPlugins.setErrorHandler(e -> {
```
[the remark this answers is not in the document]
rjuliano: All fair points. I guess my thinking was that we wouldn't want to be the ones crashing the app because of an unhandled exception. I'm definitely not crazy about the pattern myself. I like the idea of perhaps passing something in. Let me circle back on this.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -83,6 +89,34 @@ private AWSDataStorePlugin(
             @NonNull ModelSchemaRegistry modelSchemaRegistry,
             @NonNull ApiCategory api,
             @Nullable DataStoreConfiguration userProvidedConfiguration) {
+        // See https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling
+        // Without this, the host app crashes on an unhandled exception.
```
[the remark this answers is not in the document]
rjuliano: Removed

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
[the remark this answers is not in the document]
rjuliano: removed this guy too.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -97,12 +101,28 @@ public Orchestrator(
...
+                    errorOccurred.set(true);
+                }).doOnComplete(() -> {
+                    if (!errorOccurred.get()) {
+                        LOG.debug("Starting MutationProcessor and SubscriptionProcessor.");
+                        mutationProcessor.startDrainingMutationOutbox();
+                        subscriptionProcessor.startDrainingMutationBuffer();
+                    }
+                })
+                // This will prevent the host app from crashing if no API is configured.
+                .onErrorComplete()
```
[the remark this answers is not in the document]
rjuliano: Don't need this onErrorComplete here since we're handling the the errors for the individual steps.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -40,8 +43,14 @@
 
 import java.util.Collections;
 
+import io.reactivex.Completable;
+
 import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
+import static junit.framework.TestCase.assertEquals;
```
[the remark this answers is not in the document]
rjuliano: done

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -61,14 +69,14 @@
      * the SyncProcessor is running. Otherwise, either the SyncProcessor is *not*
      * running, or it is running but not functioning as we expect it to.
      */
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
rjuliano: For whatever reason, all of the `.when(mockApiPlugin).<whatever>` methods get marked as unchecked. If I explicitly cast it, I get dinged for something else :)

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: If it's ok, I'd like to undertake that as a separate effort. At the very least, we could break it up in API-mode tests vs non-API-mode tests, but it can be broken down much further.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -115,6 +115,14 @@ void startSubscriptions() {
             ));
             latch.await(SUBSCRIPTION_START_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         })
+        .doOnError(exception -> {
+            LOG.warn("An error occurred with one of the subscriptions to the remote data store for " +
```
TrekSoft, line 119: super minor nit: "datastore" vs. "data store"

jamesonwilliams: Actually, neither!

It's the _`DataStore`_.

jamesonwilliams, line 119:
```suggestion
            LOG.warn("An error occurred with one of the subscriptions to the remote DataStore for " +
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -115,6 +115,14 @@ void startSubscriptions() {
...
+                "model "+clazz.getSimpleName()+" "+subscriptionType.name(),
```
jamesonwilliams, line 120:
```suggestion
                "model " + clazz.getSimpleName() + " " + subscriptionType.name(),
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
jamesonwilliams, line 49:
```suggestion
import static org.junit.Assert.assertEquals;
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
jamesonwilliams: Sounds good to me

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
jamesonwilliams, line 72:
```suggestion
```
I'd expect it to be on the class, though -- or on one of the `@Test`. Here one the `@Before`, it doesn't seem like it suppressing anything inside of this particular method.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
jamesonwilliams, line 154:
```suggestion
        }).blockingAwait(OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
^ This won't quite work via the "accept a change" UX, unfortunately.

Anyway, this will help prevent our test from hanging indefinitely in case something goes wrong.

PR #462, opened 2020-05-14
Adds Hub events for Auth

TrekSoft:
Adds hub events to Auth to notify when these three globally relevant events happen:

- Signed out
- Signed in
- Session expired


TrekSoft:
The enum is at the Framework level and UserState is part of our current AWSMobileClient implementation so I wouldn't want to introduce that at the core level - thanks for the PR!

PR #465, opened 2020-05-14
[aws-api] Use SubscriptionListener in API subscription methods

jamesonwilliams:
The API category includes a collection of GraphQL behaviors. One of the
behaviors is to create a subscription to a GraphQL endpoint.

Currently, the signatures for these subscription methods require
numerous arguments to be provided:

  1. `Consumer<String>`, to receive notification of subscription start;
  2. `Consumer<GraphQLResponse<T>>` to receive subscription data;
  3. `Consumer<ApiException>`, when subscription terminates in failure;
  4. `Action`, called when the subscription terminates gracefully.

The user may or may not care about all of the data provided in the
various callbacks, but yet each must still be specified.

To resolve this, the callbacks are consolidated into a single base class,
an `ApiSubscriptionListener`. A user may extend this class and override any
methods for which they would like to instrument custom logic. For
example:

```
api.subscribe(BlogPost.class, SubscriptionType.CREATE,
    new ApiSubscriptionListener<BlogPost>() {
        @Override
        public void onSubscriptionStarted(String token) {
            Log.i(TAG, "Subscription started!");
        }

        @Override
        public void onSubscriptionData(GraphQLResponse<BlogPost> data) {
            Log.i(TAG, "Got some data on subscription = " + data);
        }

        @Override
        public void onSubscriptionFailure(ApiException failure) {
            Log.w(TAG, "Subscription terminated with failure", failure);
        }

        @Override
        public void onSubscriptionComplete() {
            Log.i(TAG, "Subscription terminated gracefully.");
        }
    }
);
```

Or, if only interested in subscription data:
```
api.subscribe(BlogPost.class, SubscriptionType.CREATE,
    new ApiSubscriptionListener<BlogPost>() {
        @Override
        public void onSubscriptionData(GraphQLResponse<BlogPost> data) {
            Log.i(TAG, "Got some data on subscription = " + data);
        }
    }
);
```

A `DelegatingApiSubscriptionListener` is provided for compatibility. It
accepts the four `Consumer`/`Action` arguments previously available:
```
api.subscribe(BlogPost.class, SubscriptionType.CREATE,
    DelegatingApiSubscriptionListener.create(
        startToken -> {},
        nextData -> {},
        failure -> {},
        () -> {}
    )
);
```

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -25,6 +25,7 @@ import androidx.core.util.ObjectsCompat;
 import com.amplifyframework.api.ApiException;
 import com.amplifyframework.api.ApiPlugin;
 import com.amplifyframework.api.aws.operation.AWSRestOperation;
+import com.amplifyframework.api.graphql.ApiSubscriptionListener;
 import com.amplifyframework.api.graphql.GraphQLOperation;
 import com.amplifyframework.api.graphql.GraphQLRequest;
 import com.amplifyframework.api.graphql.GraphQLResponse;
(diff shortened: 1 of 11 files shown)


jamesonwilliams:
We decided today _not_ to pursue this. Thanks @richardmcclellan for your feedback!

PR #467, opened 2020-05-15
[aws-datastore] Refactor SubscriptionProcessor

jamesonwilliams:
The logic for the `SubscriptionProcessor` was split across an anemic
class, the actual `SubscriptionProcessor`, and then another class called
`RemoteModelMutations`. Instead, this other class is combined back into
the `SubscriptionProcessor`.

Unit test coverage is added.

Previously, the processor would not wait for all of the subscriptions to
have been fully established. This resulted in situations where data
could be missed on a subscription. To resolve this, wait receipt of a
subscription ID before yielding control of a subscription observable.


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -56,15 +82,47 @@
...
+            emitter.setCancellable(cancelable::get);
+            cancelable.set(method.subscribe(
+                clazz,
+                token -> latch.countDown(),
+                emitter::onNext,
+                emitter::onError,
+                emitter::onComplete
+            ));
+            latch.await(SUBSCRIPTION_START_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        })
```
rjuliano, line 117: To incorporate, my changes all you would have to do here is:

```
        .doOnError(exception -> {
            LOG.warn("An error occurred with the subscription for model " + clazz.getSimpleName() +
                " and operation type " + subscriptionType.name(), exception.getCause());
        })
        .onErrorResumeNext((ObservableSource<GraphQLResponse<ModelWithMetadata<T>>>) Observer::onComplete)
```
This basically handles the error by logging it and completing the observable. This prevents the Orchestrator from crashing if there's no API configured or the API is not reachable.

If you think it makes sense, we can incorporate this changes here.

That's basically what I added in #459

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessorTest.java
```diff
@@ -0,0 +1,227 @@
...
+        SubscriptionProcessor.SubscriptionMethod method =
+            SubscriptionProcessor.subscriptionMethodFor(stub, subscriptionType);
+        method.subscribe(eq(clazz), anyConsumer(), anyConsumer(), anyConsumer(), anyAction());
+    }
+
+    private static void verifySubscriptions(
+            AppSync appSync, List<Class<? extends Model>> classes, SubscriptionType[] subscriptionTypes) {
+        Observable.fromIterable(classes)
+            .flatMap(modelClass -> Observable.fromArray(subscriptionTypes)
+                .map(subscriptionType -> Pair.create(modelClass, subscriptionType)))
```
rjuliano, line 208: Java's missing tuple :) I like it. I'll have to remember it's under android.util

PR #470, opened 2020-05-15
[aws-datastore] Re-enable cloud synchronization test

jamesonwilliams:
`syncUpToCloudIsWorking` is passing reliably and should be re-enabled, so
we can detect and prevent new failure modes.

`syncDownFromCloudIsWorking` needs further auditing, and is left Ignored.


jamesonwilliams:
Re CircleCI results: Oops, guess not. Back to the workshop with this one.

PR #475, opened 2020-05-16
Allow a global log level in the AndroidLoggingPlugin

jamesonwilliams:
Adds a mechanism to set a global log level. Configuring Amplify with:

    Amplify.addPlugin(new AndroidLoggingPlugin(LogLevel.VERBOSE))

Will cause any new Loggers to use `LogLevel.VERBOSE` as the default log
level.

### DIFF
--- a/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
+++ b/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
@@ -24,7 +24,6 @@ import com.amplifyframework.analytics.AnalyticsEvent;
 import com.amplifyframework.analytics.AnalyticsProperties;
 import com.amplifyframework.analytics.UserProfile;
 import com.amplifyframework.core.Amplify;
-import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.logging.Logger;
 import com.amplifyframework.testutils.Sleep;
 
@@ -51,8 +50,11 @@ import static org.junit.Assert.assertTrue;
  * Validates the functionality of the {@link AmazonPinpointAnalyticsPlugin}.
  */
 public class AnalyticsPinpointInstrumentedTest {
-    private static final Logger LOG = Amplify.Logging.forCategory(CategoryType.ANALYTICS);
 
+    /**
+     * Log tag for the test class.
+     */
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-analytics");
     private static final int EVENT_FLUSH_TIMEOUT = 60;
     private static final int EVENT_FLUSH_WAIT = 2;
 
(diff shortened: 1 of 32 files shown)


aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
```diff
@@ -50,11 +51,8 @@
  * Validates the functionality of the {@link AmazonPinpointAnalyticsPlugin}.
  */
 public class AnalyticsPinpointInstrumentedTest {
+    private static final Logger LOG = Amplify.Logging.forCategory(CategoryType.ANALYTICS);
```
jpignata, line 54: This drops the `aws-` prefix from the log slugs, which doesn't necessarily seem wrong to me, but did spark the thought that is there a reason for individual plugins to differentiate themselves? (say, for instance, you're using the `aws-analytics-pinpoint` plugin and the `foo-analytics-bar` plugin, for example.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowPredictionsService.java
```diff
@@ -35,7 +36,7 @@
  * pre-trained models to make predictions offline.
  */
 public final class TensorFlowPredictionsService {
-    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-predictions-tensorflow");
+    private static final Logger LOG = Amplify.Logging.forCategory(CategoryType.PREDICTIONS);
```
jpignata, line 39: This is an interesting one as there will be routinely two plugins -- `aws-predictions` and `aws-predictions-tensorflow` running concurrently.

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -129,6 +135,36 @@ public void noContentLoggedAtThresholdNone() {
...
+        final Map<CategoryType, String> actuals = new HashMap<>();
+        for (CategoryType categoryType : CategoryType.values()) {
+            String result = plugin.forCategory(categoryType).getNamespace();
+            actuals.put(categoryType, result);
+        }
+        final Map<CategoryType, String> expected = new HashMap<>();
+        expected.put(CategoryType.ANALYTICS, "amplify:analytics");
+        expected.put(CategoryType.API, "amplify:api");
+        expected.put(CategoryType.AUTH, "amplify:auth");
+        expected.put(CategoryType.DATASTORE, "amplify:data-store");
```
jpignata, line 155: Huh. I guess it stands to reason that `DataStore` would lowercase to `data-store`, but I hadn't seen it before.

jamesonwilliams: Hm, yea, same. `datastore` looks more natural. I can probably just update the logic to print a string of lowercase letters without punctuation, that might work.

aws-predictions-tensorflow/src/main/java/com/amplifyframework/predictions/tensorflow/service/TensorFlowPredictionsService.java
jamesonwilliams: The original format here was both like `ampilfy:plugin-name` and `amplify-category-name`. 

Unfortunately, `forPlugin(Plugin<?> plugin)` turned out not to be a realistic option. It would mean we'd have to have the plugin instance available in order to call the logger. 

I don't love putting everything under the categroy slug, but at least its consistent.

I'll redo this PR and just make it about the default log level, not about the slug stuff.

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/AnalyticsPinpointInstrumentedTest.java
jamesonwilliams: Yea, I'm not sold on it either. I'll update the PR to only include the default log level change.

PR #488, opened 2020-05-19
[aws-api] add pagination

richardmcclellan:
Three big changes in this PR:

1. Refactor GraphQLResponseFactory 
   - Moves most of logic from GraphQLResponseFactory into several JsonDeserializer objects.   
   - Eliminates duplicated logic that was in place for serializing a response with one result vs multiple results.  The new approach is more generic and can handle any type of response.
   - Allows serializing a response to any provided `Type`, instead of just `T` or `Iterable<T>`.  

2. Refactor what the type variables represent in `GraphQLBehavior`
   - `R` is a new type variable representing the type of `data` on the `GraphQLResponse`.  (this can be either `Model`, `Iterable<Model>`, or `Page<Model>`, or other types in the future)
   - `T` used to have multiple meanings - sometimes Model, sometimes type of `data` inside `GraphQLResponse`.  It now only represents Model, which makes all related code much easier to understand.
   - This is the same as the implementation on iOS.

3. Add pagination support (the smallest part of this PR)
   - Adds an abstract `Page` class which contains a list of items and also and provides paging context (`boolean hasNextPage()`, and `GraphQLRequest getRequestForNextPage()`
   - Adds an `AppSyncPage` concrete class to encapsulate AppSync specific paging logic, and associated `AppSyncPageDeserializer`

Developer Experience:
```
    private void listTodos() {
        try {
            int limit = 2;
            String nextToken = null;
            listTodos(AppSyncGraphQLRequestFactory.buildPagedQuery(Todo.class, null, limit, nextToken));
        } catch (ApiException apiFailure) {
            handleError(apiFailure);
        }
    }

    private void listTodos(GraphQLRequest<Page<Todo>> request) {
        Amplify.API.query(request, response -> {
            if (response.hasData()) {
                for (Todo todo : response.getData().getItems()) {
                    Log.d(TAG, "todo: " + todo.getName());
                }
            }
            if (response.hasErrors()) {
                errors.addAll(response.getErrors());
            }
            if(response.getData().hasNextPage()) {
                listTodos(response.getData().getRequestForNextPage());
            }
        }, apiException -> { 
             Log.e(TAG, exception.getMessage(), exception); 
        });
    }

```


aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -16,6 +16,7 @@
 package com.amplifyframework.api.aws;
 
 import android.annotation.SuppressLint;
+
```
jamesonwilliams, line 19: The current Checkstyle likes `android` and `androidx` together.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
```diff
@@ -0,0 +1,33 @@
+package com.amplifyframework.api.aws;
+
+import androidx.annotation.NonNull;
+
+import com.amplifyframework.api.graphql.GraphQLRequest;
+import com.amplifyframework.api.graphql.GraphQLResponse;
+import com.amplifyframework.api.graphql.Page;
+
+final class AppSyncPage<M> extends Page<M> {
```
jamesonwilliams, line 9: Looks reasonable, remember to add `equals`, `hashCode`, `toString` before pushing

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public boolean hasNextPage() {
+        return nextToken != null;
+    }
+
+    @Override
+    public GraphQLRequest<Page<M>> getRequestForNextPage() {
+        this.request.addVariable(NEXT_TOKEN_KEY, nextToken);
```
jamesonwilliams, line 30: I guess this will mutate the existing request. It'd be better to return a new, immutable copy.

To illustate, consider what happens when:
```
GraphQLRequest request = buildOne();
request.getVaraiables() // state A 
new AppSyncPage(response, request, token).getRequestForNextPage();
request.getVariables(); // state B.
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPageDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+
+/**
+ * Takes an input such as the following and deserializes to an AppSyncPage
+ * {
+ *   "data": {
+ *     "listTodos": {
```
jamesonwilliams, line 20: Ah, I guess this thing here, the `listTodos` and under. _This_ is the "page" right?

richardmcclellan:
For a paged response, this PR currently generates an object that looks like this:

```
Page {
    GraphQLResponse {
          List<Todo> data
          List<GraphQLResponse.Error> errors
    }
    abstract boolean getNextPage()
    abstract GraphQLRequest getRequestForNextPage()
}
```

After chatting with @jamesonwilliams and others yesterday, we all agreed that a better approach would be to return something like this:

```
GraphQLResponse {
      Page<Todo> data  {
            List<Todo> items
            abstract boolean hasNextPage()
            abstract GraphQLRequest<Page<Todo>> getRequestForNextPage()
      }
      List<GraphQLResponse.Error> errors
}
```

I'll be working on this adjustment now and should be able to push a revision at the end the day.

core/src/main/java/com/amplifyframework/api/graphql/Page.java
```diff
@@ -0,0 +1,9 @@
+package com.amplifyframework.api.graphql;
+
+public abstract class Page<M> {
```
richardmcclellan, line 3: I'm open to suggestions on other names for this class.  We actually already have a `Page` object defined in core, which is used for local pagination via DataStore, so it's kinda confusing to have this one with the same name.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
richardmcclellan: Added in latest commit!

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
richardmcclellan: Fixed!

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPageDeserializer.java
richardmcclellan: This actually was correct based on the design I had implemented yesterday.  I've updated the implementation to the design we discussed last night, which removes the top _two_ levels. (and I have updated this comment).

richardmcclellan:
> This change contains a lot of "change T to M." To reduce the size of the PR, I suggest to use `R` for the new response type, and keep `T` as the model type. I agree with you that `M` is more clear, but we could make that change in a separate unrelated PR, perhaps.

I've updated per this recommendation, so this is the new state:

 - `T` represents the `Model`
 - `R` represents the type of `GraphQLResponse.data` (e.g. `Model`, `Iterable<Model>`, `Page<Model>`, or anything we add in the future)
 - I actually realized at the end that we already use `R` in `ApiOperation` to indicate type of operation (e.g. `GraphQLRequest` or `RestRequest`, so now we're using `R` for two things, but not really in the same context so not a huge deal, but still a reason we might want to rethink what the best letters to use are.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPage.java
richardmcclellan: I added a `GraphQLRequest` copy constructor, and am returning a copy.  The value returned is still NOT immutable though, because GraphQLRequest has public methods that support some mutation (e.g. `putVariable`, `addFragment`)

I think that @drochetti might be refactoring `GraphQLRequest` right now anyway as part of making GraphQLRequest builder/helpers, so we should revisit this after that work is done.

core/src/main/java/com/amplifyframework/api/graphql/Page.java
drochetti: Does `PaginatedResult` reflect the intent? (or something along those lines)

richardmcclellan: I like that!  I'll update it.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonErrorDeserializer.java
```diff
@@ -66,7 +65,7 @@
                     message = context.deserialize(value, String.class);
                     break;
                 case LOCATIONS_KEY:
-                    Type locationsType = new TypeToken<List<GraphQLLocation>>() {}.getType();
+                    Type locationsType = TypeMaker.getParameterizedType(List.class, GraphQLLocation.class);
```
jamesonwilliams, line 68: Yess!!! Love it :-D

PR #489, opened 2020-05-19
refactor(core): change AWS-prefixed date types to Temporal.*

drochetti:
**Notes:**

- Moved `AWS*` date types to `Temporal.*`
- iOS equivalent PR: https://github.com/aws-amplify/amplify-ios/pull/316


aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
```diff
@@ -54,9 +51,9 @@ public String serialize(Map<String, Object> variables) {
      *
      * https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html
      */
-    static class AWSDateSerializer implements JsonSerializer<AWSDate> {
+    static class AWSDateSerializer implements JsonSerializer<Temporal.Date> {
```
drochetti, line 54: missed these serializers, will do another commit addressing these

drochetti: done!

core/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java
```diff
@@ -0,0 +1,474 @@
...
+import org.threeten.bp.format.DateTimeFormatterBuilder;
+import org.threeten.bp.format.DateTimeParseException;
+import org.threeten.bp.temporal.ChronoField;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Namespace for all the Date/Time related types.
+ */
+public final class Temporal {
```
richardmcclellan, line 38: I'm not crazy about having all of these crammed into one big file, though I understand there's not really another approach to achieve the naming strategy we agreed on.   

One option would be to call these `TemporalDate`, `TemporalDateTime`, ...  instead of `Temporal.Date` , `Temporal.DateTime`?   I understand we already made a decision on the naming though, so I'm fine with what's in this PR.

drochetti: I had the same conflict tbh. I didn't think about this when we decided on the naming, but then one I saw that it was less than 1k lines and the IDE helps you navigate I decided not to bring yet another naming discussion to the surface at this point...

Also, iOS has an even bigger PR with the `Temporal.*` name. The difference is that there we can do extensions like in Kotlin, so we can divide each temporal in its own file by adding an extension to the namespace class.

PR #490, opened 2020-05-20
Rjuliano/dspredicate

rjuliano:
*Issue #, if available:*

*Description of changes:*


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
```diff
@@ -42,12 +43,17 @@
     private final Class<T> classOfMutatedItem;
     private final Type mutationType;
     private final TimeBasedUuid mutationId;
+    private QueryPredicate predicate;
 
-    private PendingMutation(TimeBasedUuid mutationId, T mutatedItem, Class<T> classOfMutatedItem, Type mutationType) {
+    private PendingMutation(TimeBasedUuid mutationId,
+                            T mutatedItem, Class<T> classOfMutatedItem,
```
jamesonwilliams, line 49: Can you chomp the second var down onto a separate line

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
```diff
@@ -155,6 +173,11 @@ Type getMutationType() {
         return mutationType;
     }
 
+    @Nullable
+    QueryPredicate getPredicate() {
+        return predicate;
```
jamesonwilliams, line 178: Need to update hashCode/equals/toStirng, too.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -265,4 +211,153 @@ private Completable notifyContentAvailable() {
...
         return mutationQueue.peek();
     }
+
+    /**
+     * Encapsulate the logic to determine which actions to take based on incoming and existing
+     * mutations. Non-static so we can access instance methods of the outer class. Private because
+     * we don't want this logic called from anywhere else.
+     * @param <T> the model type
+     */
+    private final class IncomingMutationConflictHandler<T extends Model> {
```
jamesonwilliams, line 221: Should this implement our conflict handler interface ?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -84,72 +85,17 @@ public boolean hasPendingMutation(@NonNull String modelId) {
         if (existingMutation == null) {
```
jamesonwilliams, line 85: If we have a bunch of conditional updates come in, then don't we need to enqueue them all? So we would end up with a collection of mutations per model ID?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -161,6 +163,9 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api) {
             updateInput.put("_version", version);
 
             final Map<String, Object> variables = Collections.singletonMap("input", updateInput);
+            if (predicate != null) {
```
jamesonwilliams, line 166: I have always wished we had like a non-`null` `QueryPreciates.matchAll()` available. That way, we could avoid having null checks all over the place.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -191,9 +306,9 @@ private AppSyncRequestFactory() {}
      * @param <T> Type of model being mutated
      * @return Mutation doc
      */
-    private static <T extends Model> String buildMutation(Class<T> modelClass, MutationType mutationType)
-            throws DataStoreException {
-
+    private static <T extends Model> String buildMutation(Class<T> modelClass,
+                                                          boolean includePredicate,
```
jamesonwilliams, line 310: Instead of boolean flag, can we eight:
1. Pass the predicate itself in right here (even though we'll only do a null check, internally), or,
2. Use an `enum` as below
```
enum PredicateOutput {
    GENERATE,
    OMIT;
}
```

Then, when you call this method, it will look like:
```
buildMutation(clazz, PredicateOutput.OMIT, MutationType.CREATE);
```
And the meaning of the second var will be clear.

Or maybe:
```
enum GeneratePredicate {
    TRUE,
    FALSE
}
```
Which is just a fancy boolean with a special, readable name.

rjuliano:
Breaking up this PR

PR #491, opened 2020-05-20
[aws-datastore] Permit multiple mutations per model ID

jamesonwilliams:
In order to support conditional outgoing mutations, we need to allow
multiple mutations to exist in the queue at the same time.

This means there is no longer a notion of a single "pre-existing
mutation" for a model ID, there are n-many such.

All the same, we will concern the _first_ match in the queue, when we
enact business logic about the outbox state.

### DIFF
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
@@ -678,8 +678,8 @@ public final class PersistentMutationOutboxTest {
         PendingMutation<BlogOwner> firstMutation = PendingMutation.update(originalJoe, BlogOwner.class);
         storage.save(originalJoe, converter.toRecord(firstMutation));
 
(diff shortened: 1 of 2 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
```diff
@@ -650,12 +655,40 @@ public void enqueueEventPostedWhenNewOutboxItemAdded() {
...
+     */
+    @Test
+    public void nextItemForModelIdReturnsFirstEnqueued() throws DataStoreException {
+        BlogOwner originalJoe = BlogOwner.builder()
+            .name("Joe Swanson")
+            .build();
+        PendingMutation<BlogOwner> firstMutation = PendingMutation.update(originalJoe, BlogOwner.class);
+        storage.save(originalJoe, converter.toRecord(firstMutation));
+
+        BlogOwner updatedJoe = BlogOwner.builder()
```
jamesonwilliams, line 681: Ah crap this was supposed to say copyOfBuilder, to keep the old ID. Need to validate before push.

PR #495, opened 2020-05-20
[aws-datastore] Mark mutations as in-flight while processing

jamesonwilliams:
The behavior of the Mutation Outbox is changed, to include an in-memory,
"in-flight" state for its pending mutations.

Before we start to publish a mutation to AppSync, we will mark it as
"in-flight", in the Mutation Outbox. This signifies that the mutation
will be treated as immutable for the duration of its life in-memory.

The mutation will continue to be marked as "in-process" until one of two
things happen:

 1. The mutation is removed from the mutation outbox.

 2. The system restarts, and the mutation is loaded from disk, again,
    while rebuilding the mutation outbox. While loading, it will appear
    as though this mutation did _not_ finish processing successfully,
    and so should be retried.

While enqueuing new mutations, existing, in-flight mutations will be
ignored. That is to say, new mutations will be queued behind the
in-flight ones, and the in-flight ones will not be tampered with.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
@@ -115,13 +115,17 @@ final class MutationProcessor {
      * @return A Completable that emits success when the item is processed, emits failure, otherwise
      */
     private <T extends Model> Completable processOutboxItem(PendingMutation<T> mutationOutboxItem) {
-        // First, publish the mutation over the network.
-        return publishToNetwork(mutationOutboxItem)
-            .flatMapCompletable(mutation ->
-                // Once the server knows about it, it's safe to remove from the outbox.
-                // This is done before merging, because the merger will refuse to merge
-                // if there are outstanding mutations in the outbox.
-                mutationOutbox.remove(mutationOutboxItem).andThen(merger.merge(mutation))
+        // First, mark the item as in-flight.
+        return mutationOutbox.markInFlight(mutationOutboxItem.getMutationId())
+            // Then, put it "into flight"
+            .andThen(publishToNetwork(mutationOutboxItem)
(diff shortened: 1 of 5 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -115,13 +115,17 @@ private Completable drainMutationOutbox() {
...
-                mutationOutbox.remove(mutationOutboxItem).andThen(merger.merge(mutation))
+        // First, mark the item as in-flight.
+        return mutationOutbox.markInFlight(mutationOutboxItem.getMutationId())
+            // Then, put it "into flight"
+            .andThen(publishToNetwork(mutationOutboxItem)
+                .flatMapCompletable(modelWithMetadata ->
+                    // Once the server knows about it, it's safe to remove from the outbox.
+                    // This is done before merging, because the merger will refuse to merge
+                    // if there are outstanding mutations in the outbox.
+                    mutationOutbox.remove(mutationOutboxItem.getMutationId())
```
TrekSoft, line 126: Quick question - if somehow there's a failure between the server getting it and this remove from outbox step, will it gracefully ignore the duplicate transaction next time the outbox attempts to perform it?

jpignata: My suspicion is if we get into this state, the outbox will be indefinitely wedged. We have some work to do here to make the outbox a) concurrent b) resilient

TrekSoft: I would think with the conditions on Updates and Deletes (or ID check on Create) it'd be fairly simple to avoid this being a problem if we aren't already. But yeah definitely not a launch blocker if we aren't since it's an edge case we could follow up with a fix for.

PR #496, opened 2020-05-20
[aws-datastore] adding predicate parameter to the necessary APIs

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR is in preparation for adding predicates to the mutations that are sent to the remote APIs. Here is what it's expected to cover:
- Adding predicate field to the PendingMutation
- Updating the DataStore's AppSync interface to include a predicate parameters for deletes and updates.
- Updating tests to pass null into this new parameter for now

There will be a follow up PR that will wire up all the business logic of how these predicates will be applied

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
@@ -29,6 +29,7 @@ import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.async.Cancelable;
 import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.appsync.AppSync;
 import com.amplifyframework.datastore.appsync.AppSyncClient;
 import com.amplifyframework.datastore.appsync.ModelWithMetadata;
@@ -169,7 +170,7 @@ public final class AppSyncClientInstrumentationTest {
             .build();
         Long updateBlogStartTime = new Date().getTime();
 
-        ModelWithMetadata<Blog> blogUpdateResult = update(updatedBlog, 1);
+        ModelWithMetadata<Blog> blogUpdateResult = update(updatedBlog, 1, Blog.ID.eq(updatedBlog.getId()));
 
         assertEquals(updatedBlog.getName(), blogUpdateResult.getModel().getName());
         assertEquals(updatedBlog.getOwner().getId(), blogUpdateResult.getModel().getOwner().getId());
@@ -182,7 +183,7 @@ public final class AppSyncClientInstrumentationTest {
         assertTrue(updatedBlogLastChangedAt > updateBlogStartTime);
 
         // Delete one of the posts
-        ModelWithMetadata<Post> post1DeleteResult = delete(Post.class, post1.getId(), 1);
+        ModelWithMetadata<Post> post1DeleteResult = delete(Post.class, post1.getId(), 1, null);
         assertEquals(
             post1.copyOfBuilder()
                 .blog(Blog.justId(blog.getId()))
@@ -193,7 +194,8 @@ public final class AppSyncClientInstrumentationTest {
         assertEquals(Boolean.TRUE, isDeleted);
 
         // Try to delete a post with a bad version number
-        List<GraphQLResponse.Error> post2DeleteErrors = deleteExpectingResponseErrors(Post.class, post2.getId(), 0);
+        List<GraphQLResponse.Error> post2DeleteErrors =
+            deleteExpectingResponseErrors(Post.class, post2.getId(), 0, null);
         assertEquals("Conflict resolver rejects mutation.", post2DeleteErrors.get(0).getMessage());
 
         // Run sync on Blogs
@@ -234,10 +236,10 @@ public final class AppSyncClientInstrumentationTest {
      */
     @SuppressWarnings("SameParameterValue") // Keep details in the actual test.
     @NonNull
-    private <T extends Model> ModelWithMetadata<T> update(@NonNull T model, int version)
+    private <T extends Model> ModelWithMetadata<T> update(@NonNull T model, int version, QueryPredicate predicate)
             throws DataStoreException {
         return awaitResponseData((onResult, onError) ->
-            api.update(model, version, onResult, onError));
+            api.update(model, version, predicate, onResult, onError));
     }
 
     /**
@@ -251,10 +253,10 @@ public final class AppSyncClientInstrumentationTest {
      */
     @SuppressWarnings("SameParameterValue") // Reads better with details in one place
     private <T extends Model> ModelWithMetadata<T> delete(
-            @NonNull Class<T> clazz, String modelId, int version)
+            @NonNull Class<T> clazz, String modelId, int version, QueryPredicate predicate)
             throws DataStoreException {
         return awaitResponseData((onResult, onError) ->
-            api.delete(clazz, modelId, version, onResult, onError));
+            api.delete(clazz, modelId, version, predicate, onResult, onError));
     }
 
     /**
@@ -269,11 +271,11 @@ public final class AppSyncClientInstrumentationTest {
      */
     @SuppressWarnings("SameParameterValue") // It'll read better if we keep details in the call line
     private <T extends Model> List<GraphQLResponse.Error> deleteExpectingResponseErrors(
-            @NonNull Class<T> clazz, String modelId, int version)
+            @NonNull Class<T> clazz, String modelId, int version, QueryPredicate predicate)
             throws DataStoreException {
         return awaitResponseErrors(
             (Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResult, Consumer<DataStoreException> onError) ->
-                api.delete(clazz, modelId, version, onResult, onError)
+                api.delete(clazz, modelId, version, predicate, onResult, onError)
         );
     }
 
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
@@ -170,7 +170,7 @@ public final class AppSyncClientInstrumentationTest {
             .build();
         Long updateBlogStartTime = new Date().getTime();
 
-        ModelWithMetadata<Blog> blogUpdateResult = update(updatedBlog, 1, Blog.ID.eq(updatedBlog.getId()));
+        ModelWithMetadata<Blog> blogUpdateResult = update(updatedBlog, 1);
 
         assertEquals(updatedBlog.getName(), blogUpdateResult.getModel().getName());
         assertEquals(updatedBlog.getOwner().getId(), blogUpdateResult.getModel().getOwner().getId());
@@ -183,7 +183,7 @@ public final class AppSyncClientInstrumentationTest {
         assertTrue(updatedBlogLastChangedAt > updateBlogStartTime);
 
         // Delete one of the posts
-        ModelWithMetadata<Post> post1DeleteResult = delete(Post.class, post1.getId(), 1, null);
+        ModelWithMetadata<Post> post1DeleteResult = delete(Post.class, post1.getId(), 1);
         assertEquals(
             post1.copyOfBuilder()
                 .blog(Blog.justId(blog.getId()))
@@ -194,8 +194,7 @@ public final class AppSyncClientInstrumentationTest {
         assertEquals(Boolean.TRUE, isDeleted);
 
         // Try to delete a post with a bad version number
-        List<GraphQLResponse.Error> post2DeleteErrors =
-            deleteExpectingResponseErrors(Post.class, post2.getId(), 0, null);
+        List<GraphQLResponse.Error> post2DeleteErrors = deleteExpectingResponseErrors(Post.class, post2.getId(), 0);
         assertEquals("Conflict resolver rejects mutation.", post2DeleteErrors.get(0).getMessage());
 
         // Run sync on Blogs
@@ -226,6 +225,12 @@ public final class AppSyncClientInstrumentationTest {
             api.create(model, onResult, onError));
     }
 
+    @NonNull
+    private <T extends Model> ModelWithMetadata<T> update(@NonNull T model, int version)
+        throws DataStoreException {
+        return update(model, version, null);
+    }
+
     /**
      * Updates an existing item in the App Sync API, whose remote version is the expected value.
      * @param model Updated model, to persist remotely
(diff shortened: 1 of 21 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/appsync/SynchronousAppSync.java
```diff
@@ -107,11 +123,31 @@ public static SynchronousAppSync using(@NonNull AppSync appSync) {
      * @return Response data from AppSync.
      * @throws DataStoreException On failure to obtain response data
      */
+    @SuppressWarnings("LineLength")
```
jpignata, line 126: Is this necessary? Seems shorter in your revision.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/appsync/SynchronousAppSync.java
```diff
@@ -107,11 +123,31 @@ public static SynchronousAppSync using(@NonNull AppSync appSync) {
...
+     * Uses Amplify API to make a mutation which will only apply if the version sent matches the server version.
+     * @param clazz The class of the object being deleted
+     * @param objectId ID id of the object to delete
+     * @param version The version of the model we have
+     * @param predicate The condition to be applied to the delete.
+     * @param <T> The type of data in the response. Must extend Model.
+     * @return Response data from AppSync.
+     * @throws DataStoreException On failure to obtain response data
+     */
+    @SuppressWarnings("LineLength")
```
jpignata, line 143: Anyway to avoid this? (newline before the `throws`?)

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -143,16 +144,36 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api) {
...
     @Override
     public <T extends Model> Cancelable update(
             @NonNull T model,
             @NonNull Integer version,
+            @Nullable QueryPredicate predicate,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure) {
         try {
-            final String doc = AppSyncRequestFactory.buildUpdateDoc(model.getClass());
+            final String doc = AppSyncRequestFactory.buildUpdateDoc(model.getClass(), predicate != null);
```
drochetti, line 176: it feels like `buildUpdateDoc` should take the `predicate` and not a `boolean` indicating there's a predicate. 

I guess ideally `buildUpdateDocument` should return a `GraphQLDocument` or something like that, instead a `String`, the it would build the `buildDocument` would be a self-contained document factory.

I understand this can be a big undertaking for the moment, so feel free to take this feedback for a future improvement (maybe create a ticket?)

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/appsync/SynchronousAppSync.java
```diff
@@ -107,11 +123,31 @@ public static SynchronousAppSync using(@NonNull AppSync appSync) {
...
     <T extends Model> GraphQLResponse<ModelWithMetadata<T>> delete(
-            @NonNull Class<T> clazz, @NonNull String objectId, @NonNull Integer version) throws DataStoreException {
+        @NonNull Class<T> clazz, @NonNull String objectId, @NonNull Integer version) throws DataStoreException {
+        return delete(clazz, objectId, version, null);
+    }
+
+    /**
+     * Uses Amplify API to make a mutation which will only apply if the version sent matches the server version.
+     * @param clazz The class of the object being deleted
+     * @param objectId ID id of the object to delete
```
jpignata, line 136: "Id ID id" ;)

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -173,22 +197,47 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api) {
...
     public <T extends Model> Cancelable delete(
             @NonNull Class<T> clazz,
             @NonNull String objectId,
             @NonNull Integer version,
+            @Nullable QueryPredicate predicate,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure) {
         try {
-            final String doc = AppSyncRequestFactory.buildDeletionDoc(clazz);
+            final String doc = AppSyncRequestFactory.buildDeletionDoc(clazz, predicate != null);
```
jpignata, line 231: Can you extract a local here to communicate what this bool does? Alternatively, should we pass something other than a bool here to avoid the construction here on 238?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
```diff
@@ -294,7 +294,7 @@ public void existingUpdateIncomingCreationYieldsError() throws DataStoreExceptio
             .name("Tony with improvements applied")
             .build();
         PendingMutation<BlogOwner> existingUpdate =
-            PendingMutation.update(modelInExistingMutation, BlogOwner.class);
+            PendingMutation.update(modelInExistingMutation, BlogOwner.class, null);
```
drochetti, line 297: this is repeat often it seems, maybe an overload is worth it?

aws-datastore/src/test/resources/update-blog-owner-with-predicate.txt
```diff
@@ -0,0 +1,13 @@
+mutation UpdateBlogOwner($input:UpdateBlogOwnerInput!, $condition:ModelBlogOwnerConditionInput) {
+  updateBlogOwner(input:$input, condition:$condition) {
+    blog {
+      id
+    }
+    id
+    name
+    wea
```
jpignata, line 8: wea?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
```diff
@@ -691,13 +694,13 @@ public void nextItemForModelIdReturnsFirstEnqueued() throws DataStoreException {
...
             .build();
-        PendingMutation<BlogOwner> firstMutation = PendingMutation.update(originalJoe, BlogOwner.class);
+        PendingMutation<BlogOwner> firstMutation = PendingMutation.update(originalJoe, BlogOwner.class, null);
         storage.save(originalJoe, converter.toRecord(firstMutation));
 
         BlogOwner updatedJoe = originalJoe.copyOfBuilder()
             .name("Joe Swanson, MD. (He finished med school, I guess?)")
             .build();
-        PendingMutation<BlogOwner> secondMutation = PendingMutation.update(updatedJoe, BlogOwner.class);
+        PendingMutation<BlogOwner> secondMutation = PendingMutation.update(updatedJoe, BlogOwner.class, null);
```
jpignata, line 703: This is a lot of null passing. Maybe we need to support this signature sans predicate?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/appsync/SynchronousAppSync.java
```diff
@@ -93,8 +94,23 @@ public static SynchronousAppSync using(@NonNull AppSync appSync) {
...
     public <T extends Model> GraphQLResponse<ModelWithMetadata<T>> update(
         @NonNull T model, @NonNull Integer version) throws DataStoreException {
+        return update(model, version, null);
+    }
+
+    /**
+     * Uses Amplify API to make a mutation which will only apply if the version sent matches the server version.
+     * @param model An instance of the Model with the values to mutate
+     * @param version The version of the model we have
+     * @param <T> The type of data in the response. Must extend Model.
```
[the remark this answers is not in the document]
jamesonwilliams: Conventionally, the type params are usually after all other `@param`.

Checkstyle has the [`AtClause`](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocStyle) rule to compare JavaDoc `@foo` and `@bar`, but not within `@foo`s.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
```diff
@@ -249,12 +256,20 @@ public void testAllOperations() throws DataStoreException {
...
      * @throws DataStoreException If API delete call fails to render any response from AppSync endpoint
      */
+    @NonNull
+    private <T extends Model> ModelWithMetadata<T> delete(
+        @NonNull Class<T> clazz, String modelId, int version)
+        throws DataStoreException {
+        return delete(clazz, modelId, version, null);
+    }
+
     @SuppressWarnings("SameParameterValue") // Reads better with details in one place
```
jamesonwilliams, line 266: Can this be stripped, now?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/appsync/SynchronousAppSync.java
jamesonwilliams: We can swing the params down onto a new line.

My personal style is to indent +8 off the method block, which differentiates it from the method body, and the method signature.

```
@NonNull
<T extends Model> GraphQLResponse<ModelWithMetadata<T>> delete(
        @NonNull Class<T> clazz,
        @NonNull String objectId,
        @NonNull Integer version,
        @Nullable QueryPredicate predicate)
        throws DataStoreException {
    doStuffHereAtPlus4();
    coolStuffOnly();
    thatIs();
}
```

The rule is basically "Chomp down vars in a method signature, and indent by 8."

I know of _no tool_ that is able to uniquely distinguish this case, from other line-wrapping or indentation cases.

My personal style is to use +4 spaces as indentation in all other situations in a Java document.

Well,
```
final class MyClass implements
        SameTypeOfThing,
        Here {
    public void soThatYouCanTellItIsntThisMethodEasily() {
    }
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
```diff
@@ -92,11 +93,50 @@
             @NonNull Consumer<DataStoreException> onFailure
     );
 
+    /**
+     * Uses Amplify API to make a mutation which will only apply if the version sent matches the server version.
```
jamesonwilliams, line 97: "Uses Amplify API"` -- isn't true, in the interface. Even the `AppSyncClient` only needs a `GraphQLBehavior`, doesn't it? `Amplify.API` is a very specific `GraphQLBehavior`. `mock(GraphQLBehavior)` is one counter-example; this may be used to fulfill the `update` contract.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
```diff
@@ -92,11 +93,50 @@
...
+    <T extends Model> Cancelable update(
+            @NonNull T model,
+            @NonNull Integer version,
+            @Nullable QueryPredicate predicate,
+            @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
+            @NonNull Consumer<DataStoreException> onFailure
+    );
+
+    /**
+     * Uses Amplify API to make a mutation which will only apply if the version sent matches the server version.
```
jamesonwilliams, line 116: Same here

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
```diff
@@ -92,11 +93,50 @@
...
+     * @param objectId ID id of the object to delete
+     * @param version The version of the model we have
+     * @param onResponse Invoked when response data is available.
+     * @param onFailure Invoked on failure to obtain response data
+     * @param <T> The type of data in the response. Must extend Model.
+     * @return A {@link Cancelable} to provide a means to cancel the asynchronous operation
+     */
+    @NonNull
+    <T extends Model> Cancelable delete(
+        @NonNull Class<T> clazz,
```
jamesonwilliams, line 127: The other ones in this doc are using +8 off the method start

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
jamesonwilliams: ^ I agree with this.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
jamesonwilliams: Daniel and I agree we should just pass down the predicate, here. Even if it just means `predicate != null` is checked as a method-local implementation detail.

Alternately, I am always tryign to get people to do:
```
enum LabeledBoolean {
    STATE_A,
    STATE_B
}
```
with 
```
exerciseBehavior(LabeledBoolean.STATE_A);
```
It's a more legible experience to read that, than `exerciseBeahvior(false)` -- the `false` is meaningless until I read the code.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
jamesonwilliams: ^ Agree.

For these predicates, around the entire code base, we either need to:

1. Provide overloads. One does not require a predicate, the other requires a _non-null_ predicate. This makes it clear to the user what's going on, and we refuse to have our user supply a semantic `null`.

2. Don't provide an overload, just have a required param. If we do this, we need  a factory which can create a non-null sentinel, e.g.

```
QueryPredicates.matchAll() // returns something that is not `null`.
```

Then internally, we can do like:
```
if (isSentinelValue(queryPredicate)) {
    // do whatever the null meant.
}
```

in practice, that is:
```
if QueryPredicates.matchAll().equals(queryPredicate)) {
    // match all!
}
```

aws-datastore/src/test/resources/update-blog-owner-with-predicate.txt
jamesonwilliams: Heh. So, I grabbed this schema from a developer. Francisco, I think?  He was actively testing something. Anyway, our BlogOwner has a `wea` field baked into it. I don't recall where I got the schema, or what was being tested with the `wea`.

jpignata: In the jungle, the mighty jungle.. o-wimowea 
, o-wimowea.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
rjuliano: I had something like that stashed away somewhere but can't seem to find it. I'll get it changed though.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
rjuliano: Removed it.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
rjuliano: My original intent was to do just that. However, as I started going through the interactions with the API category, it looked like it was going to require a larger refactor. I'm going to create an issue so we can track this though.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
rjuliano: I addressed this by adding overloads for now.

PR #497, opened 2020-05-20
Auth Integration for Storage

TrekSoft:
Replaces Storage's direct dependence on AWSMobileClient with a provider which gets an instance of that from the Auth plugin.

For now the integration tests still directly rely on AWSMobileClient - cleared this with Jameson as being something we can address after GA.


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSAuthProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * Internal interface for providing AWS specific Auth information.
+ */
+public interface AWSAuthProvider {
+    /**
+     * Get the identity ID of the currently logged in user if they are registered in identity pools.
+     * @return the identity ID of the currently logged in user.
+     * @throws StorageException  If the proper Auth plugin isn't added or identity id is unavailable
+     */
+    String getIdentityId() throws StorageException;
+
```
raphkim, line 32: extra line

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -82,22 +80,27 @@
...
      */
     public AWSS3StoragePlugin() {
-        this(
-            (context, region, bucket) ->
-                    new AWSS3StorageService(context, region, bucket, false),
-            () -> AWSMobileClient.getInstance().getIdentityId()
-        );
+        this((context, region, bucket) ->
+                new AWSS3StorageService(context, region, bucket, new MobileClientAWSAuthProvider(), false),
+                new MobileClientAWSAuthProvider());
```
raphkim, line 85: Nit: why do both plugin and service need the reference to the auth provider? If the service is going to have access to it, then the plugin shouldn't need it right?

Also, you should keep only one reference of `MobileClientAWSAuthProvider()` rather than instantiating two separately.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/MobileClientAWSAuthProvider.java
```diff
@@ -0,0 +1,67 @@
...
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.storage.StorageException;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.mobile.client.AWSMobileClient;
+
+/**
+ * Static utility class internal to the plugin to interface with Auth.
+ */
+public final class MobileClientAWSAuthProvider implements AWSAuthProvider {
```
raphkim, line 27: Nit: `AWSMobileClientAuthProvider` instead?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
jamesonwilliams: Java's requirements that `this(...)` or `super(...)` must exist in the first line --

This is one reason I use factory methods all of the the place, cause then you can do:
```
static Gizmo create() {
    allKinds();
    ofMagical();
    thangsBefore();
    return new Gizmo(kFine, iAmReady, now);
}
```

Anyway, we are stuck with a no-args constructor as our plugin instantiation API.

To "copy" with this, you could do like:
```
class GizmoPlugin {
    GizmoPlugin() {
        this(defaultFidgetSpinner(), defaultDongler());
    }

    GizmoPlugin(FidgetSpinner spinner, Dongler dongler) {
        this(spinner, dongler, spinner, dongler); // if implementing a few needed interfaces from one component
    }
  
   GizmoPlugin(
        Fidget fidget,
        AbstractDongler abstractDongler,
        Spinner spinner,
        DonglerBase donlgerBase) {
     }
}

File: aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSAuthProvider.java [hunk 4]
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Internal interface for providing AWS specific Auth information.
+ */
+public interface AWSAuthProvider {
+    /**
+     * Get the identity ID of the currently logged in user if they are registered in identity pools.
+     * @return the identity ID of the currently logged in user.
+     * @throws StorageException  If the proper Auth plugin isn't added or identity id is unavailable
+     */
+    String getIdentityId() throws StorageException;
```
jamesonwilliams, line 31: I think the Identity ID is actually a _Cognito_ thing, not an AWS thing. So, your interface maybe should be called `CognitoIdentityProvider`.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
```diff
@@ -76,11 +81,20 @@ public void start() {
...
+
+        try {
+            serviceKey = S3Keys.createServiceKey(
+                    getRequest().getAccessLevel(),
+                    getRequest().getTargetIdentityId() != null
+                            ? getRequest().getTargetIdentityId()
+                            : awsAuthProvider.getIdentityId(),
+                    getRequest().getKey()
+            );
+        } catch (StorageException exception) {
```
jamesonwilliams, line 94: Can you scope this to only the new auth lines that throw it?

```
final String maybeVar;
try {
   maybeVar = getIt();
} catch (StorageException varNotAvailable) {
   blowUp();
}
// maybeVar is now a legit var.
```

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
TrekSoft: Yeah so I thought about both these things - I had kept them separate since the Storage Service was specifically focused on storage methods and thought it would corrupt it to have the auth methods in there as well. And since the call to the constructor has to be the first line in the method it didn't work to create a common initialization of the auth provider. 

However, since you also feel it should just be available from the service, I'll go ahead and do that.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSAuthProvider.java
TrekSoft: Fair enough

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageDownloadFileOperation.java
TrekSoft: Sure

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/MobileClientAWSAuthProvider.java
TrekSoft: Kk

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSAuthProvider.java
TrekSoft: Got it

TrekSoft:
Replaces Storage's direct dependence on AWSMobileClient with a provider which gets an instance of that from the Auth plugin.

PR #498, opened 2020-05-20
Implement conflict resolution rules for existing and incoming mutations for the same model

rjuliano:
*Issue #, if available:*

*Description of changes:*
The PR is a follow up to #496. It contains the business logic that handles conflicts when multiple mutations for the same model and enqueued. It takes into account whether a condition was included with the update/delete.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
@@ -23,6 +23,7 @@ import androidx.core.util.ObjectsCompat;
 import com.amplifyframework.AmplifyException;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.DataStoreException;
 import com.amplifyframework.datastore.storage.LocalStorageAdapter;
 import com.amplifyframework.datastore.storage.StorageItemChange;
@@ -83,19 +84,19 @@ final class PersistentMutationOutbox implements MutationOutbox {
         String modelId = incomingMutation.getMutatedItem().getId();
         @SuppressWarnings("unchecked")
         PendingMutation<T> existingMutation = (PendingMutation<T>) nextMutationForModelId(modelId);
-        if (existingMutation == null || inFlightMutations.contains(existingMutation.getMutationId())) {
-            return save(incomingMutation).andThen(notifyContentAvailable());
+        if (existingMutation == null) {
+            return save(incomingMutation)
+                .andThen(notifyContentAvailable());
+        } else {
+            return resolveConflict(existingMutation, incomingMutation);
         }
-        switch (incomingMutation.getMutationType()) {
-            case CREATE:
-                return handleCreation();
-            case UPDATE:
-                return handleUpdate(existingMutation, incomingMutation);
-            case DELETE:
-                return handleDeletion(existingMutation, incomingMutation);
-            default:
-                return unknownMutationType(incomingMutation.getMutationType());
-        }
+    }
+
+    private <T extends Model> Completable resolveConflict(@NonNull PendingMutation<T> existingMutation,
+                                                          @NonNull PendingMutation<T> incomingMutation) {
+        IncomingMutationConflictHandler<T> mutationConflictHandler =
+            new IncomingMutationConflictHandler<>(existingMutation, incomingMutation);
+        return mutationConflictHandler.resolve();
     }
 
     @VisibleForTesting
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -83,21 +84,21 @@ public boolean hasPendingMutation(@NonNull String modelId) {
...
-        switch (incomingMutation.getMutationType()) {
-            case CREATE:
-                return handleCreation();
-            case UPDATE:
-                return handleUpdate(existingMutation, incomingMutation);
-            case DELETE:
-                return handleDeletion(existingMutation, incomingMutation);
-            default:
-                return unknownMutationType(incomingMutation.getMutationType());
+        if (existingMutation == null) {
```
TrekSoft, line 87: Did you mean to remove the inFlightMutations condition? I thought that was the piece Jameson just added.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -83,21 +84,21 @@ public boolean hasPendingMutation(@NonNull String modelId) {
...
+            return save(incomingMutation)
+                .andThen(notifyContentAvailable());
+        } else {
+            return resolveConflict(existingMutation, incomingMutation);
```
TrekSoft, line 91: It looks like before, it would always handle the mutation type but now you're only handling it if the existingMutation is null so just wanted to check that's intended behavior.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -318,4 +264,158 @@ public Completable markInFlight(@NonNull TimeBasedUuid pendingMutationId) {
...
+            return Completable.error(new DataStoreException(
+                "Attempted to enqueue a model creation, but there is already a pending creation for that model ID.",
+                "Please report at https://github.com/aws-amplify/amplify-android/issues."
+            ));
+        }
+
+        private Completable modelAlreadyScheduledForDeletion() {
+            return Completable.error(new DataStoreException(
+                "Attempted to enqueue a model mutation, but that model already had a delete mutation pending.",
+                "This should not be possible. Please report on GitHub issues."
```
TrekSoft, line 402: Nit: Update suggestion to match others with a link - Jameson actually recently added a constant string to Amplify Exception for this type of suggestion you can use.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
rjuliano: I had moved that check down to one of the methods that handles conflicts. However, I think you're correct. It probably makes more sense here.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
rjuliano: I encapsulated that logic in the `IncomingMutationConflictHandler`. That class takes in both the existing and incoming mutation and applies logic similar to how iOS and JS are doing right now...at east that was the goal :)

The entry point for that is [right here](https://github.com/aws-amplify/amplify-android/pull/498/files#diff-7fc8fd67eed2c76db4c431af072805ccR293)

TrekSoft:
Sounds good - re-approved

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
jamesonwilliams: Hm, it wasn't me, actually! I thought it was you. [It was @drochetti](https://github.com/aws-amplify/amplify-android/blame/d4e60a992eee495dab09ffc0a02bfa77776b1c1d/core/src/main/java/com/amplifyframework/AmplifyException.java#L33-L41)!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -84,20 +85,20 @@ public boolean hasPendingMutation(@NonNull String modelId) {
...
+                .andThen(notifyContentAvailable());
+        } else {
+            return resolveConflict(existingMutation, incomingMutation);
         }
     }
 
+    private <T extends Model> Completable resolveConflict(@NonNull PendingMutation<T> existingMutation,
+                                                          @NonNull PendingMutation<T> incomingMutation) {
+        IncomingMutationConflictHandler<T> mutationConflictHandler =
+            new IncomingMutationConflictHandler<>(existingMutation, incomingMutation);
```
jamesonwilliams, line 98: Should we be creating a new one every time, like this?

PR #502, opened 2020-05-21
Subscription authorizer for API

raphkim:
*Issue #, if available:*

*Description of changes:*
~~**Currently not working**~~
~~Ideally, `SubscriptionWithAwsIamTest` (identical to `GraphQLInstrumentationTest` except for the API being used) should pass, but it does not.~~

**WORKS NOW**


core/src/main/java/com/amplifyframework/api/graphql/GraphQLRequest.java
```diff
@@ -69,6 +71,22 @@ public GraphQLRequest(
...
+     */
+    public String getDocument() {
+        return document;
+    }
+
+    /**
+     * Returns the immutable map of request variables.
+     * @return the map of variables
+     */
+    public Map<String, Object> getVariables() {
```
jamesonwilliams, line 86: Ah heck. I was gonna say, "remember to add this to the `hashCode()`/`equals()`/`toString()`!" But it looks like we don't even have those in this class. Which, given how prominent it is in the API signatures, is bad news.

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionAuthorizerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    private ApiAuthProviders apiAuthProviders;
+
+    /**
+     * Construct fake auth providers to override default behaviors for
+     * subscription authorizer.
+     */
+    @Before
+    public void setup() {
+        authenticationSecret = RandomString.string();
+        apiAuthProviders = ApiAuthProviders.builder()
```
jamesonwilliams, line 50: I've never liked this `ApiAuthProviders` thing, it has poor encapsulation, and muddles together several different technologies without a consistent behavioral contract.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionAuthorizer.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private URI getRequestEndpoint(boolean connectionFlag) throws ApiException {
+        try {
+            String baseUrl = configuration.getEndpoint();
+            String connectionUrl = connectionFlag ? baseUrl + "/connect" : baseUrl;
+            return new URI(connectionUrl);
+        } catch (URISyntaxException uriException) {
+            throw new ApiException(
+                    "Error constructing canonical URI for IAM request signature",
+                    uriException, AmplifyException.TODO_RECOVERY_SUGGESTION
```
jamesonwilliams, line 198: Use the `REPORT_TO_AWS`-or-whatever one that @drochetti added

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionAuthorizer.java
```diff
@@ -0,0 +1,219 @@
...
+                credentialsProvider.getCredentials()
+        );
+
+        // Extract header from signed request and return
+        Map<String, String> signedHeaders = canonicalRequest.getHeaders();
+        JSONObject authorization = new JSONObject();
+        try {
+            for (Map.Entry<String, String> headerEntry : signedHeaders.entrySet()) {
+                authorization.put(headerEntry.getKey(), headerEntry.getValue());
+            }
```
jamesonwilliams, line 176: Just, `authorization.putAll(signedHeaders()))`?

aws-api/src/androidTest/java/com/amplifyframework/api/aws/SubscriptionWithAwsIamTest.java
```diff
@@ -0,0 +1,232 @@
+/*
```
jamesonwilliams, line 1: How can you move this to `src/test`?

drochetti:
> We should get rid of the `ApiAuthProviders`.
> 
> One way to do this is to have a different OkHttp `Interceptor` implementation for each type. Each implementation would know how to append headers into the request. This way, we'd have a standard behavioral contract for how auth gets added.
> 
> I guess there's that one weird detail around IAM and changing the endpoint. I'm not sure if you can rewrite the URL using an interceptor, but possibly.
> 

That's a good call and that's how we handle on iOS. The interceptor should be able to change anything in the request object given the request was not submitted yet.

One approach could be to pass a mutable builder around to the interceptors and once the interceptor chain is exhausted `build` is called to get an immutable request object

core/src/main/java/com/amplifyframework/api/graphql/GraphQLRequest.java
raphkim: I removed `getVariables` and `getDocument` again after I realized that it isn't needed anymore. but I agree it still needs those methods you brought up.

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionAuthorizerTest.java
raphkim: Yea it was a poor implementation on my part back when I was working on it for mutate/query, but it's a legacy that should stay for until after GA...

aws-api/src/androidTest/java/com/amplifyframework/api/aws/SubscriptionWithAwsIamTest.java
raphkim: This test was removed for now

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionAuthorizer.java
raphkim: `JSONObject` doesn't have `putAll()` method, but I realized that I can just directly construct one with a map, so I'll change it.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionAuthorizer.java
raphkim: Didn't know we had that! I'll change it, and add appropriate recovery suggestion everywhere else.

PR #503, opened 2020-05-21
Auth integration with Analytics

TrekSoft:
Integrates Auth into analytics following the pattern agreed on.

AWSMobileClient is still used directly for integration tests for now but will be replaced as a fast follow hence why I didn't create a special class to handle initializing it like was done in other categories.

### DIFF
--- a/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AWSPinpointAnalyticsPlugin.java
+++ b/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AWSPinpointAnalyticsPlugin.java
@@ -20,6 +20,7 @@ import android.content.Context;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringDef;
+import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AWSPinpointAnalyticsPlugin.java
```diff
@@ -64,22 +68,30 @@
...
 
     private final Application application;
     private AutoEventSubmitter autoEventSubmitter;
     private AnalyticsClient analyticsClient;
     private AutoSessionTracker autoSessionTracker;
     private TargetingClient targetingClient;
+    private AWSCredentialsProvider credentialsProviderOverride; // Currently used for integration testing purposes
 
     /**
      * Constructs a new AmazonPinpointAnalyticsPlugin.
```
jpignata, line 81: Ah! Can you fix this and make it a reference?

TrekSoft: Done

PR #505, opened 2020-05-21
(feat) amplify-tools as a separate project

nikhname:
*Issue #, if available:*

*Description of changes:*
Adds the amplify-tools plugin as a separate project/dir in the amplify-android repo
Local publish can be tested by running ```./gradlew publishToMavenLocal``` from the amplify-tools dir
Bumped to version 1.0.0
Includes the changes requested here: https://github.com/aws-amplify/amplify-android/pull/452/


amplify-tools/.gitignore
```diff
@@ -0,0 +1,9 @@
+*.iml
```
jpignata, line 1: This is a subset of the main `.gitignore`, kill it?

amplify-tools/amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+}
+
+publishing {
+    publications {
+        pluginPublication(MavenPublication) {
+            from components.java
+            groupId 'com.amplifyframework'
+            artifactId 'amplify-tools-gradle-plugin'
+            version VERSION_NAME + "-SNAPSHOT"
```
jpignata, line 35: Sup with the "-SNAPSHOT?"

amplify-tools/amplify-tools-gradle-plugin/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=amplify-tools-gradle-plugin
+POM_NAME=Amplify Tools for Android - Gradle Plugin
```
jpignata, line 2: Lowercase t tool, please.

amplify-tools/amplify-tools-gradle-plugin/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
jpignata, line 1: Is this needed?

amplify-tools/gradle.properties
```diff
@@ -0,0 +1,36 @@
...
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app's APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
+android.useAndroidX=true
+# Automatically convert third-party libraries to use AndroidX
+android.enableJetifier=true
+
+GROUP=com.amplifyframework
+VERSION_NAME=1.0.0
+AWS_CORE_SDK_VERSION=2.16.3
```
jpignata, line 23: What is this?

amplify-tools/amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+apply from: rootProject.file('gradle-mvn-push.gradle')
+
+sourceSets.main.groovy.srcDirs = ["src/main/groovy"]
+
+dependencies {
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+}
+
+publishing {
```
jamesonwilliams, line 29: We have a default `publishing` rule available int `configuration/publishing.gradle`. If you import that file, at the top of this one, then you don't need a separate block, here.

The other modules use `gradle.properties` to define Maven publishing fields, and then include that central publishing script, which is a more elaborate version of this block, here.

-------

Update: oh, right, this is a separate _project_. Never-mind!

amplify-tools/amplify-tools-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+            from components.java
+            groupId 'com.amplifyframework'
+            artifactId 'amplify-tools-gradle-plugin'
+            version VERSION_NAME + "-SNAPSHOT"
+        }
+    }
+}
+
+sourceCompatibility = "1.7"
+targetCompatibility = "1.7"
```
jamesonwilliams, line 41: Hm, Groovy, compatible with Java 1.7? Or does "1.7" mean version 51 byte code? Are these needed?

amplify-tools/amplify-tools-gradle-plugin/proguard-rules.pro
jamesonwilliams: Doubtful, for a plugin. Since the plugin will only be deployed to a developer's machine, and not a customer's mobile phone. _If_ we succeed to integrate Proguard here, it will purely hamper our ability to debug the plugin, while providing no benefit.

TLDR we can safely remove this file.

amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/amplifytools.groovy
```diff
@@ -0,0 +1,246 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.apache.tools.ant.taskdefs.condition.Os
+import groovy.json.JsonSlurper
+import groovy.json.JsonOutput
+
+class amplifytools implements Plugin<Project> {
```
jamesonwilliams, line 7: This is a weird name for a groovy class. Does it match some other token, somewhere else? Is this intentional? I would expect this to be in Pascal case, like `Plugin<Project>` is. `AmplifyToolsPlugin implements Plugin<Project>`, something like that?

amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/amplifytools.groovy
```diff
@@ -0,0 +1,246 @@
...
+        def accessKeyId = null
+        def secretAccessKey = null
+        def region = null
+        def envName = 'amplify'
+        def syncEnabled = 'true'
+
+        project.task('verifyNode') {
+            try {
+                project.exec {
+                    commandLine 'node', '-v'
```
jamesonwilliams, line 23: You actually run `npx`, not `node`, later. So, it'd be better to check if `npx` is available.

When I used this plugin, `npx` _wasn't_ available.

I do have `node` installed, but that's only _part_ of the requirement.

amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/amplifytools.groovy
```diff
@@ -0,0 +1,246 @@
...
+            } catch (commandLineFailure) {
+                doesNodeExist = false
+                println("Node is not installed. Visit https://nodejs.org/en/download/ to install it")
+            }
+        }
+
+        project.task('createAmplifyApp') {
+            doesGradleConfigExist = project.file('amplify-gradle-config.json').isFile()
+            if (doesNodeExist && !doesGradleConfigExist) {
+                if(Os.isFamily(Os.FAMILY_WINDOWS)) {
```
jamesonwilliams, line 35: Nit: in all places, substitute `if(` with `if (`.

amplify-tools/build.gradle
```diff
@@ -0,0 +1,27 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
```
jamesonwilliams, line 2: Kill this

amplify-tools/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    repositories {
+        google()
+        jcenter()
+        
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.6.3'
+        
+        // NOTE: Do not place your application dependencies here; they belong
+        // in the individual module build.gradle files
```
jamesonwilliams, line 13: Kill this

amplify-tools/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+
+allprojects {
+    repositories {
+        google()
+        jcenter()
+
```
jamesonwilliams, line 21: kill this empty line

amplify-tools/gradle-mvn-push.gradle
```diff
@@ -0,0 +1,233 @@
+/*
+ * Copyright 2013 Chris Banes
```
jamesonwilliams, line 2: It is a little troubling to watch a second copy of this script come into the git repository. Is this identically the same as the `configuration/publishing.gradle` in the main project?

I don't think there's anything ~reasonable we can do to prevent it, unfortunately. 

We _can_ make sure that the files are identical in contents and name, thought.

This one _at least_ has a different name.

amplify-tools/gradle-mvn-push.gradle
```diff
@@ -0,0 +1,233 @@
...
+            from sourceSets.main.allSource
+        }
+
+        task javadocJar(type: Jar, dependsOn: javadoc) {
+            classifier = 'javadoc'
+            from javadoc.destinationDir
+        }
+    }
+
+    if (JavaVersion.current().isJava8Compatible()) {
```
jamesonwilliams, line 215: Ah yea, it not only has a different file name -- but the contents of this script is different from `configuration/publishing.gradle`, too.

amplify-tools/gradle.properties
```diff
@@ -0,0 +1,36 @@
...
+# The setting is particularly useful for tweaking memory settings.
+org.gradle.jvmargs=-Xmx1536m
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app's APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
+android.useAndroidX=true
```
jamesonwilliams, line 17: I don't think your plugin uses Android X. You can remove the lines about AndroidX and Jettifier.

amplify-tools/gradle.properties
```diff
@@ -0,0 +1,36 @@
...
+android.enableJetifier=true
+
+GROUP=com.amplifyframework
+VERSION_NAME=1.0.0
+AWS_CORE_SDK_VERSION=2.16.3
+
+POM_URL=https://github.com/aws-amplify/amplify-cli/
+POM_SCM_URL=https://github.com/aws-amplify/amplify-cli/
+POM_SCM_CONNECTION=scm:git:git://github.com/aws-amplify/amplify-cli.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/aws-amplify/amplify-cli.git
```
jamesonwilliams, line 28: Update these so that they are correct and refer to the current state of the project in `amplify-android`.

amplify-tools/gradle-mvn-push.gradle
jpignata: We *could* symlink it, no?

amplify-tools/gradle-mvn-push.gradle
jpignata: Oh, nevermind.

amplify-tools/amplify-tools-gradle-plugin/build.gradle
jpignata: you're not thinking fourth dimensionally

amplify-tools/amplify-tools-gradle-plugin/build.gradle
nikhname: It means the plugin is compiled with Java 7, its necessary so that we can fix down the plugin compatibility (

amplify-tools/amplify-tools-gradle-plugin/build.gradle
nikhname: Removing

amplify-tools/gradle.properties
nikhname: Removing

amplify-tools/gradle-mvn-push.gradle
jamesonwilliams: Yea, I guess I like that better. It's ~a little weird, but it solves the problem.

amplify-tools/gradle-mvn-push.gradle
jamesonwilliams: ^ I don't think it _needs_ to be. It just currently is. I suspect the differences are not intentional?

amplify-tools/gradle-mvn-push.gradle
nikhname: We would be able to use the ```configuration/publishing.gradle``` if not for the publishing block that seems to require an aar artifact. I think it makes sense to make the file identical in name and content except for that block

PR #509, opened 2020-05-22
Add AmplifyDisposables utility to each modules using Rx

jamesonwilliams:
`AmplifyDisposables` is a factory class that generates Rx `Disposable` from
an input Amplify `Cancelable`.

Three copies of this file are added. Unfortunately this is necessary
since `testutils`, `rxbinings`, and `aws-datastore` only share `core` as a
common dependency module. However, `core` does not take a dependency on
Rx, by design. So, this class could not compile, in `core`, nor is `core` an ideal
place for miscellaneous utility code.

### DIFF
--- a/rxbindings/src/main/java/com/amplifyframework/rx/RxHubBinding.java
+++ b/rxbindings/src/main/java/com/amplifyframework/rx/RxHubBinding.java
@@ -27,7 +27,6 @@ import com.amplifyframework.hub.SubscriptionToken;
 
 import io.reactivex.Completable;
 import io.reactivex.Observable;
-import io.reactivex.disposables.CompositeDisposable;
 import io.reactivex.disposables.Disposables;
 
 final class RxHubBinding implements RxHubCategoryBehavior {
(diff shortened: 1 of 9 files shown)


TrekSoft:
This sounds like the type of thing which would be useful to have in our common AWS dependencies module (or perhaps we split them up into targeted types of modules - e.g. common AWS DataStore functionality) whenever we plan that after GA.

rxbindings/src/main/java/com/amplifyframework/rx/RxHubBinding.java
```diff
@@ -53,13 +53,8 @@
...
     public Observable<HubEvent<?>> on(@NonNull HubChannel hubChannel) {
         return Observable.defer(() -> Observable.create(emitter -> {
-            final CompositeDisposable disposable = new CompositeDisposable();
-            emitter.setDisposable(disposable);
             SubscriptionToken token = hub.subscribe(hubChannel, emitter::onNext);
-            disposable.add(Disposables.fromAction(() -> {
-                hub.unsubscribe(token);
-                emitter.onComplete();
-            }));
+            emitter.setDisposable(Disposables.fromAction(() -> hub.unsubscribe(token)));
```
TrekSoft, line 57: Maybe I'm missing something but is this the type of thing you'd want to have as another method in your AmplifyDisposables factory?

jamesonwilliams: Ah, good idea. `Disposables.fromHubSubscription(hub, token)` ? Something like that?

PR #514, opened 2020-05-22
refactor(api): clean up model-based apis

drochetti:
**Notes:**

Following the decision of not having the model-based calls as part of the API Category contract, I moved them out and created semantic factories to build the `GraphQLRequest` from the models, so the functionality is preserved.

De-coupling the Model from the API category will allow us to keep a minimal and generic GraphQL client that will be open for integration with other GraphQL libraries that generate code for GraphQL backends, such as AppSync, Apollo, Prisma, etc.

**Examples:**

Mutation of type create:

```java
// previous
Amplify.API.mutate(post, MutationType.CREATE, /* result/error lambdas */);

// new (mimics the underlying GraphQL statement: createPost)
Amplify.API.mutate(ModelMutation.create(post), /* result/error lambdas */);
// or with static imports
Amplify.API.mutate(create(post), /* result/error lambdas */);
```

Queries:

```java
// previous
Amplify.API.query(Post.class, /* result/error lambdas */);

// new  (mimics the underlying GraphQL statement: listPosts)
Amplify.API.query(ModelQuery.list(Post.class), /* result/error lambdas */);
// or with static imports
Amplify.API.query(list(Post.class), /* result/error lambdas */);

// ---

// previous
Amplify.API.query(Post.class, "id", /* result/error lambdas */);

// new  (mimics the underlying GraphQL statement: getPost)
Amplify.API.query(ModelQuery.get(Post.class, "id"), /* result/error lambdas */);
// or with static imports
Amplify.API.query(get(Post.class, "id"), /* result/error lambdas */);
```

See the diff for more examples.


aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -93,18 +93,17 @@ private AppSyncGraphQLRequestFactory() { }
                     new GsonVariablesSerializer()
             );
         } catch (AmplifyException exception) {
-            throw new ApiException(
+            throw new InvalidGraphQLRequestRuntimeException(
```
TrekSoft, line 96: Change to default system runtime exception - such as InvalidParameter

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -468,4 +458,17 @@ private static String getModelFields(Class<? extends Model> clazz, int levelsDee
 
         return result.toString();
     }
+
+    static final class InvalidGraphQLRequestRuntimeException extends RuntimeException {
+
+        private static final long serialVersionUID = 1L;
+
+        InvalidGraphQLRequestRuntimeException(final String message, final Exception rootCause) {
+            super(message + "\n" + AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION, rootCause);
```
TrekSoft, line 467: Don't use Report bug for these types

aws-api/src/main/java/com/amplifyframework/api/aws/ModelPagination.java
```diff
@@ -0,0 +1,77 @@
...
+    public static ModelPagination firstPage() {
+        return nextToken(null);
+    }
+
+    /**
+     * TODO.
+     * @param nextToken todo.
+     * @return todo.
+     */
+    public static ModelPagination nextToken(@Nullable String nextToken) {
```
TrekSoft, line 48: Unused param

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
```diff
@@ -121,6 +121,7 @@ public void syncUpToCloudIsWorking() throws AmplifyException {
         publishedMutationsAccumulator.await();
 
         // Try to get Charley from the backend.
+//        BlogOwner remoteCharley = api.get(BlogOwner.class, localCharley.getId());
```
TrekSoft, line 124: Remove comment

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -50,134 +46,44 @@
...
-    @NonNull
-    @Override
-    public <T extends Model> Single<GraphQLResponse<T>> mutate(
-            @NonNull T model, @NonNull QueryPredicate mutationCriteria, @NonNull MutationType mutationType) {
-        return toSingle((onResult, onError) -> api.mutate(model, mutationCriteria, mutationType, onResult, onError));
-    }
-
     @NonNull
     @Override
     public <T> Single<GraphQLResponse<T>> mutate(@NonNull GraphQLRequest<T> graphQlRequest) {
```
richardmcclellan, line 63: Not important for this PR, but this should be R, not T.   This was something I missed in my PR.  We can fix it later without it being a breaking change though.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
jamesonwilliams: Yea, the `AmplifyException` has the strongly modeled recovery exception. I don't think we should inline the recovery suggestions into causes. I'm imagining that this `InvalidGraphQLRequestRuntimeException` gets composed into an `ApiException`, before being returned to the customer. If that's so, then we'd have _two_ recovery suggestions. This one, then the one in `ApiException`.

aws-api/src/main/java/com/amplifyframework/api/aws/ModelPagination.java
jamesonwilliams: It's a public API, the implementation needs to be updated after release to use the token. At least, I think that's whats being stated here, right, @drochetti?

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -145,7 +143,7 @@ public void graphQlQueryRendersValidResponse() throws ApiException {
 
         GraphQLResponse<Iterable<BlogOwner>> actualResponse =
             Await.<GraphQLResponse<Iterable<BlogOwner>>, ApiException>result(((onResult, onError) ->
-                plugin.query(BlogOwner.class, onResult, onError)
+                plugin.query(ModelQuery.list(BlogOwner.class), onResult, onError)
```
jamesonwilliams, line 146: `ModelQuery` is a good name, based on what the class does.

However, when a customer uses it, they might prefer to write like:

```
api.query(Where.modelClass(BlogOwner.class), ..., ...)
```

 or something like what we did for the `QueryOptions`.

Maybe just,

```
api.query(Models.ofType(BlogOwner.class)
```

That way we at least have a separate `Models.java` to put factory methods into, that know about `<T extends Model>` (I forget where it landed, but `Where` _doesn't_ know about `<T extends Model>`, right?

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -159,7 +157,7 @@ public void graphQlQueryRendersValidResponse() throws ApiException {
 
     /**
      * It should be possible to perform a successful call to
-     * {@link AWSApiPlugin#mutate(Model, MutationType, Consumer, Consumer)}.
+     * {@link AWSApiPlugin#mutate(com.amplifyframework.api.graphql.GraphQLRequest, Consumer, Consumer)}.
```
jamesonwilliams, line 160: Can you remove the full qualification here, by adding an `import` for it?

rxbindings/build.gradle
```diff
@@ -30,4 +30,5 @@ dependencies {
         // https://github.com/robolectric/robolectric/issues/5245
         exclude group: 'com.google.auto.service', module: 'auto-service'
     }
+    testImplementation project(path: ':aws-api')
```
jamesonwilliams, line 33: Let's just kill all the tests that use models, or change model to string. The rxbinding (even tests) should only know about the `core` module, not any plugin.

aws-api/src/main/java/com/amplifyframework/api/aws/ModelPagination.java
drochetti: That's correct. It was a mistake, already fixed.

rxbindings/build.gradle
drochetti: I'm afraid this has a big impact in how tests were written initially. I would be nervous to make that change given the size of this PR.

In the future I assume we would be able to change this dependency into the shared module one, like we do in iOS.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
drochetti: I think there's a difference here between that solution in DataStore and this one in API. Here we're not trying to hide the fact that it's GraphQL behind the scenes. Actually `Amplify.API.mutate()` and `Amplify.API.query()` makes that very explicit.

Therefore, the factory methods here resemble the GraphQL statements:

`ModelQuery.get(Todo.class)` is `query getTodo()`
`ModelMutation.create(todo)` is `mutation createTodo({})`
etc

With static imports the syntax becomes even closer:

```java
Amplify.API.query(get(Todo.class)); // query getTodo()
Amplify.API.mutate(update(todo)); // mutation updateTodo()
```

The PR description in the iOS repo also gives an idea of the experience there and how this Android builders were design to follow that same path: https://github.com/aws-amplify/amplify-ios/pull/470

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -468,4 +458,17 @@ private static String getModelFields(Class<? extends Model> clazz, int levelsDee
 
         return result.toString();
     }
+
+    static final class InvalidGraphQLRequestRuntimeException extends RuntimeException {
```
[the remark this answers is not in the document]
drochetti: I changed this to use `IllegalStateException`. I talked to @TrekSoft about this offline and the idea here is that those methods should only throw errors in case there's a problem with the ode during development time. Only a faulty `ModelSchema` will result in that code throwing an error which means either the CLI generated something wrong, that although it compiles, our code is not prepare to handle (a bug) or customers changed their models manually and produced something invalid (highly unlikely, can't think of a real scenario here).

Therefore, I think it's fair to have a `RuntimeException` here, no app user input can really result in a failure, at least not in this layer of the logic.

rxbindings/build.gradle
drochetti: spoke offline and I was able to understand what Jameson suggested. This was fixed.

rxbindings/src/test/java/com/amplifyframework/rx/RxApiBindingTest.java
```diff
@@ -381,4 +378,16 @@ public void httpPostEmitsResult() {
         verify(delegate)
             .post(eq(options), anyConsumer(), anyConsumer());
     }
+
+    private static <R> GraphQLRequest<R> createMockMutationRequest(Class<R> responseType) {
+        return new GraphQLRequest<R>("", responseType, null);
+    }
+
+    private static <R> GraphQLRequest<Iterable<R>> createMockListRequest(Class<R> responseType) {
```
richardmcclellan, line 386: Hmm..

R represents the `data` on a `GraphQLResponse`
T represents the `Model` type (or actually it looks like you've renamed most of these to `M` in this PR, which makes more sense than T anyway.

So, this should either return a `GraphQLRequest<R>` or a `GraphQLRequest<Iterable<T>>`, NOT a `GraphQLRequest<Iterable<R>>`

drochetti: updated everything here to `T`

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -160,20 +186,29 @@ private AppSyncGraphQLRequestFactory() { }
...
+            throw new IllegalStateException(
                     "Could not generate a schema for the specified class",
-                    exception,
-                    "Check the included exception for more details"
+                    exception
             );
         }
     }
 
+    /**
```
jamesonwilliams, line 196: Instead of all of these, you can do:

```
@SuppressWarnings("checkstyle:MissingJavadocMethod")
```

On top of the class definition. This will disable the check for the entire class.

Of rouse, better would be to write the docs. But, the suppression is still better than the not-filled-in stuff, I think.

drochetti: I added the docs. Most of them are very similar to the ones in `ModelQuery`/`ModelMutation`/`ModelSubscription`

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -65,13 +65,16 @@
...
     /**
-     * TODO.
-     * @param modelClass todo.
-     * @param objectId todo.
-     * @param <R> todo.
-     * @param <T> todo.
-     * @return todo.
-     * @throws IllegalStateException todo.
+     * Creates a {@link GraphQLRequest} that represents a query that expects a single value as a result.
+     * The request will be created with the correct correct document based on the model schema and
```
jamesonwilliams, line 69: _correct correct_ => _correct_

testutils/build.gradle
```diff
@@ -22,4 +22,8 @@ dependencies {
     implementation dependency.junit
     implementation dependency.androidx.test.core
     implementation dependency.rxjava
+
+    // dependency on Model/GraphQL integration classes
+    // remove when modules are re-organized to provide better isolation
+    compileOnly project(path: ':aws-api')
```
jamesonwilliams, line 28: Ah! It's over _here_, now. :-D This doesn't bother me as much though. Being a purist about our dependency hierarchy in test code is less important. 

Not for this PR --

Probably, though, we should migrate `SyncrhonousApi` so that it only contains methods that resemble `ApiCategoryBehavior`, sans-callbacks. 

Like:
```
void asyncFoo(Bar bar, Consumer<Ok> onOk, Conusmer<NotOk> onNotOk);
```
-->
```
Ok syncFoo(Bar bar) throws NotOk
```

drochetti: >Probably, though, we should migrate SyncrhonousApi so that it only contains methods that resemble ApiCategoryBehavior, sans-callbacks.

I like that idea! Make it implement the `ApiCategoryBehavior` so it adheres to the contract.

jpignata:
@drochetti How do you feel about merging this? I'd like to start getting @lawmicha's doc PR updated.

PR #517, opened 2020-05-22
Add equals, hashcode to VariablesSerializers

richardmcclellan:
`GraphQLRequest` already implements `equals`, `hashcode`, but it has a VariablesSerializer property which does not.  This adds `equals`, `hashcode` to those classes, so that `GraphQLRequest` should now work properly when checking for equality.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
@@ -46,11 +46,28 @@ final class GsonVariablesSerializer implements GraphQLRequest.VariablesSerialize
         return gson.toJson(variables);
     }
 
-    /**
-     * Serializer of {@link Temporal.Date}, an extended ISO-8601 Date string, with an optional timezone offset.
-     *
-     * https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html
-     */
+    @Override
+    public boolean equals(Object thatObject) {
+        if (this == thatObject) {
+            return true;
+        }
+        if (thatObject == null || getClass() != thatObject.getClass()) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        return 0;
+    }
+
+        /**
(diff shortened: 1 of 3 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
```diff
@@ -46,11 +46,28 @@ public String serialize(Map<String, Object> variables) {
...
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        return 0;
+    }
+
+        /**
+         * Serializer of {@link Temporal.Date}, an extended ISO-8601 Date string, with an optional timezone offset.
```
jpignata, line 67: Indentation?

aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
```diff
@@ -46,11 +46,28 @@ public String serialize(Map<String, Object> variables) {
...
-     * Serializer of {@link Temporal.Date}, an extended ISO-8601 Date string, with an optional timezone offset.
-     *
-     * https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html
-     */
+    @Override
+    public boolean equals(Object thatObject) {
+        if (this == thatObject) {
+            return true;
+        }
+        if (thatObject == null || getClass() != thatObject.getClass()) {
```
jpignata, line 54: Hmm, could this be simplified to eliminate the fallback `true`? Maybe something like:

```java
if (thatObject == null) {
  return false
}

return this == thatObject || getClass() == thatObject.getClass()
```

richardmcclellan: Logically, yes.

But, I think it's better to keep as is because:
1. It matches the rest of the codebase
2. It's what the IDE generates if you ask it to add equals/hashcode.

aws-api/src/main/java/com/amplifyframework/api/aws/GsonVariablesSerializer.java
richardmcclellan: Fixed!

PR #522, opened 2020-05-24
Fix Windows support in IDE integration

jpignata:
**The reformat changed every line, so please review the individual commits using [this view](https://github.com/aws-amplify/amplify-android/pull/522/files/f85ca6ea394a824d8123c65b8e3e637ccbb008de..111230d40175f92358a9556586b99a906a2972c2)**

*Issue #, if available:* n/a

*Description of changes:* Fix Windows support in IDE integration

None of this was working. It does now.

- Android Studio Reformat (⌥⌘-L)
- Use master instead of 1.0.0

### DIFF
--- a/amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
+++ b/amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
@@ -6,9 +6,6 @@ import groovy.json.JsonOutput
 
 class AmplifyTools implements Plugin<Project> {
     void apply(Project project) {
-        def doesNodeExist = true
-        def doesGradleConfigExist
-
         // profile name can be changed in amplify-gradle-config
         def profile = 'default'
         def accessKeyId = null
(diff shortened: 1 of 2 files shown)


amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -5,228 +5,222 @@ import groovy.json.JsonSlurper
...
+        def region = null
+        def envName = 'amplify'
+        def syncEnabled = 'true'
+        def gradleConfigFileName = 'amplify-gradle-config.json'
+
+        project.task('createAmplifyApp') {
+            def npx = 'npx'
+
+            if (project.file(gradleConfigFileName).exists()) {
+                return
```
jamesonwilliams, line 22: Should we update this file, across multiple plugin runs, in case the project config has changed in between? Either way, is it warranted to log an Info for this condition?

jpignata: I don't think we want to run it each time. Since this runs `amplify-app`, it could conceivably clobber other state. I'm also slightly skeptical on logging as I think this will be noise as first time users won't encounter this state. I think!

amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -5,228 +5,230 @@ import groovy.json.JsonSlurper
...
-
-		project.task('verifyNode') {
-			try {
-				project.exec {
-					commandLine 'npx', '-v'
-					standardOutput = new ByteArrayOutputStream()
-				}
-			} catch (commandLineFailure) {
-				doesNodeExist = false
-				println("Node is not installed. Visit https://nodejs.org/en/download/ to install it")
```
nikhname, line 28: Maybe we should say the npx was not found instead since as node could be installed without npx

amplify-tools/amplify-tools-gradle-plugin/src/main/groovy/com/amplifyframework/tools/gradle/plugin/AmplifyTools.groovy
```diff
@@ -5,228 +5,230 @@ import groovy.json.JsonSlurper
...
+            if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+                npx += '.cmd'
+            }
+
+            try {
+                project.exec {
+                    commandLine npx, 'amplify-app', '--platform', 'android'
+                }
+            } catch (commandLineFailure) {
+                throw new Exception('Node.js is not installed. Visit https://nodejs.org/en/download/ to install it.')
```
nikhname, line 34: Maybe we should say that npx was not found instead since node could be installed without npx

jpignata: Hm, I don't agree, because most Android developers don't know what npx is. We have a better shot if we say node.js at comprehension.

jamesonwilliams: Does npx come bundled with node? I thought my laptop had been in a state at one point where I had `node`, but not `npx`?

jpignata: npx comes bundled with npm (after version 5.something), which comes bundled with node. In Node 10+, all users who have node installed should also have npx installed. I tested this on a new system with Windows, installed Node 12 installed, and have npx available.

Issue #532, opened 2020-05-26
Application name occlusions with DataStore and custom Application subclass


jamesonwilliams:
I am following the Project Setup guide for Android, in the docs.

I subclass the `Application` class, and initialize Amplify, there. I name this subclass as my application name in `AndroidManifest.xml`:
```
<application
    android:name=".MyAmplifyApp"
    ...
</application>
```
_Additionally_, I include several modules in my application's `build.gradle`, particularly the _DataStore_:
```
implementation 'com.amplifyframework:core:master'
implementation 'com.amplifyframework:aws-datastore:master'
implementation 'com.amplifyframework:aws-api:master'
```

When I do this, and go to build the app, I get:
```
Manifest merger failed : Attribute application@name value=(com.amplifyframework.myamplifyapp.MyAmplifyApp) from AndroidManifest.xml:12:9-37
	is also present at [com.amplifyframework:aws-datastore:master] AndroidManifest.xml:25:18-70 value=(androidx.multidex.MultiDexApplication).
	Suggestion: add 'tools:replace="android:name"' to <application> element at AndroidManifest.xml:5:5-20:19 to override.
```

I can do as the suggestion says, and override the application name, with my own. But then, will I have multi-dex support < API 21? TODO: research this. Also, would like to find a repair to Amplify, so that the app doesn't need to use the override, at all.

jamesonwilliams:
Closing this as the issue has gone a stale issue of low consequence.

PR #533, opened 2020-05-26
Move MultiDexApplication declaration from main to androidTest

richardmcclellan:
Potential solution to https://github.com/aws-amplify/amplify-android/issues/532

This does eliminate the need for clients to add `'tools:replace="android:name"'` to the `<application>` element in their AndroidManifest.xml.  

I'm not sure if this breaks the integration tests though, because our integration tests have been failing generally for about a week now.


richardmcclellan:
Hmm, this is actually now causing this compilation error in my sample app: 

```
Cannot fit requested classes in a single dex file (# methods: 69208 > 65536)
```

Looks like this is actually required, and is not just for integration tests.

PR #539, opened 2020-05-27
fix(api): Fix #538 by providing a default empty byte array for REST payload

### LINKED ISSUE #538: REST API converts PUT and POST with no body into GET request in-flight
OkHttp seems to convert in-flight POST and PUT requests to GET requests if the body provided is null. This is surprising.

In order to address this, `RestOptions` should be changed to:

1. Ensure that the constructor cannot be called directly, rather we should prefer `RestOptions.Builder`
1. In the constructor, if provided `data` is `null`, pass an empty `byte[]`

TrekSoft:
*Issue #, if available:* #538 

*Description of changes:*

OkHTTP converts POST/PUT requests to GET requests if the payload is null. Addresses this by sending an empty byte array as the payload if the Customer does not specify one.

Also privatizes the constructor of RestOptions since the builder is what should be used


TrekSoft:
**Not ready to merge yet**: Need the option of a null payload for GET requests so will move the logic to set content only for posts and puts

PR #540, opened 2020-05-27
fix(core): use sentinel for null query predicate

jamesonwilliams:
Around the code-base, a `null` value for a `QueryPredicate` has the
meaning of "don't apply any filter," or, "match everything."

That's sort of clear, maybe.

Instead of having a meaningful `null` value, use a sentinel, instead,
which declares its purpose as `QueryPredicates.matchAll()`.


jamesonwilliams:
> LGTM! One thought I had was perhaps we should consider `all()` versus `matchAll()` in the spirit of brevity.

@drochetti mentions that iOS uses `all()`, and JS uses `ALL`. So, I agree with _all_ of you!

PR #543, opened 2020-05-28
Remove old user-agent header from API interceptor

raphkim:
*Issue #, if available:*

*Description of changes:*
There was a redundant user agent being added to API auth interceptors. Removed it to prevent clashing with the actual user agent being added by `UserAgentInterceptor`.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
@@ -25,7 +25,6 @@ import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.http.HttpMethodName;
 import com.amazonaws.util.IOUtils;
-import com.amazonaws.util.VersionInfoUtils;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


raphkim:
> It is possible to write a component (`src/test`) test for this.
> 
> Now that we are past GA, we should make this part of our standard process.
> 
> It would like this:
> 
> 1. Use `MockWebServer` to setup a fake backend.
> 2. Configure an `AWSApiPlugin`, with the mock web server as the endpoint URL.
> 3. Make a request via the plugin.
> 4. Ask the mock web server what request it captured.
>    a. Inspect the headers, and peel out the user agent.
>    b. Validate the user agent has the format used in our `com.amplifyframework.util.UserAgent`.

Thanks for the feedback! I will update this PR with appropriate tests

raphkim:
@jamesonwilliams Add unit test as recommended!

Turns out the incorrect user agents were already being overriden by the correct user agent interceptor. Still, This PR should be useful for removing useless code and increasing test coverage.

jpignata:
Can you squash this into a message that follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)?

aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Make an API request to the mock server and check the user agent
+     * header format on the request. Verify that the Android version is
+     * accurately reflected on the user agent.
+     * @throws Exception if API call fails or thread is interrupted while
+     *          waiting for request
+     */
+    @Test
+    @Config(sdk = 16)
```
jamesonwilliams, line 98: Nice. Good test idea.

PR #545, opened 2020-05-29
chore: Replace threeten backport dependency with java.time

richardmcclellan:
By updating to Android Gradle Plugin 4.0.0, we can now take advantage of [Java 8+ desugaring support](https://developer.android.com/studio/write/java8-support#library-desugaring), which allows us to use `java.time`.  This PR removes the threetenabp dependency and replaces it with `java.time`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,7 +78,6 @@ ext {
             translate: "com.amazonaws:aws-android-sdk-translate:$awsSdkVersion"
         ],
 
-
         okhttp: 'com.squareup.okhttp3:okhttp:4.7.2',
         gson: 'com.google.code.gson:gson:2.8.6',
         rxandroid: 'io.reactivex.rxjava2:rxandroid:2.1.1',
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ ext {
     awsSdkVersion = '2.16.12'
     dependency = [
         android: [
-            desugar_jdk_libs: 'com.android.tools:desugar_jdk_libs:1.0.5',
+            desugartools: 'com.android.tools:desugar_jdk_libs:1.0.5',
         ],
         androidx: [
             v4support: 'androidx.legacy:legacy-support-v4:1.0.0',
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -49,6 +49,9 @@ ext {
 
     awsSdkVersion = '2.16.12'
     dependency = [
+        android: [
+            desugar_jdk_libs: 'com.android.tools:desugar_jdk_libs:1.0.5',
```
jamesonwilliams, line 53: For better or worse, I've avoided `_` in the other ones, so far.

How about just:
```suggestion
            desugartool: 'com.android.tools:desugar_jdk_libs:1.0.5',
```

build.gradle
```diff
@@ -75,12 +78,12 @@ ext {
             translate: "com.amazonaws:aws-android-sdk-translate:$awsSdkVersion"
         ],
 
+
```
jamesonwilliams, line 81:
```suggestion
```

core/build.gradle
```diff
@@ -17,10 +17,18 @@ apply plugin: 'com.android.library'
...
 
+android {
+    compileOptions {
+        coreLibraryDesugaringEnabled true
+    }
+}
+
 dependencies {
+    coreLibraryDesugaring dependency.android.desugar_jdk_libs
+
```
jamesonwilliams, line 28: We can move these into the `afterEvaluate` block inside the root project `build.gradle`, so that they get applied to all modules, and keep the module `build.gradle` slim and consistent.

core/build.gradle
```diff
@@ -17,10 +17,18 @@ apply plugin: 'com.android.library'
...
     implementation dependency.androidx.v4support
     implementation dependency.androidx.annotation
-    implementation dependency.threetenabp
+
```
jamesonwilliams, line 31:
```suggestion
```

core/build.gradle
richardmcclellan: Good idea, updated!

PR #546, opened 2020-05-29
chore: dont run proguard only publish lib rules

jamesonwilliams:
The current configuration is to run Proguard on the library, before
generating an artifact.

Instead, we will vend rules to a consumer, so that _they_ can run
Proguard.

This approach is more in line with the behavior of other Android
libraries.


build.gradle
```diff
@@ -128,12 +128,5 @@ private void configureAndroidLibrary(Project project) {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
-
-        buildTypes {
-            release {
-                minifyEnabled true
-                proguardFiles 'proguard-rules.pro'
-            }
-        }
```
raphkim, line 137: will this fix the issue where our private/protected items become public?

jamesonwilliams: Since Proguard won’t be run, field visibility won’t be modified, true. I think this change will also fix our stack traces and “step-in” capability when debugging from a consuming application.

jamesonwilliams:
> I just tested this, and I CAN step in with the debugger now! This is great!
> 
> I'm confused though because I would expect our proguard rules to still exists in `configuration/proguard-rules.pro`, but that file looks empty in this PR. Am I missing something?

Awesome!

I looked at the vended library rules for a few other projects. They tend to be very slim:
1.  [OkHttp's `okhttp3.pro`](https://github.com/square/okhttp/blob/master/okhttp/src/main/resources/META-INF/proguard/okhttp3.pro) has just four rules;
2. [RxJava's `rxjava3.pro`](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/resources/META-INF/proguard/rxjava3.pro) has just 1 customization.

We might end up with one or two ourselves. But for now, I'm not sure what we need to put there. Approximately 0 rules seems like a good start.

richardmcclellan:
Don’t we at least something like `-keepClassMembers` so that any of our objects that are serialized with Gson don’t become obfuscated?

PR #557, opened 2020-06-04
Remove nextToken from ModelPagination since nextToken is never exposed

richardmcclellan:
`nextToken` is an AppSync detail, which should not be exposed in the public interface.   There is no way to obtain it from a response directly, so we can remove it from the `ModelPagination` object.

To make a request to get a "next page" of results, customers should obtain a `GraphQLRequest` via the `getRequestForNextPage()` method on the `PaginatedResult` object.  This request object is a copy of the `GraphQLRequest` used in the previous query, but with the `nextToken` added as a variable.   Then, customers can pass this GraphQLRequest directly to `Amplify.API.query(...)`

This also adds an additional factory method, so customers can do `ModelPagination.limit(1000)`, instead of `ModelPagination.firstPage().withLimit(1000)`

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPaginatedResult.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPaginatedResult.java
@@ -16,6 +16,7 @@
 package com.amplifyframework.api.aws;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.api.graphql.GraphQLRequest;
@@ -27,7 +28,7 @@ final class AppSyncPaginatedResult<T extends Model> extends PaginatedResult<T> {
     private final Iterable<T> items;
 
     AppSyncPaginatedResult(@NonNull Iterable<T> items,
-                           @NonNull GraphQLRequest<PaginatedResult<T>> requestForNextResult) {
+                           @Nullable GraphQLRequest<PaginatedResult<T>> requestForNextResult) {
         this.requestForNextResult = requestForNextResult;
         this.items = items;
     }
--- a/aws-api/src/main/java/com/amplifyframework/api/graphql/model/ModelQuery.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/graphql/model/ModelQuery.java
@@ -122,7 +122,7 @@ public final class ModelQuery {
             @NonNull ModelPagination pagination
     ) {
         return AppSyncGraphQLRequestFactory.buildPaginatedResultQuery(
-                modelType, null, pagination.getLimit());
+                modelType, QueryPredicates.all(), pagination.getLimit());
     }
 
 }
--- a/aws-api/src/main/java/com/amplifyframework/api/graphql/model/ModelQuery.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/graphql/model/ModelQuery.java
@@ -93,7 +93,6 @@ public final class ModelQuery {
      * @param pagination the pagination settings.
      * @param <M> the concrete model type.
      * @return a valid {@link GraphQLRequest} instance.
-     * @see ModelPagination#nextToken(String)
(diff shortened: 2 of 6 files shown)


raphkim:
> To make a request to get a "next page" of results, customers should obtain a `GraphQLRequest` via the `getRequestForNextPage()` method on the `PaginatedResult` object. This request object is a copy of the `GraphQLRequest` used in the previous query, but with the `nextToken` added as a variable. Then, customers can pass this GraphQLRequest directly to `Amplify.API.query(...)`

Is this behavior tested anywhere? 

Also, one random thing I noticed while looking through the code: the constructor for `AppSyncPaginatedResult` seems to take in `@NonNull GraphQLRequest<PaginatedResult<T>> requestForNextResult` but its implementation of `hasNextResult()` is `return requestForNextResult != null;`, which always evaluates to `true`. Will `requestForNextResult` always be non-null as the annotation suggests?

richardmcclellan:
> > To make a request to get a "next page" of results, customers should obtain a `GraphQLRequest` via the `getRequestForNextPage()` method on the `PaginatedResult` object. This request object is a copy of the `GraphQLRequest` used in the previous query, but with the `nextToken` added as a variable. Then, customers can pass this GraphQLRequest directly to `Amplify.API.query(...)`
> 
> Is this behavior tested anywhere?
> 

Good callout, turns out it's not.   I just added a test, and it actually caught a bug where an exception was thrown whenever a paginated query was made.  I fixed the bug as well.  Yay tests!

> Also, one random thing I noticed while looking through the code: the constructor for `AppSyncPaginatedResult` seems to take in `@NonNull GraphQLRequest<PaginatedResult<T>> requestForNextResult` but its implementation of `hasNextResult()` is `return requestForNextResult != null;`, which always evaluates to `true`. Will `requestForNextResult` always be non-null as the annotation suggests?

Good call, it can be null, so I changed it to `@Nullable`

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -158,6 +161,32 @@ public void graphQlQueryRendersValidResponse() throws ApiException {
         );
     }
 
+    /**
+     * Same as graphQlQueryRendersValidResponse, except with pagination.   Expect a PaginatedResult&lt;BlogOwner&gt;
```
jamesonwilliams, line 165:
```suggestion
     * Same as {@link #graphQlQueryRendersValidResponse()}, except with pagination.
     * Expect a PaginatedResult&lt;BlogOwner&gt;
```

aws-api/src/main/java/com/amplifyframework/api/graphql/model/ModelQuery.java
```diff
@@ -122,7 +122,7 @@ private ModelQuery() {}
             @NonNull ModelPagination pagination
     ) {
         return AppSyncGraphQLRequestFactory.buildPaginatedResultQuery(
-                modelType, null, pagination.getLimit());
+                modelType, QueryPredicates.all(), pagination.getLimit());
```
jamesonwilliams, line 125: Woop!

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncPaginatedResult.java
```diff
@@ -27,7 +28,7 @@
     private final Iterable<T> items;
 
     AppSyncPaginatedResult(@NonNull Iterable<T> items,
-                           @NonNull GraphQLRequest<PaginatedResult<T>> requestForNextResult) {
+                           @Nullable GraphQLRequest<PaginatedResult<T>> requestForNextResult) {
```
jamesonwilliams, line 31: Is it possible to have a paginated result, that has no request associated with it?

richardmcclellan: Yes, if `nextToken` is null (the last page), then the `requestForNextResult` will also be null

PR #559, opened 2020-06-04
feat(API): Add AuthRule annotation to support @auth directive

richardmcclellan:
This adds an `@AuthRule` annotation, so that it can be used in codegen model files.   The annotation can represent an authorization rule if any are specified in a GraphQL schema with the [@auth directive](https://docs.amplify.aws/cli/graphql-transformer/directives#auth).  `ModelConfig` and `ModelField` now include an array of @AuthRule objects.   

Note that this replaces the `boolean hasOwnerAuthorization()` method previously on `ModelConfig`, which was intended to support the basic owner based auth use case.  This PR adds support for *all* use cases allowed by the @auth directive.

Sample `schema.graphql`:
```
type Todo
  @model
  @auth(rules: [
    # Owner authorization, simplest case
    { allow: owner },

    # Owner authorization, the long form way
    { allow: owner, ownerField: "owner", operations: [create, update, delete, read] },

    # Owner authorization, allow others to read
    { allow: owner, operations: [create, delete, update] },

    # Static group authorization
    { allow: groups, groups: ["Admin"] },

    # Dynamic group authorization
    { allow: groups, groupsField: "groups" },

    # public authorization
    { allow: public },

    # public authorization with provider override
    { allow: public, provider: iam },

    # private authorization
    { allow: private },

    # private authorization with provider override
    { allow: private, provider: iam },

    # owner authorization with provider override
    { allow: owner, provider: oidc, identityClaim: "sub" }

    # custom claims (owner)
    { allow: owner, identityClaim: "user_id" },

    # custom claims (groups)
    { allow: groups, groups: ["Moderator"], groupClaim: "user_groups" }
    ])
{
  id: ID!
  name: String!
  description: String
    @auth (rules: [
    { allow: private, provider: iam, operations: [create, update] }
    ])
}
```

Sample expected codegen output: `Todo.java`:
```
package com.amplifyframework.datastore.generated.model;

import androidx.core.util.ObjectsCompat;

import com.amplifyframework.core.model.AuthStrategy;
import com.amplifyframework.core.model.ModelOperation;
import com.amplifyframework.core.model.annotations.AuthRule;
import com.amplifyframework.core.model.annotations.ModelConfig;
import com.amplifyframework.core.model.annotations.ModelField;
import com.amplifyframework.core.model.query.predicate.QueryField;

import java.util.Objects;
import java.util.UUID;

import static com.amplifyframework.core.model.query.predicate.QueryField.field;

/** This is an auto generated class representing the Todo type in your schema. */
@SuppressWarnings("all")
@ModelConfig(
        pluralName = "Todos",
        authRules = {
                // Owner authorization, simplest case
                @AuthRule(
                        allow = AuthStrategy.OWNER
                ),

                // Owner authorization, the long form way
                @AuthRule(
                        allow = AuthStrategy.OWNER,
                        ownerField = "owner",
                        operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ }
                ),

                // Owner authorization, allow others to read
                @AuthRule(
                        allow = AuthStrategy.OWNER,
                        operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE }
                ),

                // Static group authorization
                @AuthRule(
                        allow = AuthStrategy.GROUPS,
                        groups = { "Admin" }
                ),

                // Dynamic group authorization
                @AuthRule(
                        allow = AuthStrategy.GROUPS,
                        groupsField = "groups"
                ),

                // Public authorization
                @AuthRule(
                        allow = AuthStrategy.PUBLIC
                ),

                // Public authorization with provider override (provider does not need to be specified in codegen)
                @AuthRule(
                        allow = AuthStrategy.PUBLIC
                ),

                // private authorization
                @AuthRule(
                        allow = AuthStrategy.PRIVATE
                ),

                // private authorization with provider override (provider does not need to be specified in codegen)
                @AuthRule(
                        allow = AuthStrategy.PRIVATE
                ),

                // owner authorization with provider override (provider does not need to be specified in codegen)
                @AuthRule(
                        allow = AuthStrategy.OWNER,
                        identityClaim = "sub"
                ),

                // custom claims (owner)
                @AuthRule(
                        allow = AuthStrategy.OWNER,
                        identityClaim = "user_id"
                ),

                // custom claims (groups)
                @AuthRule(
                        allow = AuthStrategy.GROUPS,
                        groups = { "Moderator" },
                        groupClaim = "user_groups"
                ),
        }
)
public final class Todo implements Model {
  public static final QueryField ID = field("id");
  public static final QueryField NAME = field("name");
  public static final QueryField DESCRIPTION = field("description");
  private final @ModelField(targetType="ID", isRequired = true) String id;
  private final @ModelField(targetType="String", isRequired = true) String name;
  private final @ModelField(
    targetType = "String",
    isRequired = true,
    authRules = {
      @AuthRule(
        allow = AuthStrategy.PRIVATE,
        operations = { ModelOperation.CREATE, ModelOperation.UPDATE }
      )
    }
  ) String description;

```

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
@@ -15,8 +15,9 @@
 
 package com.amplifyframework.core.model;
 
+import androidx.core.util.ObjectsCompat;
+
 import java.util.Arrays;
-import java.util.Objects;
 
 /**
  * {@link AuthRule} is used define an authorization rule for who can access and operate against a
@@ -125,17 +126,17 @@ public final class AuthRule {
         AuthRule authRule = (AuthRule) object;
 
         return allow == authRule.allow &&
-                Objects.equals(ownerField, authRule.ownerField) &&
-                Objects.equals(identityClaim, authRule.identityClaim) &&
-                Objects.equals(groupClaim, authRule.groupClaim) &&
+                ObjectsCompat.equals(ownerField, authRule.ownerField) &&
+                ObjectsCompat.equals(identityClaim, authRule.identityClaim) &&
+                ObjectsCompat.equals(groupClaim, authRule.groupClaim) &&
                 Arrays.equals(groups, authRule.groups) &&
-                Objects.equals(groupsField, authRule.groupsField) &&
+                ObjectsCompat.equals(groupsField, authRule.groupsField) &&
                 Arrays.equals(operations, authRule.operations);
     }
 
     @Override
     public int hashCode() {
-        int result = Objects.hash(allow, ownerField, identityClaim, groupClaim, groupsField);
+        int result = ObjectsCompat.hash(allow, ownerField, identityClaim, groupClaim, groupsField);
         result = 31 * result + Arrays.hashCode(groups);
         result = 31 * result + Arrays.hashCode(operations);
         return result;
--- a/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
@@ -54,7 +54,7 @@ public final class AuthRule {
      * Returns the type of strategy for this {@link AuthRule}.
      * @return the type of strategy for this {@link AuthRule}
      */
-    public AuthStrategy allow() {
+    public AuthStrategy getAllow() {
         return this.allow;
     }
 
@@ -63,7 +63,7 @@ public final class AuthRule {
      *
      * @return name of a {@link ModelField} of type String which specifies the user which should have access
      */
-    public String ownerField() {
+    public String getOwnerField() {
         return this.ownerField;
     }
 
@@ -72,7 +72,7 @@ public final class AuthRule {
      *
      * @return identity claim
      */
-    public String identityClaim() {
+    public String getIdentityClaim() {
         return this.identityClaim;
     }
 
@@ -81,7 +81,7 @@ public final class AuthRule {
      *
      * @return group claim
      */
-    public String groupClaim() {
+    public String getGroupClaim() {
         return this.groupClaim;
     }
 
@@ -90,7 +90,7 @@ public final class AuthRule {
      *
      * @return array of groups which should have access
      */
-    public String[] groups() {
+    public String[] getGroups() {
         return this.groups;
     }
 
@@ -100,7 +100,7 @@ public final class AuthRule {
      * @return name of a {@link ModelField} of type String or array of Strings which specifies a group or list of groups
      * which should have access.
      */
-    public String groupsField() {
+    public String getGroupsField() {
         return this.groupsField;
     }
 
@@ -109,7 +109,7 @@ public final class AuthRule {
      * the list are not protected by default.
      * @return list of {@link ModelOperation}s for which this {@link AuthRule} should apply.
      */
-    public ModelOperation[] operations() {
+    public ModelOperation[] getOperations() {
         return this.operations;
     }
 
--- a/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
@@ -27,7 +27,7 @@ import java.util.Arrays;
  * documentation.</a>
  */
 public final class AuthRule {
-    private final AuthStrategy allow;
+    private final AuthStrategy authStrategy;
     private final String ownerField;
     private final String identityClaim;
     private final String groupClaim;
@@ -41,7 +41,7 @@ public final class AuthRule {
      * @param authRule an {@link com.amplifyframework.core.model.annotations.AuthRule} annotation.
      */
     public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
-        this.allow = authRule.allow();
+        this.authStrategy = authRule.allow();
         this.ownerField = authRule.ownerField();
         this.identityClaim = authRule.identityClaim();
         this.groupClaim = authRule.groupClaim();
@@ -54,8 +54,8 @@ public final class AuthRule {
      * Returns the type of strategy for this {@link AuthRule}.
      * @return the type of strategy for this {@link AuthRule}
      */
-    public AuthStrategy getAllow() {
-        return this.allow;
+    public AuthStrategy getAuthStrategy() {
+        return this.authStrategy;
     }
 
     /**
@@ -125,7 +125,7 @@ public final class AuthRule {
 
         AuthRule authRule = (AuthRule) object;
 
-        return allow == authRule.allow &&
+        return authStrategy == authRule.authStrategy &&
                 ObjectsCompat.equals(ownerField, authRule.ownerField) &&
                 ObjectsCompat.equals(identityClaim, authRule.identityClaim) &&
                 ObjectsCompat.equals(groupClaim, authRule.groupClaim) &&
@@ -136,7 +136,7 @@ public final class AuthRule {
 
     @Override
     public int hashCode() {
-        int result = ObjectsCompat.hash(allow, ownerField, identityClaim, groupClaim, groupsField);
+        int result = ObjectsCompat.hash(authStrategy, ownerField, identityClaim, groupClaim, groupsField);
         result = 31 * result + Arrays.hashCode(groups);
         result = 31 * result + Arrays.hashCode(operations);
         return result;
@@ -145,7 +145,7 @@ public final class AuthRule {
     @Override
     public String toString() {
         return "AuthRuleImpl{" +
-                "allow=" + allow +
+                "authStrategy=" + authStrategy +
                 ", ownerField='" + ownerField + '\'' +
                 ", identityClaim='" + identityClaim + '\'' +
                 ", groupClaim='" + groupClaim + '\'' +
--- a/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * {@link AuthRule} is used define an authorization rule for who can access and operate against a
+ * {@link Model} or a {@link ModelField}.
+ *
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth")>GraphQL Transformer @auth directive
+ * documentation.</a>
+ */
+public final class AuthRule {
+    private final AuthStrategy allow;
+    private final String ownerField;
+    private final String identityClaim;
+    private final String groupClaim;
+    private final String[] groups;
+    private final String groupsField;
+    private final ModelOperation[] operations;
+
+    /**
+     * Constructor to create an {@link AuthRule} from an {@link com.amplifyframework.core.model.annotations.AuthRule}
+     * annotation.
+     * @param authRule an {@link com.amplifyframework.core.model.annotations.AuthRule} annotation.
+     */
+    public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
+        this.allow = authRule.allow();
+        this.ownerField = authRule.ownerField();
+        this.identityClaim = authRule.identityClaim();
+        this.groupClaim = authRule.groupClaim();
+        this.groups = authRule.groups();
+        this.groupsField = authRule.groupsField();
+        this.operations = authRule.operations();
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
```diff
@@ -458,14 +485,15 @@ public Builder pluralName(@Nullable String pluralName) {
...
-         * @param hasOwnerAuthorization true if the model has owner based authorization, false otherwise
+         *
+         * @param authRules list of {@link AuthRule}s for this {@link Model}
          * @return the builder object
          */
         @NonNull
-        public Builder hasOwnerAuthorization(boolean hasOwnerAuthorization) {
-            this.hasOwnerAuthorization = hasOwnerAuthorization;
+        public Builder authRules(List<AuthRule> authRules) {
+            this.authRules = authRules;
```
jamesonwilliams, line 496: To keep `this.authRules` as non-`null`, maybe change this to:
```
if (authRules != null) {
     this.authRules.addAll(authRules);
}
```
Then on L455 you can mark the declaration as `final`, as well.

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
```diff
@@ -396,7 +423,7 @@ public int hashCode() {
         return ObjectsCompat.hash(
                 getName(),
                 getPluralName(),
-                hasOwnerAuthorization(),
+                authRules(),
```
jamesonwilliams, line 426: Rename as `getAuthRules()`, the other ones use the `get<Foo>` prefix.

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
```diff
@@ -242,12 +254,27 @@ public String getPluralName() {
...
+    public List<AuthRule> authRules() {
+        return authRules;
+    }
+
+    /**
+     * Returns true if this model has owner based authorization which changes the parameters for subscriptions.
+     * @return true if owner authorization is present on this model and false if not or if not explicitly annotated
      */
+
     public boolean hasOwnerAuthorization() {
```
jamesonwilliams, line 271: Is this still necessary? Is it kept because it is `public` and you don't want to change the contract? Should we mark this `@Deprecated`, even if we'll keep it and update its implementation? (I'm not sure?)

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
```diff
@@ -113,14 +114,20 @@ public static ModelSchema fromModelClass(@NonNull Class<? extends Model> clazz)
...
             // Set the model name and plural name (null if not provided)
             ModelConfig modelConfig = clazz.getAnnotation(ModelConfig.class);
             final String modelName = clazz.getSimpleName();
             final String modelPluralName = modelConfig != null && !modelConfig.pluralName().isEmpty()
                     ? modelConfig.pluralName()
                     : null;
-            final boolean hasOwnerAuthorization = modelConfig != null && modelConfig.hasOwnerAuthorization();
+
+            if (modelConfig != null) {
+                for (com.amplifyframework.core.model.annotations.AuthRule ruleAnnotation : modelConfig.authRules()) {
```
jamesonwilliams, line 127: Remove the import qualification

core/src/main/java/com/amplifyframework/core/model/ModelOperation.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package com.amplifyframework.core.model;
+
+/**
+ * Used as a property of {@link AuthRule} to specify which operations the rule applies to.
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth")>GraphQL Transformer @auth directive
+ * documentation.</a>
+ */
+public enum ModelOperation {
```
jamesonwilliams, line 23: Ah man, I have like 3-4 more enums like this in the DataStore category/plugin. In future work, we can probably come back and just use yours everywhere.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -0,0 +1,156 @@
...
+        this.groups = authRule.groups();
+        this.groupsField = authRule.groupsField();
+        this.operations = authRule.operations();
+    }
+
+    /**
+     * Returns the type of strategy for this {@link AuthRule}.
+     * @return the type of strategy for this {@link AuthRule}
+     */
+    public AuthStrategy allow() {
```
jamesonwilliams, line 56: Should we use the `get<Foo>` on new getters? It makes property accessors work right from Kotlin. On the other hand, we already missed the boat on that, and we're going to have to wrap the API surface, anyway. So, if that's true, maybe we should just write the best possible Java API - which would _not_ use `get<Foo>`.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -0,0 +1,156 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model;
+
+import java.util.Arrays;
+import java.util.Objects;
```
jamesonwilliams, line 19: `Objects` requires minSdk of 19, doesn't it? We are using `ObjectsCompat` for this reason in much of the code base. Nevermind that there are currently issues with APIs 16-19... might as well not put gas on the bonfire. 😄 🚀

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
jpignata: I'm not sure we should assume the Kotlin extensions we write will wrap every parameter. The ones I've looked at are actually pretty thin. My POV: `getFoo` is still probably the right place for us to land for Java, and will automagically work in Kotlin.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
richardmcclellan: Fixed!   This was for the IDE-generated implementation of `equals()`.  I'll make note to replace with `ObjectsCompat` next time.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
richardmcclellan: I agree `getAllow()` is probably the best pattern for this class, and will make the change.  Of note, there is also an `@interface AuthRule`, for the @AuthRule annotation. I think `allow()` still makes sense for the annotation, otherwise using the annotation would be weird (`getAllow = AuthStrategy.OWNER`)

jamesonwilliams: Ahh! Yes!

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
richardmcclellan: Hmm, I actually can't remove it because there are references to the `AuthRule` class and the `@AuthRule` annotation within ModelSchema.java, so this differentiates which one I am referring to.   I suppose the qualifications could be eliminated with a larger refactor of some business logic, but I don't see a strong reason to do that at this point.

Of note, there are also fully qualified class names for the `@ModelField` annotation within this ModelSchema.java file as well for the same reason (there's a `@ModelField` annotation, and a `@ModelField` class)

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
richardmcclellan: This is actually a helper method now, and is being used as such.  As I dive in deeper to figure out how to annotate each GraphQLRequest with auth stuff (my next task), I might refactor/move this logic elsewhere, but not totally sure on that yet.

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
richardmcclellan: I've updated to the approach below.  Similar to what you suggested, but matches other similar methods on this `Builder` a bit better.

```
        public Builder authRules(@NonNull List<AuthRule> authRules) {
            Objects.requireNonNull(authRules);
            this.authRules.clear();
            this.authRules.addAll(authRules);
            return this;
        }
```

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
richardmcclellan: Done!

richardmcclellan:
All comments are addressed!

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -0,0 +1,157 @@
...
+
+import androidx.core.util.ObjectsCompat;
+
+import java.util.Arrays;
+
+/**
+ * {@link AuthRule} is used define an authorization rule for who can access and operate against a
+ * {@link Model} or a {@link ModelField}.
+ *
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth")>GraphQL Transformer @auth directive
```
jamesonwilliams, line 26: `")>` -- is the `)` supposed to be there?

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -0,0 +1,157 @@
...
+        this.groupsField = authRule.groupsField();
+        this.operations = authRule.operations();
+    }
+
+    /**
+     * Returns the type of strategy for this {@link AuthRule}.
+     * @return the type of strategy for this {@link AuthRule}
+     */
+    public AuthStrategy getAllow() {
+        return this.allow;
```
jamesonwilliams, line 58: Why not:
```
public AuthStrategy getAuthStrategy() {
    return this.authStrategy;
}
```

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model;
+
+/**
+ *  The type of strategy for an @auth directive rule.
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth")>GraphQL Transformer @auth directive
```
jamesonwilliams, line 20: Same here, is the `")>` intentional?

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Arrays;
+
+/**
+ * {@link AuthRule} is used define an authorization rule for who can access and operate against a
+ * {@link Model} or a {@link ModelField}.
+ *
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth")>GraphQL Transformer @auth directive
+ * documentation.</a>
+ */
+public final class AuthRule {
```
jamesonwilliams, line 29: This class seems like it is "modal."  Like, depending on the `AuthStrategy`, the `AuthRule` will have different fields become null, by design, right?

Instead, would it make sense to have an `interface AuthRule`, and then do `GroupBasedAuthRule`, which only returned group info, `UserBasedAuthRule`, which only returned user info, etc.?

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
richardmcclellan: Good catch, removed!

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
richardmcclellan: Removed!

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
richardmcclellan: For the @AuthRule annotation, `allow` is used since that is the term used in the schema.graphql.  For this model though, I agree `authStrategy` makes more sense, and have made this change.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
richardmcclellan: You're right, it is somewhat "modal".   I did consider a similar approach to what you are a suggesting, but did not implement because I wasn't quite sure how I would handle an invalid rule definition.

For example if a customer defines a rule of: `{ allow: owner, groupsField: "groups" }`, (`groupsField` should only be used when `allow: groups`), what should happen?

Interestingly, `amplify push` does *not* fail with a schema like this. 

We could:
1. Annotate requests with a mismatch of rule metadata, and perhaps end up with an error in the `GraphQLResponse`.
2. Silently ignore the groupsField at runtime.
3. Throw a `RuntimeException` and suggest to the customer to fix their schema.

Anyway, I *think* refactoring to a "modal" design could make sense, but would like to address that in my next PR which will annotate requests with these details, after I've had a chance to dive deeper on how this really works.

jamesonwilliams: You could even make an `InvalidAuthRule` subclass!

richardmcclellan:
> > I think refactoring to a "modal" design could make sense, but would like to address that in my next PR which will annotate requests with these details, after I've had a chance to dive deeper on how this really works.
> 
> Sounds good!
> 
> Are there any opportunities to add tests with this current work? Or will that come in a subsequent PR?
> 
> Aside from these ^ looks good, approving!

Thanks - tests will definitely be coming in my next PR!

Issue #458, opened 2020-05-13
SQLiteException: no such table 


bryannevega:
Hello,

Why when I use "EQ" filter my query executes but when I use "contains" it gives me the error mentioned above?

```
com.amplifyframework.datastore.DataStoreException: Error in querying the model.
        at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.lambda$query$4$SQLiteStorageAdapter(SQLiteStorageAdapter.java:64)
        at com.amplifyframework.datastore.storage.sqlite.-$$Lambda$SQLiteStorageAdapter$jJRyJ1jI1PnOmSf78Qfq5UwDppI.run(Unknown Source:10)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:457)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: android.database.sqlite.SQLiteException: no such table: type_id (code 1): , while compiling:
```

jamesonwilliams:
Hi @bryannevega Can you please include relevant portions of your schema and a code snippet? Also, looks like the stack trace is truncated a bit. Thanks!

Andrew-Dickinson:
I'll chime in, since I'm having exactly the same issue

**My Schema**
```GraphQL
type GroceryItem @model {
  id: ID!
  checked: Boolean!
  name: String
  position: Int!
  aisle: Aisle! @connection(name: "AisleGroceries")
}

type Aisle @model {
  id: ID!
  name: String!
  position: Int!
  groceries: [GroceryItem!]! @connection(name: "AisleGroceries")
}
```

**WORKS**
```kotlin
 Amplify.DataStore.query(
            GroceryItem::class.java, QueryField.field("GroceryItem.name").eq(queryString),
            ...
```

**DOESN'T WORK**
```kotlin
 Amplify.DataStore.query(
            GroceryItem::class.java, QueryField.field("GroceryItem.name").contains(queryString),
            ...
```
**The full stacktrace**
```java
E/SQLiteLog: (1) no such table: GroceryItem.name
E/GetStarted: Query failed.
    com.amplifyframework.datastore.DataStoreException: Error in querying the model.
[...]

PR #584, opened 2020-06-17
Fix for contains predicate

rjuliano:
*Issue #, if available: #458 

*Description of changes:*
The issue is that the `contains` predicate assumes that the field we're querying is a list. However, when used on a String field, it results in invalid SQL syntax. So for example:

```
Blog.TAGS.contains("something");
```

yields

```sql
"something" IN tags
```  
which results in an error.

At first, I thought fixing it would just be a matter of determining whether the field we’re querying against is a List<> or a String. However, that information is not available from inside the `SQLPredicate` class as it exists today, so I need to explore some options.

## Update
Decided to go with the [instr](https://www.sqlite.org/lang_corefunc.html#instr) function to handle string pattern matching operation.

------------

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
@@ -141,12 +141,25 @@ public final class SQLPredicate {
                         .append("?");
             case CONTAINS:
                 ContainsQueryOperator containsOp = (ContainsQueryOperator) op;
-                addBinding(containsOp.value());
-                return builder.append("?")
+                //TODO: Need to somehow the the underlying model type to determine whether the
+                //      field we're querying against is a string field of a list of primitive types
+                //
+                JavaFieldType javaType = TypeConverter.getJavaFieldTypeFromValue(containsOp.value());
+                if (JavaFieldType.STRING.equals(javaType)) {
+                    addBinding("%" + containsOp.value() + "%");
+                    return builder.append(field)
+                        .append(SqlKeyword.DELIMITER)
+                        .append(SqlKeyword.LIKE)
+                        .append(SqlKeyword.DELIMITER)
+                        .append("?");
+                } else {
+                    addBinding(containsOp.value());
+                    return builder.append("?")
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
```diff
@@ -141,12 +141,25 @@ private StringBuilder parsePredicateOperation(QueryPredicateOperation<?> operati
                         .append("?");
             case CONTAINS:
                 ContainsQueryOperator containsOp = (ContainsQueryOperator) op;
-                addBinding(containsOp.value());
-                return builder.append("?")
+                //TODO: Need to somehow the the underlying model type to determine whether the
```
rjuliano, line 144:
```suggestion
                //TODO: Need to somehow get the underlying model type to determine whether the
```

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SQLPredicateTest.java
```diff
@@ -0,0 +1,46 @@
+package com.amplifyframework.datastore.storage.sqlite;
```
jamesonwilliams, line 1: Please add the copyright header

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SQLPredicateTest.java
```diff
@@ -0,0 +1,46 @@
...
+     * @throws DataStoreException
+     */
+    @Test
+    public void testContainsForStringList() throws DataStoreException {
+        QueryPredicateOperation<String> predicate = Blog.TAGS.contains("something");
+        SQLPredicate sqlPredicate = new SQLPredicate(predicate);
+        assertStringResults(sqlPredicate, "tags");
+    }
+
+    private void assertStringResults(SQLPredicate sqlPredicate, String fieldName) {
```
jamesonwilliams, line 41: Can we get a more descriptive name for this helper method?

rjuliano:
Fix will be included in the next release.

PR #587, opened 2020-06-17
chore: add code maintainers for default reviewers

jamesonwilliams:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*   @aws-amplify/android-team


jamesonwilliams:
> Do we also need to change some permissions associated with the @aws-amplify/android-team itself?  I ask because currently, I cannot manually assign that group as a reviewer.

I think so. I’m not sure this PR achieved what I wanted, anyway. I think your idea on group chat was more what I was after. Not sure what’s up with the permissions either, sigh.

Issue #186, opened 2019-12-10
[API] Subscribe fails if model has owner authorization


TrekSoft:
If a model has @auth(rules: [{allow: owner}]) AppSync requires an owner parameter be passed to the subscribe function but we currently are not generating this when we build the GraphQL Request for subscribe on these models. 

Furthermore, the CLI model gen does not indicate which models have this type of auth so we would not be able to know when to generate it.

jamesonwilliams:
Some work has been started on this in https://github.com/aws-amplify/amplify-android/pull/559.

a-melnyk:
Hi all,

is there a workaround for this issue (besides disabling authentication)? if my app doesn't use `DataStore.observe` would it be possible to skip the subscription request?

Another more general question: Could someone please give an estimate when the framework will be production-ready? in terms of stability (offline mode compatibility, not crash the app) and support for all Amplify features (like this one and AppSync Scalar types).
Background: my team builds an android application which has to be production-ready in 2 months and we have to decide if we can count on Amplify Framework or if we have to re-evaluate other options (apollo-android+AppSync & aws-android-sdk-cognitoidentityprovider+Cognito vs Firebase)

Thanks!

richardmcclellan:
Support for owner based auth has been added for `AWSApiPlugin` and `AWSDataStorePlugin`, in version `1.1.0`, released today.   Full end to end functionality also requires an update to the CLI though, so that codegen will add `@AuthRule` annotations to each generated model.    This work is in progress here: https://github.com/aws-amplify/amplify-cli/pull/4717

PR #596, opened 2020-06-24
GraphQLRequestFactory refactor & add owner based auth support to API category

richardmcclellan:
#### Summary:

- Fix for https://github.com/aws-amplify/amplify-android/issues/186.  If the model has `@AuthRule` with an `AuthStrategy.OWNER` auth strategy, then:
   - ~~ownerField (defaults to `owner`) is added to selection set if it isn't explicitly set in the schema.  This is done in `SelectionSet.java`~~ Removed per feedback from @lawmicha.
   - ownerField (defaults to `owner`) is added as a variable on subscriptions, with a value of `CognitoUserPoolsAuthProvider.getUsername()`, iff the `SubscriptionType` is protected by the `@AuthRule`.  For example, an `onCreate` subscription request only adds username if `create` mutations are protected (specified via `modelOperations`). This is done in `AWSApiPlugin.subscription` and `AppSyncGraphQLRequest.setAuthProvider`.   If the auth plugin is not configured or cognito user pools are not configured on the API, an exception is thrown on request construction. 
   - For mutations, the model object is serialized as a request variable.  If ownerField (defaults to `owner`) is defined, but set to null, we filter it out of the request object since it will be hydrated automatically by AppSync.
- Refactored `AppSyncGraphQLRequestFactory` so that request construction is more extensible, and no longer relies on long `StringBuilder`s.
   - `GraphQLRequest` is now abstract.  `SimpleGraphQLRequest` is a concrete implementation with the same constructors previously in `GraphQLRequest`, and is still used by `aws-datastore`.  `AppSyncGraphQLRequest` is a concrete implementation in `aws-api` where the bulk of the GraphQL document construction takes place now. 
   - `OperationType` is a new interface implemented by `QueryType`, `MutationType`, and `SubscriptionType`.   To build an `AppSyncGraphQLRequest`, simply pass the `OperationType` to the builder.
   - Some minor whitespace changes were made in the expected `request.json` files due to the request string construction refactor.


core/src/main/java/com/amplifyframework/util/Wrap.java
```diff
@@ -18,10 +18,10 @@
 import androidx.annotation.Nullable;
 
 /**
- * A utility to wrap strings into quotes.
+ * A utility to wrap strings.
  */
-public final class Quotes {
-    private Quotes() {}
+public final class Wrap {
```
jamesonwilliams, line 23: Nice adaptation, here!!

jpignata:
*slaps forehead*

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+     *      subscription OnCreatePerson(owner: String!, nextToken: String) {
+     *            onCreatePerson(owner: $owner, nextToken: $nextToken) {
+     *               age dob first_name id last_name relationship owner
+     *            }
+     *       }
+     *
+     * @return String value used for GraphQL "query" in HTTP request body
+     */
+    @Override
+    public String getQuery() {
```
rjuliano, line 166: I noticed that `nextToken` was removed as an input from the query in one of the test cases (`query-person-by-predicate.txt `) and I don't see that accounted for anywhere. Is that by design?

richardmcclellan: Good question - it is by design.  That test case is a request for a first page of results, so `nextToken` isn't needed.  For requesting subsequent pages, the `nextToken` variable is added in `AppSyncPaginatedResultDeserializer.java` when creating the `requestForNextPage`.  

Prior to this refactor, it was included within the request document String because the `requestForNextPage` used the same request document `String`, and just set the `nextToken` in the `variables` json object.  This is possible now because `GraphQLRequest` now knows how to add inputs and regenerate a document String after the object is created.

aws-api/src/test/java/com/amplifyframework/api/aws/SelectionSetTest.java
```diff
@@ -0,0 +1,50 @@
...
+@RunWith(RobolectricTestRunner.class)
+public class SelectionSetTest {
+    /**
+     * Test that selection set serialization works as expected.
+     * @throws AmplifyException if a ModelSchema can't be derived from Post.class
+     */
+    @Test
+    public void selectionSetSerializesToExpectedValue() throws AmplifyException {
+        SelectionSet.Node selectionSet = SelectionSet.fromModelClass(Post.class, 2);
+        assertEquals(Resources.readAsString("selection-set-post.txt").trim(), selectionSet.toString().trim());
```
jamesonwilliams, line 38: Are you able to get rid of the `.trim()` by modifying the contents of the `selection-set-post.txt`?

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -222,6 +237,63 @@ public void graphQlMutationGetsResponse() throws JSONException, ApiException {
...
+    @Test
+    public void graphQLSubscriptionDoesntAddOwnerFieldIfTypeIsNotProtected() throws JSONException {
+        GraphQLRequest<OwnerAuthReadUpdateOnly> request = ModelSubscription.onUpdate(OwnerAuthReadUpdateOnly.class);
+        plugin.subscribe(request,
+            subscriptionId -> { },
+            response -> { },
+            exception -> { },
+            EmptyAction.instance());
+
+        JSONAssert.assertEquals(Resources.readAsString("request-ownerauthreadupdateonly.json"),
```
jamesonwilliams, line 292: Can you add `-` chars to this file name, so its easier to read?

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -222,6 +237,63 @@ public void graphQlMutationGetsResponse() throws JSONException, ApiException {
...
+     * @throws JSONException from JSONAssert.assertEquals
+     */
+    @Test
+    public void graphQLSubscriptionAddsCustomOwnerField() throws JSONException {
+        GraphQLRequest<OwnerAuthCustomField> request = ModelSubscription.onUpdate(OwnerAuthCustomField.class);
+        plugin.subscribe(request,
+            subscriptionId -> { },
+            response -> { },
+            exception -> { },
+            EmptyAction.instance());
```
jamesonwilliams, line 269: Do we have an `EmptyConsumer` or `NoOpConsumer` somewhere, right now? I think so. If you use the `EmptyAction.instance()`, it would be nice to keep all of these in a similar manner of expression. (Same for above, in this test file, too.)

aws-api/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -0,0 +1,185 @@
...
+        @Override
+        public String toString() {
+            List<String> fieldsList = new ArrayList<>();
+            StringBuilder builder = new StringBuilder();
+
+            if (value != null) {
+                builder.append(value);
+            }
+
+            if (nodes != null && nodes.size() > 0) {
```
jamesonwilliams, line 155:
```java
if (!Empty.check(nodes)) {
```

aws-api/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -0,0 +1,185 @@
...
+
+        private Node(String value, Set<Node> nodes) {
+            this.value = value;
+            this.nodes = nodes;
+        }
+
+        /**
+         * Generate the String value of the SelectionSet used in the GraphQL query document.
+         *
+         * Sample return value:
```
jamesonwilliams, line 133: Nice docs

aws-api/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -0,0 +1,185 @@
...
+        private final String value;
+        private final Set<Node> nodes;
+
+        /**
+         * Copy constructor.
+         * @param node node to copy
+         */
+        Node(Node node) {
+            this.value = node.value;
+            this.nodes = new HashSet<>(node.nodes);
```
jamesonwilliams, line 122:
```suggestion
            this(value, new HashSset<>(node.nodes));
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+    public static AppSyncGraphQLRequest.Builder builder() {
+        return new Builder();
+    }
+
+    static final class Builder {
+        private Class<? extends Model> modelClass;
+        private OperationType operationType;
+        private Type responseType;
+        private Map<String, Object> variables;
+        private Map<String, String> variableTypes;
```
jamesonwilliams, line 258:
```suggestion
        private final Map<String, Object> variables;
        private final Map<String, String> variableTypes;
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    @Override
+    public String toString() {
+        return "AppSyncGraphQLRequest{" +
+                "modelSchema=" + modelSchema +
+                ", operationType=" + operationType +
+                ", selectionSet=" + selectionSet +
+                ", variables=" + variables +
+                ", variableTypes=" + variableTypes +
+                super.toString() +
```
jamesonwilliams, line 241: Hm, is this gonna work? It'll output something like `GraphQLRequest { ... }` in the middle. Would it be better to just access the `super.field`s directly to flatten them inline with these others?

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+                inputParameters.add(key + ": $" + key);
+            }
+
+            inputTypeString = Wrap.inParentheses(TextUtils.join(", ", inputTypes));
+            inputParameterString = Wrap.inParentheses(TextUtils.join(", ", inputParameters));
+        }
+
+        String modelName = modelSchema.getName();
+        String operationString = new StringBuilder()
+                .append(Casing.from(Casing.CaseType.SCREAMING_SNAKE_CASE).to(Casing.CaseType.CAMEL_CASE)
```
jamesonwilliams, line 186: In hindsight, I was having too much fun when I made this `Casing` utility. But I think my intention was that you could import `CaseType` directly, for the most readability:
```java
Casing.from(CaseType.SCREAMING_SNAKE_CASE, ...)
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Used to add owner argument from authProvider if needed.
+     * @param authProvider CognitoUserPoolsAuthProvider for obtaining the username to set as the owner field.
+     * @throws ApiException if request requires owner argument and authProvider or authProvider.getUsername() is null.
+     */
+    public void setAuthProvider(CognitoUserPoolsAuthProvider authProvider) throws ApiException {
+        for (AuthRule authRule : modelSchema.getAuthRules()) {
+            if (isOwnerArgumentRequired(authRule.getOperationsOrDefault())) {
+                setVariable(authRule.getOwnerFieldOrDefault(), "String!", getUsername(authProvider));
```
lawmicha, line 118: So I can't explain why we have to set `owner: String!` and the value on graphQL document/variables. We already intercept the request with the token which contains all of this information, so shouldn't the service already know how to retrieve the token and check if the user is authorized? Why do both? @SwaySway tried to explain this to me once and it had something to do with how subscriptions work. There is the initial connection that contains the auth token on the request header and then on each subscription message, it probably requires the `owner` value to be used as a filter to check if the data can be passed back to the subscriber?

So from trying out different auth rules with "allow: owner", it turns out, as long as the operation is specified then the correlating operation's subscription requires the owner value to be present. This is what you have written here. The callout here is that this does not work for all auth rules. I don't think this should block this PR from shipping as this is just a limitation that we can document and increase in functionality later. 

The problem is when there are auth rules with "allow: groups, groups: ["SomeGroup"]". 
```
type ApprovedBlog @model @auth(rules: [{allow: groups, groups: ["ApprovedBloggers"], operations: [read, create]}]) {
  id: ID!
  blogId: String
}
```
ApprovedBlog can only be read and created by users in the "ApprovedBloggers" group. (Uh, i'm not sure what this means for others not in the group, maybe they can update and delete ApprovedBlog objects if they have access to the id of the object? not sure)

None of the generated subscription operations require the owner field
```
onCreateApprovedBlog: ApprovedBlog
onUpdateApprovedBlog: ApprovedBlog
onDeleteApprovedBlog: ApprovedBlog
```
I have `user1` not in the group and `user2` in the group.

Both logged in users create a subscription without the owner:
```
subscription onCreaeApprovedBlog {
	onCreateApprovedBlog {
    id
    blogId
  }  
}
```

user1 fails with 
```
Error: {
    "errors": [
        {
            "message": "Connection failed: {\"errors\":[{\"errorType\":\"Unauthorized\",\"message\":\"Not Authorized to access onCreateApprovedBlog on type Subscription\"}]}"
        }
    ]
}
```
user2 is able to create a subscription connection.

Since models with static groups auth rules do not require the `owner` to be passed in, the check for `isOwnerArgumentRequired` should be: authRule's allow == owner && `SubscriptionType` match `ModelOperation`

aws-api/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -0,0 +1,185 @@
...
+                    result.add(new Node(fieldName, fields));
+                }
+            } else {
+                result.add(new Node(fieldName, null));
+            }
+            for (AuthRule authRule : schema.getAuthRules()) {
+                if (AuthStrategy.OWNER.equals(authRule.getAuthStrategy())) {
+                    result.add(new Node(authRule.getOwnerFieldOrDefault(), null));
+                    break;
+                }
```
lawmicha, line 106: on iOS, when generating the selection set, it's a bit more generic and only adds the fields from the Model schema to the selection set. We don't actually add it in based on the auth rule's owner field. The reason why we do not add it in is because it's up to the developer to decide whether that data is returned in their Model

For example
1. Developer has annotated the Model with Auth directive with "allow: owner" and by default it sets ownerField to "owner". They do not add it to their model like
```
type Model @model @auth { allow: owner } {
   id: ID!
   content: String
   // missing owner
}
```

Model only contains id and content, so if the selection set had the owner field added, it would just be dropped when deserializing
DataStore sync to cloud does custom deserialization to separate out into Model and MutationSyncMetadata (containinig version and deleted). Even if ownerField is added to the selection set, and the value is returned in the response, it would just be ignored

2. Developer sets it on their model
```
type Model @model @auth { allow: owner } {
   id: ID!
   content: String
   owner: String
}
```
Codegen generates Model with id, content, and owner. SelectionSet contains all three fields because it's generated from the Model's fields. Does not care about the auth rules

3. Developer sets a custom ownerField, but not on their model
```
type Model @model @auth { allow: owner, ownerField: "customOwner" } {
   id: ID!
   content: String
   // missing customOwner
}
```
I don't think a developer would ever do this since they are trying to reference a custom owner field in the Model but have not specified it in their model. If they want a custom owner field, they should specify it like in scenario 4 below

4. Developer sets a custom ownerField and adds it to the model
```
type Model @model @auth { allow: owner, ownerField: "customOwner" } {
   id: ID!
   content: String
   customOwner: String
}
```
Codegen generates Model with id, content, and customOwner. SelectionSet contains all three fields because it's generated from the Model's fields. Does not care about the auth rules.

**owner field on the Model should be optional**
when you specify the owner field on the Model, whether it's "owner" or "customOwner", it has to be optional since it's automatically managed (API intercept the request with the token). Making it required means the caller needs to add it in the request somehow.


For scenario 2 and 4, does that mean this code is adding duplicate owner fields to the selection set? I think this code can just be removed if it the selection set is already being constructed by model fields.

lawmicha: this may not be relevant on Android depending on how the graphql variables are generated from the model instance, but on iOS, for scenario 2 and 4, we would generate the graphQL variables based on the model instance. Since the model instance contains the owner field, requests with the input with null owner failed.

Model instance created like `let model = Model(id = "111", owner = nil)` would generate a graphql variables like
```
{
   "input": {
      "id": "111",
      "owner": null
   }
}
```

So, we actually [filter it out](https://github.com/aws-amplify/amplify-ios/blob/eb5b7ff98ae9ac090dc8c71e374b8f3bcb0ee472/AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDecorator.swift#L31)

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+            }
+        }
+    }
+
+    private String getUsername(CognitoUserPoolsAuthProvider authProvider) throws ApiException {
```
lawmicha, line 123: This is what iOS is currently doing as well, using the username as the identityClaim and we have opened a [feature request](https://github.com/aws-amplify/amplify-ios/issues/485) to support more than just the default identityClaim type when using the auth directive

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+        this.variables.put(key, value);
+        this.variableTypes.put(key, type);
+    }
+
+    /**
+     * Used to add owner argument from authProvider if needed.
+     * @param authProvider CognitoUserPoolsAuthProvider for obtaining the username to set as the owner field.
+     * @throws ApiException if request requires owner argument and authProvider or authProvider.getUsername() is null.
+     */
+    public void setAuthProvider(CognitoUserPoolsAuthProvider authProvider) throws ApiException {
```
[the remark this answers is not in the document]
richardmcclellan: Fair point, will change to `setOwner`.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+            throw new ApiException(
+                    "Attempted to subscribe to a model with owner based authorization without a Cognito provider",
+                    "Did you add the AWSCognitoAuthPlugin to Amplify before configuring it?"
+            );
+        }
+        String username = authProvider.getUsername();
+        if (username == null) {
+            throw new ApiException(
+                    "Attempted to subscribe to a model with owner based authorization without a username",
+                    "Make sure that a user is logged in before subscribing to a model with owner based auth"
```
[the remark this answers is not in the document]
richardmcclellan: Can't take credit, @TrekSoft wrote these!  I just moved them around a bit :).

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -255,7 +257,20 @@ public void configure(
...
+        if (graphQLRequest instanceof AppSyncGraphQLRequest) {
+            try {
+                AppSyncGraphQLRequest<R> request = (AppSyncGraphQLRequest<R>) graphQLRequest;
+                CognitoUserPoolsAuthProvider cognitoProvider = authProvider.getCognitoUserPoolsAuthProvider();
+                if (cognitoProvider == null) {
+                    cognitoProvider = new DefaultCognitoUserPoolsAuthProvider();
+                }
+                request.setAuthProvider(cognitoProvider);
+            } catch (ApiException exception) {
+                onSubscriptionFailure.accept(exception);
```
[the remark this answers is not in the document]
richardmcclellan: Good catch!   Except that the return type for this method is `GraphQLOperation`, not `Cancelable`, so I will just return `null` here.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+            );
+        }
+        return username;
+    }
+
+    private boolean isOwnerArgumentRequired(List<ModelOperation> operations) {
+        if (SubscriptionType.ON_CREATE.equals(operationType) && operations.contains(ModelOperation.CREATE)) {
```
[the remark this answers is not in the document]
richardmcclellan: :). I kinda went down a similar train of thought, but I do think this way is the most readable.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Constructor for AppSyncGraphQLRequest.
+     * @throw AmplifyException if a ModelSchema can't be derived from the provided model class
+     */
+    private AppSyncGraphQLRequest(Builder builder) throws AmplifyException {
+        super(builder.responseType, new GsonVariablesSerializer());
+        this.modelSchema = ModelSchema.fromModelClass(builder.modelClass);
+        SelectionSet.Node set = SelectionSet.fromModelClass(builder.modelClass, DEFAULT_LEVEL_DEPTH);
+        if (QueryType.LIST.equals(builder.operationType)) {
```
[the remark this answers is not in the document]
richardmcclellan: I like that idea.  Updated.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
richardmcclellan: It would work in the sense that it would print out all the properties, but I agree, with a weird nested `GraphQLRequest { }`.  I just updated it just print them all inline with the others.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
richardmcclellan: I'll go ahead and go a step further, and use static imports so I can just do this: `Casing.from(SCREAMING_SNAKE_CASE).to(PASCAL_CASE).convert(operationType.toString()))`

richardmcclellan: Hmm, our checkstyle rules don't allow static imports, so I'm gonna leave this as is for now.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
richardmcclellan: > Since models with static groups auth rules do not require the `owner` to be passed in, the check for isOwnerArgumentRequired should be: authRule's allow == owner && `SubscriptionType` match `ModelOperation`

Thanks for this clarification, I just added a check to make sure the owner argument is only added for the `owner` strategy.

> there are other related [issues](https://github.com/aws-amplify/amplify-ios/issues/576) to keep in mind which may or may not be relevant on Android related to having both "allow: owner" and "allow: groups"

I will keep this in mind, as my next task is to incorporate the auth directive into `DataStore`.

aws-api/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardmcclellan: > For scenario 2 and 4, does that mean this code is adding duplicate owner fields to the selection set?

Good question - no, it doesn't allow duplicates because the data structure is a `Set`.

That said, thank you for that explanation.  I agree we should just allow the developer to decide whether to include the `owner` field in the selection set, so I've removed the logic I had for adding it.

> So, we actually [filter the owner out](https://github.com/aws-amplify/amplify-ios/blob/eb5b7ff98ae9ac090dc8c71e374b8f3bcb0ee472/AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDecorator.swift#L31).

Thanks for sharing this.  I didn't have this before, but have added similar logic in this PR to filter out the owner field, if null, for mutations.

core/src/main/java/com/amplifyframework/util/Empty.java
```diff
@@ -49,9 +48,9 @@ public static boolean check(@Nullable Map<?, ?> map) {
      * Returns true if str is null or empty.
      *
      * TextUtils.isEmpty() provides the same behavior, but throws an exception when called from a unit test, since the
-     *  Android library is not available in unit tests.  Empty.check() is preferred to allow for unit testing.
-     *
-     * @see <a href="http://tools.android.com/tech-docs/unit-testing-support#TOC-Method-...-not-mocked.">documentation</a> for more details.
+     *  Android library is not available in unit tests.  Empty.check() is preferred to allow for unit testing.  See
+     *  <a href="http://tools.android.com/tech-docs/unit-testing-support#TOC-Method-...-not-mocked.">Android
+     *  tools documentation</a> for details.
```
jamesonwilliams: The "not mocked" error can be fixed by using `@RunWith(RobolectricTestRunner.class)`.

The Robolectric test runner includes a version of the Android APIs that can run on the host JVM -- not the phone's ART/JVM.

The downside is that Robolectric tests are slower.

To make this comment accurate, we should really go through hand update all references to `TextUtils.isEmpty`, and try to strip off the test-runners, to see if the tests will run after the change.

Overall it seems reasonable -- avoiding `TextUtils` basically means faster unit tests (_if_ the source code doesn't use any other Android-y stuff, and so Robolectric annotations could be removed from the class.)

core/src/main/java/com/amplifyframework/api/graphql/GraphQLRequest.java
```diff
@@ -86,73 +60,33 @@ public GraphQLRequest(
      * @param <R> The type of data contained in the GraphQLResponse expected from this request.
      * @return Copy of the GraphQLRequest object
      */
-    public <R> GraphQLRequest<R> copy() {
-        return new GraphQLRequest<R>(this);
-    }
+    public abstract <R> GraphQLRequest<R> copy();
```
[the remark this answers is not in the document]
richardmcclellan: Yeah, as you said, I think it's better to avoid `clone()` since I can't ensure subclasses will implement it.

aws-api/src/test/java/com/amplifyframework/api/aws/SelectionSetTest.java
richardmcclellan: I was trimming the expected value because `selection-set-post.txt` has a newline at the end of the file, which can't be removed.  I just eliminated the need for this by appending "\n" to the actual value.

I was trimming the actual value because it had a single space at the beginning of the String.  I just eliminated the need for this by modifying the contents of `selection-set-post.txt`

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
richardmcclellan: I just found `NoOpConsumer` :). I've updated to use it.

core/src/main/java/com/amplifyframework/util/Empty.java
richardmcclellan: Thanks for that explanation!   There were 3 other usages of `TextUtils.isEmpty` in the codebase, and I've updated them to use `Empty.check`.  As far as I can tell, those usages aren't covered by unit tests (unfortunately), so there aren't any related `Robolectric` annotations to be removed.

core/src/main/java/com/amplifyframework/api/graphql/OperationType.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.amplifyframework.api.graphql;
+
+/**
+ * Represents a GraphQL operation type.
+ */
+public interface OperationType {
+    /**
+     * Returns the name of the operation to be used in the GraphQL document (i.e. query, subscription, or mutation).
```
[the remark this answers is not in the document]
richardmcclellan: I like the idea to create an enum for the operations (query, mutation, subscription), so that the possible operations are constrained, so I've made that change.

As for the second part of this suggestion, I do see the benefit of guaranteeing that `Operation` is only implemented by an enum of a type constrained to a `OperationType`.   However, I think it's overly complicated for now, since we'd have to introduce a new set of objects (Query, Mutation, Subscription), that wrap the type enums (QueryType, MutationType, SubscriptionType).  Not opposed to exploring again in the future though.

richardmcclellan:
@jamesonwilliams I've addressed all your feedback now - could you take a final look?

lawmicha:
FYI https://github.com/aws-amplify/amplify-cli/issues/4182 logic for adding owner id may have changed, need to confirm

core/src/main/java/com/amplifyframework/api/graphql/Operation.java
```diff
@@ -0,0 +1,50 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.graphql;
+
+/**
+ * Represents a GraphQL operation.
+ */
+public enum Operation {
```
jamesonwilliams, line 21: Are these named in reverse? Should an `Operation` have a `getOperationType()`, which returns an `OperationType.SUBSCRIPTION`, etc.?

core/src/main/java/com/amplifyframework/api/graphql/SimpleGraphQLRequest.java
```diff
@@ -0,0 +1,118 @@
...
+     * @param document document String for request
+     * @param responseType Type of R, the data contained in the GraphQLResponse expected from this request
+     * @param variablesSerializer an object which can take a map of variables and serialize it properly
+     */
+    public SimpleGraphQLRequest(
+            String document,
+            Type responseType,
+            VariablesSerializer variablesSerializer
+    ) {
+        this(document, new HashMap<>(), responseType, variablesSerializer);
```
jamesonwilliams, line 43: `Collections.emptyMap()`

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -255,7 +257,21 @@ public void configure(
...
+            try {
+                AppSyncGraphQLRequest<R> request = (AppSyncGraphQLRequest<R>) graphQLRequest;
+                CognitoUserPoolsAuthProvider cognitoProvider = authProvider.getCognitoUserPoolsAuthProvider();
+                if (cognitoProvider == null) {
+                    cognitoProvider = new DefaultCognitoUserPoolsAuthProvider();
+                }
+                request.setOwner(cognitoProvider);
+            } catch (ApiException exception) {
+                onSubscriptionFailure.accept(exception);
+                return null;
```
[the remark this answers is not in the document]
richardmcclellan: All the others return `null`, so I'm leave this as is for now, but that could be done as a separate change.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -0,0 +1,299 @@
...
+            this.operationType = Objects.requireNonNull(operationType);
+            return Builder.this;
+        }
+
+        Builder responseType(@NonNull Type responseType) {
+            this.responseType = Objects.requireNonNull(responseType);
+            return Builder.this;
+        }
+
+        Builder setVariable(@NonNull String key, String type, Object value) {
```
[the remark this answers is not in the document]
richardmcclellan: > Or (the verb, 2p imperative) "do set" a thing called a variable? 

This one.  I agree, `.variable` makes sense, will change.

core/src/main/java/com/amplifyframework/api/graphql/Operation.java
richardmcclellan: I went back and forth on this as well, and can't remember why I landed where I did.  Thinking about it again now though, I think flipping them makes sense.  Will change.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -255,7 +257,21 @@ public void configure(
...
 
-        SubscriptionOperation<T> operation = SubscriptionOperation.<T>builder()
+        if (graphQLRequest instanceof AppSyncGraphQLRequest) {
+            try {
+                AppSyncGraphQLRequest<R> request = (AppSyncGraphQLRequest<R>) graphQLRequest;
+                CognitoUserPoolsAuthProvider cognitoProvider = authProvider.getCognitoUserPoolsAuthProvider();
+                if (cognitoProvider == null) {
+                    cognitoProvider = new DefaultCognitoUserPoolsAuthProvider();
+                }
+                request.setOwner(cognitoProvider);
```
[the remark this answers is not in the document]
richardmcclellan: I'm going to try and address this in https://github.com/aws-amplify/amplify-android/pull/635.

richardmcclellan:
Whoops, I merged this before pushing the changes to flip `Operation` / `OperationType`, and rename `setVariable to `variable`.  I've gone ahead and incorporated them into the Datastore PR (https://github.com/aws-amplify/amplify-android/pull/635) though, so they won't be lost.

Issue #589, opened 2020-06-17
Error on re-subscription


chiragmittal19:
After cancelling a subscription in `onDestroy` of `MainActivity` and restarting the application, if we subscribe again, the following error is thrown -

```console
AmplifyException {message=Subscription not acknowledged., cause=null, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}
        at com.amplifyframework.api.aws.SubscriptionEndpoint$Subscription.awaitSubscriptionReady(SubscriptionEndpoint.java:2)
        at com.amplifyframework.api.aws.SubscriptionEndpoint.requestSubscription(SubscriptionEndpoint.java:99)
        at com.amplifyframework.api.aws.SubscriptionOperation.lambda$start$0(SubscriptionOperation.java:4)
        at com.amplifyframework.api.aws.-$$Lambda$SubscriptionOperation$FXgdmzMG9ubNhC1f2NTwXyG53uo.run(Unknown Source:2)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:457)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
```

I assure you that there is no issue with internet connectivity.

Also, the same error is thrown when trying to subscribe after the internet connection is re-established.

Steps to reproduce -

1. Initialize Amplify and subscribe, in the `onCreate` method of activity.
2. Turn off the internet connection. `onSubscriptionFailure` is called.
3. Turn on the internet connection. Once the connection is established, then subscribe again.
4. After 10 seconds, the above mentioned error is thrown.

rjuliano:
Apologies for the delay on this. I am actively focused on a fix for this issue. Will provide an update ASAP.

rjuliano:
A new version (1.1.0) has just been released that includes a fix (a retry strategy with backoff when subscription fails) for this. If you encounter this error, feel free to re-open this issue.

Issue #581, opened 2020-06-17
App crashes when doing Amplify.DataStore.clear(...) after initilization


btdo:
Hi,

I get an exception when I try to clear the cache of the datastore right after initialization.  Here is my code 

```kotlin
try {
    val authProviders = ApiAuthProviders.builder()
        .oidcAuthProvider(AuthenticationServiceImpl)
        .build()
    Amplify.addPlugin(AWSDataStorePlugin())
    Amplify.addPlugin(AWSApiPlugin(authProviders)) // If using remote model synchronization
    Amplify.configure(applicationContext)
    Log.i("MyAmplifyApp", "Initialized Amplify")
    Amplify.DataStore.clear(Action { Timber.d("Cache Cleared") }, Consumer { Timber.e("Error while clearing cache") })
} catch (error: AmplifyException) {
    Log.e("MyAmplifyApp", "Could not initialize Amplify", error)
}.
```

Here is my exception:
```console
2020-06-15 21:51:25.086 31152-31216/com.wwm.trackappsyncapp E/AndroidRuntime: FATAL EXCEPTION: pool-2-thread-2
    Process: com.wwm.trackappsyncapp, PID: 31152
    java.lang.NullPointerException: next is null
        at io.reactivex.internal.functions.ObjectHelper.requireNonNull(ObjectHelper.java:39)
        at io.reactivex.Completable.andThen(Completable.java:1181)
        at com.amplifyframework.datastore.AWSDataStorePlugin.initialize(AWSDataStorePlugin.java:6)
        at com.amplifyframework.core.category.Category.initialize(Category.java:15)
        at com.amplifyframework.core.Amplify.lambda$beginInitialization$0(Amplify.java:1)
```

Thanks

eeatonaws:
I get the same exception when completing the "Integrate in your app" section of the Amplify Getting Started tutorial, specifically when attempting to save an item to the datastore. The code (from the tutorial) is:
` Amplify.DataStore.save(
         item,
         success -> Log.i("Tutorial", "Saved item: " + success.item().getName()),
         error -> Log.e("Tutorial", "Could not save item to DataStore", error)
 );`

rjuliano:
Apologies for the delay on this. I am actively focused on a fix for this issue. Will provide an update ASAP.

rjuliano:
A new version (1.1.0) has just been released that includes a fix for this. If you encounter this error, feel free to re-open this issue.

Issue #563, opened 2020-06-05
AWSDataStorePlugin fail to initialize and app crashes


btdo:
Hello,

I have a demo android app using Amplify Data store and once a while when I boot up the app it crashes and I would see an error in the log that seems to be related to the plugin.  If I start the app again then it would be fine.  

Here's is how I initialize the plugin:

```
 implementation 'com.amplifyframework:core:1.0.0'
 implementation 'com.amplifyframework:aws-api:1.0.0' // If using cloud sync
 implementation 'com.amplifyframework:aws-datastore:1.0.0'
```

```
try {
    val authProviders = ApiAuthProviders.builder()
        .oidcAuthProvider(MyOicAuthServiceImpl)
        .build()
    Amplify.addPlugin(AWSDataStorePlugin())
    Amplify.addPlugin(AWSApiPlugin(authProviders)) // If using remote model synchronization
    Amplify.configure(applicationContext)
    Log.i("MyAmplifyApp", "Initialized Amplify")
} catch (error: AmplifyException) {
    Log.e("MyAmplifyApp", "Could not initialize Amplify", error)
}
```

Here's the stack trace 

```
2020-06-05 16:41:33.786 8343-8407/com.wwm.trackappsyncapp E/AndroidRuntime: FATAL EXCEPTION: pool-2-thread-2
    Process: com.wwm.trackappsyncapp, PID: 8343
    java.lang.RuntimeException: java.util.concurrent.TimeoutException: The source did not signal an event for 5000 milliseconds and has been terminated.
        at io.reactivex.internal.util.ExceptionHelper.wrapOrThrow(ExceptionHelper.java:46)
        at io.reactivex.internal.observers.BlockingMultiObserver.blockingGetError(BlockingMultiObserver.java:153)
        at io.reactivex.Completable.blockingGet(Completable.java:1302)
        at com.amplifyframework.datastore.AWSDataStorePlugin.initialize(AWSDataStorePlugin.java:7)
        at com.amplifyframework.core.category.Category.initialize(Category.java:15)
        at com.amplifyframework.core.Amplify.lambda$beginInitialization$0(Amplify.java:1)
        at com.amplifyframework.core.-$$Lambda$Amplify$EB8h7rBH19uqUcPr0ZtYS0ww0r4.run(Unknown Source:4)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[...]

Issue #541, opened 2020-05-28
DataStore crashes app when device loses network connection


jamesonwilliams:
While running an application that uses the DataStore (with cloud synchronization _enabled_), I see the DataStore initialize successfully, and begin to process subscription data:
```
I/amplify:aws-datastore: Starting processing subscription data buffer.
```
At this point, my device is connected to the Internet.

When I place the device into Airplane Mode, however, the application _crashes_, with the exception below.

Instead, I would expect the DataStore to operate in an "offline" mode, where-in mutations are queued into the `PersistentMutationOutbox`, to be dispatched later. When the device regains its connectivity, within a reasonable amount of time (a minute?) the Sync Engine should re-form the subscriptions, and start draining the mutation outbox.

**Amplify Android Core, API, DataStore @ 1.0.0.**

```
E/AndroidRuntime: FATAL EXCEPTION: OkHttp https://REDACTED.appsync-realtime-api.us-east-1.amazonaws.com/...
    Process: com.amplifyframework.datastore.sample, PID: 27020
    io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | AmplifyException {message=Error during subscription., cause=AmplifyException {message=Subscription completion not acknowledged., cause=null, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}, recoverySuggestion=Evaluate details.}
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
        at io.reactivex.internal.operators.observable.ObservableCreate$CreateEmitter.onError(ObservableCreate.java:73)
        at com.amplifyframework.datastore.syncengine.-$$Lambda$e4FRR6iHUV0ToSpRUkOM1VPObso.accept(Unknown Source:4)
        at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$subscription$3(AppSyncClient.java:1)
        at com.amplifyframework.datastore.appsync.-$$Lambda$AppSyncClient$8gXTT9bgLZGP51blif6hkh7EwIc.accept(Unknown Source:4)
        at com.amplifyframework.api.aws.SubscriptionEndpoint$Subscription.awaitSubscriptionCompleted(SubscriptionEndpoint.java:3)
[...]

PR #599, opened 2020-06-25
DataStore race condition fix and other stability-related fixes

rjuliano:
*Issue #, if available:* #541 #581 Maybe: #589 #563 

*Description of changes:*
1 - Modified clear() operation so it doesn’t invoke beforeOperation like all the other API calls. This is because we need to stop the Orchestrator anyways before we can delete the database. After we remove the database and re-initialize the storage adapter, we then make a call to re-initialize the orchestrator asynchronously.

2 - Modified beforeOperation so it doesn't call initializeOrchestrator. My thought is that operations against the DataStore should have little to no dependency on what the orchestrator is currently doing (with the exception of clear). I believe this was in place as a mechanism to ensure the synchronization processes remain alive or are restarted as needed. I will put a separate PR together to fill this gap.

3 - Added a semaphore to the `Orchestrator` to prevent race conditions when calling `start()`. This shouldn't happen often except for cases where `clear()` is called immediately after Amplify initializes.

4 - The change in `SubscriptionProcessor` is probably the most abstract one of the bunch. It has to do with the issues mentioned above, which are related to how RxJava lifecycle rules work. I'm basically wrapping the envent emitter so I can check whether I can check whether the downstream subscriber has already been disposed, which happens when one of the AppSync subscriptions fails. I added a comment to `AmplifyDisposables` to go into more detail.

As I mentioned, there will be a follow-up to this PR to fill the gap I mentioned in item 2. A more long term solution would be to re-structure how we wire the streams, but it would require a fair amount of refactoring. I'm hoping this will help with the fatal app crashes in the short term.


jpignata:
Can we do a synchronous code review? Given that it's concurrency, I think that might help versus me leaving a bunch of reckons here.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -115,15 +119,21 @@ public boolean isStarted() {
...
      */
     @NonNull
-    public Completable start() {
-        // Only start if it's stopped.
-        if (!OrchestratorStatus.STOPPED.equals(status.get())) {
-            return initializationCompletable;
+    public synchronized Completable start() {
+        boolean permitAcquired = acquirePermit(OrchestratorAction.START);
+        // Only start if it's stopped AND if we can get a permit.
+        if (!permitAcquired || !status.compareAndSet(OrchestratorStatus.STOPPED, OrchestratorStatus.STARTING)) {
```
jpignata, line 125: Augustus De Morgan sez:

```suggestion
        if (!(permitAcquired && status.compareAndSet(OrchestratorStatus.STOPPED, OrchestratorStatus.STARTING)) {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -115,15 +119,21 @@ public boolean isStarted() {
      * @return A Completable operation to start the sync engine orchestrator
      */
     @NonNull
-    public Completable start() {
-        // Only start if it's stopped.
-        if (!OrchestratorStatus.STOPPED.equals(status.get())) {
```
jpignata, line 120: It seems like `OrchestratorStatus` needs some helpers like `isStopped`, `isStarting`, etc.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -147,25 +157,50 @@ public Completable start() {
...
             storageObserver.stopObservingStorageChanges();
             mutationProcessor.stopDrainingMutationOutbox();
             status.compareAndSet(OrchestratorStatus.STOPPING, OrchestratorStatus.STOPPED);
             LOG.debug("Stopped remote synchronization.");
+        } finally {
+            startStopSemaphore.release();
         }
+    }
 
+    private boolean acquirePermit(OrchestratorAction action) {
```
jpignata, line 192: Can we wrapped the `compareAndSet` operation in the same method that grabs the semaphore so it can release it as needed?

rjuliano:
> Can we do a synchronous code review? Given that it's concurrency, I think that might help versus me leaving a bunch of reckons here.

We can definitely do that

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -449,7 +450,10 @@ private Completable initializeOrchestrator() {
         if (api.getPlugins().isEmpty()) {
             return Completable.complete();
         } else {
-            return orchestrator.start();
+            // Let's prevent the orchestrator startup from possibly running in main.
+            return orchestrator.start()
+                .observeOn(Schedulers.io())
+                .subscribeOn(Schedulers.io());
```
richardissuperman, line 459: I saw this pattern in multiple place in Amplify Android's source code. Is there any reason why we want perform observeOn and subscribeOn at the same place with the same scheduler?

Should'nt 

```
orchestrator.start()
                  .subscribeOn(Schedulers.io());
```

Do the work? Unless you were using a Subject which may change the threading so you have to reset the scheduler in observeOn but i dont see any Subject in orchestrator

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -425,19 +426,22 @@ public void observe(
     @Override
     public void clear(@NonNull Action onComplete,
                       @NonNull Consumer<DataStoreException> onError) {
-        beforeOperation(() -> {
-            orchestrator.stop();
-            sqliteStorageAdapter.clear(onComplete, onError);
-        });
+        // We shouldn't call call beforeOperation when clearing the DataStore. The
```
jamesonwilliams, line 429: nit: `call call`

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: The thinking behind that was to ensure it was not in main since we have the occasional blockingAwait/Get call throughout the process. Not sure if it makes a difference, but the subscriber of the `SubscriptionProcessor` is a `ReplaySubject`

richardissuperman: @rjuliano yeah then i dont think its needed to call observeOn again, it will just wrap the lower stream runnable into the same thread pool. 

 we only need to switch to same scheduler if we have below use case , which is to transform a callback style code into rxjava observable 

```
 Observable.<StorageItemChange<? extends Model>>create(emitter ->
                localStorageAdapter.observe(emitter::onNext, emitter::onError, emitter::onComplete)
            )
            .subscribeOn(Schedulers.single())
            .observeOn(Schedulers.single())

```
The above code makes sense cause the emitter will be emitting the onNext event in a different thread other than the one we specify in Schedulers.single(), so it makes sense to quickly switch back to it by calling single,

similar use case : https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L137

nit: the above link code i think it need to be changed as well as the onErrorResume() will be executed not in IO() thread scheduler

rjuliano: That makes sense. That's the change you made in #612 right @richardmcclellan ?

richardmcclellan: I think you meant @richardissuperman, not me :)

richardissuperman: @rjuliano yep, exactly, but that PR was messed up with my previous commit, will revise all two later tonight

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -117,14 +117,19 @@ void startSubscriptions() {
...
                 emitter::onNext,
-                emitter::onError,
+                AmplifyDisposables.onErrorConsumerWrapperFor(emitter),
                 emitter::onComplete
             );
             // When the observable is disposed, we need to call cancel() on the subscription
             // so it can properly dispose of resources if necessary. For the AWS API plugin,
             // this means means closing the underlying network connection.
             emitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable));
-            latch.await(SUBSCRIPTION_START_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
jamesonwilliams, line 127: Even before, we weren't checking the return value here -- wether or not the latch actually counted down. Should we be, now?

aws-datastore/src/main/java/com/amplifyframework/datastore/AmplifyDisposables.java
```diff
@@ -61,4 +71,47 @@ public boolean isDisposed() {
...
+     *                    value of this function.
+     * @param <T> The type of model handled by the emitter.
+     * @param <E> The type of exception for the onError consumer
+     * @return A {@link Consumer} that proxies the {@link ObservableEmitter#onError(Throwable)} call
+     * to the {@code realEmitter} if it's not disposed or logs a warning.
+     * @see <a href="https://github.com/aws-amplify/amplify-android/issues/541">GitHub issue #541</a>
+     *
+     */
+    @NonNull
+    public static <T extends Model, E extends AmplifyException> Consumer<E> onErrorConsumerWrapperFor(
```
jamesonwilliams, line 107: Is it an `AmplifyDisposable`s thing, really? It doesn't build a `Disposable`. If you only use it from the `SubscriptionProcessor`, maybe it should live as a static utility method on that class, instead?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -425,19 +426,22 @@ public void observe(
...
+        // only thing we have to wait for is the category initialization latch.
+        try {
+            categoryInitializationsPending.await();
```
jamesonwilliams, line 432: Should we be checking the return value here to tell whether or not the latch did count down? What are the appropriate actions to take if so, vs. if not?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -425,19 +426,22 @@ public void observe(
...
+        } catch (InterruptedException exception) {
+            LOG.warn("Execution interrupted while waiting for DataStore to be initialized.");
+        }
+        orchestrator.stop();
+        sqliteStorageAdapter.clear(() -> {
+            initializeOrchestrator().subscribe();
+            onComplete.call();
```
jamesonwilliams, line 439: Should `onComplete.call()` be fired _as a result_ of the subscription, when the initialization of the orchestrator completes, later? That is,
```java
initializeOrchestrator().subscribe(onComplete::call);
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -147,25 +149,56 @@ public Completable start() {
...
             storageObserver.stopObservingStorageChanges();
             mutationProcessor.stopDrainingMutationOutbox();
             status.compareAndSet(OrchestratorStatus.STOPPING, OrchestratorStatus.STOPPED);
             LOG.debug("Stopped remote synchronization.");
+        } finally {
+            startStopSemaphore.release();
         }
+    }
 
+    private boolean performAction(OrchestratorAction intendedAction) {
```
jamesonwilliams, line 176: Can you get rid of the `OrchestratorAction` entirely, if you make this:
```java
private boolean transitionToState(OrchestratorState targetState) {
``` 
?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -147,25 +149,56 @@ public Completable start() {
...
+            LOG.debug(String.format("Trying to %s the orchestrator.", intendedAction.name()));
+            boolean permitAcquired = startStopSemaphore.tryAcquire(ACQUIRE_PERMIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            if (!permitAcquired) {
+                LOG.warn(String.format("Unable to acquire permit to %s the orchestrator. ", intendedAction.name()));
+                return false;
+            }
+            boolean statusSet = status.compareAndSet(intendedAction.expectedCurrentStatus, intendedAction.targetStatus);
+            // only stop if it's started AND if we can get a permit.
+            if (!statusSet) {
+                LOG.warn(String.format("Failed %s orchestrator. Current status: %s",
```
jamesonwilliams, line 189: This might be an error, if the user is expecting cloud sync to function? Should we wire this back to the user error handler somehow, too?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Cool...that's super helpful and thanks for updating the PR. I just merged it.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: I have an isStarted right now. That being said, I did end up changing how we do this a little bit: rather than check the start and perform action, I'm using the `compareAndSet` method of the `AtomicReference`.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Let me think about this one. Just realized also that we don't have a timeout on this await. We'll need to do some testing on this though.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Done

aws-datastore/src/main/java/com/amplifyframework/datastore/AmplifyDisposables.java
jamesonwilliams: ❤️ was granted for the wiki link! :-D

rjuliano:
### Latest round of changes

- Orchestrator now starts asynchronously and no longer blocks interactions with the local DataStore. Only exception is for `DataStore.clear()` since we need to stop the orchestrator before we can delete the database.
- Increased the timeout when waiting to acquire a permit to start/stop the orchestrator.
- Created a RetryStrategy class with a simple exponential backoff strategy for now. This is probably something we can iterate on, but this is a starting point.
- Refactored the `SubscriptionProcessor` since that appeared to be the culprit of a few unexpected crash issues reported by the users. We now have a retries when there are problems with the subscriptions. I anticipate I'll need to do something similar for `MutationProcessor`, but I'd like to do that in a separate PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: I added a timeout and it's also now checking the result of the await operation.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: I added some logic here to invoke the user-provided onError callback (if one is provided)

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
```diff
@@ -248,5 +249,25 @@ public SyncConfigurator mockSuccessResponses() {
...
+        public <T extends Model> SyncConfigurator mockFailure() {
+            doAnswer(invocation -> {
+                final int errorConsumerPosition = 3;
+                final Consumer<DataStoreException> consumer = invocation.getArgument(errorConsumerPosition);
+                consumer.accept(new DataStoreException("Something timed out during sync.", "Nothing to do."));
+                return new NoOpCancelable();
+            }).when(appSync).sync(
+                any(), // Item class to sync
+                any(), // last sync time
+                any(), // Consumer<Iterable<ModelWithMetadata<T>>>
```
jamesonwilliams, line 267: Do you have `anyConsumer()` available here? I forget where that utility lives.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
```diff
@@ -248,5 +249,25 @@ public SyncConfigurator mockSuccessResponses() {
             );
             return SyncConfigurator.this;
         }
+
+        /**
+         * Triggers an exception when invoking the sync method.
+         * @param <T> Type of models for which a response is mocked
+         * @return The same Configurator instance, to enable chaining of calls
+         */
+        public <T extends Model> SyncConfigurator mockFailure() {
```
jamesonwilliams, line 258: Can you change to accept a `DataStoreException`, and then pass in the failure when you make the call? That way, the `AppSyncMocking` can remain as utility software, with any specific values codified external to its implementation.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -242,6 +247,8 @@ private void assertRemoteSubscriptionsCancelled() {
     }
 
     private void assertRemoteSubscriptionsStarted() {
+        // Introduce a delay to allow the orchestrator to start since this is done in a non-blocking manner now.
+        Single.just(true).delay(ASSERTION_TIMEOUT_MS, TimeUnit.MILLISECONDS).blockingGet();
```
jamesonwilliams, line 251: We should keep this event driven, even from test. There is an initialization event that is published, currently. The intent of that event was to prevent the need for waits, in test. Between it, and the `HubAccumulator`, you should be able to await the initialization results. It's a more determinist mechanism than tuning `ASSERTION_TIMEOUT_MS`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -113,9 +114,18 @@ Completable hydrate() {
                         .flatMapCompletable(merger::merge)
                     )
                     .andThen(syncTimeRegistry.saveLastSyncTime(modelClass, SyncTime.now()))
-                    .doOnError(failureToSync ->
-                        LOG.warn("Initial cloud sync failed.", failureToSync)
-                    )
+                    .doOnError(failureToSync -> {
+                        LOG.warn("Initial cloud sync failed.", failureToSync);
+                        DataStoreErrorHandler dataStoreErrorHandler =
+                            dataStoreConfigurationProvider.getConfiguration().getDataStoreErrorHandler();
```
jamesonwilliams, line 120: Can `getConfiguration()` ever return `null`? Is there a need to guard against that scenario, here, to prevent a `NullPointerException`?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
rjuliano: It's a private static function buried in [SubscriptionProcessorTest](https://github.com/aws-amplify/amplify-android/blob/6a4848025c79e928a9ed3d69267a924628241188/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessorTest.java#L231) It probably should be moved elsewhere.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
rjuliano: Done

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: That makes sense. I just changed it so the orchestrator emits a hub event on start and stop. Then modified the test to use the HubAccumulator. It will be in my next commit.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: I don't think so. That parameter comes from the [plugin's constructor](https://github.com/aws-amplify/amplify-android/blob/074fae16941849ecac7b83cfe3bb04b50bcb4a54/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L101) and passes the plugin configuration that's initialized [here](https://github.com/aws-amplify/amplify-android/blob/074fae16941849ecac7b83cfe3bb04b50bcb4a54/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L174). Since the orchestrator is only started after the plugin is configured, I think we're safe here.

jamesonwilliams:
I was able to verify #541 is fixed. The others may be as well. Subscriptions are a lot more stable at this point. Ship it.

PR #600, opened 2020-06-25
fix(aws-api): Fix path with leading slash breaking REST

raphkim:
*Issue #, if available:*

*Description of changes:*
* Re-enable the REST API integration tests that were previously turned off.
* REST URL constructor will now strip leading slashes in provided path to avoid the [following OkHttp behavior](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.Builder.html#addPathSegments-java.lang.String-):
  * > If `pathSegments` starts with a slash, the resulting URL will have empty path segment.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
@@ -51,17 +51,16 @@ public final class RestRequestFactory {
             @Nullable String path,
             @Nullable Map<String, String> queryParameters
     ) throws MalformedURLException {
-        // If path isn't specified, then use empty string
-        if (path == null) {
-            path = "";
-        }
 
         URL url = new URL(endpoint);
         HttpUrl.Builder builder = new HttpUrl.Builder()
             .scheme(url.getProtocol())
             .host(url.getHost())
-            .addPathSegment(stripLeadingSlashes(url.getPath()))
-            .addPathSegments(stripLeadingSlashes(path));
+            .addPathSegment(stripLeadingSlashes(url.getPath()));
+
+        if (path != null) {
+            builder.addPathSegments(stripLeadingSlashes(path));
+        }
 
         if (queryParameters != null) {
             for (Map.Entry<String, String> entry : queryParameters.entrySet()) {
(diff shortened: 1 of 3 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
```diff
@@ -125,12 +129,20 @@ public static Request createRequest(
...
         if (data != null) {
             strategy.buildRequest(builder, data);
         }
     }
 
+    // Segment separator can be either '/' or '\'.
+    // HttpUrl.Builder assumes an empty URL if path segments
+    // begin with either character. Strip them before appending.
+    private static String stripLeadingSlashes(final String path) {
+        return path.replaceAll("^[\\\\/]+", "");
```
jpignata, line 143: Suggestion for readability (eschew regular expression):

```suggestion
        if (path.startsWith("/") || path.startsWith("\\")) {
            return path.substring(1);
        }
        
        return path;
```

aws-api/src/test/java/com/amplifyframework/api/aws/utils/RestRequestFactoryTest.java
```diff
@@ -83,9 +97,9 @@ public void createValidURLWithQuery() throws MalformedURLException {
      * @throws MalformedURLException Should throw since the URL is invalid.
      */
     @Test(expected = MalformedURLException.class)
-    public void createInValidURL() throws MalformedURLException {
+    public void createInvalidURL() throws MalformedURLException {
```
jpignata, line 100: Why is this invalid? No scheme?

aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
raphkim: I know this is excessive, but it works for any combination of any number of slashes in the beginning. This way I *REALLY* ensure that the path doesn't begin with "/" or "\" or any combination of those two.

This way the method name actually makes sense, since it wants to completely strip the left side of any slashes.

aws-api/src/test/java/com/amplifyframework/api/aws/utils/RestRequestFactoryTest.java
raphkim: I did not write this test, but I think OkHttp doesn't like it when the endpoint is missing the http

jpignata:
Should we use package names as scopes (e.g. aws-api) or category names? (e.g. API)

Issue #590, opened 2020-06-19
Temporal date types don't work with DataStore category


tomasfelder:
Hi,
I'm having an issue while using the AWS DataStore. I need my datastore to sync with the cloud. In my schema I have some types that have a field of `AWSDateTime` data type. When I generate my models for android in Java it creates the objects and the `AWSDateTime` is parsed to a `Temporal.DateTime` object. I create the object and it is stored correctly to the de DataStore, but the problem occurs when it wants to sync it with the cloud. It seems like the automatic generated mutation is not able to parse the `DateTime` object correctly and it fails. I get the following error:

```console
AmplifyException {message=Mutation failed. Failed mutation = PendingMutation{mutatedItem=TecpronProject {id=0becde35-31e2-408e-bbab-6607a30162ed, name=TEC2020, startDate=Temporal.DateTime{offsetDateTime='2020-06-19T13:20:37.951Z'}}, classOfMutatedItem=class com.amplifyframework.datastore.generated.model.TecpronProject, mutationType=CREATE, mutationId=aa6cb76d-b22f-11ea-9c05-873b0e7d83c4, predicate=null}. AppSync response contained errors = [GraphQLResponse.Error{message='Variable 'startDate' has an invalid value. Unable to parse `{offsetDateTime={dateTime={date={day=19, month=6, year=2020}, time={hour=13, minute=20, nano=951000000, second=37}}, offset={totalSeconds=0}}}` as a valid DateTime Object.', locations='[GraphQLLocation{line='1', column='31'}]', path='null', extensions='null'}], cause=null, recoverySuggestion=Verify that your AppSync endpoint is able to store TecpronProject models.}
        at com.amplifyframework.datastore.syncengine.MutationProcessor.lambda$null$11(MutationProcessor.java:7)
        at com.amplifyframework.datastore.syncengine.-$$Lambda$MutationProcessor$w1EW1gU2W9mWN5QztizneJ-zhAE.accept(Unknown Source:8)
        at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$mutation$4$AppSyncClient(AppSyncClient.java:2)
        at com.amplifyframework.datastore.appsync.-$$Lambda$AppSyncClient$nKAFu77xjCuzuWw9lR7faFIdN9U.accept(Unknown Source:8)
        at com.amplifyframework.api.aws.AppSyncGraphQLOperation$OkHttpCallback.onResponse(AppSyncGraphQLOperation.java:16)
[...]

PR #601, opened 2020-06-25
Create aws-appsync module, fix Datastore serialization bug (part 1)

richardmcclellan:
#### Summary
- Created `aws-appsync` module to hold AppSync specific logic, such as for request serialization and deserialization.   Depended on by `aws-api` and `aws-datastore`.
- Fixed part of date serialization bug (https://github.com/aws-amplify/amplify-android/issues/590).    The API category supports serialization and deserialization of the date serialization types, but Datastore does not.  This PR adds support for serializing dates.  Deserializing dates will be added in a subsequent PR.


aws-appsync/build.gradle
```diff
@@ -0,0 +1,13 @@
+apply plugin: 'com.android.library'
```
jamesonwilliams, line 1: Nit: add header

aws-datastore/src/test/resources/update-meeting.json
```diff
@@ -0,0 +1,14 @@
+{"query":"mutation UpdateMeeting($input:UpdateMeetingInput!) {
```
jamesonwilliams, line 1: Hm, I don't think this is valid JSON.

Maybe either - 

1. Rename the file to `update-meeting.txt`, or;
2. Rework the string value here so that the embedded GraphQL is properly escaped? (at least `{`, possibly other chars?)

aws-appsync/build.gradle
richardmcclellan: Added!

aws-datastore/src/test/resources/update-meeting.json
richardmcclellan: Renamed to `update-meeting.txt`

richardmcclellan:
@jamesonwilliams I just moved `Temporal`, and `JavaFieldType` from `core` to `aws-appsync`.  I will plan on moving more from API in a subsequent PR.

jpignata:
Idle thought: There's a convention to our modules that @raphkim  once explained to me. This is something we should document. I think it's provider-category-service. I know this kind of straddles categories, but perhaps it would be better off named differently so readers don't think it's a plugin in? Or, go the full nine as `aws-api-appsync`.

aws-appsync/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
jpignata, line 1: Nit: Several of these warning glyphs.

aws-appsync/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=appsync
+POM_NAME=Amplify Framework for Android - AppSync
+POM_DESCRIPTION=Amplify Framework for Android - AppSync
```
jpignata, line 3: We can likely be more descriptive here. We should discuss its usage - shared components used by x, y categories for interacting with AWS AppSync.

richardmcclellan: updated

aws-appsync/.gitignore
richardmcclellan: updated

richardmcclellan:
> Idle thought: There's a convention to our modules that @raphkim once explained to me. This is something we should document. I think it's provider-category-service. I know this kind of straddles categories, but perhaps it would be better off named differently so readers don't think it's a plugin in? Or, go the full nine as `aws-api-appsync`.

Thanks for the naming suggestion.  I think `aws-api-appsync` is better than `aws-appsync`, so I've renamed it.

PR #608, opened 2020-06-26
ci: improve integration test retry logic

richardmcclellan:
With these changes, I was able to get a green checkmark on the `integrationtest` step 4 out of 5 times.  The [one failure](https://app.circleci.com/pipelines/github/aws-amplify/amplify-android/1915/workflows/351584cc-d22c-42e4-a9a3-bfc061a5909c/jobs/3716/steps) was because the emulator didn't start.

 - Integration tests now run one module at a time (e.g. `./gradlew aws-api:connectedDebugAndroidTest` instead of `./gradlew connectedDebugAndroidTest`, and retry logic is applied at the module level.  So, if an `aws-api` integration test fails, it only retries `aws-api`, instead of the whole suite. 
 - Updated hash for codecov.io/bash script, since there was an update to the script.

I have a hunch that the main reason for the improvement here is actually not because of modularizing retries, but because the tasks are not running in parallel anymore. 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ jobs:
             .circleci/copy-configs
       - run:
           name: Run integration test
-          command: .circleci/retry.sh 3 ./gradlew connectedDebugAndroidTest --no-daemon
+          command: .circleci/run-integration-tests.sh
       - run:
           name: Store coverage data
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,10 +91,9 @@ jobs:
       - run:
           name: Build the project
           command: |
-            .circleci/retry.sh 3 900 ./gradlew build
+            .circleci/retry.sh 3 ./gradlew build
             .circleci/aggregate-test-results.sh ~/amplify-android-home
             .circleci/aggregate-build-artifacts.sh
-          no_output_timeout: 30m
       - store_test_results:
           path: test-results
       - store_artifacts:
@@ -120,11 +119,10 @@ jobs:
       - run:
           name: Run integration test
           command: .circleci/run-integration-tests.sh
-          no_output_timeout: 30m
       - run:
           name: Store coverage data
           command: |
-              .circleci/retry.sh 3 900 ./gradlew jacocoFullReport --no-daemon
+              .circleci/retry.sh 3 ./gradlew jacocoFullReport --no-daemon
               tmp_file="$(mktemp)"
               curl -s https://codecov.io/bash > "$tmp_file"
               actual_hash="$(cat $tmp_file | cksum | cut -f 1 -d\ )"
@@ -138,7 +136,7 @@ jobs:
               chmod +x "$tmp_file"
               bash "$tmp_file"
               rm "$tmp_file"
-          no_output_timeout: 30m
+
       - store_artifacts:
           path: logcat.log
 workflows:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -118,15 +118,15 @@ jobs:
...
+          command: .circleci/run-integration-tests.sh
       - run:
           name: Store coverage data
           command: |
               .circleci/retry.sh 3 ./gradlew jacocoFullReport --no-daemon
               tmp_file="$(mktemp)"
               curl -s https://codecov.io/bash > "$tmp_file"
               actual_hash="$(cat $tmp_file | cksum | cut -f 1 -d\ )"
-              expected_hash='2175291078'
+              expected_hash='261829867'
```
jpignata, line 129: So this might change more than we expected. Should we vendor this? What's the right (and secure) way? @jamesonwilliams

.circleci/retry.sh
```diff
@@ -1,23 +1,25 @@
 #!/bin/bash
 # Usage: retry.sh <max_tries> <command to run>
 
+tag="RETRY"
 readonly max_tries=$1
 readonly command="${@: 2}"
```
jpignata, line 6: I know this wasn't your code, but I think you need parentheses here. See [shellcheck's wiki](https://github.com/koalaman/shellcheck/wiki/SC2124). Might be a good thing to run on shell scripts we modify.

.circleci/run-integration-tests.sh
```diff
@@ -0,0 +1,22 @@
...
+return_code=0
+
+# List all available gradle tasks, grep for the integration test tasks, and then use cut to strip the task description
+# and just return the name of the task, one for each module (e.g. aws-api:connectedDebugAndroidTest)
+tasks=($(./gradlew tasks --all | grep connectedDebugAndroidTest | cut -d " " -f 1))
+
+# Run the integration tests for each module.  Make up to 3 attempts on each module before failing.
+for task in "${tasks[@]}"; do
+  echo "$tag: Starting $task"
+  .circleci/retry.sh 3 ./gradlew $task --no-daemon
```
jpignata, line 13: Nit: missing double quotes on the variable

richardmcclellan:
> Worth a shot! I don't quite get your hunch, but I have no way to prove or disprove it. I'm like a caveman looking at the sun and speculating about its internals.

The hunch is really only based on observing all the builds for each commit in this PR.  With everything running serially, I'm seeing lots of successes which the retry isn't even needed, the tests just seem to pass on the first attempt more frequently.

.circleci/run-integration-tests.sh
richardmcclellan: Fixed!

.circleci/retry.sh
richardmcclellan: It is my code actually, I'm just not an expert with bash scripts :).  Thanks for sharing the `shellcheck` tool - I just used it to verify the problem and confirm a fix!

.circleci/config.yml
richardmcclellan: Yeah, I had that thought too.  I went ahead and copied it into our repo for now.  The downside is it will probably require manual updates, but not sure of another option that is safe.

richardmcclellan: I made a separate PR for this: https://github.com/aws-amplify/amplify-android/pull/609

PR #612, opened 2020-06-27
fix(DataStore): remove unnecessary observeOn call

richardissuperman:
calling observeOn with the same schedulers wont change the below rxjava stream's running thread.


richardissuperman:
> Looks like your other unrelated changes from #613 got added accidentally to this current #612?

yeah not sure why, i was using two different branch for these 2 PR, let me check and revise them. :)

richardissuperman:
@jamesonwilliams turns out i push that wrong commit in this branch, fixed it already.

PR #613, opened 2020-06-28
fix(DataStore): use a LinkedList and HashMap to improve the time comp…

richardissuperman:
Previously we need to traverse the entire mutation list to perform lots of pending mutation operation, which results in a O(N) time complexity
1. updateExistingQueueItemOrAppendNew
2. removeFromQueue
3. findPendingMutationById

Use above mentioned data structure to optimise them , which uses O(1) time complexity

All tests passed.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
@@ -1,0 +1,219 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.syncengine;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.amplifyframework.core.model.Model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The {@link MutationQueue} is a LinkedHashMap like container , the goal of using this container is to
+ * achieve O(1) time complexity for both getting a {@link PendingMutation} and update an existing mutation with
+ * valid id.
+ * MutationQueue is implementing the Queue interface and provide most of the queue operations,
+ */
+public final class MutationQueue {
+
+    private final Map<TimeBasedUuid, Node> mutationMap = new HashMap<>();
+    private final Node dummyHead;
+    private final Node dummyTail;
+
+    /**
+     * Default constructor for {@link #MutationQueue}.
+     */
+    public MutationQueue() {
+        dummyHead = new Node();
+        dummyTail = new Node();
+        dummyHead.next = dummyTail;
+        dummyTail.prev = dummyHead;
+    }
+
+    /**
+     * Find the first Pending Mutation which its model has the same id.
+     *
+     * @param modelId the model id
+     * @return the {@link PendingMutation} instance
+     */
+    synchronized PendingMutation<? extends Model> nextMutationForModelId(String modelId) {
+        Node head = dummyHead.next;
+        if (head == dummyTail) {
+            return null;
+        }
+        while (head != dummyTail) {
+            if (head.mutation.getMutatedItem().getId().equals(modelId)) {
+                return head.mutation;
+            }
+            head = head.next;
+        }
+        return null;
+    }
+
+    /**
+     * Remove the {@link PendingMutation} from {@link MutationQueue} by its Id.
+     * this operation should be consuming constant time.
+     *
+     * @param timeBasedUuid the {@link TimeBasedUuid} UUID of a PendingMutation in @{@link #mutationMap}
+     * @return {@link Boolean} return true if remove is successful
+     */
+    synchronized boolean removeById(TimeBasedUuid timeBasedUuid) {
+        if (!mutationMap.containsKey(timeBasedUuid)) {
+            return false;
+        }
+        Node removingNode = mutationMap.get(timeBasedUuid);
+        Node removingPrev = removingNode.prev;
+        Node removingNext = removingNode.next;
+        //remove from map
+        mutationMap.remove(timeBasedUuid);
+        //remove from current linkedlist
+        removingPrev.next = removingNext;
+        removingNext.prev = removingPrev;
+        return true;
+    }
+
+    /**
+     * Add a {@link PendingMutation} to the tail of this Queue.
+     *
+     * @param pendingMutation {@link PendingMutation}
+     * @return true if successfully added a pending mutation
+     */
+    private synchronized boolean addToTail(@NonNull PendingMutation<? extends Model> pendingMutation) {
+        //constructing a new node
+        Node pendingNode = new Node();
+        pendingNode.id = pendingMutation.getMutationId();
+        pendingNode.mutation = pendingMutation;
+        //insert the new node into the current linkedlist
+        Node addingPrev = dummyTail.prev;
+        addingPrev.next = pendingNode;
+        dummyTail.prev = pendingNode;
+        pendingNode.prev = addingPrev;
+        pendingNode.next = dummyTail;
+        //put the mutation into the mutationmap
+        mutationMap.put(pendingMutation.getMutationId(), pendingNode);
+        return true;
+    }
+
+    /**
+     * Replace an existing {@link PendingMutation} inside the Queue.
+     *
+     * @param timeBasedUuid              the UUID of a pending mutation
+     * @param pendingMutation the pending mutation's instance
+     */
+    synchronized void updateExistingQueueItemOrAppendNew(@NonNull TimeBasedUuid timeBasedUuid,
+                                                         @NonNull PendingMutation<? extends Model> pendingMutation) {
+        // If there is already a mutation with same ID in the queue,
+        // we'll go find it, and then update it, with this contents.
+        if (mutationMap.containsKey(timeBasedUuid)) {
+            mutationMap.get(timeBasedUuid).mutation = pendingMutation;
+        } else {
+            // Otherwise, just add it to the end of the queue.
+            addToTail(pendingMutation);
+        }
+    }
+
+    /**
+     * Get a {@link PendingMutation} instance by its UUID from {@link #mutationMap}.
+     *
+     * @param timeBasedUuid the UUID of a pending mutation
+     * @return an instance of {@link PendingMutation}
+     */
+    @Nullable // When there is no match.
+    synchronized PendingMutation<? extends Model> getMutationById(TimeBasedUuid timeBasedUuid) {
+        if (mutationMap.containsKey(timeBasedUuid)) {
+            return mutationMap.get(timeBasedUuid).mutation;
+        }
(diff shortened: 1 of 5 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -137,16 +132,7 @@ public boolean hasPendingMutation(@NonNull String modelId) {
     }
 
     private <T extends Model> void updateExistingQueueItemOrAppendNew(PendingMutation<T> pendingMutation) {
```
jamesonwilliams, line 134: After your refactor, you have several wrapper methods left inside of `PersistentMutationOutbox`, which are ~one line long. I would remove these, and just call your dat astructure directly, instead of through these wrapper methods:
- `updateExitingQueueItemOrAppendNew`
- `nextMutationForModelId`
- `removeFromQueue`
- `findPendingMutationById`

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -55,15 +55,15 @@
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
 
     private final LocalStorageAdapter storage;
-    private final LinkedList<PendingMutation<? extends Model>> mutationQueue;
+    private final MutationLinkedList mutationQueue;
```
jamesonwilliams, line 58: How about `private final MutationQueue mutationQueue`, as the data structure's name.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationLinkedList.java
```diff
@@ -0,0 +1,112 @@
...
+
+import com.amplifyframework.core.model.Model;
+import java.util.HashMap;
+
+/*
+ * The {@link MutationLinkedList} is a LinkedHashMap like container , the goal of using this container is to
+ * achieve O(1) time complexity for both getting a {@link PendingMutation} and update an existing mutation with
+ * valid id.
+ */
+public class MutationLinkedList {
```
jamesonwilliams, line 11: Should you implement any of the standard Java contracts? `extends LinkedHashMap`,  `implements List`, ... `implements Comparable<T>`, implements `Collection<T>`, etc.?

------------

Also for name, I'd use `MutationQueue`. The `PersistentMutationOutbox` documentation should be updated to desribe how it differs from a `MutationQueue` -- that is, the outbox applies (what?) rules about populating the queue, etc.

--------

This new class should have a corresponding `MutationQueueTest.java` which tests it as an isolated unit.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationLinkedList.java
```diff
@@ -0,0 +1,112 @@
...
+        pendingNode.id = pendingMutation.getMutationId();
+        pendingNode.mutation = pendingMutation;
+        //insert the new node into the current linkedlist
+        Node addingPrev = dummyTail.prev;
+        addingPrev.next = pendingNode;
+        dummyTail.prev = pendingNode;
+        pendingNode.prev = addingPrev;
+        pendingNode.next = dummyTail;
+        //put the mutation into the mutationmap
+        mutationMap.put(pendingMutation.getMutationId(), pendingNode);
```
jamesonwilliams, line 71: Instead of maintaining your own head and tail, can you use a linked list/map or some kind directly? "Mutation linked list _has a_ (some type of) list, composed into it?"

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
richardissuperman: @jamesonwilliams sure will revise

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
richardissuperman: @jamesonwilliams good call. will revise

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationLinkedList.java
richardissuperman: @jamesonwilliams will add some tests for this new class. I didnt implement collections' interface methods as they are not needed at MutationOutBox's use case, but i can add them

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationLinkedList.java
richardissuperman: @jamesonwilliams i was thinking about using LinkedListHashMap which internally does the same thing as my new data structure , but LinkedListHashMap doesnt expose the api to iterate the linkedlist from head to tail. Also same for LinkedList java standard implementation , if you try to remove it it will traverse through the entire LinkedList and find the target which is again O(N) time complexity.
This data structure is actually inspired by a leetcode question i did before:
https://leetcode.com/problems/lru-cache/

Its for LRU cache but does the same thing.

richardissuperman:
@jamesonwilliams there are some coding styles failure in the CI, let me know if they are hard requirement i can revise

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
```diff
@@ -0,0 +1,317 @@
...
+import io.reactivex.functions.Function;
+
+/*
+ * The {@link MutationQueue} is a LinkedHashMap like container , the goal of using this container is to
+ * achieve O(1) time complexity for both getting a {@link PendingMutation} and update an existing mutation with
+ * valid id.
+ *
+ * MutationQueue is implementing the Queue interface and provide most of the queue operations,
+ */
+public class MutationQueue implements Queue<PendingMutation<? extends Model>> {
```
jamesonwilliams, line 23: Can be a `final` class, most likely.

The `Queue<...>` ends up bringing in a lot of methods, actually. Now that you've investigate that route, do you think they add value? Or not? It was an interesting idea, to implement the `Queue` interface, but perhaps it is overkill. Particularly if many will just throw an exception.

richardissuperman: @jamesonwilliams  will revise according to your comments, you are actually right that using a Queue interface is a overkill, especially we cant even reference this object in MutationOutBox as a Queue because we need the findMutationById() method to be exposed which is not a standard Queue operation,

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
```diff
@@ -764,7 +764,7 @@ public void nextItemForModelIdReturnsFirstEnqueued() throws DataStoreException {
 
         assertEquals(
             firstMutation,
-            mutationOutbox.nextMutationForModelId(originalJoe.getId())
+            mutationOutbox.getMutationQueue().nextMutationForModelId(originalJoe.getId())
```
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams agree this is a better pattern, especially when it comes to dependency injection in testing, but it requires us to update the constructor which is a breaking change to all its consumers.  It also make all its consumer import the ```MutationQueue``` class hence couple with it, which those consumers may not even care about. Let me know what you think,

Do you think its worth it to do ?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
```diff
@@ -0,0 +1,317 @@
...
+
+    private HashMap<TimeBasedUuid, Node> mutationMap = new HashMap<>();
+    private Node dummyHead = new Node();
+    private Node dummyTail = new Node();
```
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams Make sense, will revise

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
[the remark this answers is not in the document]
jamesonwilliams: > but it requires us to update the constructor which is a breaking change to all its consumers. It also make all its consumer import the `MutationQueue` class hence couple with it, which those consumers may not even care about.

```java
final class Foo {
    private final Dependency dependency;

    Foo() {
        this(new ProductionConfiguredDependency());
    }

    @VisibleForTesting
    Foo(Dependency dependency) {
        this.dependency = dependency;
    }
}

final class Bar {
    private final Foo foo;

    Bar() {
        this.foo = new Foo();
    }
}

public final class FooTest {
    @Before public void setup() {
        this.mockDependency = mock(Dependency.class);
        this.foo = new Foo(mockDependency);
    }
}   
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
```diff
@@ -0,0 +1,317 @@
...
+        mutationMap.put(pendingMutation.getMutationId(), pendingNode);
+        return true;
+    }
+
+    /**
+     * replace an existing {@link PendingMutation} inside the Queue
+     * @param id
+     * @param pendingMutation
+     */
+    public synchronized void updateExistingQueueItemOrAppendNew(TimeBasedUuid id, PendingMutation<? extends Model> pendingMutation) {
```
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams will revise

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /**
+     * remove the {@link PendingMutation} from {@link MutationQueue} by its Id.
+     * this operation should be consuming constant time
+     * @param id uuid
+     * @return
+     */
+    public synchronized boolean removeById(TimeBasedUuid id) {
+        if (mutationMap.containsKey(id)) {
```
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams revise already

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
```diff
@@ -0,0 +1,317 @@
...
+import androidx.annotation.Nullable;
+
+import com.amplifyframework.core.model.Model;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Queue;
+import io.reactivex.Observable;
```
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams thanks, have revised all the checkstyle erros

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutboxTest.java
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams revise as suggested

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java
richardissuperman: @jamesonwilliams made needed change in the latest commit :)

richardissuperman:
@rjuliano I see the merge is still blocked due to change requested, should i make other changes? thanks

PR #615, opened 2020-06-29
Changes related to subscription connectivity problems

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR is an attempt to fix a couple issues that have an impact on DataStore

1 - Subscriptions that failed on startup were left in memory. To address this, I'm keeping track of the `Future` returned by the executorService (in `SubscriptionOperation`) so we can cancel it before releasing the subscription. I'm also keeping track of the subscriptionIds that are pending initial connectivity (in `SubscriptionEndpoint`). Once a failure occurs for a subscription, it gets removed and the client needs to re-create it. I attempted to swap out the underlying websocket connection so I would make that transparent, but didn't have much luck with that approach.

2a - The websocket connection we were using was not recovering once a failure occurred. This is the behavior of the OkHttp library. Recommendation is to recreate the websocket and listener once the onFailure event is triggered. If you don't do that, any interactions with the socket from that point forward are just ignored without emitting errors.

2b - The CountdownLatch was being initialized in the `SubscriptionEndpoint` constructor. The `SubscriptionEndpoint` is only instantiated once from the API plugin. If an app started up without issues, the latch would countdown as expected. If the app lost connectivity after that, the latch was already counted down so it "assumed" the connection was still good.

To address the websocket-related issues, I refactored `SubscriptionEndpoint` and created a subclass of `WebsocketListenener` (called it `AmplifyWebSocketListener` for now) that encapsulates some of the logic necessary to reset the connection. All we have to do is call newWebSocket passing in a new instance of `AmplifyWebSocketListener`.

~~I tried to add a retry with backoff from the DataStore side, but that surfaced another issue. Because the initial setup never completed, we don't have a way to cancel those stale subscriptions because we don't have their subscription IDs (subscriptionId is only returned when the subscription is successfully established.) To address this, I'm generating the subscriptionId in `SubscriptionOperation` and passing it to `SubscriptionEndpoint`. This allows us to cancel the subscriptions in situations like that.~~

~~This way, when cancel is invoked, the `SubscriptionOperation` can try to call `releaseSubscription` with that id. This surfaced another problem: if the app attempts to invoke `clear()` in the DataStore before the initial subscription setup completes, the pending call will proceed because it's never really canceled. New subscriptions are added in addition to the ones that are there. They eventually fail because they're trying to invoke a callback that's no longer available, but they could potentially leak resources or keep another user's subscriptions active.~~ NOT APPLICABLE ANYMORE


aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -76,14 +79,12 @@ public synchronized void start() {
             ));
             return;
         }
-        executorService.submit(() -> {
+        subscriptionFuture = executorService.submit(() -> {
             LOG.debug("Requesting subscription: " + getRequest().getContent());
             subscriptionEndpoint.requestSubscription(
+                subscriptionId,
```
richardmcclellan, line 85: Is there a reason you need to pass in `subscriptionId`?  It seems like that could be handled internally within `SubscriptionEndpoint`.  It doesn't appear that `subscriptionId` is actually used anywhere else in `SubscriptionOperation`, except for checking if it's null in `cancel()`, which doesn't seem like a necessary check anyway.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -95,6 +96,9 @@ public synchronized void start() {
     public synchronized void cancel() {
         if (subscriptionId != null && !canceled) {
             canceled = true;
+            if (subscriptionFuture.cancel(true)) {
```
richardmcclellan, line 99: Possible NPE since subscriptionFuture could be null?

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -62,67 +67,86 @@
     private final GraphQLResponse.Factory responseFactory;
     private final TimeoutWatchdog timeoutWatchdog;
     private final CountDownLatch connectionResponse;
+    private final Set<String> pendingSubscriptions;
```
jamesonwilliams, line 70: Naming: `pendingSubscriptionIds`?

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -305,34 +335,42 @@ private void notifySubscriptionData(String subscriptionId, String data) throws A
...
         }
 
-        subscription.awaitSubscriptionCompleted();
-        subscriptions.remove(subscriptionId);
-
         // If we have zero subscriptions, close the WebSocket
         if (subscriptions.size() == 0) {
             timeoutWatchdog.stop();
             webSocket.close(NORMAL_CLOSURE_STATUS, "No active subscriptions");
+            webSocket = null; //force creation of a new websocket
```
jamesonwilliams, line 373: Ahh. Nice catch.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -41,6 +43,7 @@
 
     private String subscriptionId;
```
jamesonwilliams, line 44: Can this be `final` now that you assign it in the operation constructor?

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -53,6 +57,7 @@
  * and multiple GraphQL subscriptions that work on top of it.
  */
 final class SubscriptionEndpoint {
+    private static final Logger LOG = Amplify.Logging.forNamespace("aws-api:websocket");
```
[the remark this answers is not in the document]
rjuliano: fixed

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
rjuliano: Nope...I reverted that change.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
rjuliano: Yup...accounting for that now.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
rjuliano: Changed

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
rjuliano: It ended up looking a little different, but the spirit is the same. Basically need to force the creation of a new websocket.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
rjuliano: Unfortunately no because I can't assign it until the subscription actually starts.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -40,7 +42,8 @@
     private final Action onSubscriptionComplete;
 
     private String subscriptionId;
-    private boolean canceled;
+    private AtomicBoolean canceled;
```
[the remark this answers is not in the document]
rjuliano: done

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -85,21 +87,31 @@ public synchronized void start() {
...
                 onNextItem,
-                onSubscriptionError,
+                apiException -> {
+                    // Guard against calling something that's been cancelled already.
+                    if (!canceled.get()) {
+                        canceled.set(true);
+                        onSubscriptionError.accept(apiException);
+                    }
+                },
                 onSubscriptionComplete
```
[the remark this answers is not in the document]
rjuliano: Right...I feel like we're missing a representation of a state here. In my mind, we're missing a representation of the type of cancelation that took place: was it gracefully canceled, or was it a result of some error condition. 

I started doing some refactoring here, but ended up going down a rabbit hole so I shelved those changes for now. Probably something that needs to be looked at more broadly.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -134,11 +140,15 @@
                 .toString()
             );
         } catch (JSONException | ApiException exception) {
-            onSubscriptionError.accept(new ApiException(
+            // If the subscriptionId was still pending, then we can call the onSubscriptionError
```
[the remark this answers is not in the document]
rjuliano: For us to get here, we know that the subscription was added to the pendingSubscriptionIds set and we have a good connection. These two exceptions can happen either because of some malformed value passed into the `put` calls, or if `createHeadersForSubscription` throws an `ApiException`.

If the subscription was not pending, it's likely because `cancel()` was invoked some time after the connection was successfully established, but before the subscription handshake completed.

PR #617, opened 2020-06-30
Display Developer Menu only in Debug Build

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Only display the developer menu when in a debug (debuggable) build, not in a release build.


core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -135,12 +136,13 @@ public static void configure(@NonNull final AmplifyConfiguration configuration,
                     beginInitialization(category, context);
                 }
             }
-
-            // Start activity for developer menu
-            Intent mainIntent = new Intent(context, DeveloperMenuActivity.class);
-            mainIntent.setAction(Intent.ACTION_MAIN);
-            mainIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-            context.startActivity(mainIntent);
+            if ((context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0) {
```
TrekSoft, line 139: Is there ever a real life scenario where getApplicationInfo could be null and thus throw a null exception here? If not this is fine, otherwise might be worth checking that ahead of accessing the flags attribute.

eeatonaws: I don't think getApplicationInfo could return null because the ApplicationInfo object returned by getApplicationInfo corresponds to the application tag (and its contents) in AndroidManifest.xml, which is required to run the app.

PR #622, opened 2020-07-02
Show and hide dev menu on device shake or keyboard shortcut

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* When the device is shaken or ctrl (or command) + m is pressed, then show the developer menu if it is currently hidden, or hide the developer menu if it is currently visible.


core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -16,16 +16,112 @@
...
+                }
+            }
+        }
+
+        @Override
+        public void onAccuracyChanged(Sensor sensor, int accuracy) { }
+    };
+
+    // Manager for the device's sensors.
+    private SensorManager sensorManager;
```
TrekSoft, line 65: These should probably be listed above the event listener, especially since the listener references shakeStart.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -16,16 +16,112 @@
...
+
+    @Override
+    public boolean onKeyUp(int keyCode, KeyEvent event) {
+        // KEYCODE_MENU is the code for pressing ctrl
+        // (or command) + m
+        if (keyCode == KeyEvent.KEYCODE_MENU) {
+            changeVisibility();
+            return true;
+        } else {
+            return super.onKeyUp(keyCode, event);
```
TrekSoft, line 112: Should super.onKeyUp always be called?

eeatonaws: In examples in the Android documentation, they override onKeyUp in a similar manner (only returning super.onKeyUp if the keycode doesn't match a certain value). Also, if I change the code to always return super.onKeyUp, even if the keycode matches the one for ctrl+m, then the visibility of the developer menu is only changed the first two times I press ctrl+m.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -16,16 +16,112 @@
...
+        devMenuLayout = findViewById(R.id.devLayout);
+        devMenuLayout.setFocusable(true);
+        devMenuLayout.setVisibility(View.GONE);
+        shakeStart = 0;
+    }
+
+    @Override
+    protected void onResume() {
+        if (accelerometer != null) {
+            sensorManager.registerListener(sensorEventListener, accelerometer, SensorManager.SENSOR_DELAY_NORMAL);
```
jamesonwilliams, line 91: If I'm reading this correctly, Amplify would keep the accelerometer active at all times, to receive sensor data, so it can detect a shake. This will drain the battery, so we need a different solution. What mechanisms can you put in place to limit the lifecycle of the accelerometer? How can you prevent this from running "most of the time", when Amplify is in production?

eeatonaws: When you say "Amplify is in production", are you referring to when the developer is using Amplify in their app and their app is in production (release build)?

jamesonwilliams: The second one!

eeatonaws: The developer menu will only be activated for a debuggable build, not for release builds (I added a check in Amplify.java in a previous PR that only starts the developer menu activity if the build is debuggable).

eeatonaws:
> Good start! I'd recommend separating the mechanical components (working with sensors) from the presentation stuff (displaying different UI states).
> 
> The Activity should be about the _display_ of the data. You could pull all of your sensor management logic into a `ShakeDetector.java`, or something like that.
> 
> ```java
> final class ShakeDetector {
>     private final Context context;
>     private final ShakeDetector.Listener listener;
> 
>     ShakeDetector(Context context, ShakeDetector.Listener listener) {
>         this.context = context;
>         this.listener = listener;
>     }
> 
>     void startDetecting() {
>         ...
>        // when you detect one, call listener.onShakeDetected().
>     }
> 
>     void stopDetectring() {
>         ...
>     }
> 
>     interface Listener {
>         void onShakeDetected();
>     }
> }
> ```
> 
> The activity would start it, something like this:
> 
> ```java
> ShakeDetector detector = new ShakeDetector(context, () -> {
>     // Update the presentation, when a shake is detected.
> });
> detector.startDetecting();
> ```

Moving the sensor management logic out of the activity class and into a separate file definitely makes sense, thanks for pointing that out. Following your suggestion, I moved the sensor management logic into ShakeDetector.java.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
jamesonwilliams: Okay, great. Thanks for reminding me. It's my job to be paranoid about it. :-D

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -16,16 +16,65 @@
 package com.amplifyframework.core;
 
 import android.app.Activity;
+import android.content.Context;
+import android.hardware.SensorManager;
```
jamesonwilliams, line 20: Still sort of "leaking" a detail of the `ShakeDetector`, here. I'd grab this off the `context `internally.

core/src/main/java/com/amplifyframework/core/ShakeDetector.java
```diff
@@ -0,0 +1,110 @@
...
+import android.hardware.SensorEvent;
+import android.hardware.SensorEventListener;
+import android.hardware.SensorManager;
+
+import com.amplifyframework.util.Time;
+
+/**
+ * Detects a device shake event.
+ */
+public final class ShakeDetector {
```
jamesonwilliams, line 28: You'll want to add a unit test for this.

PR #626, opened 2020-07-06
Add buttons to main screen of developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Add buttons for view environment information, view device information, view logs, and file issue to the main screen of the developer menu.


core/src/main/res/values/strings.xml
```diff
@@ -16,4 +16,8 @@
 
 <resources >
```
jamesonwilliams, line 17: There's an extra ` ` space char here

---------


There is no _formal_ convention around this, but it is more common in my experience to see user-provided XML ids and names as `name="something_in_lower_snake_case` as opposed to `name="somethingInCamelCase"`.

You can tell from the Android view attributes that even Android itself does not impose a coherent convention.

Same for IDs, though: `@+id/some_id`

eeatonaws: Thanks for pointing that out, I'll change the camel case to lower snake case.

PR #632, opened 2020-07-09
Develop screen navigation and header for dev menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Adds a header to each screen of the developer menu (that includes a back button), overrides the Android phone back button to allow navigation between developer menu screens, and navigates to a new view when a button on the main developer menu screen is pressed.

Main developer menu view:

Dev Menu Main Screen New

Environment information view (the views displayed when the other three buttons are pressed look the same as this view, only with a different title):

Dev Menu Env Info New


core/src/main/res/layout/activity_dev_menu.xml
```diff
@@ -25,46 +25,101 @@
...
+
+            <Button
+                android:id="@+id/file_issue_button"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:text="@string/file_issue_button_title"
+                android:textAllCaps="false" />
+        </LinearLayout>
+
+        <LinearLayout
```
richardmcclellan, line 97: Rather than having all of these subpages all in one big xml layout, it'd be better for each one of them to have their own xml layout, which is inflated by the fragment.  If the subpages are similar, you could just have a single xml file that works for all subpages, and then after it is inflated, programmatically set properties specific to that subpage, like the title.

core/src/main/res/layout/activity_dev_menu.xml
```diff
@@ -25,46 +25,101 @@
...
+            android:theme="@android:style/Widget.DeviceDefault.ActionBar.Solid"
+            android:layout_marginBottom="15dp">
+
+            <ImageButton
+                android:id="@+id/back_button"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:background="@drawable/ic_baseline_arrow_back_24"
+                android:layout_marginStart="20dp"
+                android:layout_marginLeft="20dp"
```
richardmcclellan, line 43: I think you can remove `android:layout_marginLeft` since you've defined `android:layout_marginStart`.  `start` is better, since it allows support for RTL languages.   

Actually, I'd suggest replacing `Left` with `Start` and `Right` with `End` everywhere in this file.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -74,4 +122,30 @@ private void changeVisibility() {
...
+        buttons.setVisibility(View.GONE);
+        newView.setVisibility(View.VISIBLE);
+        backButton.setVisibility(View.VISIBLE);
+        titleText.setText(titleTextId);
+    }
+
+    /**
+     * Hide the current view and display the main developer menu view.
+     */
+    private void displayMainView() {
```
jamesonwilliams, line 142: These individual views have a parent view, right? Can you set it `GONE` and `VISIBLE`, instead of doing it on all of the individual children?

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -63,6 +100,17 @@ public boolean onKeyUp(int keyCode, KeyEvent event) {
         }
     }
 
+    @Override
```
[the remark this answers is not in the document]
jamesonwilliams: Fragments are one approach. You can also look at the (~new-ish) [Navigation component in Jet Pack](https://developer.android.com/guide/navigation).

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -74,4 +122,30 @@ private void changeVisibility() {
...
         }
     }
+
+    /**
+     * Hides the main developer menu view, displays the given View, and
+     * changes the title text on the action bar.
+     * @param newView View to display.
+     * @param titleTextId ID of the string resource for the action bar title text.
+     */
+    private void leaveMainView(View newView, int titleTextId) {
```
jamesonwilliams, line 132: It's a subtle point here, but the `Activity` ought to be about visual display. So I would phrase this as "display" and "hide" the main view, instead of the navigation-sounding (control) names you're using.

core/src/main/res/layout/activity_dev_menu.xml
```diff
@@ -25,46 +25,101 @@
...
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:text="@string/logs_button_title"
-            android:textAllCaps="false" />
+            android:orientation="vertical"
+            android:visibility="gone" />
 
-        <Button
-            android:id="@+id/file_issue"
+        <LinearLayout
```
jamesonwilliams, line 118: It's best to avoid nesting of layout elements like this (see [discussion](https://developer.android.com/training/improving-layouts/optimizing-layout).)

Three isn't the end of the world. But checkout the `ConstraintLayout` and see if you could fit all of your individual items in it?

core/src/main/res/layout/activity_dev_menu.xml
eeatonaws: If I remove android:layout_marginLeft, it says "To support older versions than API 17 (project specifies 16) you should also add android:layout_marginLeft="20dp"". Should I be using a higher minimum API version (16 is the default)?

core/src/main/res/layout/activity_dev_menu.xml
```diff
@@ -25,46 +25,101 @@
         android:layout_height="wrap_content"
         android:background="@android:color/background_light"
         android:orientation="vertical"
-        android:layout_alignParentBottom="true"
-        android:padding="15dp">
+        android:layout_alignParentBottom="true">
 
-        <TextView
-            android:id="@+id/title_text_view"
+        <RelativeLayout
```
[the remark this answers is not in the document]
eeatonaws: I thought about using a Toolbar, but it requires a minimum of SDK 21, so I chose to implement my own header to avoid assuming the customer has at least SDK 21.

core/src/main/res/layout/activity_dev_menu.xml
TrekSoft: Yeah no we want to maintain a minimum version of 16

richardmcclellan: Ahh, yeah, we don't want to raise the minimum above 16.   Disregard!

core/src/main/res/layout/activity_dev_menu.xml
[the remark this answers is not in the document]
richardmcclellan: Hmm, right, `android.widget.Toolbar` does have that limitation.  I think you should be able to use  [`androidx.appcompat.widget.Toolbar`](https://developer.android.com/reference/androidx/appcompat/widget/Toolbar) though, as it is backported for older APIs.

To use, you would just add this in your layout xml:

```
    <androidx.appcompat.widget.Toolbar
        android:id="@+id/toolbar"
        android:layout_width="match_parent"
        android:layout_height="?android:attr/actionBarSize"
        app:title="Amplify Developer Menu"/>
```

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -63,6 +74,16 @@ public boolean onKeyUp(int keyCode, KeyEvent event) {
         }
     }
 
+    @Override
+    public void onBackPressed() {
```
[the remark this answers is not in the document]
eeatonaws: The DeveloperMenuActivity is finished by default (if I don't override onBackPressed) if the user presses the phone's back button, but that means the user can no longer activate the developer menu.

core/src/main/res/layout/activity_dev_menu.xml
```diff
@@ -25,46 +27,40 @@
...
-            android:text="@string/device_button_title"
-            android:textAllCaps="false" />
-
-        <Button
-            android:id="@+id/logs"
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:text="@string/logs_button_title"
-            android:textAllCaps="false" />
+            <TextView
```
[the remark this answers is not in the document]
eeatonaws: I added the TextView in order to center the title text on the toolbar.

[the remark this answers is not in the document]
richardmcclellan: Hmm, by convention, titles are typically left aligned on Android, and centered on iOS.   I'd suggest that we prefer platform conventions (left aligned) where possible.   One major benefit of using the built in title TextView is that the Navigation framework will automatically update the title when moving between fragments.   

That said, if there's a reason it needs to be centered, I think the solution you have now works.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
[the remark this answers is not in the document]
richardmcclellan: Ahh, I just took at look at the changes you made in `Amplify.configure()`.   By design, an [`Activity`](https://developer.android.com/reference/android/app/Activity) should really only be used when there is UI to be displayed, so I don't think it's a good idea to create an `Activity` until the dev menu is actually supposed to be shown.    

Another issue I think you will have with the current approach is this scenario:
 - In the customer's app, they call `Amplify.configure()`, which starts the `DeveloperMenuActivity`
 - Then the customer creates one or more of their own activities for their app. 
 - Then, the customer shakes the phone.  Now, `DeveloperMenuActivity` is not the foreground activity, so even if you change visibility of elements on the dev menu, it won't be shown, since it is behind other activities.

Instead, perhaps you could turn `ShakeDetector` into a singleton, which is initialized in `Amplify.configure`, and is always listening for shake events.  Then, when the ShakeDetector detects a shake, it could start the `DeveloperMenuActivity`?

build.gradle
```diff
@@ -45,6 +45,7 @@ ext {
     buildToolsVersion = "29.0.2"
     compileSdkVersion = 29
     minSdkVersion = 16
+    nav_version = "2.3.0"
```
richardmcclellan, line 48: For no particular reason, none of the variables in this file have underscores, but to be consistent, I'd just call this `navigationversion`.  Also, it's probably best to move it right below `awsSdkVersion`, so that it's with the other dependencies.  These values here at the top (`minSdkVersion`, `targetSdkVersion`, etc), aren't related to dependencies.

build.gradle
```diff
@@ -57,6 +58,11 @@ ext {
             annotation: 'androidx.annotation:annotation:1.1.0',
             appcompat: 'androidx.appcompat:appcompat:1.1.0',
             core: 'androidx.core:core:1.3.0',
+            navfragment: "androidx.navigation:navigation-fragment:$nav_version",
+            navfragmentktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
+            navui: "androidx.navigation:navigation-ui:$nav_version",
+            navuiktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
+            navfeatfragment: "androidx.navigation:navigation-dynamic-features-fragment:$nav_version",
```
jamesonwilliams, line 65: How about:
```
nav: [
    fragment: "...",
    fragmentktx: "...",
    ui: "...",
    uiktx: "...",
    dynamicfeatures: "..."
]
```

When you can then reference as `androidx.nav.ui`, etc.

build.gradle
```diff
@@ -57,6 +58,11 @@ ext {
             annotation: 'androidx.annotation:annotation:1.1.0',
             appcompat: 'androidx.appcompat:appcompat:1.1.0',
             core: 'androidx.core:core:1.3.0',
+            navfragment: "androidx.navigation:navigation-fragment:$nav_version",
+            navfragmentktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
+            navui: "androidx.navigation:navigation-ui:$nav_version",
+            navuiktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
```
jamesonwilliams, line 64: This is included twice, once as `navfragmentktx`, and once as ` navuiktx`.

build.gradle
```diff
@@ -45,10 +45,10 @@ ext {
     buildToolsVersion = "29.0.2"
     compileSdkVersion = 29
     minSdkVersion = 16
-    nav_version = "2.3.0"
     targetSdkVersion = 29
 
     awsSdkVersion = '2.16.12'
+    navigationversion = "2.3.0"
```
richardmcclellan, line 51:
```suggestion
    navigationversion = '2.3.0'
```
Single quotes, not double.

build.gradle
```diff
@@ -58,11 +58,13 @@ ext {
             annotation: 'androidx.annotation:annotation:1.1.0',
             appcompat: 'androidx.appcompat:appcompat:1.1.0',
             core: 'androidx.core:core:1.3.0',
-            navfragment: "androidx.navigation:navigation-fragment:$nav_version",
-            navfragmentktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
-            navui: "androidx.navigation:navigation-ui:$nav_version",
-            navuiktx: "androidx.navigation:navigation-fragment-ktx:$nav_version",
-            navfeatfragment: "androidx.navigation:navigation-dynamic-features-fragment:$nav_version",
+            nav: [
+                    fragment: "androidx.navigation:navigation-fragment:$navigationversion",
```
richardmcclellan, line 62: Single quotes for all of these as well, not double quotes.

eeatonaws: I need to use double quotes in order to reference the $navigationversion variable (with single quotes, it sees $navigationversion as part of the string rather than a variable).

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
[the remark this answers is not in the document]
eeatonaws: Thanks for pointing this out. I just tested it and you're right that the DeveloperMenuActivity isn't visible even when shaking the phone if another activity is started.

build.gradle
richardmcclellan: Oh you're right, got it!

jamesonwilliams, line 62: minor: +4 spaces is the current indentation level in this file -- can you update before inclusion?

build.gradle
jamesonwilliams: uppercase `V`, no? `navigationVersion`?

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -29,27 +27,34 @@
  * This is the activity to display the developer menu.
  */
 public final class DeveloperMenuActivity extends FragmentActivity {
+
+    // Indicates whether this activity has started.
+    private static boolean activityStarted;
```
richardmcclellan, line 32: Maintaining state in a static variable should generally be avoided because it can be accessed/modified from anywhere, and thus can lead to bugs and code that's hard to maintain.  Looks like you are referencing this in `Amplify.configure` to see whether DeveloperMenuActivity should be started.  Instead, I'd suggest this:
 - When a shake is detected, stop listening for shakes, and then start DeveloperMenuActivity using the `startActivityForResult()` method (https://developer.android.com/training/basics/intents/result). 
 - Then, implement `onActivityResult()` to know when DeveloperMenuActivity is finished.  When it is finished, start listening for shakes again.

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -137,11 +137,16 @@ public static void configure(@NonNull final AmplifyConfiguration configuration,
...
                 Intent mainIntent = new Intent(context, DeveloperMenuActivity.class);
                 mainIntent.setAction(Intent.ACTION_MAIN);
                 mainIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-                context.startActivity(mainIntent);
+                ShakeDetector shakeDetector = new ShakeDetector(context, () -> {
+                    if (!DeveloperMenuActivity.isActivityStarted()) {
+                        context.startActivity(mainIntent);
+                    }
+                });
+                shakeDetector.startDetecting();
```
jamesonwilliams, line 149: In the interest of keeping the `configure(...)` method as lean as possible, can we state this block as:
```
if (isDebuggableApplication()) {
    enableDeveloperMenu();
}
```
And then put the details into helpers?

core/src/main/java/com/amplifyframework/core/DevMenuLogsFragment.java
```diff
@@ -0,0 +1,41 @@
...
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import androidx.fragment.app.Fragment;
+
+/**
+ * A {@link Fragment} subclass representing the view
+ * to display the logs on the developer menu.
+ */
+public final class DevMenuLogsFragment extends Fragment {
```
jamesonwilliams, line 28: This template is repeated a few times.

Can you make just one of these fragments, and give it a constructor which accepts:
```
public final class DevMenuLogsFragement extends Fragment {
    @IdRes private final int layoutResourceId;

    public DevMenuLogsFragment(@IdRes int layoutResourceId) {
        this.layoutResourceId = layoutResourceId;
    }
    ...
}
```

___Update___: oh, duh. Nope. Your comment says the no-args constructor is required. Nevermind!

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -15,16 +15,22 @@
...
-public final class DeveloperMenuActivity extends Activity {
-    // The parent layout for the developer menu.
+public final class DeveloperMenuActivity extends FragmentActivity {
+
+    // Indicates whether this activity has started.
+    private static boolean activityStarted;
+    // Parent layout for the developer menu.
     private View devMenuLayout;
     // Detect and handle shake events.
     private ShakeDetector detector;
```
jamesonwilliams, line 36: Hm, you now have two instances of the `ShakeDetector`, one in `Amplify`, and one in this Activity. Ideally, you could have one detector that broadcast to multiple listeners. Instead of the simple callback mechanism we are using right now, you could use `Amplify.Hub` instead. Then, just listen for a hub event in both locations -- from one detector instance.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
```diff
@@ -29,27 +27,34 @@
...
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_dev_menu);
         devMenuLayout = findViewById(R.id.dev_layout);
         devMenuLayout.setFocusable(true);
-        devMenuLayout.setVisibility(View.GONE);
-        detector = new ShakeDetector(getApplicationContext(), this::changeVisibility);
+        detector = new ShakeDetector(getApplicationContext(), this::finish);
```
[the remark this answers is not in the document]
TrekSoft: I believe she supports both which I think is the best customer experience.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
eeatonaws: For our use case, would this approach be equivalent to starting to listen for shakes again in the DeveloperMenuActivity onStop()?

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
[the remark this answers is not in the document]
richardmcclellan: I was a little concerned that if a customer does a particularly strong shake, it could be recognized as 2 shakes, which would show and then immediately hide the menu, which I don't think would be an expected experience.   I don't feel too strongly though, I'm good to try it with this functionality.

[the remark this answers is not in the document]
eeatonaws: I added a duration threshold in ShakeDetector that only calls the given callback method if a minimum amount of time has passed since the shake started, in order to prevent the developer menu from appearing and then immediately disappearing if there's a strong shake.

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -179,6 +170,20 @@ private static void beginInitialization(@NonNull Category<? extends Plugin<?>> c
...
+     * Starts listening for a shake event and shows the developer menu on device shake.
+     * @param context an Android Context
+     */
+    public static void enableDeveloperMenu(Context context) {
+        // Start activity for developer menu on device shake
+        Intent mainIntent = new Intent(context, DeveloperMenuActivity.class);
+        mainIntent.setAction(Intent.ACTION_MAIN);
+        mainIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        ShakeDetector shakeDetector = new ShakeDetector(context, () -> context.startActivity(mainIntent),
+                true);
```
jamesonwilliams, line 183: Just an idea -- feel free to ignore --

Instead of using the boolean, you could encode this behavior in your callback:
```java
final class ShakeDetector {
    private final ShakeDetector.Listener listener;

    ShakeDetector(ShakeDetector.Listener listener) {
        this.listener = listener;
    }

    // Change this
    @FunctionalInterface
    interface Listener {
        void onShakeDetected(ShakeDetector source);
    }
}
```

Then you'll have an instance to call `stopDetecting()` on, in your lambda:
```java
ShakeDetector detector = new ShakeDetector(context, it -> {
    context.startActivity(mainIntent);
    it.stopDetecting();
});
```

Without changing the listener, you could already do:
```java
final AtomicReference<ShakeDetector.Listener> listener = new AtomicReference<>();
ShakeDetector detector = new ShakeDetector(context, () -> {
    context.startActivity(mainIntent);
    listener.get().stopDetecting();
});
```

^ This would give you a `final` reference inside of your lambda. The `AtomicReference` thing is sort of a kludge that I try to avoid if I can, though.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
richardmcclellan: Hmm, how are you referencing the ShakeDetector defined in Amplify.configure from `DeveloperMenuActivity.onStop()`

One concern is that the logic for starting and stopping listening is not encapsulated in one place, and thus harder to maintain/understand - starting is in Amplify.configure, and stopping is in `DeveloperMenuActivity`. 

Another concern is the following scenario:
 - Amplify.configure starts listening for shakes.
 - Customer shakes device, listening stops, DeveloperMenuActivity is launched.
 - Customer receives a push notification, taps it, which launches a new Activity to display that push notification.  `onStop` is called, so listening starts again.
 - Customer taps back, and is now on DeveloperMenuActivity, and listener is still active. 
 - Customer shakes devices again, and now a new DeveloperMenuActivity instance is started on top of the existing one.

core/src/main/java/com/amplifyframework/core/Amplify.java
eeatonaws: Thanks for the suggestion. Using the FunctionalInterface works well and it does seem cleaner than using a boolean.

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
eeatonaws: In DeveloperMenuActivity.onStop(), I call Amplify.enableDeveloperMenu, which handles starting and stopping listening for shake events when the developer menu activity has not started (essentially creating a new ShakeDetector instance).

richardmcclellan: I see - hmm, we really should avoid making `Amplify.enableDeveloperMenu` a public method, as that is now exposed to customers, and I don't think there's a strong customer use case to support that.  If there was a customer use case, this is the kind of thing that would need to go through a design review with all platforms as well.

eeatonaws: I didn't realize that, thanks for pointing it out. I was trying to avoid repeating the same code in DeveloperMenuActivity. If I move the method to DeveloperMenuActivity, but still keep it public so I can call it in Amplify.configure, would it no longer be exposed to customers?

richardmcclellan: My previous suggestion was to implement `onActivityResult` in the class that starts `DeveloperMenuActivity`, but I just realized that doesn't work because that has to be implemented in an Activity, and `Amplify` is not an `Activity`. 

Still thinking about this...

eeatonaws: Since I'm starting the developer menu activity from outside of an activity, I have to use FLAG_ACTIVITY_NEW_TASK. The documentation for this flag says "if a task is already running for the activity you are now starting, then a new activity will not be started; instead, the current task will simply be brought to the front of the screen with the state it was last in". Does this mean that in the scenario you described, a new developer menu activity would not be started?

richardmcclellan: > Does this mean that in the scenario you described, a new developer menu activity would not be started?

Yes, that's right.  

If you are okay with not supporting shake to dismiss, using the `FLAG_ACTIVITY_SINGLE_TOP` would probably give you a complete solution.    

If you do want to support shake to dismiss, it is a bit more complex.  I think you could accomplish it though with something like the logic below.  It's not my favorite idea, because it still involves maintaining a static state, but I'm not sure of a different solution - it's definitely a tricky problem.

```
class DevMenuManager implements ShakeDetector.Listener {
    private boolean visible;
 
    private static DevMenuManager sInstance = new DevMenuManager();

    public static sharedInstance() {
         return sInstance;
    }

    public start() {
         // Subscribe to shake events
    }

    public void onShakeDetected() {
         if(visible) {
            Amplify.Hub.sendMessage("HideDeveloperMenu");  // this syntax is probably wrong, but the idea is to broadcast a message on Hub.
         } else {
            Intent mainIntent = new Intent(context, DeveloperMenuActivity.class);
            mainIntent.setAction(Intent.ACTION_MAIN);
            mainIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
            context.startActivity(mainIntent);
            visible = true;
       }
   }

   public void setHidden() {
       this.visible = false;
   }
}
```

In `Amplify.configure()`:

```
DevMenuManager.sharedInstance().startListening()
```

In `DeveloperMenuActivity.onStop()`:

```
DevMenuManager.sharedInstance.setHidden();
```

core/src/main/java/com/amplifyframework/core/DeveloperMenuActivity.java
eeatonaws: After rewriting the code, I now have only one instance of ShakeDetector (in DeveloperMenuManager). I also tried using Amplify.Hub, but that would require adding a new CategoryType for the developer menu in the HubChannel enum.

core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
```diff
@@ -0,0 +1,100 @@
...
+            context.startActivity(mainIntent);
+            visible = true;
+        }
+    }
+
+    /**
+     * Set the visibility of the developer menu.
+     * @param visible boolean indicating whether the developer menu is visible.
+     */
+    public void setVisibility(boolean visible) {
```
richardmcclellan, line 78: The name `setVisibility` makes me think this method would take a `Visibility` enum parameter, (GONE, INVISIBLE, VISIBLE).  

Maybe `setVisible`?

core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
```diff
@@ -0,0 +1,100 @@
...
+import android.content.Context;
+import android.content.Intent;
+
+/**
+ * Manage the display and shake detection behavior for the
+ * developer menu.
+ */
+public final class DeveloperMenuManager implements ShakeDetector.Listener {
+    // An instance of DeveloperMenuManager.
+    private static DeveloperMenuManager sInstance = new DeveloperMenuManager();
```
richardmcclellan, line 27: It'd be better to lazily initialize this in sharedInstance() like this:
```
public static DeveloperMenuManager sharedInstance() {
    if(sInstance == null) {
        sInstance = new DeveloperMenuManager();
    }
    return sInstance;
}
```

core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    public void onShakeDetected() {
+        if (visible) {
+            if (hideAction != null) {
+                hideAction.hideDeveloperMenu();
+            }
+            visible = false;
+        } else {
+            Context context = shakeDetector.getContext();
```
richardmcclellan, line 65: It'd probably be better for DeveloperMenuManager to just keep its own `Context` property, which gets set in `startListening()`.  It isn't really the job of shakeDetector to be providing Context to others.

eeatonaws: I initially had a Context field in DeveloperMenuManager, but there was a warning that this caused a memory leak since you can access a static instance of DeveloperMenuManager.

richardmcclellan: Storing `Context` within a static variable is a valid concern for memory leaks.   The same concern is also valid though for storing it in ShakeDetector, since it is still stored in a static instance (DeveloperMenuManager.sharedInstance().shakeDetector.context).    The same concern is even valid for your first iteration of this PR, where you were starting the DeveloperMenuManager from a callback inside of the `Amplify.configure` method.    In all cases, you are holding onto a reference of Context indefinitely.

It's important to understand there are two types of `Context` in Android:
1. The Application context, which lives for the lifetime of your app
2. An Activity `Context`, which lives for the lifecycle of an Activity

The important thing is not to keep references to any Activity `Context` instances, beyond the lifecycle of that Activity because that will prevent the Activity from being removed from memory when it is finished (a memory leak).  Using the Application Context is okay though, since is available for the lifecycle of your app.  The only caveat to this is that the Application Context has more limited functionality than an Activity Context - you can't use it to show a dialog view, for example. I don't think that's actually a concern for this use case though. 

Customers are *probably* passing in an Application Context to `Amplify.configure()`, since [that's what we say to do in the setup docs](https://docs.amplify.aws/lib/project-setup/create-application/q/platform/android#n4-initialize-amplify-in-the-application).   There's no guarantee though - an Activity class is actually a subclass of Context, so they might be passing in an Activity.

To account for this, if an Activity is passed in, you can get an Application context from it by calling `getApplicationContext()` on the Activity.  This is something we should probably do to ensure we ensure we aren't causing any memory leaks.    Here's [an example of how Firebase does something similar](https://github.com/firebase/firebase-android-sdk/blob/db66120ce7926205975a2dd50fad2f8798227a5b/firebase-common/src/main/java/com/google/firebase/FirebaseApp.java#L305-L311) in their equivalent of Amplify.configure.  For now, I'd suggest putting some similar logic inside the `startListening()` method.

Regarding the warning itself - where do you see that specifically?  Is it flagged by our checkstyle rules, or is it actually preventing the build from compiling?

eeatonaws: The warning appears when I open DeveloperMenuManager.java and it also prevents the build from succeeding. Using Context.getApplicationContext() and adding a Context parameter to the constructor makes the warning disappear (the warning is still present if I set the context field to context.getApplicationContext() inside the startListening method).

PR #633, opened 2020-07-09
feat(DataStore) add sorting

richardmcclellan:
### Summary 
 - This PR adds support for sorting on the `DataStore.query` API.  
 - Also, currently customers can only query for pagination if they also provide a predicate.  This PR adds a new `paginated` method on the `Where` builder so that a customer can specify just pagination if desired (`Where.paginated(Page.firstPage())`).

Sample DX:
```
        Amplify.DataStore.query(Todo.class,
                Where
                        .matches(Todo.NAME.beginsWith("foo").and(Todo.DESCRIPTION.beginsWith("bar")))
                        .paginated(Page.firstPage())
                        .sorted(Todo.NAME.ascending(), Todo.DESCRIPTION.descending(), Todo.ID.ascending()),
                response -> { },
                error -> { });
```

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SqlKeyword.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SqlKeyword.java
@@ -17,6 +17,7 @@ package com.amplifyframework.datastore.storage.sqlite;
 
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.core.model.query.QuerySortOrder;
 import com.amplifyframework.core.model.query.predicate.QueryOperator;
 import com.amplifyframework.core.model.query.predicate.QueryPredicateGroup;
 
@@ -149,10 +150,26 @@ public enum SqlKeyword {
     /**
      * SQL keyword to specify a size limit of the result set (used for paginating results).
      */
-    LIMIT("LIMIT");
+    LIMIT("LIMIT"),
+
+    /**
+     * SQL keyword to specify a column to sort the results by.
+     */
+    ORDER_BY("ORDER BY"),
+
+    /**
+     * SQL keyword meaning to sort in ascending order, for use with ORDER_BY.
+     */
+    ASC("ASC"),
+
+    /**
+     * SQL keyword meaning to sort in descending order, for use with ORDER_BY.
+     */
+    DESC("DESC");
 
     private static final Map<QueryOperator.Type, SqlKeyword> QUERY_OPERATOR_TO_SQL = new HashMap<>();
     private static final Map<QueryPredicateGroup.Type, SqlKeyword> QUERY_PREDICATE_GROUP_TO_SQL = new HashMap<>();
+    private static final Map<QuerySortOrder, SqlKeyword> QUERY_SORT_BY_TO_SQL = new HashMap<>();
 
     private final String stringValue;
 
@@ -167,6 +184,9 @@ public enum SqlKeyword {
         QUERY_PREDICATE_GROUP_TO_SQL.put(QueryPredicateGroup.Type.AND, SqlKeyword.AND);
         QUERY_PREDICATE_GROUP_TO_SQL.put(QueryPredicateGroup.Type.OR, SqlKeyword.OR);
         QUERY_PREDICATE_GROUP_TO_SQL.put(QueryPredicateGroup.Type.NOT, SqlKeyword.NOT);
+
+        QUERY_SORT_BY_TO_SQL.put(QuerySortOrder.ASCENDING, SqlKeyword.ASC);
+        QUERY_SORT_BY_TO_SQL.put(QuerySortOrder.DESCENDING, SqlKeyword.DESC);
     }
 
     SqlKeyword(String stringValue) {
@@ -198,6 +218,21 @@ public enum SqlKeyword {
         if (null == sqlKeyword) {
             throw new IllegalArgumentException(
                     "No SQL keyword mapping defined for query predicate group type = " + groupType.toString()
+            );
+        }
+        return sqlKeyword;
+    }
+
+    /**
+     * Retrieve the SQL specific keyword for the query sort order type.
+     * @param sortOrder the query sort order type
+     * @return the SQL specific keyword
+     */
+    public static SqlKeyword fromQuerySortOrder(@NonNull QuerySortOrder sortOrder) {
+        final SqlKeyword sqlKeyword = QUERY_SORT_BY_TO_SQL.get(Objects.requireNonNull(sortOrder));
+        if (null == sqlKeyword) {
+            throw new IllegalArgumentException(
+                    "No SQL keyword mapping defined for query sort order type = " + sortOrder.toString()
             );
         }
         return sqlKeyword;
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SqlKeyword.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SqlKeyword.java
@@ -198,6 +198,7 @@ public enum SqlKeyword {
      * @param queryOperatorType the query operator type
      * @return the SQL specific keyword
      */
+    @NonNull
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -0,0 +1,35 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model.query;
+
+/**
+ * A simple data structure that holds sort information that can be applied queries.
+ */
+public class QuerySortBy {
```
jamesonwilliams, line 21:
```suggestion
public final class QuerySortBy {
```

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Constructor for {@code QuerySortBy}.
+     *
+     * @param field name of field to sort by.
+     * @param sortOrder order to sort by, either ASCENDING or DESCENDING.
+     */
+    public QuerySortBy(String field, QuerySortOrder sortOrder) {
+        this.field = field;
+        this.sortOrder = sortOrder;
+    }
```
jamesonwilliams, line 34: field accessors, equals, toString, hashCode, all that Jazz? (If only we could use Kotlin data classes, sigh.)

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -0,0 +1,35 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model.query;
+
+/**
+ * A simple data structure that holds sort information that can be applied queries.
```
jamesonwilliams, line 19: Probably room for rewording

core/src/main/java/com/amplifyframework/core/model/query/QueryOptions.java
```diff
@@ -60,6 +78,17 @@ public QueryOptions paginated(@NonNull final QueryPaginationInput paginationInpu
         return this;
     }
 
+    /**
+     * Updates the current options with a given {@code sortBy}.
+     *
+     * @param sortBy sorting information
+     * @return current options with an updated {@code sortBy}.
+     */
+    public QueryOptions sorted(@NonNull final QuerySortBy... sortBy) {
```
jamesonwilliams, line 87: The new `sorted` and `matches` change `QueryOptions` into a mutable class. 

One ~easy solution would be to have these generate clones, that "patch" the field of interest.

e.g.
```java
    public QueryOptions sorted(@NonNull final QuerySortBy... sortBy) {
        return new QueryOptions(queryPredicate, paginationInput, Immutable.of(Arrays.asList(sortBy)));
    }
```
Also, it would certainly be ___cool___* if we could build up the list of `QuerySortBy` through subsequent appended calls:
```java
Where.sorted(User.LAST_NAME.ascending())
    .sorted(User.FIRST_NAME.ascending())
    .sorted(User.CREATED_AT.descending())
```

If we were to support that, I guess the impl of this method would be more like:
```java
    public QueryOptions sorted(@NonNull final QuerySortBy... sortBy) {
        List<QuerySortBy> augmentedList = new ArrayList<>(this.sortBy);
        augmentedList.addAll(Immutable.of(sortBy));
        return new QueryOptions(queryPredicate, paginationInput, Immutable.of(augmentedList));
    }
```

core/src/main/java/com/amplifyframework/core/model/query/QueryOptions.java
```diff
@@ -22,28 +22,46 @@
...
+    }
+
+    /**
+     * Update the current options with a given {@code queryPredicate}.
+     *
+     * @param queryPredicate predicate.
+     * @return current options with an updated {@code queryPredicate}.
+     */
+    @NonNull
+    public QueryOptions matches(@NonNull final QueryPredicate queryPredicate) {
```
jamesonwilliams, line 62: Mostly same discussion on this one as `sorted(...)` I think, except without the list building crap. TLDR, See if you can return a new copy of `QueryOptions` instead of mutating this instance. This will keep the options object immutable.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLiteTable.java
```diff
@@ -173,6 +175,22 @@ public String getPrimaryKeyColumnName() {
         return Immutable.of(foreignKeys);
     }
 
+    /**
+     * Returns the SQLiteColumn for the provided columnName.
+     * @param columnName columnName of the column
```
eeatonaws, line 180: Nit: Instead of "columnName of the column" it may be more clear to say "name of the column"

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SqlKeyword.java
```diff
@@ -203,6 +223,21 @@ public static SqlKeyword fromQueryPredicateGroup(@NonNull QueryPredicateGroup.Ty
         return sqlKeyword;
     }
 
+    /**
+     * Retrieve the SQL specific keyword for the query sort order type.
+     * @param sortOrder the query sort order type
+     * @return the SQL specific keyword
+     */
```
eeatonaws, line 230: Since the return value of this method will not be null, you could annotate this method with @NonNull.

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -0,0 +1,92 @@
...
+import com.amplifyframework.core.model.query.predicate.QueryField;
+import com.amplifyframework.datastore.DataStoreCategoryBehavior;
+
+import java.util.Objects;
+
+/**
+ * A data structure representing a model field and an order to sort by (ascending or descending), used to specify the
+ * order of results from {@link DataStoreCategoryBehavior#query(Class, QueryOptions, Consumer, Consumer)}.
+ *
+ * The preferred way to create a QuerySortBy is with the @{@link QueryField#ascending()} and
```
eeatonaws, line 31: Should the first "@" in "@{@link QueryField#ascending()}" be removed?

richardmcclellan: Good catch!

core/src/main/java/com/amplifyframework/core/model/query/QueryOptions.java
```diff
@@ -56,8 +73,17 @@
      */
     @NonNull
     public QueryOptions paginated(@NonNull final QueryPaginationInput paginationInput) {
-        this.paginationInput = paginationInput;
-        return this;
+        return new QueryOptions(queryPredicate, paginationInput, sortBy);
+    }
+
+    /**
+     * Returns an immutable copy of the current query options with the the given {@code sortBy} arguments.
```
eeatonaws, line 80: Nit: "the" is repeated twice in this comment ("with the the given")

core/src/main/java/com/amplifyframework/core/model/query/QueryOptions.java
```diff
@@ -56,8 +73,17 @@
...
+     *
+     * @param querySortBy sorting information.
+     * @return current options with an updated {@code sortBy}.
+     */
+    public QueryOptions sorted(@NonNull final QuerySortBy... querySortBy) {
+        return new QueryOptions(queryPredicate, paginationInput, Arrays.asList(querySortBy));
```
eeatonaws, line 86:
```suggestion
        return new QueryOptions(queryPredicate, paginationInput, Arrays.asList(Objects.requireNonNull(querySortBy)));
```

PR #639, opened 2020-07-15
chore: update api levels and dependencies

jamesonwilliams:
desugaring tools -> 1.0.9

The library will not currently run on API levels below 21. Reflect this
reality in the minSdkVersion as well as in the project README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Guide](https://docs.amplify.aws/start/q/integration/android).
 
 ## Platform Support
 
-The Amplify Framework supports Android API level 16 (Android 4.1) and above.
+The Amplify Framework supports Android API level 21 (Android 5.0) and above.
 
 ## Using Amplify from Your App
 
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
@@ -95,8 +95,8 @@ public final class UserAgentTest {
      *          waiting for request
      */
     @Test
-    @Config(sdk = 16)
-    public void testUserAgentWithApi16() throws Exception {
+    @Config(sdk = 21)
+    public void testUserAgentWithApi21() throws Exception {
         String userAgent = checkUserAgent();
 
         // Assert the correct format and content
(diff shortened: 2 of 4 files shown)


TrekSoft:
Whoah! Did I miss a conversation on this? 16 to 21 seems like a big jump and we've had a bunch of complaints about that on the SDK.

jpignata:
> The library will not currently run on API levels below 21.

Do we have issues with < 21 today? If so, we'll need to invest in fixing those rather than bumping the `minSdkVersion`. We've agreed to support 16+.

jamesonwilliams:
> Do we have issues with < 21 today? If so, we'll need to invest in fixing those rather than bumping the `minSdkVersion`. We've agreed to support 16+.

Yes; the library will only run on 21+ right now, and the `minSdkVersion` is incorrect from a technical standpoint, independent of market factors.

TrekSoft:
Which parts of the library only run on 21+ now? Is it true we can't use any libraries under 21? Is it only certain methods?

jpignata:
I'll add some work to the board to do analysis here. I don't think we've gotten any inquiries from customers here yet (but please reach out if you're reading this and hitting a problem with <21!)

jamesonwilliams:
> Which parts of the library only run on 21+ now? Is it true we can't use any libraries under 21? Is it only certain methods?

There are a few different issues; at least the API and DataStore categories are impacted. Currently we use a secure, modern release of OkHttp, which requires API 21. We are also using some SQL features that require a more recent API level than 16. I have not created an exhaustive list, though. Refactoring the API category to make use of the `HttpUrlConnection` would lower the floor considerably. That refactoring work would also decrease the dex count for our customers, considerably.

TrekSoft:
Maybe rather than mark core as min sdk 21+ we mark the plugins we know don't currently run on less than 21 (so API and DataStore for now) as min sdk 21. Because I agree it's better to be upfront if a plugin isn't going to work on less than 21 until we address the underlying cause.

jamesonwilliams:
> Maybe rather than mark core as min sdk 21+ we mark the plugins we know don't currently run on less than 21 (so API and DataStore for now) as min sdk 21. Because I agree it's better to be upfront if a plugin isn't going to work on less than 21 until we address the underlying cause.

I agree that this would be an improvement. iOS API and DataStore are [only available on iOS 13+](https://github.com/aws-amplify/amplify-ios/search?q=%22%40available%28iOS+13.0%2C+%2A%29%22&type=Code). Our business has requested that we support older platforms on Android.

PR #647, opened 2020-07-16
fix(api) Fix tests broken by PR #599

rjuliano:
*Issue #, if available:*

*Description of changes:*
My changes for #599 ended up causing some subscription related tests to fail. This PR fixes that.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -68,8 +68,8 @@ final class SubscriptionEndpoint {
     private final GraphQLResponse.Factory responseFactory;
     private final TimeoutWatchdog timeoutWatchdog;
     private final Set<String> pendingSubscriptionIds;
-    private final String subscriptionUrl;
     private final OkHttpClient okHttpClient;
+    private String subscriptionUrl;
     private WebSocket webSocket;
     private AmplifyWebSocketListener webSocketListener;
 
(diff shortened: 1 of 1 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -103,6 +102,14 @@
         Objects.requireNonNull(onSubscriptionError);
         Objects.requireNonNull(onSubscriptionComplete);
 
+        if (subscriptionUrl == null) {
+            try {
+                subscriptionUrl = buildConnectionRequestUrl();
+            } catch (ApiException apiException) {
+                onSubscriptionError.accept(apiException);
```
jamesonwilliams, line 109: Should you add `return;` after this line, to cede the flow of control from `requestSubscription(...)`?

rjuliano: Yikes...good eye on that one. I debated just making the function throw, but it started betting messy upstream.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -107,10 +105,16 @@
...
-            webSocket = okHttpClient.newWebSocket(new Request.Builder()
-                .url(subscriptionUrl)
-                .addHeader("Sec-WebSocket-Protocol", "graphql-ws")
-                .build(), webSocketListener);
+            try {
+                webSocket = okHttpClient.newWebSocket(new Request.Builder()
+                    .url(buildConnectionRequestUrl())
+                    .addHeader("Sec-WebSocket-Protocol", "graphql-ws")
+                    .build(), webSocketListener);
+            } catch (ApiException apiException) {
```
jamesonwilliams, line 113: What actually throws the `ApiException`? The `buildConnectionRequestUrl()` bit? The rest looks likes its pure OkHttp, and wouldn't throw one of our internal types.

rjuliano: Yeah...it's buried deep down into the code that adds the security headers.

PR #648, opened 2020-07-16
Create device information UI for developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Creates the developer menu device information UI (shown below).

Dev Menu Device Info Screen


core/src/main/res/layout/fragment_device.xml
```diff
@@ -14,10 +14,17 @@
    permissions and limitations under the License.
 -->
 
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
+    android:layout_height="150dp"
```
TrekSoft, line 20: Am I correctly reading the activity_dev_menu.xml code to have been designed to have its height vary based on the height of its content (such as this)? 

If so, I think it'd be better to have the outer visible dev menu set to a fixed height so that we don't inadvertently have a child  with a taller height than another and the dev menu grows and shrinks in height as you go from one option to another. But maybe I'm missing something in your code and this is not what's happening.

eeatonaws: You're correct and that is a good point. I'll set the developer menu to have a consistent height across screens.

core/src/main/res/values/strings.xml
```diff
@@ -24,4 +24,5 @@
     <string name="device_view_title">Device Information</string>
     <string name="logs_view_title">Logs</string>
     <string name="file_issue_view_title">File Issue</string>
+    <string name="placeholder_device_info">Temporary text to be replaced by device info</string>
```
jamesonwilliams, line 27: Hm. Maybe something like "Loading device info..." ?

eeatonaws: That String resource will be removed when I write the code to show the actual device information.

PR #649, opened 2020-07-16
Developer menu navigation unit tests

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Added unit tests that test the developer menu navigation when pressing a button on the main screen of the developer menu.


core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Tests the navigation behavior for {@link DevMenuMainFragment}.
+ */
+public final class DevMenuMainViewInstrumentationTest {
+    // Amount of time (in milliseconds) to wait before starting
+    // a test, to allow any animations to complete.
+    private static final int START_TEST_DELAY_MS = 500;
+    // A navigation host controller for testing.
+    private static TestNavHostController navHostController;
```
raphkim, line 42: does this variable have to be static?

core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
```diff
@@ -0,0 +1,107 @@
...
+        testNavigationOnButtonPress(R.id.file_issue_button, R.id.file_issue_fragment);
+    }
+
+    /**
+     * Test that the app is navigated to the destination with the given ID when
+     * the button with the given ID is pressed.
+     * @param buttonId ID of the button pressed
+     * @param destinationId ID of the navigation destination
+     */
+    private void testNavigationOnButtonPress(int buttonId, int destinationId) {
```
raphkim, line 100: Not a big deal since it's a private helper function, but `@RawRes` annotation helps clarify that the param is looking for a resource ID :)

core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
eeatonaws: No. I think that may have been left over from a previous iteration of these tests, so I'll remove "static".

build.gradle
```diff
@@ -63,13 +64,16 @@ ext {
                 fragmentktx: "androidx.navigation:navigation-fragment-ktx:$navigationVersion",
                 ui: "androidx.navigation:navigation-ui:$navigationVersion",
                 uiktx: "androidx.navigation:navigation-ui-ktx:$navigationVersion",
-                dynamicfeatures: "androidx.navigation:navigation-dynamic-features-fragment:$navigationVersion"
+                dynamicfeatures: "androidx.navigation:navigation-dynamic-features-fragment:$navigationVersion",
+                navtesting: "androidx.navigation:navigation-testing:$navigationVersion",
+                fragmenttesting: "androidx.fragment:fragment-testing:$fragmentVersion"
```
jamesonwilliams, line 69: I guess these should go under `test`, not `androidx`? I _do_ currently have Android X dependencies down there, as well. I'll admit that the layout of this `ext` block makes no sense. I just _did some stuff_. We can re-organize things if you can think of a better standardization. Currently I am trying to minimize the average length of `dep.foo.bar.whatever`, while still having descriptive and meaningful phrasing. A linguistic optimization challenge of sorts, information science at its core! Aye.

core/build.gradle
```diff
@@ -47,4 +47,25 @@ dependencies {
     androidTestImplementation dependency.androidx.test.core
     androidTestImplementation dependency.androidx.test.runner
     androidTestImplementation dependency.androidx.test.junit
+    androidTestImplementation dependency.androidx.test.espresso
+    androidTestImplementation dependency.androidx.nav.navtesting
+
+    debugImplementation dependency.androidx.nav.fragmenttesting
```
jamesonwilliams, line 53: Shouldn't this say `androidTestImplementation`? `debugImplementation` is an `implementation` for the `debug` build. If you need it in the main artifact (not the test artifact), I'd sort this up int othe `implementation` section, as opposed to the instrument test dependencies section, where it is now.

core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Test;
+
+import static androidx.test.espresso.Espresso.onView;
+
+/**
+ * Tests the navigation behavior for {@link DevMenuMainFragment}.
+ */
+public final class DevMenuMainViewInstrumentationTest {
+    // Amount of time (in milliseconds) to wait before starting
+    // a test, to allow any animations to complete.
```
jamesonwilliams, line 40: You can also try:
```groovy
android {

  ...

  testOptions {
    animationsDisabled = true
  }
}
```
to disable animations, and prevent the need for this sleep. (Sleeps in test code are brittle, since they're outside of the deterministic flow of the source code's logic.)

It would probably be appropriate, for `amplify-android` anyway, to cook that `testOption` right into the project-level `build.gradle`. I say this because, this is not a UI-centric project, so enablement of animations isn't the ethos of what this project needs to be testing. If it were a dedicated UI library, we probably would consider more refinement with that.

core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Go to the main screen of the developer menu.
+     */
+    @Before
+    public void resetView() {
+        navHostController = new TestNavHostController(ApplicationProvider.getApplicationContext());
+        navHostController.setGraph(R.navigation.dev_menu_nav_graph);
+        FragmentScenario<DevMenuMainFragment> mainMenuScenario =
+                FragmentScenario.launchInContainer(DevMenuMainFragment.class);
+        mainMenuScenario.onFragment(fragment -> Navigation.setViewNavController(fragment.requireView(),
```
jamesonwilliams, line 54: Passing thought: personally, I like to wrap on `->` so I can see the action behavior all in one line,
like:

```java
setSomeFunctionalBehavior(() ->
    doTheStuff(withComplextParams, orWhatever, maybeItsALongLine));
```

core/src/androidTest/java/com/amplifyframework/core/DevMenuMainViewInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+     * the button with the given ID is pressed.
+     * @param buttonId ID of the button pressed
+     * @param destinationId ID of the navigation destination
+     */
+    private void testNavigationOnButtonPress(@RawRes int buttonId, @RawRes int destinationId) {
+        Sleep.milliseconds(START_TEST_DELAY_MS);
+        onView(ViewMatchers.withId(buttonId)).perform(ViewActions.click());
+        NavDestination curDestination = navHostController.getCurrentDestination();
+        Assert.assertNotNull(curDestination);
+        Assert.assertEquals(curDestination.getId(), destinationId);
```
jamesonwilliams, line 106: The predominant style in the repo is to use static imports for jUnit and Mockito utility code. That is, here, just `assertEquals`, `assertFoo`, `assertBar`, without the `Assert` class ref.

PR #656, opened 2020-07-20
chore(datastore): Use AppSyncGraphQLRequest instead of SimpleGraphQLRequest for sync queries

richardmcclellan:
This is the first step towards adding pagination support in `DataStore`.  We need to use `AppSyncGraphQLRequest` instead of `SimpleGraphQLRequest`, so that we can use the deserialize the response as an `AppSyncPaginatedResult` using `AppSyncPaginatedResultDeserializer`.


core/src/main/java/com/amplifyframework/api/graphql/QueryType.java
```diff
@@ -27,7 +27,12 @@
     /**
      * GraphQL query list.
      */
-    LIST;
+    LIST,
+
+    /**
+     * GraphQL query sync.
+     */
+    SYNC;
```
jamesonwilliams, line 35: Almost certainly not a comment for this PR, but while I'm here:

I hadn't thought about this at the time, but _get_, _list_, _sync_: these are really verbs from the AppSync domain, aren't they? GraphQL itself has no such pre-defined operation types. Should we aim to move the `QueryType` enum into `aws-api-appsync`?

aws-datastore/src/test/resources/delta-sync-request-document-for-post.txt
```diff
@@ -1,6 +1,9 @@
-query SyncPosts {
-  syncPosts(lastSync: 123123123) {
+query SyncPosts($lastSync: AWSTimestamp) {
```
jamesonwilliams, line 1: An `AWSTimestamp` in the wild! How cool :-D

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactoryTest.java
```diff
@@ -69,10 +73,10 @@ public void validateRequestGenerationForDeltaSync() throws DataStoreException {
     @Test
     public void validateRequestGenerationForPagination() throws DataStoreException {
         final String nextToken = Resources.readAsString("base-sync-request-next-token-value.txt").trim();
-        assertEquals(
-            Resources.readAsString("base-sync-request-paginating-blog-owners.txt"),
-            AppSyncRequestFactory.buildSyncDoc(BlogOwner.class, null, nextToken)
-        );
+        final GraphQLRequest<Iterable<Post>> request = AppSyncRequestFactory
+                .buildSyncRequest(BlogOwner.class, null, nextToken);
```
jamesonwilliams, line 77: nit-pick: consider wrapping on the `=`, not on the `.`.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactoryTest.java
```diff
@@ -30,6 +31,8 @@
 
 import java.util.Map;
 
+import edu.emory.mathcs.backport.java.util.Collections;
```
jamesonwilliams, line 34: Just `java.util.Collections` here, not the Emory repackaged one.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -85,70 +84,32 @@ private AppSyncRequestFactory() {}
      * @throws DataStoreException On Failure to inspect
      */
     @NonNull
-    static <T extends Model> String buildSyncDoc(
-            @NonNull final Class<T> modelClass,
+    static <T, M extends Model> AppSyncGraphQLRequest<T> buildSyncRequest(
```
jamesonwilliams, line 87: What are the possible resolutions for `T`, here -- `Iterable<String>`, and importantly, some upcoming type of paginated result?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -85,70 +84,32 @@ private AppSyncRequestFactory() {}
...
-        // Outer container, e.g. query SyncBlogPost {
-        doc.append("query Sync").append(capitalizedPluralName).append(" {\n");
-
-        // Inner directive, e.g. syncBlogPosts(
-        doc.append(padBy(++indent)).append("sync").append(capitalizedPluralName);
+            AppSyncGraphQLRequest.Builder builder = AppSyncGraphQLRequest.builder()
+                    .modelClass(modelClass)
+                    .operation(QueryType.SYNC)
+                    .requestOptions(new DataStoreGraphQLRequestOptions())
+                    .responseType(new TypeToken<Iterable<String>>(){}.getType());
```
jamesonwilliams, line 98: You have your little type factory gizmo where you can do like `TypeFactory.of(Iterable.class, String.class)` right? That'd remove the `gson` import from the top of the doc, which would be cool from an encapsulation standpoint.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
richardmcclellan: Yep, right now it is `Iterable<String>`, but my next PR will change it to be `PaginatedResult<ModelWithMetadata<M>>`

core/src/main/java/com/amplifyframework/api/graphql/QueryType.java
richardmcclellan: That makes sense, I will move it to `aws-api-appsync`.  I think there's several more things like this that I will also be moving over soon.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
richardmcclellan: Updated!

richardmcclellan:
@jamesonwilliams feedback addressed!

Issue #644, opened 2020-07-16
AWS api gateway, returning {"message":"Forbidden"}


tendynewbies:
Hello n' Greetings!!

I am using **implementation 'com.amplifyframework:aws-api:1.0.0'** as suggested for [API(REST)](https://docs.amplify.aws/lib/restapi/fetch/q/platform/android#accessing-query-parameters--body-in-lambda-proxy-function).

My **amplifyconfiguration.json** looks like:
```json
"api": {
    "plugins": {
        "awsAPIPlugin": {
            "VODApi": {
                "endpointType": "REST",
                "endpoint": "https://p4wfal4mb8.execute-api.ap-south-1.amazonaws.com/POC/vod-with-aws",
                "region": "ap-south-1",
                "authorizationType": "AWS_IAM"
            }
        }
    }
},
"auth": .........
```

My **Amplify initialization** looks like:
```java
try {
    Amplify.addPlugin(new AWSCognitoAuthPlugin());
    Amplify.addPlugin(new AWSS3StoragePlugin());
    Amplify.addPlugin(new AWSApiPlugin());
    Amplify.configure(getApplicationContext());
    Log.i("Media Analysis", "Initialized Amplify");
} catch (AmplifyException error) {
    Log.e("Media Analysis", "Could not initialize Amplify", error);
}
```

My **API call** looks like:
```java
private void fetchPreparedStreams() {
    RestOptions options = RestOptions.builder()
        .addPath("convert?userId="+Amplify.Auth.getCurrentUser().getUserId())
        .build();
    Amplify.API.get("VODApi", options,
        response -> Log.i("MyAmplifyApp", response.getData().asString()),
        error -> Log.e("MyAmplifyApp", "GET failed", error)
    );
}
```

**The Unexpected Response**, I am facing looks like:
`SUCCESS: {"message":"Forbidden"}`

I think, this might be similar to https://github.com/aws-amplify/aws-sdk-android/issues/968.

Please help, urgent client-delivery at stake.

jpignata:
Hi! I think you want to call `addQueryParameters` to add the query string, which is likely why the query string isn't being processed correctly. See https://github.com/aws-amplify/amplify-android/blob/12ec8e9819228f63bd677a0431a5ab191f79d9e1/core/src/main/java/com/amplifyframework/api/rest/RestOptions.java#L184 for details.

tendynewbies:
Hi @jpignata,
Greetings!!

[...]

PR #658, opened 2020-07-21
fix(api): now correctly signs requests with query string params

raphkim:
*Issue #, if available:*
#650 #644 

*Description of changes:*
* The request was being SigV4 signed without adding the query parameters. This was fixed.

*Minor changes:*
* URL is now decoded and re-encoded inside `RestRequestFactory` to ensure correct encoding.
* Signer now distinguishes AppSync and API gateway when signing

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
@@ -29,7 +29,6 @@ import com.amazonaws.util.IOUtils;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLDecoder;
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
@@ -25,12 +25,14 @@ import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.http.HttpMethodName;
 import com.amazonaws.util.IOUtils;
+import com.amazonaws.util.StringUtils;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


adil-uddin:
@raphkim thanks for this. I had this issue when I had encoded path params too.
For ex: /users/user%231145 -> ( /users/user#1145)
does this PR fix this too ?

aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
```diff
@@ -25,9 +25,15 @@
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.http.HttpMethodName;
 import com.amazonaws.util.IOUtils;
+import com.amazonaws.util.StringUtils;
```
jamesonwilliams, line 28: This enters the project from a transitive dependency. IMO we should minimize our use of `com.amazonaws.*` overall, and try to stick to public APIs related to the per-category functionalities.

Since you're only using it for an `isBlank(...)` check, use our `Empty.check(...)` instead.

aws-api/src/main/java/com/amplifyframework/api/aws/sigv4/AppSyncSigV4SignerInterceptor.java
```diff
@@ -221,23 +236,23 @@ public Response intercept(Chain chain) throws IOException {
...
+    // Source: https://stackoverflow.com/questions/13592236/parse-a-uri-string-into-name-value-collection
+    @NonNull
+    private static Map<String, String> splitQuery(URL url) throws UnsupportedEncodingException {
+        Map<String, String> queryPairs = new LinkedHashMap<>();
+        String query = url.getQuery();
+        if (StringUtils.isBlank(query)) {
+            return Collections.emptyMap();
+        }
+        String[] pairs = query.split("&");
+        for (String pair : pairs) {
```
jamesonwilliams, line 249: I had some [similar code](https://github.com/aws-amplify/aws-sdk-android/blob/5dd0186faf85ad1b7482b641c3d69d70a31db465/aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java#L217-L250) recently and @richardmcclellan raised [some good points](https://github.com/aws-amplify/aws-sdk-android/pull/2008#discussion_r453857894): what would happen here if the URL were malformed? Is a null pointer or out-of-range error thrown?

adil-uddin:
> > @raphkim thanks for this. I had this issue when I had encoded path params too.
> > For ex: /users/user%231145 -> ( /users/user#1145)
> > does this PR fix this too ?
> 
> @adil-uddin can you clarify that -- what did you input (in Java code), what did you expect to happen, and what should happen? Maybe @raphkim can add a unit test to make sure it is covered.

https://github.com/aws-amplify/amplify-android/issues/655 This issue is exactly my use case.

jpignata:
Hi @adil-uddin -- anything after a hash mark is part of the fragment uri which is not sent to the server. I don't expect this to work with the REST category and would suggest using a query parameter instead.

PR #660, opened 2020-07-22
Show device info on developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Displays the actual device information on the developer menu. Also adds a small amount of padding to the left of the title for each screen on the developer menu.

Device information screen for a physical device:
Device info -- real device

Device information screen for an emulator:
Device info -- emulator


core/src/main/java/com/amplifyframework/core/DevMenuDeviceFragment.java
```diff
@@ -36,6 +37,9 @@ public DevMenuDeviceFragment() { }
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         // Inflate the layout for this fragment
-        return inflater.inflate(R.layout.fragment_device, container, false);
+        View deviceInfoView = inflater.inflate(R.layout.fragment_device, container, false);
+        // Set the device information text.
+        ((TextView) deviceInfoView.findViewById(R.id.device_info_text)).setText(new DeviceInfo().toString());
```
jamesonwilliams, line 42:
```suggestion
        View deviceInfoView = inflater.inflate(R.layout.fragment_device, container, false);
        // Set the device information text.
        TextView deviceInfoText = deviceInfoView.findViewById(R.id.device_info_text);
        deviceInfoText.setText(new DeviceInfo().toString());
```

core/src/main/java/com/amplifyframework/core/DeviceInfo.java
```diff
@@ -0,0 +1,67 @@
...
+        return Build.DEVICE.toLowerCase(Locale.getDefault()).contains("generic")
+                && MODEL.toLowerCase(Locale.getDefault()).contains("sdk");
+    }
+
+    /**
+     * Returns a String representation of the device information.
+     * @return a String containing device information.
+     */
+    public String toString() {
+        String result = "Device Manufacturer: " + MANUFACTURER + "\nDevice Model: " + MODEL
```
jamesonwilliams, line 58: I suggest using `String.format(...)` instead, to make the main message more readable. This will also allow you to split out your `isEmulator()` foramtting business:

```
String.format(
    ".. blah ... is emu? %s",
    isEmulator() ? "Yes" : "No"
);
```

core/src/main/java/com/amplifyframework/core/DeviceInfo.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public static final String MODEL = Build.MODEL;
+    /**
+     * SDK version running on the device.
+     */
+    public static final int SDK_VERSION = Build.VERSION.SDK_INT;
+    /**
+     * Android system version running on the device.
+     */
+    public static final String SYSTEM_VERSION = Build.VERSION.RELEASE;
```
jamesonwilliams, line 42: Just use the `Build.FOO` values directly. I don't think this provides more clarity, or much encapsulation.

core/src/main/java/com/amplifyframework/core/DeviceInfo.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Returns true if the device is an emulator and false if not.
+     * @return a boolean indicating whether the device is an emulator or not
+     */
+    public boolean isEmulator() {
+        return Build.DEVICE.toLowerCase(Locale.getDefault()).contains("generic")
+                && MODEL.toLowerCase(Locale.getDefault()).contains("sdk");
```
jamesonwilliams, line 50: Is this a standard solution? Can you link to StackOverflow in a comment, if so?

eeatonaws: When researching how to detect if a device is an emulator, there were many different approaches and I came up with my own approach based on multiple answers to a StackOverflow post, but I can add a link to that post in a comment.

core/src/main/java/com/amplifyframework/core/DeviceInfo.java
```diff
@@ -0,0 +1,47 @@
...
+        return Build.DEVICE.toLowerCase(Locale.getDefault()).contains("generic")
+                && Build.MODEL.toLowerCase(Locale.getDefault()).contains("sdk");
+    }
+
+    /**
+     * Returns a String representation of the device information.
+     * @return a String containing device information.
+     */
+    public String toString() {
+        return String.format(Locale.getDefault(), "Device Manufacturer: %s\nDevice Model: %s\nAndroid System Version: "
```
jamesonwilliams, line 43: I think you can hardcode `Locale.US` here, since most of the text is. That might be more defensive against unexpected behavior.

PR #661, opened 2020-07-22
Sends proper user agent for Auth

TrekSoft:
Currently Auth isn't sending the proper Amplify user agent because AWSMobileClient overrides it. This works in conjunction with this PR on the SDK branch to enable us to override that user agent with our own: https://github.com/aws-amplify/aws-sdk-android/pull/2024

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -124,10 +124,12 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     ) throws AuthException {
         final CountDownLatch latch = new CountDownLatch(1);
         final AtomicReference<Exception> asyncException = new AtomicReference<>();
+        JSONObject mobileClientConfig;
         LogFactory.setLevel(LogFactory.Level.OFF);
 
         try {
-            pluginConfiguration.put("UserAgentOverride", UserAgent.string());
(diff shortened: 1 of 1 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -124,6 +126,14 @@ public void configure(
         final AtomicReference<Exception> asyncException = new AtomicReference<>();
         LogFactory.setLevel(LogFactory.Level.OFF);
 
+        try {
+            pluginConfiguration.put("UserAgentOverride", UserAgent.string());
```
jamesonwilliams, line 130: It's a bad practice to mutate method arguments. Here, the following occurs:
1. `pluginConfiguration` has state A.
2. `configure(..., pluginConfiguration, ...)` runs.
3. `pluginConfiguration` has state B.

Instead, I would do:
```java
JSONObject mobileClientConfig = new JSONObject(pluginConfiguration);
mobileClientConfig.put("UserAgentOverride", UserAgent.string());
```

TrekSoft: Done

raphkim:
> It would be possible to use a `MockWebServer` and capture the outgoing request from this component, and snoop the `User-Agent` header. I think @raphkim might have done a test like that, not to long ago?

Yes! Here's the file that does that:
https://github.com/aws-amplify/amplify-android/blob/main/aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java

Issue #636, opened 2020-07-13
[DataStore] Retry initial sync after login


a-melnyk:
**Describe the bug**
DataStore doesn't retry initial sync from AppSync after login

**Use cases**
* user removes the app & reinstalls it, doesn't see any data after login
* user switches to a new android device, doesn't see any data after login on the new device
* user logged out a device (or Cognito token expired on the device), changed his data on another device and receives stale data on the first device after a login

**Environment**
* Amplify API with Cognito User Pool authorisation
* com.amplifyframework:core:1.0.0
* com.amplifyframework:aws-datastore:1.0.0
* com.amplifyframework:aws-api:1.0.0
* com.amplifyframework:aws-auth-cognito:1.0.0
* amplify-cli 4.23.1

**Details**
When the framework doesn't have a valid Cognito token (during the initial start or in case of token expiration), it can't authenticate requests to AppSync and throws AmplifyExceptions during initial sync:
```
2020-07-13 14:18:27.626 17661-17785/org.example.myapp D/AWSMobileClient: Inspecting user state details
2020-07-13 14:18:27.630 17661-17785/org.example.myapp D/AWSMobileClient: waitForSignIn: userState:SIGNED_OUT
2020-07-13 14:18:27.637 17661-17784/org.example.myapp W/amplify:aws-datastore: Failed to sync down cloud state for model type = SampleMedication
    AmplifyException {message=Failure performing sync query to AppSync., cause=AmplifyException {message=Could not retrieve the response body from the returned JSON, cause=java.io.IOException: Failed to retrieve Cognito User Pools token., recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}
        at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$sync$1(AppSyncClient.java:1)
        at com.amplifyframework.datastore.appsync.-$$Lambda$AppSyncClient$suvEma05s-p8v-66iE4cswLKYac.accept(Unknown Source:4)
        at com.amplifyframework.api.aws.AppSyncGraphQLOperation$OkHttpCallback.onFailure(AppSyncGraphQLOperation.java:1)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:510)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[...]

PR #662, opened 2020-07-23
fix: start orchestrator before any datastore operations

jamesonwilliams:
fix: daemon style lifecycle for orchestrator
    
    The state machine in the orchestrator is changed. The three possible
    states are now:
    
      a. STOPPED      No sync component is running, and local changes are not
                      being queued into an offline mutations queue.
      b. LOCAL_ONLY   Local changes are being buffered into an offline
                      mutations queue, but are not being transacte with the
                      cloud
      c. SYNC_VIA_API Data is syncrhonizing to and from the cloud.
    
    Logic is added to transition safely between any pair of these states
    (there are six possible transitions.)
    
    If the Orchestrator fails to start up cleanly, then tear down any/all
    copmonents that were started, and enter offline-only mode.
    
    0. Retry logic is _removed_ from the subscription processor. Instead, if
       a subscription fails, the failure is communicated back to the
       Orchestrator. The Orchestrator should then exist API sync mode.
    
    1. The AWSDataStorePlugin had strated to take on sync engine
       orchestration concerns. All orchestration concerns are removed from
       the plugin, and moved into the Orchestrator, again.
    
    2. Surface exceptions from the hydrate() call on the SyncProcessor. This
       way, if the sync fails, the orchestrator will catch the exception,
       and begin a graceful shutdown sequence.
    
    Some changes amde in the test code include:
    
    0. Mock AppSync behaviors in the OrchestratorTest. Previously, the
       subcription code was not being included in the test. Once exercised,
       it became necessarily to fulfill the missing mocks.
    
    1. Start listening for HubEvents in AWSDataStorePluginTest _before_ they
       will be published. This will help to ensure we always can catch them,
       in th tests.
    
    2. Since the SyncEngine errors are no longer gobbled, the behavior of
       userProvidedErrorCallbackInvokedOnFailure has changed. We now
       expect only a single call to the user-provided error callback,
       instead of _many_ consecutives ones, as before.
    
    3. Add a version of Latch#await(int, TimeUnit) which allows a custom
       wait time to be provided. This is useful in the orchestrator code,
       where longer-than-usual timeouts are encountered.
    
    Future work will address transient network failures, further. Instead of
    individual sync engine components owning their own retries, the
    Orchestrator's lifecycle *as a whole* should be retried. When the
    network goes down, or when auth fails -- any transient failure -- the
    Orchestrator should back off and enter LOCAL_ONLY mode. Additional retry
    logic added to the orchestrator will ensure that it intelligently
    attempts to reach the target mode, again.
    
    Resolves: https://github.com/aws-amplify/amplify-android/issues/636

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -15,7 +15,6 @@
 
 package com.amplifyframework.datastore;
 
-import android.annotation.SuppressLint;
 import android.content.Context;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -52,9 +51,9 @@ import java.util.Iterator;
 import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.reactivex.Completable;
+import io.reactivex.disposables.Disposable;
 import io.reactivex.schedulers.Schedulers;
 
 /**
@@ -62,8 +61,8 @@ import io.reactivex.schedulers.Schedulers;
  */
 public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
-    private static final long PLUGIN_INIT_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(5);
-    private static final long PLUGIN_TERMINATE_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(5);
+    private static final long LIFECYCLE_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(5);
+
     // Reference to an implementation of the Local Storage Adapter that
     // manages the persistence of data on-device.
     private final LocalStorageAdapter sqliteStorageAdapter;
@@ -75,11 +74,6 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
     // Keeps track of whether of not the category is initialized yet
     private final CountDownLatch categoryInitializationsPending;
 
-    private final AtomicBoolean isOrchestratorReady;
-
-    // Used to interrogate plugins, to understand if sync should be automatically turned on
-    private final ApiCategory api;
-
     // User-provided configuration for the plugin.
     private final DataStoreConfiguration userProvidedConfiguration;
 
@@ -87,7 +81,6 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
     // overrides provided via the userProvidedConfiguration
     private DataStoreConfiguration pluginConfiguration;
 
-    @SuppressLint("CheckResult")
     private AWSDataStorePlugin(
             @NonNull ModelProvider modelProvider,
             @NonNull ModelSchemaRegistry modelSchemaRegistry,
@@ -95,14 +88,14 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
             @Nullable DataStoreConfiguration userProvidedConfiguration) {
         this.sqliteStorageAdapter = SQLiteStorageAdapter.forModels(modelSchemaRegistry, modelProvider);
         this.categoryInitializationsPending = new CountDownLatch(1);
-        this.isOrchestratorReady = new AtomicBoolean(false);
-        this.api = api;
+        // Used to interrogate plugins, to understand if sync should be automatically turned on
         this.orchestrator = new Orchestrator(
             modelProvider,
             modelSchemaRegistry,
             sqliteStorageAdapter,
             AppSyncClient.via(api),
-            () -> pluginConfiguration
+            () -> pluginConfiguration,
+            () -> api.getPlugins().isEmpty() ? Orchestrator.Mode.LOCAL_ONLY : Orchestrator.Mode.SYNC_VIA_API
         );
         this.userProvidedConfiguration = userProvidedConfiguration;
     }
@@ -167,7 +160,6 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
     /**
      * {@inheritDoc}
      */
-    @SuppressLint("CheckResult")
     @Override
     public void configure(
             @NonNull JSONObject pluginConfiguration,
@@ -198,18 +190,14 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
     @Override
     public void initialize(@NonNull Context context) throws AmplifyException {
         Throwable initError = initializeStorageAdapter(context)
-            .blockingGet(PLUGIN_INIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            .blockingGet(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         if (initError != null) {
-            throw new AmplifyException("Failed to initialize the local storage adapter for the DataStore plugin.",
-                                        initError,
-                                        AmplifyException.TODO_RECOVERY_SUGGESTION);
+            throw new AmplifyException(
+                "Failed to initialize the local storage adapter for the DataStore plugin.",
+                initError, AmplifyException.TODO_RECOVERY_SUGGESTION
+            );
         }
-        // Kick off orchestrator asynchronously.
-        synchronized (isOrchestratorReady) {
-            initializeOrchestrator()
-                .subscribeOn(Schedulers.io())
-                .subscribe();
-        }
+        orchestrator.start();
     }
 
     /**
@@ -226,14 +214,12 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
 
     /**
      * Terminate use of the plugin.
-     * @throws AmplifyException On failure to terminate use of the plugin
      */
     @SuppressWarnings("unused")
-    synchronized void terminate() throws AmplifyException {
+    synchronized void terminate() {
         Throwable throwable = orchestrator.stop()
-            .andThen(
-                Completable.fromAction(sqliteStorageAdapter::terminate)
-            ).blockingGet(PLUGIN_TERMINATE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            .andThen(Completable.fromAction(sqliteStorageAdapter::terminate))
+            .blockingGet(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         if (throwable != null) {
             LOG.warn("An error occurred while terminating the DataStore plugin.", throwable);
         }
@@ -438,6 +424,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
      * @param onComplete Invoked if the call is successful.
      * @param onError Invoked if not successful
      */
+    @SuppressWarnings("unused")
     @Override
     public void clear(@NonNull Action onComplete,
                       @NonNull Consumer<DataStoreException> onError) {
@@ -445,7 +432,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
         // only thing we have to wait for is the category initialization latch.
         boolean isCategoryInitialized = false;
         try {
-            isCategoryInitialized = categoryInitializationsPending.await(PLUGIN_INIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            isCategoryInitialized = categoryInitializationsPending.await(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         } catch (InterruptedException exception) {
             LOG.warn("Execution interrupted while waiting for DataStore to be initialized.");
         }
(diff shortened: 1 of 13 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -472,14 +471,16 @@ public void clear(@NonNull Action onComplete,
 
     private void beforeOperation(@NonNull final Runnable runnable) {
         Throwable throwable = Completable.fromAction(categoryInitializationsPending::await)
+            .timeout(PLUGIN_INIT_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+            .andThen(initializeOrchestrator())
```
jamesonwilliams, line 475: This is the single most important line -- the core purpose -- of this PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -165,7 +167,14 @@ public synchronized Completable start(Action onLocalStorageReady) {
                 LOG.debug("Orchestrator started.");
                 announceRemoteSyncStarted();
             })
-        ).doFinally(startStopSemaphore::release);
+        )
+        .doOnError(error -> {
+            LOG.warn("Orchestrator failed to start.", error);
+            startStopSemaphore.release();
+            //noinspection ResultOfMethodCallIgnored
+            stop().blockingAwait(ACQUIRE_PERMIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
jamesonwilliams, line 175: If we don't have a clean start, we should teardown again, so we're back to a deterministic stop state.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -118,18 +118,15 @@ Completable hydrate() {
                         LOG.warn("Initial cloud sync failed.", failureToSync);
                         DataStoreErrorHandler dataStoreErrorHandler =
                             dataStoreConfigurationProvider.getConfiguration().getDataStoreErrorHandler();
-                        if (dataStoreErrorHandler != null) {
```
jamesonwilliams, line 121: According to Lint, the handler is never null.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -118,18 +118,15 @@ Completable hydrate() {
...
+                        dataStoreErrorHandler.accept(
+                            new DataStoreException(
+                                "Initial cloud sync failed.",
+                                failureToSync,
+                                "Check your internet connection."));
                     })
                     .doOnComplete(() ->
                         LOG.info("Successfully sync'd down model state from cloud.")
                     )
-                    .onErrorComplete()
```
jamesonwilliams, line 132: I need the `hydrate()` call to surface its failures, so that the orchestrator can respond to them by cleaning up and shutting down.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -499,9 +500,10 @@ private Completable initializeOrchestrator() {
                 // This callback is invoked when the local storage observer gets initialized.
                 isOrchestratorReady.set(true);
             })
-            .repeatUntil(() -> isOrchestratorReady.get())
+            .repeatUntil(isOrchestratorReady::get)
             .observeOn(Schedulers.io())
-            .subscribeOn(Schedulers.io());
+            .subscribeOn(Schedulers.io())
+            .onErrorComplete();
```
jamesonwilliams, line 506: (This used to be at the end of `hydrate()`.)

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: What's the effect of the timeout operator? Does it wait for the `categoryInitializationsPending` to be counted down up to `PLUGIN_INIT_TIMEOUT_MS`ms and then tries to start the orchestrator?

If there's no connectivity, will this be retried on any interaction with the DataStore?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: This makes sense. Only concern I have is similar to what I mentioned earlier. If a connection is not available what the impact on calls to the DataStore APIs?

jamesonwilliams: > If a connection is not available what the impact on calls to the DataStore APIs?

Thanks for bringing this case to my attention. This is an indeed an issue:
1. Device is offline;
2. User is not signed in;
3. A call to `save(...)` is made;
4. No response is rendered until `initializeOrchestrator()` fails.
5. At that point, the `onResult` consumer _is_ triggered as desired, indicating successful use of the local storage.

(4) is a problem.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: > What's the effect of the timeout operator? Does it wait for the `categoryInitializationsPending` to be counted down up to `PLUGIN_INIT_TIMEOUT_MS`ms and then tries to start the orchestrator?

I think that's how it's working, yes. That was the intent, anyway.

> If there's no connectivity, will this be retried on any interaction with the DataStore?

Yes, offline or not signed in. I guess we probably don't want to _block_ on its completion?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: To build on my response to the other comment, I think we only need to block interactions with the local DataStore until the StorageObserver is ready.

Maybe we need to account for more than just started/stopped for the orchestrator. 

```
if Orchestrator is STOPPED => // need to start everything and signal when StorageObserver is ready.
   triggerStartAndWaitForStorageObserver
   if StorageObserver STARTED =>
      emit storageObserverReady //This unlocks callers
   else
      Set Orchestrator back to STOPPED
      throw error back to caller.      
   [MutationProcessor, SubscriptionProcessor, SyncProcessor].asyncStart 
   if [MutationProcessor, SubscriptionProcessor, SyncProcessor] STARTED
     set Orchestrator STARTED
   else
     set Orchestrator FAILED

if STARTED => nothing to do, it's ready to go

if FAILED => trigger non-blocking retry.
   [MutationProcessor, SubscriptionProcessor, SyncProcessor].asyncStart 
   if [MutationProcessor, SubscriptionProcessor, SyncProcessor] STARTED
     set Orchestrator STARTED
   else
     set Orchestrator FAILED
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -75,34 +74,28 @@
...
-        this.api = api;
+        // Used to interrogate plugins, to understand if sync should be automatically turned on
         this.orchestrator = new Orchestrator(
             modelProvider,
             modelSchemaRegistry,
             sqliteStorageAdapter,
             AppSyncClient.via(api),
-            () -> pluginConfiguration
+            () -> pluginConfiguration,
+            api.getPlugins().isEmpty() ? Orchestrator.Mode.LOCAL_ONLY : Orchestrator.Mode.SYNC_VIA_API
```
rjuliano, line 98: One issue I ran into here was that if I add the DataStore plugin before the API plugin, the Orchestrator goes into LOCAL_ONLY mode:
```
// Starts in LOCAL_ONLY
 Amplify.addPlugin(AWSDataStorePlugin())
Amplify.addPlugin(AWSApiPlugin())
```
as compared to:
```
// Starts in SYNC_VIA_API
Amplify.addPlugin(AWSApiPlugin())
Amplify.addPlugin(AWSDataStorePlugin())
```


 It's kind of a similar situation that we ran into for pluginConfiguration, where we ended up going with a provider-style configuration to delay evaluation of the expression.

jamesonwilliams: Ahhh, yup, absolutely. Great catch.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -109,168 +106,231 @@ public Orchestrator(
...
+        .observeOn(Schedulers.io());
+    }
+
+    private void stopApiSyncBlocking() {
+        Throwable failure = stopApiSync().blockingGet(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        if (failure != null) {
+            LOG.warn("Failed to stop API sync.", failure);
+        }
+    }
+    /**
```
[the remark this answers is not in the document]
jamesonwilliams: Thanks!

jamesonwilliams, line 298:
```suggestion

    /**
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: Resolution: I ended up doing the second thing re: "atomic entity."

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
jamesonwilliams: Resolution: I decoupled the initialization of the API synchronization code, from the code which observes the storage. Now, there is always a storage observer when `beforeOperation()` completes, and so any customer-provided data will be observed. The API sync may start at a later time.

jamesonwilliams:
> I definitely like the paradigm shift here. I had one comment which I noted in the code and I'm also running into one of those nasty RxJava exceptions if I toggle airplane mode on/off
>
> ...
>
> io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with.

Thanks for catching this, @rjuliano. I believe I have resolved it in the latest push. Left [some discussion on your Gist, here](https://gist.github.com/rjuliano/a8808ab98f4bbc7d3a53d15b92bdf8c3#gistcomment-3394006).

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -438,70 +424,50 @@ public void observe(
...
         // We shouldn't call beforeOperation when clearing the DataStore. The
         // only thing we have to wait for is the category initialization latch.
         boolean isCategoryInitialized = false;
         try {
-            isCategoryInitialized = categoryInitializationsPending.await(PLUGIN_INIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            isCategoryInitialized = categoryInitializationsPending.await(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         } catch (InterruptedException exception) {
             LOG.warn("Execution interrupted while waiting for DataStore to be initialized.");
         }
         if (!isCategoryInitialized) {
```
rjuliano, line 439: TLDR version: The DataStore.clear call may have to wait for the orchestrator to start. We might need to do something like this:

```suggestion
        if (!isCategoryInitialized || !orchestrator.isStarted()) {
```

If the clear API is called immediately after calling `Amplify.configure` but before the Orchestrator has a chance to fully start (i.e. it's in LOCAL_ONLY mode and in the process doing stuff before going into SYNC_VIA_API), the call to `DataStore.clear()` does the following:

- Call `orchestrator.stop()` (which is correct); however, the call to `transitionToStopped` only stops the storage observer because the current status of the orchestrator is LOCAL_ONLY. 
- Subscription attempt [is interrupted](https://github.com/aws-amplify/amplify-android/blob/da16e44a92d64f941fda4adb291cf88cd63c8d94/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L118) but the rest of the `startApiSync` continues because we're not re-throwing the interrupted exception.
- Orchestrator status is set to STOPPED.
- DataStore is cleared. (also correct)
- [Original call to `orchestrator.start()`] Somehow, the `startApiSync` call completes and sets the status SYNC_VIA_API
- Calls `orchestrator.start()` but this results in a no-op because the status is already SYNC_VIA_API. The problem is that at this point, the SubscriptionProcessor and MutationProcessor are in an unknown state. 

Using my sample app, the app does not crash but mutations are not being sent and subscriptions are not started.

jamesonwilliams: Thanks for this detailed analysis!

I started looking at additional steps that could be taken to tighten up the serialization guarantees in the Orchestrator. I think we can still do that.

But, in the interest of delivering this current fix, I will go with the `!orchestrator.isStarted()` approach you suggest. It isn't ideal, but as you note, we've solved most of the problems by now and are into the long tail.

rjuliano: Totally agree with this. I think we can continue to iterate on this as needed, but I also feel that with that change, this PR should be considered ready to go.

PR #665, opened 2020-07-23
chore(datastore) Remove all API deserialization logic in DataStore, and rely on API instead

richardmcclellan:
*This PR is a prerequisite for adding pagination support to DataStore.   The API category response deserializer knows how to extract pagination details from an AppSync response.  This PR refactors the AppSyncClient in DataStore to rely on API's response deserializer, instead of having its own implementation.* 

### Summary 
 - Moved `ModelMetadata` and `ModelWithMetadata` from `aws-datastore` to `aws-api-appsync`, and added a `ModelWithMetadataDeserializer` in API.
 - Refactored `AppSyncRequestFactory` (in DataStore) to build the complete GraphQLRequest objects. Instead of just the query document, it also handles the variables too now.
 - Deleted AppSyncResponseDeserializer (in DataStore), since we are now relying on the one in API.


aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelWithMetadataDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+
+/**
+ * Deserializer for ModelWithMetadata.
+ */
+public final class ModelWithMetadataDeserializer implements JsonDeserializer<ModelWithMetadata<Model>> {
```
jamesonwilliams, line 31: Should it be `<? extends Model>`?

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -204,7 +204,11 @@ private AppSyncGraphQLRequestFactory() { }
             }
 
             if (!QueryPredicates.all().equals(predicate)) {
-                String conditionType = "Model" + graphQlTypeName + "ConditionInput";
+                String conditionType = new StringBuilder()
+                    .append("Model")
```
jamesonwilliams, line 208: Not the subclass simple name?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactoryTest.java
```diff
@@ -43,68 +44,80 @@
...
-        assertEquals(
+    public void validateUpdateWithPredicateGeneration() throws DataStoreException, JSONException {
+        String blogOwnerId = "926d7ee8-4ea5-40c0-8e62-3fb80b2a2edd";
+        BlogOwner blogOwner = BlogOwner.builder().name("John Doe").id(blogOwnerId).build();
+        JSONAssert.assertEquals(
             Resources.readAsString("update-blog-owner-with-predicate.txt"),
-            AppSyncRequestFactory.buildUpdateDoc(BlogOwner.class, true)
+            AppSyncRequestFactory.buildUpdateRequest(
+                    blogOwner, 42, BlogOwner.WEA.contains("ther")).getContent(),
+            true
```
jamesonwilliams, line 98: formatting here is weird

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactoryTest.java
```diff
@@ -43,68 +44,80 @@
...
     @Test
-    public void validateDeleteWithPredicateGeneration() throws DataStoreException {
-        assertEquals(
+    public void validateDeleteWithPredicateGeneration() throws DataStoreException, JSONException {
+        JSONAssert.assertEquals(
             Resources.readAsString("delete-person-with-predicate.txt"),
-            AppSyncRequestFactory.buildDeletionDoc(Person.class, true)
+            AppSyncRequestFactory.buildDeletionRequest(
+                    Person.class, "123", 456, Person.AGE.gt(40)).getContent(),
+            true
```
jamesonwilliams, line 113: ditto

aws-datastore/src/test/resources/on-delete-request-for-blog-owner.txt
```diff
@@ -11,3 +12,6 @@ subscription OnDeleteBlogOwner {
     wea
   }
 }
+",
+  "variables": null
+}
```
jamesonwilliams, line 17: Does this need to be explicitly present and null? Or can we figure out how to omit this block from the output, if the value is null?

core/src/main/java/com/amplifyframework/api/graphql/PaginatedResult.java
```diff
@@ -15,15 +15,13 @@
...
-import com.amplifyframework.core.model.Model;
-
 /**
  * Represents a page of results returned from an API.  Specifically, contains the list of items in the page, as well as
  * a GraphQLRequest which can be used to obtain the next page.
  *
- * @param <T> Type of the items in the list.  Must extend Model.
+ * @param <T> Type of the items in the list.
  */
-public abstract class PaginatedResult<T extends Model> {
```
jamesonwilliams, line 26: Hm. Are there any remaining types we want/need to paginate over, that are not models?

`Iterable<String>`, `<String>`, `ModelWithMutation<T>`, `CustomGizmoNotFromCodeGen`?

aws-datastore/src/test/resources/on-delete-request-for-blog-owner.txt
richardmcclellan: I'm not sure - I was trying to maintain behavior of the existing code, in case AppSync does require it explicitly.  We can try to remove it later, after doing some more thorough end to end testing.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
richardmcclellan: graphQLTypeName is actually the same as the simpleName(), it's just built in a more roundabout way:

```
String graphQLTypeName = ModelSchema.fromModelClass(model.getClass()).getName()
```

In `ModelSchema.fromModelClass`, `name` is obtained via `modelClass.getSimpleName()`

core/src/main/java/com/amplifyframework/api/graphql/PaginatedResult.java
richardmcclellan: The reason I changed this from <T extends Model> to <T> is so we can paginate over ModelWithMetadata<M>, since ModelWithMetadata is not a Model itself.

rjuliano: Maybe it can be `PaginatedResult<ModelWithMetadata<T extends Model>>`?

richardmcclellan: > Maybe it can be `PaginatedResult<ModelWithMetadata<T extends Model>>`?

That would work for DataStore, but not for API, because API expects a PaginatedResult<? extends Model>.

rjuliano: Ah...makes sense. I guess the only other alternative I can think of would be to make ModelWithMetadata a Model.

```
public final class ModelWithMetadata<M extends Model> implements Model {
     private final M model;

    @NonNull
    @Override
    public String getId() {
        return model.getId();
    }
}
```

Not sure if it's worth it though unless it would simplify all the other bits of code.

richardmcclellan: Yeah, it's not really a `Model` though.  We could create a "PaginatedResultItem` interface, which is implemented by `Model` and `ModelWithMetadata` though.  Not sure if that's really worth it right now, but I'll think about that for my next PR, which is going to get a little deeper into this code.

richardmcclellan:
> These have been clean. I like the way you broke up the work on this pagination effort.

Yeah, my previous PRs have been too big. Trying to keep them more bite size like this so they can be reviewed easier and faster :)

PR #669, opened 2020-07-24
chore(api): Fix flakey UserAgentTest

richardmcclellan:

### DIFF
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
@@ -47,7 +47,7 @@ import static org.junit.Assert.assertTrue;
 @RunWith(RobolectricTestRunner.class)
 public final class UserAgentTest {
     // This was previously 200ms, but resulted in flaky tests because server.takeRequest would sometimes return null.
-    private static final long REQUEST_TIMEOUT_MS = 5_000;
+    private static final long REQUEST_TIMEOUT_SECONDS = 5;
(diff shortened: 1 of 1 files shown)


aws-api/src/test/java/com/amplifyframework/api/aws/UserAgentTest.java
```diff
@@ -46,7 +46,8 @@
  */
 @RunWith(RobolectricTestRunner.class)
 public final class UserAgentTest {
-    private static final long REQUEST_TIMEOUT_MS = 200;
+    // This was previously 200ms, but resulted in flaky tests because server.takeRequest would sometimes return null.
+    private static final long REQUEST_TIMEOUT_MS = 5_000;
```
jamesonwilliams, line 50: Maybe, `REQUEST_TIMEOUT_SECONDS = 5`, and then use `TimeUnit.SECONDS` down below?

Is it enough to just go to 1 full second? `5_000` seems like a lot for just stuff that's running on an application processor, without any disk or network involved.

richardmcclellan: > Maybe, `REQUEST_TIMEOUT_SECONDS = 5`, and then use `TimeUnit.SECONDS` down below?

:+1:

> Is it enough to just go to 1 full second? `5_000` seems like a lot for just stuff that's running on an application processor, without any disk or network involved.

It probably is, but I would have thought that about 200ms as well. It's difficult to test what the best value is, so I picked something big to see if this fix even works, without spending too much time tweaking the value.

PR #670, opened 2020-07-24
chore(datastore) Represent lastChangedAt as Temporal.Timestamp instead of long

richardmcclellan:
`lastChangedAt`, part of `ModelMetadata` is represented as an `AWSTimestamp` by AppSync, so we should represent it as such on the client with `Temporal.Timestamp`, instead of `Long`.


aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
```diff
@@ -37,7 +38,7 @@
     private final @ModelField(targetType = "ID", isRequired = true) String id;
     private final @ModelField(targetType = "Boolean") Boolean _deleted;
     private final @ModelField(targetType = "Int") Integer _version;
-    private final @ModelField(targetType = "AWSTimestamp") Long _lastChangedAt;
```
jamesonwilliams, line 40: Will this require the SQL table to be reprovisioned?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
```diff
@@ -101,7 +101,7 @@ public static CreateConfigurator onCreate(AppSync mock) {
 
                 // Pass back a ModelWithMetadata. Model is the one provided.
                 ModelMetadata metadata =
-                    new ModelMetadata(capturedModel.getId(), false, 1, Time.now());
+                    new ModelMetadata(capturedModel.getId(), false, 1, new Temporal.Timestamp());
```
jamesonwilliams, line 104: Why not your cool `Temporal.now()`, in all places, instead?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncResponseDeserializerTest.java
```diff
@@ -61,7 +61,7 @@ public void awsDateTypesCanBeDeserialized() {
                 .build();
         Boolean deleted = false;
         Integer version = 42;
-        Long lastChangedAt = Long.valueOf(1594858827);
+        Temporal.Timestamp lastChangedAt = new Temporal.Timestamp(1594858827, TimeUnit.SECONDS);
```
jamesonwilliams, line 64: You could make a `Temporal.timestamp(long, TimeUnit)` util, too, that might be cool.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
richardmcclellan: Oops, missed this one!  Will change.

aws-api-appsync/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java
```diff
@@ -434,6 +434,14 @@ public Timestamp(@NonNull java.util.Date date) {
             this(date.getTime(), TimeUnit.MILLISECONDS);
         }
 
+        /**
+         * Returns a new Timestamp instance that represents the current system time.
+         * @return a new Timestamp instance that represents the current system time.
+         */
+        public static Timestamp now() {
+            return new Timestamp();
+        }
```
raphkim, line 443: nice 👍

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
```diff
@@ -37,7 +38,7 @@
...
+    private final @ModelField(targetType = "AWSTimestamp") Temporal.Timestamp _lastChangedAt;
```
raphkim, line 41: Just making sure: this change is scoped only to model metadata right? Codegen won't be affected?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
raphkim: ^If so, will this be considered a breaking change?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
richardmcclellan: Correct, no impact on codegen.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncResponseDeserializerTest.java
richardmcclellan: Agreed, that would be nice.  Maybe in a subsequent PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
richardmcclellan: No, it will not require the SQL table to be reprovisioned, since it will still be stored as a `Long` either way.  That said, thanks for asking this question - I realized that the SQLiteModelFieldTypeConverter doesn't support `AWSTimestamp` yet, so I will add that.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/TestModelWithMetadataInstances.java
```diff
@@ -53,12 +61,18 @@
...
+                    new Temporal.Timestamp(123123L, TimeUnit.SECONDS))
         );
     public static final ModelWithMetadata<Post> DELETED_DRUM_POST =
         new ModelWithMetadata<>(
             DRUM_POST.getModel(),
-            new ModelMetadata("83ceb757-c8c8-4b6a-bee0-a43afb53a73a", Boolean.TRUE, 5, 123123L)
+            new ModelMetadata("83ceb757-c8c8-4b6a-bee0-a43afb53a73a",
+                    Boolean.TRUE,
+                    5,
+                    new Temporal.Timestamp(123123L, TimeUnit.SECONDS))
```
jamesonwilliams, line 75: We could potentially create `Temporal.Timestamp.seconds(123_123L)` and `Temporal.Timestamp.milliseconds(123_123_000L)` utility factories?

jamesonwilliams:
Blargh, still one more:
> [ant:checkstyle] [ERROR] /home/circleci/amplify-android-home/aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java:33:8: Unused import - com.amplifyframework.util.Time. [UnusedImports]

PR #673, opened 2020-07-27
fix(amplify-tools) added a null check for RunManagerNode by parsing o…

### LINKED ISSUE #672: Importing of a project into Android Studio fails if Amplify Gradle plugin is applied
When importing a Gradle project with Amplify plugin into the Android Studio **with no run configurations** defined, the build fails with the error message:

> Failed to apply plugin [id 'com.amplifyframework.amplifytools']
   > Cannot invoke method children() on null object

Reproducable on Android Studio 4.0.1, Android Studio 3.6.3

ivpdev:
…f workspace.xml

Amplify Gradle plugin tries to parse the workspace file of Android Studio.
If in the Android Studio no run configurations are defined, there is
no RunManagerNode in the workspace.xml which causes a crash of the plugin.

Fixes #672


raphkim:
LGTM. Thank you for your contribution @ivpdev !

PR #675, opened 2020-07-27
Store logs to be retrieved by developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Store the logs generated by AndroidLogger.java so they can be retrieved by the developer menu. Also moves all classes related to the developer menu into a new devmenu package.

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -29,6 +29,7 @@ import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.plugin.Plugin;
 import com.amplifyframework.datastore.DataStoreCategory;
 import com.amplifyframework.hub.HubCategory;
+import com.amplifyframework.logging.AndroidLoggingPlugin;
 import com.amplifyframework.logging.LoggingCategory;
 import com.amplifyframework.predictions.PredictionsCategory;
 import com.amplifyframework.storage.StorageCategory;
@@ -127,6 +128,11 @@ public final class Amplify {
                 );
             }
 
+            boolean debuggableApp = isDebuggableApplication(context);
+            if (debuggableApp && Logging.getPlugins().isEmpty()) {
+                addPlugin(new AndroidLoggingPlugin());
+            }
+
             for (Category<? extends Plugin<?>> category : CATEGORIES.values()) {
                 if (category.getPlugins().size() > 0) {
                     CategoryConfiguration categoryConfiguration =
@@ -135,7 +141,8 @@ public final class Amplify {
                     beginInitialization(category, context);
                 }
             }
-            if (isDebuggableApplication(context)) {
+
+            if (debuggableApp) {
                 DeveloperMenuManager.sharedInstance(context).startListening();
             }
 
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -27,6 +27,7 @@ import com.amplifyframework.core.category.CategoryConfiguration;
 import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.plugin.Plugin;
 import com.amplifyframework.datastore.DataStoreCategory;
+import com.amplifyframework.devmenu.DeveloperMenu;
 import com.amplifyframework.hub.HubCategory;
 import com.amplifyframework.logging.LoggingCategory;
 import com.amplifyframework.predictions.PredictionsCategory;
@@ -126,6 +127,8 @@ public final class Amplify {
                 );
             }
 
+            DeveloperMenu.singletonInstance(context).enableDeveloperMenu();
+
             for (Category<? extends Plugin<?>> category : CATEGORIES.values()) {
                 if (category.getPlugins().size() > 0) {
                     CategoryConfiguration categoryConfiguration =
@@ -134,8 +137,6 @@ public final class Amplify {
                     beginInitialization(category, context);
                 }
             }
-
-            DeveloperMenuManager.sharedInstance(context).enableDeveloperMenu();
 
             CONFIGURATION_LOCK.set(true);
         }
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -16,7 +16,6 @@
 package com.amplifyframework.core;
 
 import android.content.Context;
-import android.content.pm.ApplicationInfo;
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.AmplifyException;
@@ -29,7 +28,6 @@ import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.plugin.Plugin;
 import com.amplifyframework.datastore.DataStoreCategory;
 import com.amplifyframework.hub.HubCategory;
-import com.amplifyframework.logging.AndroidLoggingPlugin;
 import com.amplifyframework.logging.LoggingCategory;
 import com.amplifyframework.predictions.PredictionsCategory;
 import com.amplifyframework.storage.StorageCategory;
@@ -128,11 +126,6 @@ public final class Amplify {
                 );
             }
 
-            boolean debuggableApp = isDebuggableApplication(context);
-            if (debuggableApp && Logging.getPlugins().isEmpty()) {
-                addPlugin(new AndroidLoggingPlugin());
-            }
-
             for (Category<? extends Plugin<?>> category : CATEGORIES.values()) {
                 if (category.getPlugins().size() > 0) {
                     CategoryConfiguration categoryConfiguration =
@@ -142,9 +135,7 @@ public final class Amplify {
                 }
             }
 
-            if (debuggableApp) {
-                DeveloperMenuManager.sharedInstance(context).startListening();
-            }
+            DeveloperMenuManager.sharedInstance(context).enableDeveloperMenu();
 
             CONFIGURATION_LOCK.set(true);
         }
@@ -214,15 +205,6 @@ public final class Amplify {
     private enum RegistryUpdateType {
         ADD,
         REMOVE
-    }
-
-    /**
-     * Returns true if the application is in a debuggable build, false otherwise.
-     * @param context an Android Context
-     * @return a boolean indicating whether the application is in a debuggable build
-     */
-    private static boolean isDebuggableApplication(Context context) {
-        return (context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;
     }
 
 }
--- a/core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
+++ b/core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
@@ -17,6 +17,7 @@ package com.amplifyframework.core;
 
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.ApplicationInfo;
 
 /**
  * Manage the display and shake detection behavior for the
@@ -50,6 +51,16 @@ public final class DeveloperMenuManager implements ShakeDetector.Listener {
             sInstance = new DeveloperMenuManager(context);
         }
         return sInstance;
+    }
+
+    /**
+     * Allows the developer menu to be activated if the app is in a debuggable build.
+     */
+    public void enableDeveloperMenu() {
+        if ((context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0) {
+            Amplify.Logging.shouldStoreLogs(true);
+            startListening();
+        }
     }
 
     /**
--- a/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
@@ -18,6 +18,7 @@ package com.amplifyframework.devmenu;
 import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.logging.LogLevel;
 
@@ -103,19 +104,26 @@ public final class LogEntry implements Comparable<LogEntry> {
 
     @Override
     public boolean equals(Object object) {
-        if (!(object instanceof LogEntry)) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
             return false;
         }
-        LogEntry otherEntry = (LogEntry) object;
-        return dateTime.equals(otherEntry.getDateTime()) && logLevel.equals(otherEntry.getLogLevel())
-                && namespace.equals(otherEntry.getNamespace()) && message.equals(otherEntry.getMessage())
-                && throwable.equals(otherEntry.getThrowable());
+        LogEntry logEntry = (LogEntry) object;
+        return dateTime.equals(logEntry.getDateTime()) && ObjectsCompat.equals(namespace, logEntry.getNamespace())
+                && ObjectsCompat.equals(message, logEntry.getMessage()) && logLevel == logEntry.getLogLevel()
+                && ObjectsCompat.equals(throwable, logEntry.getThrowable());
     }
 
     @Override
     public int hashCode() {
-        return dateTime.hashCode() + logLevel.hashCode() + namespace.hashCode() + message.hashCode()
-                + throwable.hashCode();
+        int result = getDateTime().hashCode();
+        result = 31 * result + (getNamespace() != null ? getNamespace().hashCode() : 0);
+        result = 31 * result + (getMessage() != null ? getMessage().hashCode() : 0);
+        result = 31 * result + (getThrowable() != null ? getThrowable().hashCode() : 0);
+        result = 31 * result + getLogLevel().hashCode();
+        return result;
     }
 
     /**
--- a/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
@@ -23,12 +23,13 @@ import com.amplifyframework.logging.LogLevel;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Locale;
 import java.util.Objects;
 
 /**
  * A representation of a log.
  */
-public final class LogEntry {
+public final class LogEntry implements Comparable<LogEntry> {
     // The format for the log's date and time.
     private static final String DATE_TIME_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
     private final LocalDateTime dateTime;
@@ -95,13 +96,36 @@ public final class LogEntry {
         return logLevel;
     }
 
+    @Override
+    public int compareTo(LogEntry logEntry) {
+        return getDateTime().compareTo(logEntry.getDateTime());
+    }
+
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof LogEntry)) {
+            return false;
+        }
+        LogEntry otherEntry = (LogEntry) object;
+        return dateTime.equals(otherEntry.getDateTime()) && logLevel.equals(otherEntry.getLogLevel())
+                && namespace.equals(otherEntry.getNamespace()) && message.equals(otherEntry.getMessage())
+                && throwable.equals(otherEntry.getThrowable());
+    }
+
+    @Override
+    public int hashCode() {
+        return dateTime.hashCode() + logLevel.hashCode() + namespace.hashCode() + message.hashCode()
+                + throwable.hashCode();
+    }
+
     /**
      * Returns a String representation of this log.
      * @return a String representing this log.
      */
     public String toString() {
-        return String.format("[%s] %s %s: %s \n%s", logLevel.name(),
-                dateTime.format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT)), namespace, message,
-                throwable == null ? "" : Log.getStackTraceString(throwable));
+        String dateString = dateTime.format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT));
+        String exceptionTrace = throwable == null ? "" : Log.getStackTraceString(throwable);
+        return String.format(Locale.US, "[%s] %s %s: %s \n%s",
+                logLevel.name(), dateString, namespace, message, exceptionTrace);
     }
 }
--- a/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
@@ -13,11 +13,13 @@
  * permissions and limitations under the License.
  */
 
-package com.amplifyframework.logging;
+package com.amplifyframework.devmenu;
 
 import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
+import com.amplifyframework.logging.LogLevel;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
@@ -34,19 +36,6 @@ public final class LogEntry {
     private final String message;
     private final Throwable throwable;
     private final LogLevel logLevel;
-
-    /**
-     * Creates a new LogEntry representing a log with the given time, tag,
-     * and message that was logged at the given level.
-     * @param dateTime the date and time of the log.
-     * @param namespace the namespace of the logger that emitted the log.
-     * @param message the message for the log.
-     * @param logLevel the level the log was logged at.
-     */
-    public LogEntry(@NonNull LocalDateTime dateTime, @Nullable String namespace, @Nullable String message,
-                    @NonNull LogLevel logLevel) {
-        this(dateTime, namespace, message, null, logLevel);
-    }
 
     /**
      * Creates a new LogEntry representing a log with the given time, tag, message,
--- a/core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
@@ -22,6 +22,7 @@ import androidx.annotation.Nullable;
 import com.amplifyframework.logging.Logger;
 import com.amplifyframework.logging.LoggingCategoryBehavior;
 import com.amplifyframework.logging.LoggingPlugin;
+import com.amplifyframework.util.Immutable;
 
 import org.json.JSONObject;
 
@@ -87,6 +88,6 @@ public final class PersistentLogStoragePlugin extends LoggingPlugin<Void> {
             logs.addAll(logger.getLogs());
         }
         Collections.sort(logs);
-        return logs;
+        return Immutable.of(logs);
     }
 }
--- a/core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
@@ -67,9 +67,7 @@ public final class PersistentLogStoragePlugin extends LoggingPlugin<Void> {
     }
 
     @Override
-    public void configure(
-            JSONObject pluginConfiguration,
-            @NonNull Context context) {
+    public void configure(JSONObject pluginConfiguration, @NonNull Context context) {
     }
 
     @Nullable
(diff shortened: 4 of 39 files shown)


core/src/main/java/com/amplifyframework/logging/LogEntry.java
```diff
@@ -0,0 +1,78 @@
...
+    // The date and time of the log.
+    private LocalDateTime dateTime;
+    // The namespace of the logger that emitted the log.
+    private String namespace;
+    // The message for the log.
+    private String message;
+    // The Throwable (if any) associated with the log.
+    private Throwable throwable;
+    // The level the log was logged at.
+    private LogLevel logLevel;
```
raphkim, line 38: these can be `final`

core/src/main/java/com/amplifyframework/logging/LogEntry.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Creates a new LogEntry representing a log with the given time, tag, message,
+     * and throwable that was logged at the given level.
+     * @param dateTime the date and time of the log.
+     * @param namespace the namespace of the logger that emitted the log.
+     * @param message the message for the log.
+     * @param throwable the Throwable associated with the log.
+     * @param logLevel the level the log was logged at.
+     */
+    public LogEntry(LocalDateTime dateTime, String namespace, String message, Throwable throwable, LogLevel logLevel) {
```
raphkim, line 61: Since this is a public constructor, it would be helpful to specify `@NonNull` vs `@Nullable` for each of them and enforce it with `Objects.requireNonNull()`. Otherwise, `toString()` will throw NPE for either `logLevel` or `dateTime` being null.

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -131,6 +137,39 @@ public void noContentLoggedAtThresholdNone() {
...
+        Context mockContext = mock(Context.class);
+        ApplicationInfo mockAppInfo = mock(ApplicationInfo.class);
+        mockAppInfo.flags = ApplicationInfo.FLAG_DEBUGGABLE;
+        when(mockContext.getApplicationInfo()).thenReturn(mockAppInfo);
+        AndroidLoggingPlugin plugin = new AndroidLoggingPlugin();
+        plugin.configure(new JSONObject(), mockContext);
+        Logger logger = plugin.forNamespace("logging-test");
+        logger.info("Info log");
+        assertNotNull(plugin.getLogs());
+        assertEquals(plugin.getLogs().size(), 1);
```
raphkim, line 170: Nit: It would be a stronger test to assert that the content of the log matches what you expect after the size assertion

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -131,6 +137,39 @@ public void noContentLoggedAtThresholdNone() {
...
     }
 
+    /**
+     * When the application is not in a debuggable build, logs are not stored.
+     */
+    @Test
+    public void noLogsStored() {
+        Context mockContext = mock(Context.class);
+        ApplicationInfo mockAppInfo = mock(ApplicationInfo.class);
+        mockAppInfo.flags = 0;
```
raphkim, line 147: I think it defaults to 0 :)

core/src/main/java/com/amplifyframework/logging/LogEntry.java
```diff
@@ -0,0 +1,78 @@
...
+import android.util.Log;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A representation of a log.
+ */
+public final class LogEntry {
+    // The format for the log's date and time.
```
raphkim, line 27: Should these variables be readable? Maybe you should create getters for each of these, and move the comments to an actual javadoc for each method

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -127,6 +128,11 @@ public static void configure(@NonNull final AmplifyConfiguration configuration,
                 );
             }
 
+            boolean debuggableApp = isDebuggableApplication(context);
+            if (debuggableApp && Logging.getPlugins().isEmpty()) {
+                addPlugin(new AndroidLoggingPlugin());
+            }
+
```
jamesonwilliams, line 135: `Amplify` is a "façade." A downside to the façade as a pattern is that it can become a god object coupled to all classes of an app. [ref](https://refactoring.guru/design-patterns/facade).

We want to "fan out" quickly. Strive for a ~one-line integration of your Dev Menu code.

Would it be possible to have your entire integration be a single line?
```java
DevMenu.attach(this);
```

All of your logic, like whether or not its a debug build, whether or not you need a logging plugin, etc. etc. would then be in your `DevMenu`, not in `Amplify`.

core/src/main/java/com/amplifyframework/logging/AndroidLoggingPlugin.java
```diff
@@ -66,6 +77,7 @@ public String getPluginKey() {
     public void configure(
             JSONObject pluginConfiguration,
             @NonNull Context context) {
+        storeLogs = (context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;
```
jamesonwilliams, line 80: Can you centralize this check into some helper class? I think you perform this check in a couple places now.
```java
final class IsDebuggable {
    private IsDebuggable() {}

    boolean check(Context context) {
        return context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE != 0;
    }
}
```
(You can get creative with the naming, this is just one idea.)

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -131,6 +137,39 @@ public void noContentLoggedAtThresholdNone() {
         assertTrue(systemLog.getLines().isEmpty());
     }
 
+    /**
+     * When the application is not in a debuggable build, logs are not stored.
+     */
+    @Test
+    public void noLogsStored() {
+        Context mockContext = mock(Context.class);
+        ApplicationInfo mockAppInfo = mock(ApplicationInfo.class);
```
jamesonwilliams, line 146: Do you need a mock for this one, or can you just create a real instance?

You probably do need a mock for the `context`.

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -131,6 +137,39 @@ public void noContentLoggedAtThresholdNone() {
...
+        mockAppInfo.flags = 0;
+        when(mockContext.getApplicationInfo()).thenReturn(mockAppInfo);
+        AndroidLoggingPlugin plugin = new AndroidLoggingPlugin();
+        plugin.configure(new JSONObject(), mockContext);
+        Logger logger = plugin.forNamespace("logging-test");
+        logger.info("Info log");
+        assertNull(plugin.getLogs());
```
jamesonwilliams, line 153: Better to return an empty collection than null 
```
assertEquals(Collections.emptyList(), plugin.getLogs())
```

https://www.thefinestartist.com/effective-java/43

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
jamesonwilliams, line 170: These are backwards --

```java
// It should be ...
assertEquals(expectedValue, actualValue);
```

core/src/test/java/com/amplifyframework/logging/AndroidLoggingPluginTest.java
```diff
@@ -131,6 +137,39 @@ public void noContentLoggedAtThresholdNone() {
...
+
+    /**
+     * When the application is in a debuggable build, logs are stored.
+     */
+    @Test
+    public void logsAreStored() {
+        Context mockContext = mock(Context.class);
+        ApplicationInfo mockAppInfo = mock(ApplicationInfo.class);
+        mockAppInfo.flags = ApplicationInfo.FLAG_DEBUGGABLE;
+        when(mockContext.getApplicationInfo()).thenReturn(mockAppInfo);
```
jamesonwilliams, line 164: Elsewhere, I suggest making a utility to tell whether or not the app is debuggable. When I wrote it, I had it as a static utility class. However, it might simplify your test code if it is just package-local, open-for-extension, regular class:
```java
class AmplifyApp {
    private final WeakReference<Context> context;

    AmplifyApp(WeakReference<Context> context) {
        this.context = context;
    }

    boolean isDebuggable() {
        Context currentContext = context.get();
        return currentContext.blahblah();
    }
}
```
Or its test form:
```java
AmplifyApp amplifyApp = mock(AmplifyApp.class);
when(amplifyApp.isDebuggable()).thenReturn(true);
```

eeatonaws: After changing the code to add a BroadcastLogger and manage the loggers inside of LoggingCategory, I now only check if the app is in a debuggable build inside DeveloperMenuManager.enableDeveloperMenu, so I did not add a utility class for this.

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -142,9 +135,7 @@ public static void configure(@NonNull final AmplifyConfiguration configuration,
                 }
             }
 
-            if (debuggableApp) {
-                DeveloperMenuManager.sharedInstance(context).startListening();
-            }
+            DeveloperMenuManager.sharedInstance(context).enableDeveloperMenu();
```
jamesonwilliams, line 138: This is way cleaner, thanks!

I'd ditch the `Manager` part, you could just call it a `DeveloperMenu`. (Ref: https://blog.codinghorror.com/i-shall-call-it-somethingmanager/)

The `sharedInstance` verbiage is idiomatic to the iOS platform, but not to Android. On Android, this would probably be expressed as just `instance` (if you can have multiple) or `singleton`, or `singletonInstance`, if the code internally prevents multiple instances from being created.

core/src/main/java/com/amplifyframework/core/DeveloperMenuManager.java
```diff
@@ -52,6 +53,16 @@ public static DeveloperMenuManager sharedInstance(Context context) {
         return sInstance;
     }
 
+    /**
+     * Allows the developer menu to be activated if the app is in a debuggable build.
+     */
+    public void enableDeveloperMenu() {
+        if ((context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0) {
+            Amplify.Logging.shouldStoreLogs(true);
```
jamesonwilliams, line 61: I think you still want to strive to have the category not know about log storage. This is leaking your dev menu logic into the category API.

Instead, I would aim to express this as:
```java
Amplify.addPlugin(new PersistentLogStoragePlugin());
```

core/src/main/java/com/amplifyframework/logging/AndroidLoggingPlugin.java
```diff
@@ -55,16 +49,14 @@ public AndroidLoggingPlugin() {
...
 
     @NonNull
     @Override
     public Logger forNamespace(@Nullable String namespace) {
         String usedNamespace = namespace == null ? AMPLIFY_NAMESPACE : namespace;
-        AndroidLogger newLogger = new AndroidLogger(usedNamespace, defaultLoggerThreshold, storeLogs);
-        loggers.add(newLogger);
-        return newLogger;
+        AndroidLogger androidLogger = new AndroidLogger(usedNamespace, defaultLoggerThreshold);
+        return Amplify.Logging.createBroadcastLogger(androidLogger);
```
jamesonwilliams, line 59: Create the broadcast logger inside of `LoggingCategory`:

```java
@NonNull
@Override
public Logger forNamespace(@NonNull String namespace) {
    List<Logger> delegates = new ArrayList<>();
    for (LoggingPlugin plugin : pluginsAddedViaApi) {
        delegates.add(plugin.forNamespace(namespace));
    }
    return new BroadcastLogger(delegates);
}
```

with:
```java
public final class BroadcastLogger {
    private final List<Logger> delegates;
   
    public BroadcastLogger(@Nullable List<Logger> delegates) {
        this.delegates = new ArrayList<>();
        if (delegates != null) {
            this.delegates.addAll(delegates);
        }
    }

    void info(String message) {
        log(LogLevel.INFO, message);
    }

    ...

    void log(LogLevel level, String message) {
        for (Logger delegate : delegates) {
            delegate.log(level, message);
        }
    }
}
```

core/src/main/java/com/amplifyframework/logging/BroadcastLogger.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * An implementation of {@link Logger} that can store logs and
+ * uses a provided logger to emit logs.
+ */
+final class BroadcastLogger implements Logger {
+    // Maximum number of logs to store.
+    private static final int MAX_NUM_LOGS = 500;
```
jamesonwilliams, line 35: This is logic that you'll put in your `PersistentLogger`.

The `BroadcastLogger` just farms work out to an `AndroidLogger` (that emits to logcat) and a `PersistentLogger`, which buffers content for display in the dev menu.

core/src/main/java/com/amplifyframework/logging/LogEntry.java
```diff
@@ -16,26 +16,24 @@
...
+import androidx.annotation.Nullable;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Objects;
 
 /**
  * A representation of a log.
  */
 public final class LogEntry {
```
jamesonwilliams, line 29: You could bundle together all of your code for the dev menu into `core/src/main/java/com/amplifyframework/devmenu`. This `LogEntry` is more a thing that's used by the dev menu, I'd say, than it is part of the core spec of the logger.

`core/src/main/java/com/amplifyframework/logging` is meant to be a core spec for the logger category.

It does also include the default Android implementations of the plugin and logger. Which isn't 100% architecturally pure, true. In other categories, those plugins would be separate Gradle modules entirely.

core/src/main/java/com/amplifyframework/logging/LoggingCategory.java
```diff
@@ -30,6 +34,10 @@
  */
 public final class LoggingCategory extends Category<LoggingPlugin<?>> implements LoggingCategoryBehavior {
     private final LoggingPlugin<?> defaultPlugin;
+    // List of BroadcastLoggers created.
+    private final List<BroadcastLogger> loggers;
+    // Indicates whether the logs from all loggers should be stored.
+    private boolean storeLogs;
```
jamesonwilliams, line 40: If you have a Persistent logging plugin included, then this should be implicitly true, based on the presence of the plugin. In other words, you wouldn't need this state boolean.

core/src/main/java/com/amplifyframework/logging/LogEntry.java
eeatonaws: Would PersistentLogStoragePlugin and PersistentLogger go in the devmenu package or logging package?

core/src/main/java/com/amplifyframework/logging/BroadcastLogger.java
```diff
@@ -0,0 +1,110 @@
...
+        this.delegates = new ArrayList<>();
+        if (delegates != null) {
+            this.delegates.addAll(delegates);
+        }
+    }
+
+    @NonNull
+    @Override
+    public LogLevel getThresholdLevel() {
+        throw new UnsupportedOperationException("Cannot get threshold level for BroadcastLogger.");
```
jamesonwilliams, line 46: Does it make sense to align the behavior to `getNamespace()`? Maybe,
```java
if (delegates.isEmpty()) {
    throw ...
} else {
    return delegates.get(0).getThresholdLevel();
}
```

An alternative is that you enforce all of the delegates  in the `BroadcastLogger` to have the same namespace and threshold level, and then you can just return them as normal.

core/src/main/java/com/amplifyframework/logging/LoggingCategory.java
```diff
@@ -53,15 +56,23 @@ public CategoryType getCategoryType() {
     @NonNull
     @Override
     public Logger forNamespace(@Nullable String namespace) {
-        return getLoggingPlugin().forNamespace(namespace);
+        List<Logger> delegates = new ArrayList<>();
+        boolean defaultPluginAdded = false;
```
jamesonwilliams, line 60: If you use a `Set`, you might be able to avoid the need to record this state boolean. You could add a plugin twice, and the Set should be able to ignore the duplicate, on the basis of its `hashCode()` value.

core/src/main/java/com/amplifyframework/logging/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public PersistentLogStoragePlugin() {
+        loggers = new ArrayList<>();
+    }
+
+    @NonNull
+    @Override
+    public Logger forNamespace(@Nullable String namespace) {
+        String usedNamespace = namespace == null ? AMPLIFY_NAMESPACE : namespace;
+        PersistentLogger logger = new PersistentLogger(usedNamespace);
```
jamesonwilliams, line 48: Do you want to use a `Map<String, PersistentLogger>`, that way you can check:
```java
PersistentLogger preExistingLogger = map.get(usedNamespace);
if (preExistingLogger != null) {
    return preExistingLogger;
} else {
   PersistentLogger newLogger = new PeristentLogger(usedNamespace);
   map.put(usedNamespace, newLogger);
   return newLogger;
}
```

core/src/main/java/com/amplifyframework/logging/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,82 @@
...
+        loggers.add(logger);
+        return logger;
+    }
+
+    @NonNull
+    @Override
+    public String getPluginKey() {
+        return "PersistentLogStoragePlugin";
```
jamesonwilliams, line 56:
```suggestion
        return PersistentLogStoragePlugin.class.getSimpleName();
```

core/src/main/java/com/amplifyframework/logging/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Returns the logs stored by all of the {@link PersistentLogger}s.
+     * @return a list of LogEntry.
+     */
+    public List<LogEntry> getLogs() {
+        List<LogEntry> logs = new ArrayList<>();
+        for (PersistentLogger logger : loggers) {
+            logs.addAll(logger.getLogs());
```
jamesonwilliams, line 78: Hm. I think you'll need to sort these by timestamp somehow, right? Right now, they are grouped into chunks by namespace.

You can add like:
```java
Collections.sort(logs, (left, right) ->
    left.getTimestamp().compareTo(right.getTimestamp()));
```
before you return.

core/src/main/java/com/amplifyframework/logging/PersistentLogger.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.logging;
```
jamesonwilliams, line 16: I'd put this one with the `.devmenu` stuff.

core/src/main/java/com/amplifyframework/logging/LoggingCategory.java
eeatonaws: Using a Set does not work as Loggers of the same type and namespace as well as LoggingPlugins of the same type have different hash codes (these classes don't override the hashCode method and I think the default one only guarantees that the hash codes for two objects will be equal if they represent the same instance).

core/src/main/java/com/amplifyframework/logging/LogEntry.java
jamesonwilliams: /md One could argue either way. I'd say _devmenu_, since its the only consumer we currently expect for the plugin.

If we find another use for it, we can always move it back to the more general location, later.

core/src/main/java/com/amplifyframework/logging/LoggingCategory.java
jamesonwilliams: Ah yes true. But, _you could_ add the `equals` and `hashCode` implementations! I think they didn't exist before, since there was only ever one of each, and they didn't need to be "collection safe."

core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
```diff
@@ -0,0 +1,107 @@
...
+import com.amplifyframework.logging.LogLevel;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+
+/**
+ * A representation of a log.
+ */
+public final class LogEntry {
```
jamesonwilliams, line 31: Ah crap, actually. In my last round of feedback, I suggested you do:
```java
Collection.sort(listOfLogEntry, (left, right) -> { /* blah balh */ });
```
With the comparator specified at the call site.

However, cleaner and better encapsulated would be to have `public final class LogEntry implements Comparable<LogEntry> {`. Then, you would only need to do `Collections.sort(listOfLogEntry)`, and it would Just Work™.

core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Returns a String representation of this log.
+     * @return a String representing this log.
+     */
+    public String toString() {
+        return String.format("[%s] %s %s: %s \n%s", logLevel.name(),
+                dateTime.format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT)), namespace, message,
+                throwable == null ? "" : Log.getStackTraceString(throwable));
```
jamesonwilliams, line 105:
```suggestion
        String dateString = dateTime.format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT));
        String exceptionTrace = throwable == null ? "" : Log.getStackTraceString(throwable);
        return String.format(Locale.US,
            "[%s] %s %s: %s \n%s",
            logLevel.name(), dateString, namespace, message, exceptionTrace
        );
```

core/src/main/java/com/amplifyframework/logging/LoggingCategory.java
eeatonaws: Are you suggesting adding equals and hashCode implementations for Logger or LoggingPlugin?

core/src/test/java/com/amplifyframework/devmenu/PersistentLogStoragePluginTest.java
```diff
@@ -0,0 +1,54 @@
...
+        String message = "Error log";
+        Throwable throwable = new Throwable("error message");
+        logger.error(message, throwable);
+        List<LogEntry> logs = plugin.getLogs();
+        assertEquals(1, logs.size());
+        LogEntry log = logs.get(0);
+        assertEquals(logger.getNamespace(), log.getNamespace());
+        assertEquals(message, log.getMessage());
+        Assert.assertEquals(LogLevel.ERROR, log.getLogLevel());
+        assertEquals(throwable, log.getThrowable());
```
jamesonwilliams, line 52: Another strategy here, which I prefer, is to compare the top-level types with a single `assertEquals` statement. That is:
```java
assertEquals(
    Collections.singletonList(new LogLine(
        yourExcpected,
        valuesHere
    )),
    plugin.getLogs()
)
```

core/src/main/java/com/amplifyframework/logging/BroadcastLogger.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Creates a new BroadcastLogger.
+     * @param delegates the list of loggers to emit logs to
+     */
+    public BroadcastLogger(@Nullable List<Logger> delegates) {
+        this.delegates = new ArrayList<>();
+        if (delegates != null) {
+            this.delegates.addAll(delegates);
+        }
```
jamesonwilliams, line 40: Hm. Actually, you could refuse to create a `BroadcastLogger` if `delegates.size() < 1`. That'd remove the need for your `if (delegates.isEmpty())` checks, anyway.

core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,94 @@
...
+    @NonNull
+    @Override
+    public String getPluginKey() {
+        return PersistentLogStoragePlugin.class.getSimpleName();
+    }
+
+    @Override
+    public void configure(
+            JSONObject pluginConfiguration,
+            @NonNull Context context) {
```
jamesonwilliams, line 72: trivial nit: would join this up onto a single line

core/src/main/java/com/amplifyframework/logging/BroadcastLogger.java
eeatonaws: Good point. Also, delegates.size() should never be less than 1 since a logger from the defaultPlugin is always added.

core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof LogEntry)) {
+            return false;
+        }
+        LogEntry otherEntry = (LogEntry) object;
+        return dateTime.equals(otherEntry.getDateTime()) && logLevel.equals(otherEntry.getLogLevel())
+                && namespace.equals(otherEntry.getNamespace()) && message.equals(otherEntry.getMessage())
+                && throwable.equals(otherEntry.getThrowable());
```
jamesonwilliams, line 112: Since some of these can be `null`, you risk throwing `NullPointerException` from this code. The `ObjectsCompat.equals(foo, bar)` performs a `null`-safe check, and will return `true` when both args are `null`.

core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return dateTime.hashCode() + logLevel.hashCode() + namespace.hashCode() + message.hashCode()
+                + throwable.hashCode();
```
jamesonwilliams, line 118: For both `equals()` and `hashCode()`, you can have the IDE generate them for you. There's a standard recipe for `hashCode()` in Java which involves offsetting individual component hash codes with a prime.

In Android Studio, right click in the main document of the `LogEntry` class, and click "Generate..." then select "equals() and hashCode()." For "Template," I use "IntelliJ Default." The next page will ask you which values can be null or non-null, and you can follow along with what you put in your constructor params.

core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,92 @@
...
+     * in order from oldest to newest in terms of timestamp.
+     * @return a sorted list of LogEntry.
+     */
+    public List<LogEntry> getLogs() {
+        List<LogEntry> logs = new ArrayList<>();
+        for (PersistentLogger logger : loggers.values()) {
+            logs.addAll(logger.getLogs());
+        }
+        Collections.sort(logs);
+        return logs;
```
jamesonwilliams, line 90: `return Immutable.of(logs)` ?

core/src/main/java/com/amplifyframework/devmenu/PersistentLogger.java
```diff
@@ -0,0 +1,118 @@
...
+    // Maximum number of logs to store.
+    private static final int MAX_NUM_LOGS = 500;
+    // Namespace for this logger.
+    private final String namespace;
+    // The logs stored by this logger.
+    private final List<LogEntry> logs;
+
+    PersistentLogger(@NonNull String namespace) {
+        this.namespace = Objects.requireNonNull(namespace);
+        logs = new LinkedList<>();
```
jamesonwilliams, line 44:
```suggestion
        this.logs = new LinkedList<>();
```

Issue #679, opened 2020-07-28
RFC: Align Android Storage APIs to iOS


jamesonwilliams:
- [Current Gaps](#current-gaps)
- [Proposed Changes](#proposed-changes)
  * [Add new `uploadFile` and `downloadFile` overloads to Android.](#add-new--uploadfile--and--downloadfile--overloads-to-android)
  * [Add new `uploadData` and `downloadData` methods to Android.](#add-new--uploaddata--and--downloaddata--methods-to-android)
- [Side-by-side comparison](#side-by-side-comparison)
  * [Get URL of a remote resource ✅](#get-url-of-a-remote-resource--)
  * [Download `Data` from a remote resource ❌](#download--data--from-a-remote-resource--)
  * [Download remote resource into local `File` ❌](#download-remote-resource-into-local--file---)
  * [Upload `Data` to a remote resource ❌](#upload--data--to-a-remote-resource--)
  * [Upload `File` to remote resource ❌](#upload--file--to-remote-resource--)
  * [Remove a remote resource ✅](#remove-a-remote-resource--)
  * [List remote resources ✅](#list-remote-resources--)

## Current Gaps

Currently, there are two functional gaps between the [iOS Storage APIs](https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Categories/Storage/StorageCategoryBehavior.swift#L11) and the [Android Storage APIs](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/storage/StorageCategoryBehavior.java#L43):

1. Android’s [`uploadFile(...)`](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/storage/StorageCategoryBehavior.java#L118-L153) and [`downloadFile(...)`](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/storage/StorageCategoryBehavior.java#L79-L116) do not include a facility to listen for progress updates. This gap is noted in https://github.com/aws-amplify/amplify-android/issues/646.
[...]

PR #680, opened 2020-07-30
feature: adds progress callbacks to storage

jamesonwilliams:
The Storage category in Amplify iOS includes progress listeners, which
can be used to monitor the status of an active download or upload. The
original APIs launched for Amplify Android did not include this
capability.

This commit adds overloads for `uploadFile(...)` and `downloadFile(...)`, in a
non-breaking way. The old signatures are still available. However, a caller may
optionally tap into progress updates by using these newly available forms:

```kotlin
Amplify.Storage.downloadFile(fromRemoteKey, toLocalFile,
    StorageDownloadFileOptions.defaultInstance(),
    { Log.i("download", "Progress = ${it.fractionCompleted}") },
    { Log.i("download", "Result = ${it.file.canonicalPath}") },
    { Log.e("download", "Error = ${it.localizedMessage}") }
)
```

Or:
```kotlin
Amplify.Storage.uploadFile(toRemoteKey, fromLocalFile,
    StorageUploadFileOptions.defaultInstance(),
    { Log.i("upload", "Progress = ${it.fractionCompleted}") },
    { Log.i("upload", "Result = ${it.key}") },
    { Log.e("upload", "Error = ${it.localizedMessage}") }
)
```

While not part of the public API, the `UPLOAD_PROGRESS` and
`DOWNLOAD_PROGRESS` Hub events are being removed. Previously they were
published without any token to signify which download or which upload
to which they meant to refer. In a system with multiple uploads and
downloads, this would become a source of defects.

Resolves: https://github.com/aws-amplify/amplify-android/issues/679

### DIFF
--- a/core/src/main/java/com/amplifyframework/storage/result/StorageTransferProgress.java
+++ b/core/src/main/java/com/amplifyframework/storage/result/StorageTransferProgress.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.storage.result;
+
+import androidx.annotation.Nullable;
+
+/**
+ * Represents the current amount of progress that has made during a storage transfer operation.
+ */
+public final class StorageTransferProgress {
+    private final long currentBytes;
+    private final long totalBytes;
+
+    /**
+     * Creates a new StorageTransferProgress instance.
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/amplifyframework/storage/result/StorageTransferProgress.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Gets the fraction of the transfer that has been completed, so far. This is a value
+     * greater than or equal to 0, and less than or equal to 1. When the value is 0, the transfer
+     * has not started. When the value is 1, the transfer is complete. If 5 bytes have been transferred,
+     * and 10 bytes are expected, this value is (5.0f/10.0f) = 0.5f.
+     * @return Fraction of transfer that has been completed, a value between 0 and 1, inclusive.
+     */
+    public double getFractionCompleted() {
+        return ((double) currentBytes) / totalBytes;
```
richardmcclellan, line 66: Maybe add a check here to prevent a possible divide by zero exception?  Might be a good idea because the previous logic in AWSS3StorageUploadFileOperation.java did have a check like this.

jamesonwilliams:
> Good addition overall!
> 
> I see that we are currently ignoring the related integration tests (`AWSS3StorageDownloadTest` and `AWSS3StorageUploadTest`) because they "Contain tests that hang, or hang the suite overall.". Are you able to run those locally at all, or perhaps use some portion of them to verify this new PR doesn't break anything?

Yes! Those tests do pass, locally, when run individually.

PR #684, opened 2020-07-30
chore(automation)Leverage fastlane to automate deployments

rjuliano:
This PR adds some basic components so we can use fastlane (and existing fastlane components used by other teams[1]) to push releases to maven. This is largely modeled after the approach used by iOS[2] where we have two "lanes": unstable and release. This PR deals mostly the unstable lane. 

Keep in mind that this is not currently wired in as a step in the circle-ci workflow, so it will not execute automatically. Once configured setup, this is going to execute on every commit to `main` after all the tests completed successfully.

To execute you can run `bundle exec fastlane android unstable`, which will perform the following:
- Get latest release tag. (Example: `release_v1.1.1`)
- Calculate next versionName and versionCode based on that. (Result: `1.1.2-unstable.0` and `101020` respectively)
- Modify `gradle.properties` to update VERSION_NAME and VERSION_CODE (`1.1.2-unstable.0` and 
- Execute the gradle `assemble` task for the `Release` build variant.
- Create a git tag. (Currently not pushing to remote so we're not polluting the repo during this phase)
- Execute the gradle `publishToMavenLocal` task (This will be updated to publish to sonatype)

`VERSION_CODE` is calculated with the following formula:
`version_code = (<major> * 100000) + (<minor> * 1000) + (<patch> * 10) + (<build_number> or 0)`

**NOTE**: This PR depends on the PR I opened in the amplify-ci-support repo [1]

References:
[1] https://github.com/aws-amplify/amplify-ci-support/pull/58
[2] https://github.com/aws-amplify/amplify-ios/blob/v1.0.5-unstable.0/fastlane/Fastfile


Gemfile.lock
```diff
@@ -0,0 +1,187 @@
+GIT
+  remote: https://github.com/aws-amplify/amplify-ci-support
+  revision: c1687ac33bc2d5420a29cd1dc4c5e90aeb36b701
+  branch: master
+  glob: src/fastlane/release_actions/*.gemspec
+  specs:
+    fastlane-plugin-release_actions (1.0.4)
+
+GEM
```
jamesonwilliams, line 9: Is this `Gemfile.lock` auto-generated? How was it generated, out of curiosity?

fastlane/Fastfile
```diff
@@ -0,0 +1,85 @@
...
+      task: "assemble",
+      build_type: "Release"
+    )
+
+    # # Create tag and push to origin
+    add_tag(version: RELEASE_TAG_PREFIX + next_version.to_s)
+
+    # # Push to maven
+    UI.message("Pushing #{next_version.to_s} to maven")
+    publish_to_maven
```
jamesonwilliams, line 28: Is the idea here that the different lanes have different publications? private -> local, public -> maven central?

What's the use case for Maven local? Would we run the Fastlane script on our local machines?

Gemfile.lock
rjuliano: I can give that a try. The only catch is that the `Fastfile` has to live under a `fastlane` subdirectory relative to where the `Gemfile` is (as far as I can tell). Maybe I can move the `Gemfile` under `.circleci` and move the `fastlane` folder there.? We'd end up with:
```
|[.circleci]
|-Gemfile
|-Gemfile.lock
|-[fastlane]
|-...fastlane stuff...
```

fastlane/Fastfile
rjuliano: Maven local was just a placeholder. I wanted to get the bits in place up to the point where I'd be publishing to maven. There will be a follow-up PR for that with the changes to publish to sonatype.

Regarding the lanes, my understanding is that "public" lanes (in our case unstable and release) are the only ones that can be invoked when using the fastlane cli. `bundle exec fastlane <platform> <public lane name>`. "Private" lanes cannot be invoked directly and basically serve as a way to encapsulate some functionality you might use from different public lanes. 

Commits to the `main` branch will trigger a step in the circleci workflow (after all the testing) which will run the `unstable` lane: `bundle exec fastlane android unstable`. The artifacts of a successful build will then be published to maven with version `1.1.2-unstable.0` and a tag will be created with `release_v1.1.2-unstable.0`.

The release lane is similar, but will also generate release notes and will be published to maven as the final release with version `1.1.2` and tagged as `release_v1.1.2`. I'm not sure what the trigger will be (iOS is using a release branch). The first step is to get unstable builds published so developers can use those unstable builds to verify fixes and things of that sort.

Gemfile.lock
rjuliano: I moved it to the `.circleci` subdir. We can always change it if that doesn't feel like the right place.

.circleci/Gemfile.lock
```diff
@@ -0,0 +1,187 @@
+GIT
+  remote: https://github.com/aws-amplify/amplify-ci-support
+  revision: bea40e79f1c871f1fb5ceb86aeb6b1f2a284ed87
+  branch: rjuliano/android-deploy
```
rjuliano, line 4: Using the branch for the PR I have against the amplify-ci-support repo. That will hopefully give us a chance to test things out

jpignata:
Should we close this for now as we rework to test out CodeBuild?

rjuliano:
Will revisit this when we implement release automation

PR #690, opened 2020-08-03
Display logs on the developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Display any stored logs on the developer menu.

Logs screen when there are logs to display:
Dev Menu Logs Displayed

Logs screen when there are no logs to display:
Dev Menu No Logs


core/src/main/java/com/amplifyframework/devmenu/RetrieveLogs.java
```diff
@@ -0,0 +1,44 @@
...
+public class RetrieveLogs {
+
+    /**
+     * Returns a String representation of all of the stored logs.
+     * @return the stored logs as a String.
+     */
+    public String getLogs() {
+        String persistentPluginKey = new PersistentLogStoragePlugin().getPluginKey();
+        LoggingPlugin<?> persistentLogPlugin = Amplify.Logging.getPlugin(persistentPluginKey);
+        List<LogEntry> logs = ((PersistentLogStoragePlugin) persistentLogPlugin).getLogs();
```
jamesonwilliams, line 37: This seems like overkill. Instead, I'd recommend  doing away with the `RetrieveLogs`. Instead, in your `DevMenu`, hang on to the instance of the `PersistentLogStoragePlugin`, as an instance variable. Then, whenever you want the logs, just do `plugin.getLogs()`, directly.

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
```diff
@@ -38,6 +39,10 @@ public DevMenuLogsFragment() { }
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         // Inflate the layout for this fragment
-        return inflater.inflate(R.layout.fragment_logs, container, false);
+        View logsView = inflater.inflate(R.layout.fragment_logs, container, false);
+        // Display the logs (if any).
+        TextView logsText = logsView.findViewById(R.id.logs_text);
+        logsText.setText(DeveloperMenu.singletonInstance(getContext()).getLogs());
```
richardmcclellan, line 45: This is probably not in scope for this PR, but it would be great if the log menu updated in real time, instead of just `onCreateView`.   Is that possible?

eeatonaws:
> From the screenshot you posted, the log window seems really small. Any thoughts about making the dev menu full screen?

We made the developer menu half screen to match React Native's developer menu and also because it might be less invasive for the user to see a smaller menu overlaid on the current screen of their app rather than a full, new screen.

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
eeatonaws: It might be possible using either Android's ObservableList to store the logs in each PersistentLogger or making PersistentLogger observable.

PR #700, opened 2020-08-05
Implement developer menu file issue functionality

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Implements the functionality of the file issue screen and makes changes to the file issue UI for the developer menu.

File issue screen after pressing the "File Issue" button on the main screen of the developer menu:
Dev Menu New File Issue Screen

Error message that appears when trying to copy the issue body or file an issue and the issue description does not satisfy the minimum character count requirement:
Dev Menu New File Issue Description Error

When the user clicks the "File Issue" button on the file issue screen, they are taken to the new issue page on GitHub (after logging in, if necessary). The issue body on GitHub looks as follows when clicking the "Preview" button on the GitHub new issue page:
Dev Menu Create New Issue


core/src/main/java/com/amplifyframework/devmenu/DevMenuFileIssueFragment.java
```diff
@@ -28,6 +32,18 @@
  * file issue view for the developer menu.
  */
 public final class DevMenuFileIssueFragment extends Fragment {
+    // Minimum number of characters required for the issue description.
+    private static final int MIN_DESCRIPTION_LENGTH = 140;
```
jamesonwilliams, line 36: If this will actually _prevent_ upload, we may want to bring it down to like 50. We can always triage/delete crappy issues, once we get them. But I could see myself getting frustrated if I had something really simple to share, but needed to pad it out to 140.

core/src/main/res/values/dev_menu_strings.xml
```diff
@@ -35,5 +35,6 @@
     <string name="copy_issue_text">Copy to Clipboard</string>
     <string name="file_issue_warning_text">
         Warning: device and environment information will be included in the issue description.
```
jamesonwilliams, line 37: re @richardmcclellan's observation - we could remove the explicit "Warning: " label, too? What's left will certainly still _be_ and _act_ as a warning, it just won't trigger the same psychological alarm, if we remove the declaration.
```suggestion
        Device and environment information will be included in the issue description.
```

core/src/main/java/com/amplifyframework/devmenu/DevMenuFileIssueFragment.java
eeatonaws: I think 140 characters is a reasonable requirement to help make sure the customer is actually providing a description of the issue. For reference, the following description (which still seems pretty short) is 148 characters:
"Some description of the issue would be added here. This description meets the minimum character count requirement. More information about the issue."

jamesonwilliams: Okay, cool! Let's run with it.

PR #703, opened 2020-08-06
feature(core): allow additional platforms in amplify config

raphkim:
*Description of changes:*
* Update `AmplifyConfiguration` to use a builder pattern and deprecate the other publicly exposed constructors. 
* Allow passing in additional platforms as parameters to building an Amplify configuration
* Add global UserAgent `configure()` method


core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -189,5 +227,101 @@ public CategoryConfiguration forCategoryType(@NonNull CategoryType categoryType)
...
+
+    /**
+     * Loads a builder for an {@link AmplifyConfiguration} from an amplifyconfiguration.json file.
+     *
+     * @param context Context needed for reading JSON file
+     * @return An Amplify configuration builder instance
+     * @throws AmplifyException If there is a problem in the config file
+     */
+    @NonNull
+    public static Builder loadConfigFile(@NonNull Context context) throws AmplifyException {
```
rjuliano, line 239: Maybe consider moving the `loadConfigFile` method into the `Builder` and create a builder method for `AmplifyConfiguration` that creates an instance of that `Builder`. So the callers would do something like this:

```
AmplifyConfiguration
    .builder() 
    .fromConfigFile(context)
    .forCategoryType(<category type>)
    .build();
```

It would be a bit more consistent with how other builders are structured throughout the framework.

core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -20,26 +20,83 @@
...
-            instance = new UserAgent.Builder()
-                .libraryName("amplify-android")
+            LOG.debug("User-Agent is not yet configured. Returning default Android user-agent.");
+            return androidUserAgent();
+        }
+
+        return instance;
+    }
+
+    private static String androidUserAgent() {
```
jamesonwilliams, line 97: should this just be part of `string()`?

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
raphkim: The reason for this implementation approach is that we want user to choose one specific way to load in details for the builder. Otherwise, a user may do 

```
AmplifyConfiguration.builder()
    .fromConfigFile(context)
    .fromJson(json) // what does this do?
    .fromConfigFile(anotherContext) // ????
    .build();
```

My implementation guarantees that one and only one loading approach is used for building a configuration.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/TestApiCategory.java
```diff
@@ -61,7 +61,8 @@ static ApiCategory fromConfiguration(@RawRes int resourceId) throws AmplifyExcep
                 )
         );
         CategoryConfiguration apiConfiguration =
-            AmplifyConfiguration.fromConfigFile(context, resourceId)
+            AmplifyConfiguration.loadConfigFile(context, resourceId)
```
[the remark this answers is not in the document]
raphkim: I like the 2nd approach! I'll do that.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginUserAgentTest.java
```diff
@@ -45,7 +45,7 @@
  * in {@link AWSApiPlugin}.
  */
 @RunWith(RobolectricTestRunner.class)
-public final class UserAgentTest {
+public final class AWSApiPluginUserAgentTest {
```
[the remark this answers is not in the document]
raphkim: Haha yea. It's an unnecessary change I made that made its way to this PR, but I think its better to be descriptive anyways.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -147,7 +148,9 @@ public void setup() throws AmplifyException {
     public void testConfigure() throws AmplifyException, JSONException {
         UserStateDetails userStateDetails = new UserStateDetails(UserState.SIGNED_OUT, null);
         Context context = getApplicationContext();
-        JSONObject pluginConfig = new JSONObject().put("TestKey", "TestVal");
+        JSONObject pluginConfig = new JSONObject()
+                .put("TestKey", "TestVal")
+                .put("UserAgentOverride", UserAgent.string());
```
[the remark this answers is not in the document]
raphkim: This test was broken after user-agent override PR for auth was merged in, not from this PR. I was just correcting it while I'm at it.

core/src/main/java/com/amplifyframework/Platform.java
```diff
@@ -0,0 +1,46 @@
...
+package com.amplifyframework;
+
+/**
+ * Enum to represent various platforms that use Amplify library for
+ * tracking usage metrics.
+ */
+public enum Platform {
+    /**
+     * Represents the Android platform.
+     */
```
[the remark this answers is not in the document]
raphkim: This enum should not be used actually haha. It is only used internally by `UserAgent` to construct the default android portion. I should actually document that a user should NOT use this, since it's always accounted for already.

core/src/main/java/com/amplifyframework/Platform.java
```diff
@@ -0,0 +1,46 @@
...
+    ANDROID("amplify-android"),
+
+    /**
+     * Represents the Flutter platform.
```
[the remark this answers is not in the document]
raphkim: Yep I'll add a bit more to make it more descriptive.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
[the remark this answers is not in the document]
jamesonwilliams: Ah, okay, cool!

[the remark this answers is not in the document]
raphkim: re-removed from this PR since David merged in a fix

core/src/main/java/com/amplifyframework/util/UserAgent.java
raphkim: its used twice in the code so i just made it a standalone method. renamed to `forAndroid()` though. (`UserAgent.forAndroid()` sounds better than `UserAgent.androidUserAgent()`)

core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -20,26 +20,83 @@
...
+import com.amplifyframework.logging.Logger;
+
+import java.util.Map;
 
 /**
  * A utility to construct a User-Agent header, to be sent with all network operations.
  */
 public final class UserAgent {
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:core");
+    private static final int SIZE_LIMIT = 254; // VARCHAR(254)
```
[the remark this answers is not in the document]
raphkim: Documented in the code where the size limit is being checked.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/TestApiCategory.java
```diff
@@ -61,7 +61,8 @@ static ApiCategory fromConfiguration(@RawRes int resourceId) throws AmplifyExcep
                 )
         );
         CategoryConfiguration apiConfiguration =
-            AmplifyConfiguration.fromConfigFile(context, resourceId)
+            AmplifyConfiguration.builder(context, resourceId)
+                .build()
```
jamesonwilliams, line 65: For these ones, I think you can keep around the factory method. Just have it proxy to the builder.

e.g.

```java
public final class AmplfiyConfiguration {
    private AmplifyConfiguration(..., ..,) { /* make one, privately */ }

    public static AmplifyConfiguration fromConfigFile(Context context, \@ResId int resourceId) {
        return builder(conext, resourceId).build();
    }

    public static Builder builder(Context context, \@ResId int resourceId) {
        reutrn new Builder(contetx, resourceId);
    }

    ...
    public static final class Builder {
        ...
        AmplifyConifguration build() { ... }
    }
}
```

If you do this, you'll cleanup the PR diff quite a bit too. There's a number of ~unnecessary changes to use the new builder, when you could just keep the factory, for most uses.

core/src/main/java/com/amplifyframework/Platform.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework;
```
jamesonwilliams, line 16: Is this the right package? Is there a more specific one you can use? What about `.core`, at least?

Or what about moving to an inner enum of the `AmplifyConfiguration`?

core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -20,26 +20,83 @@
...
+                    AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION
+            );
+        }
+
+        instance = userAgent.toString();
+    }
+
     /**
      * Gets a String to use as the value of a User-Agent header.
+     * Uses default android user-agent if {@link #configure(Map)} was not called yet.
```
jamesonwilliams, line 83:
```suggestion
     * Uses default Android user-agent if {@link #configure(Map)} has not been called yet.
```

core/src/main/java/com/amplifyframework/util/UserAgent.java
```diff
@@ -20,26 +20,83 @@
...
 
     private UserAgent() {}
 
+    /**
+     * Configure User-Agent singleton with Amplify configuration instance.
+     * @param platformVersions A map of additional platforms that are calling this library
+     *                         and their respective versions.
+     * @throws AmplifyException If called twice or user-agent exceeds size limit.
+     */
+    public static synchronized void configure(@NonNull Map<Platform, String> platformVersions)
```
jamesonwilliams, line 48: If you move `Platform` to be `AmplifyConifguration.Platform`, then do something like:

```
LinkedHashMap<String, String> paltformVersions.= new LinkedHashMap<>();
for (Map.Entry<Platform, String> provided : inputVersions.entrySet()) {
    platformVersion.put(provided.getKey().name(), provided.getValue());
}
```
Then you could break the type dependency and just have String-to-String map on this API.

core/src/main/java/com/amplifyframework/Platform.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Enum to represent various platforms that use Amplify library for
+ * tracking usage metrics.
+ */
+public enum Platform {
+    /**
+     * This is the default platform that will be included in every user-agent
```
jamesonwilliams, line 24: If we're talking about  "user agents" in this `Platform.java`, maybe it should be an inner enum of `UserAgent.java`?

core/src/main/java/com/amplifyframework/Platform.java
```diff
@@ -0,0 +1,51 @@
...
+     * that is generated by this library.
+     * A user should never be using this enum to specify their platform version
+     * to avoid redundancy.
+     */
+    ANDROID("amplify-android"),
+
+    /**
+     * The Flutter library calls on Android Amplify. This enum should be specified
+     * during the construction of an Amplify configuration object to indicate that
```
jamesonwilliams, line 33: If you talk about construction, include a link to the method used to do so.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -53,81 +57,115 @@
     private static final String DEFAULT_IDENTIFIER = "amplifyconfiguration";
 
     private final Map<String, CategoryConfiguration> categoryConfigurations;
+    private final Map<Platform, String> platformVersions;
```
jamesonwilliams, line 60: If order matters, you may want to use `LinkedHashMap` on the LHS of all uses. The Liskov principle is about using the type of least specificity. But for your use case, if ordering is essential, `Map` is not the type of least specificity.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -53,81 +57,115 @@
...
      * Creates an AmplifyConfiguration from an amplifyconfiguration.json file.
+     *
      * @param context Context needed for reading JSON file
      * @return An Amplify configuration instance
      * @throws AmplifyException If there is a problem in the config file
+     * @deprecated This method is now deprecated in favor of
+     *          {@link AmplifyConfiguration#builder(Context)}.
      */
     @SuppressWarnings("WeakerAccess")
+    @Deprecated
```
jamesonwilliams, line 106: I think it's still useful; I'd leave it, and use it around the code base.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -53,81 +57,115 @@
...
      * @param context Android Context
      * @param configFileResourceId
      *        The Android resource ID of a raw resource which contains
      *        an amplify configuration as JSON
      * @return An Amplify configuration instance
      * @throws AmplifyException If there is a problem in the config file
+     * @deprecated This method is now deprecated in favor of
+     *          {@link AmplifyConfiguration#builder(Context, int)}.
      */
+    @Deprecated
```
jamesonwilliams, line 124: Ditto, you have `builder(..., ...).build()` around the codebase, so why not just keep this and use it, since it's cleaner looking.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -53,81 +57,115 @@
...
+        return builder(context, configFileResourceId).build();
+    }
+
+    /**
+     * Obtain a map of additional platforms that are using this library
+     * for tracking usage metrics.
+     * @return A map of additional platforms and their versions.
+     */
+    @NonNull
+    public Map<Platform, String> getPlatformVersions() {
```
jamesonwilliams, line 137: Order matters right? `LinkedHashMap` as a return type here? Also: will `Immutable` respect that? Or does it return a generic `Map`? May need to use generics and change it to liek:
```java
public static <T, M extends Map<T>> M of(M map) {
    ...
}
```

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -19,8 +19,10 @@
 import android.content.res.Resources;
```
jamesonwilliams, line 19: Be sure to include `@Nullable` and `@NonNull` on every input and output of any public method in this file. Kotlin will suck, if you don't, and most of our customers use Kotlin.

core/src/main/java/com/amplifyframework/Platform.java
raphkim: I didn't want to move it as an inner enum, since that makes the user need to specify `AmplifyConfiguration.Platform.FLUTTER` instead of just `Platform.FLUTTER`. It also keeps it more hidden from plain sight.

Also, `UserAgent` is currently configured globally with `Map<Platform, String>`. Making `Platform` a part of `AmplifyConfiguration` would couple it with `UserAgent`, which you mentioned you would rather not do.

raphkim: As for whether it belongs inside `core` package or the outermost package, I'm not sure what the best approach would be. Personally, I thought that `Platform` specification is one of the most broad concepts in the library, so it might as well belong outside of the core package. I don't think there's a big distinction between `com.amplifyframework` and `com.amplifyframework.core` packages as of right now.

core/src/main/java/com/amplifyframework/Platform.java
raphkim: Oh that's not a terrible idea actually. I initially wanted `Platform` to be a standalone thing, but I can't really think of it being used outside of the context of User-Agent any time in the future.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
raphkim: Replacing it with `@VisibleForTesting` for now because it really shouldn't be used by the customers.

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
raphkim: Every public method should already have the annotation.

aws-api/src/androidTest/java/com/amplifyframework/api/aws/TestApiCategory.java
raphkim: Will do!

core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
raphkim: Wouldn't it automatically obtain an iterator from `LinkedHashMap` implementation everywhere I am referencing it because it is instantiated as one?

core/src/main/java/com/amplifyframework/util/UserAgent.java
raphkim: Moving to `UserAgent` instead

core/src/main/java/com/amplifyframework/Platform.java
raphkim: Added sample usage code

README.md
```diff
@@ -63,15 +63,15 @@ dependencies section:
...
+    implementation 'com.amplifyframework:core:1.1.2'
 
     // Only specify modules that provide functionality your app will use
-    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.0.0'
-    implementation 'com.amplifyframework:aws-api:1.0.0'
-    implementation 'com.amplifyframework:aws-auth-cognito:1.0.0'
-    implementation 'com.amplifyframework:aws-datastore:1.0.0'
-    implementation 'com.amplifyframework:aws-predictions:1.0.0'
-    implementation 'com.amplifyframework:aws-storage-s3:1.0.0'
+    implementation 'com.amplifyframework:aws-analytics-pinpoint:1.1.2'
```
jamesonwilliams, line 69: Uh oh, looks like a rebasing issue.

raphkim: ugh

raphkim:
> Looks good to me! What did you do for testing?

Oof. No tests have been added in this PR.

I initially had some test suites in `core` package set up with the same code from `aws-api`'s `UserAgentTest` set up to test it, but I removed it because it felt strange depending on the API category to test the user agent.

PR #710, opened 2020-08-08
feat(datastore) Trigger hub events during sync

rjuliano:
*Issue #, if available:*

*Description of changes:*
Trigger the following events during initial sync:

- syncQueriesStarted: triggered when the sync process starts.
- syncQueriesReady: triggered when all models have been sync'd
- modelSynced: triggered for each individual model upon completion of the sync process for each model. Includes the sync type (FULL or DELTA), model name and number of records added, updates and deleted during the sync.


core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -52,7 +52,22 @@
     /**
      * The remote synchonization processes stopped.
      */
-    REMOTE_SYNC_STOPPED("remote_sync_stopped");
+    REMOTE_SYNC_STOPPED("remote_sync_stopped"),
```
jamesonwilliams, line 55: Should we also remove `REMOTE_SYNC_STARTED` and `REMOTE_SYNC_STOPPED` as part of this? Right now, I think we have some duplicative code. A good goal would be to align the events to iOS and Android. In so doing, that means we have to rename/remove the events we have, that already provide a similar functionality.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -52,7 +52,22 @@
...
+
+    /**
+     * The DataStore is about to start the Sync Queries.
+     */
+    SYNC_QUERIES_STARTED("syncQueriesStarted"),
+
+    /**
+     * All models have been synced.
+     */
+    SYNC_QUERIES_READY("syncQueriesReady"),
```
jamesonwilliams, line 65: I'm not sure what "sync queries ready" means, just from the name. Is this a good name?

The Javadoc is the same on L63 and L68 so I think at least one needs to be updated.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -52,7 +52,22 @@
...
+
+    /**
+     * All models have been synced.
+     */
+    MODEL_SYNCED("modelSynced");
```
jamesonwilliams, line 70: One says model (singular) and one says models (plural). From the Javadoc, it looks like it means the plural form. If that's so, can we name this `ALL_MODELS_SYCNED`? This probably means going back to the cross-platform design doc and making sure the other guys do likewise. Or, if JS already has these publicly documented and live, we could at least align some better names with iOS.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -92,12 +106,67 @@ public static Builder builder() {
...
+                .map(this::filterOutOldSyncTimes)
+                .blockingGet();
+            metricsByModel.put(clazz.getSimpleName(), new ModelSyncMetrics(lastSyncTime));
             hydrationTasks.add(createHydrationTask(modelWithMetadataComparator, clazz));
         }
-        return Completable.concat(hydrationTasks);
+
+        // We'll collect metrics for the initial sync by observing changes to the local DataStore.
+        Cancelable metricsObserver =
+            localStorageAdapter.observe(itemChange -> {
```
jamesonwilliams, line 121: Since we have the observer pattern here, I don't think you even need to put this in the `SyncProcessor` at all, do you? You could have this be like a `SyncMetricsPublisher.java` or a `SyncMetricsObserver.java`, and then cleanly separate all of this logic into its own little place.

Overall though, I'm worry about how long `hydrate()` is getting, here. Any approaches you can think of to break this apart into some smaller bits would probably useful, towards keeping the code clear. -- Keep in mind, the reader expects `hydrate()` to be _mostly_ about populating data, not broadcasting events.

The other benefit to breaking out a separate observer class is that you will no longer have `SyncProcessor` depending on the `LocalStorageAdapter`. I don't think I ever _completely_ got there, but you can se several mid-level utilities like the `VersionRepository`, etc., which abstract away the `LocalStorageAdapter` from the sync processor. All in all, I think the `LocalStorageAdapter` is a fairly low level detail for the sync processor and -- where possible -- it would be best to avoid direct dependency on it.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncType.java
```diff
@@ -0,0 +1,31 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.syncengine;
+
+/**
+ * Enum representing the type of initial sync.
+ */
+public enum SyncType {
```
jamesonwilliams, line 21: Ahh. Hm, yea.

So, this looks reasonable. Can you create a separate isolated PR for this?

If we treat this seriously, I think we can make use of `SyncType` in a number of places where delta vs. base sync logic is managed currently. If I'm not mistaken, it is a nullability check in most places right now. Like, do I have a timestamp yes vs. no? However, it would be cleaner to be like:
```java
if (lastSyncTime == null) {
    syncType = SyncType.FULL;
}
```
etc.

TLDR there's a bunch of places you can use this, and it'll probably be significant enough for its own (~low complexity, but potentially verbose) PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/ModelSyncedEvent.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public int getDeleted() {
+        return deleted;
+    }
+
+    /**
+     * Getter for the isFullSync property.
+     * @return True if it's a full sync, false otherwise.
+     */
+    public boolean isFullSync() {
```
jamesonwilliams, line 90: Or, alternately, you could just use your enum type? `getSyncType()`? The way you have it is fine too.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/ModelSyncedEvent.java
```diff
@@ -0,0 +1,112 @@
...
+package com.amplifyframework.datastore.syncengine.events;
+
+import androidx.annotation.NonNull;
+
+import com.amplifyframework.datastore.syncengine.SyncType;
+
+/**
+ * Hub event payload emitted when the initial sync completes for a given model.
+ */
+public final class ModelSyncedEvent {
```
jamesonwilliams, line 25: Since you'll send these back to the customer, be sure that all of your event structures are:
1. Immutable;
2. Implement equals, toString, and hashCode
3. If applicable, implementing `Comparable` might be appropriate as well?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/SyncQueriesStartedEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public final class SyncQueriesStartedEvent {
+    private String[] models;
+
+    /**
+     * Constructs a SyncQueriesStartedEvent object.
+     * @param models An array of model names.
+     */
+    public SyncQueriesStartedEvent(String[] models) {
+        this.models = models;
```
jamesonwilliams, line 29: If you use the array primitive, be sure to do an `Array.copy` on the input, so that your DTO won't be effected if the input hcanges. https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#copyOf(U[],%20int,%20java.lang.Class)

same deal on this one -- `equals`, `toString`, and `hashCode`, please

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
rjuliano: Definitely agree on the naming. That being said, I think we're supposed to align with the names currently used in JS [1]; however, I'll check and see what iOS is currently doing.

Regarding the javadoc, that's my trademarked **Clipboard Inheritance** pattern [2] 🤓 

[1] - https://github.com/aws-amplify/amplify-js/blob/a7feacea4ed506340d250249d0b15286fe3ef5fa/packages/datastore/src/sync/index.ts#L71-L82
[2] - There's a wikipedia page for it, so it's real. https://en.wikipedia.org/wiki/Copy-and-paste_programming

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
rjuliano: Makes sense. We're using this for a slightly different purpose I believe (it's more of a orchestrator start and stop), but I don't believe anyone is using this as it is not documented.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
rjuliano: This event is emitted once for each individual model, so it makes sense for it to be singular.

Regarding the event names in general, I'll check in with iOS to see how they're going about implementing. My understanding from one of the meeting we had is that they were going to follow JS though.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: LOL...I should have followed my gut on this one. I was going to do just that and talked myself out of it. 🤣  The moment I had to add the local `LocalStorageAdapter` as a dependency, it felt wrong. Glad to refactor this though.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
undefobj: This was the one that we talked about in the working group that we could possible rename but if we do it then should be done quickly since the JS versions aren't documented yet. Suggest you sync up with Ivan this week.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncType.java
rjuliano: Just posted #713 

Note that in that PR, I'm using `BASE` instead of `FULL` since it appears to be more consistent with what I've seen in the docs. 

I have some other changes to this PR which should address some of the other issues. Once we get #713 merged, I can update this PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: I refactored this on the latest commit.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/ModelSyncedEvent.java
rjuliano: The event payload calls for isFullSync and isDeltaSync, so I just kept it consistent with what JS is doing.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
rjuliano: Fixed the javadoc on this one.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/ModelSyncedEvent.java
rjuliano: 1 and 2 should be addressed. I'm not sure it makes sense to implement comparable here since none of the fields really stand out in terms of sorting.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/events/SyncQueriesStartedEvent.java
rjuliano: Done. Added tests using `EqualsToStringHashValidator`

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
rjuliano: I removed REMOTE_SYNC_STOPPED and REMOTE_SYNC_STARTED

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -97,6 +103,11 @@ public Orchestrator(
         Merger merger = new Merger(mutationOutbox, versionRepository, localStorageAdapter);
         SyncTimeRegistry syncTimeRegistry = new SyncTimeRegistry(localStorageAdapter);
 
+        this.localStorageAdapter = localStorageAdapter;
+        syncableModels = Observable.fromIterable(modelProvider.models())
+            .toList()
+            .blockingGet();
```
jamesonwilliams, line 109: This seems to be a frivolous construct. Isn't this just equivalent to using `modelProvider.models()` directly?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -106,6 +117,9 @@ public Orchestrator(
             .merger(merger)
             .dataStoreConfigurationProvider(dataStoreConfigurationProvider)
             .build();
+        this.syncObserver = new SyncMetricsObserver(localStorageAdapter,
```
jamesonwilliams, line 120: It gets assigned here, but where is it used? Do you have behavior in the constructor? If so, can you move it to `start()` and `stop()` methods, and keep the constructor only for dependency wiring?

Other thought: it would be kinda nice if this could be `syncProcessor.addObserver(new SyncMtericsObserver())`, or if the `syncProcessor` just managed this thing internally, itself._

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Class that encapsulates the logic of listening to local DataStore changes
+ * and emitting metrics when base/delta sync completes.
+ */
+class SyncMetricsObserver {
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
+    private Map<String, ModelSyncMetrics> metricsByModel;
```
jamesonwilliams, line 43:
```suggestion
    private final Map<String, ModelSyncMetrics> metricsByModel;
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Class that encapsulates the logic of listening to local DataStore changes
+ * and emitting metrics when base/delta sync completes.
+ */
+class SyncMetricsObserver {
```
jamesonwilliams, line 41:
```suggestion
final class SyncMetricsObserver {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
```diff
@@ -0,0 +1,135 @@
...
+        itemChangeObserver.cancel();
+    }
+
+    private void onError(DataStoreException dataStoreException) {
+        LOG.warn("Unable to gather metrics for remote sync.", dataStoreException);
+    }
+
+    private void onItemChange(StorageItemChange<? extends Model> itemChange) {
+        boolean isSyncProcessChange = StorageItemChange.Initiator.SYNC_ENGINE.equals(itemChange.initiator());
+        if (isSyncProcessChange) {
```
jamesonwilliams, line 77: Maybe invert this, so you can save 4 columns for the main body of the method.
e.g.
```java
if (!isSyncProcessChange) {
    return;
}
ahaSweetISavedSomeCols();
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -206,6 +204,10 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
      */
     @Test
     public void clearStopsSyncUntilNextInteraction() throws AmplifyException, JSONException {
+//        HubAccumulator subscriptionsStarted =
```
jamesonwilliams, line 207: Is this still needed? Can you remove the commented out stuff?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -307,16 +309,17 @@ private void assertRemoteSubscriptionsStarted() {
...
+     * method.
+     * @param mockApi Mock or spy of the ApiCategory being used for the test.
      */
-    private void assertSyncProcessorNotStarted() {
-        boolean syncProcessorNotInvoked = mockingDetails(modelProvider)
-            .getInvocations()
-            .stream()
-            .noneMatch(invocation -> invocation.getLocation().getSourceFile().contains("SyncProcessor"));
-        assertTrue(syncProcessorNotInvoked);
+    private void assertSyncProcessorNotStarted(ApiCategory mockApi) {
```
jamesonwilliams, line 317: Can we do this by checking for an event on the hub, instead of by using mockito invocation verification?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -69,13 +79,18 @@
 public final class SyncProcessorTest {
     private static final long OP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(2);
     private static final long BASE_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
+    private static final List<String> SYSTEM_MODEL_NAMES =
+        Observable.fromIterable(SystemModelsProviderFactory.create().models())
+            .map(m -> m.getSimpleName()).toList().blockingGet();
```
jamesonwilliams, line 84: `m -> m.getSimpleName()` I think you can write `Class::getSimpleName`, instead?

The stream APIs are a little nicer for simple collection mapping like this aren't they? Oh well.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -116,6 +140,91 @@ public void setup() throws AmplifyException {
...
+     * During a base sync, there are a series of events that should be emitted.
+     * This test verifies that these events are published via Amplify Hub depending
+     * on actions takes for each available model.
+     * @throws DataStoreException Not expected.
+     */
+    @Test
+    public void dataStoreHubEventsTriggered() throws DataStoreException {
+        // Arrange - BEGIN
+        // We're only emitting events for Post and Blogger
+        int expectedModelCount = 2;
```
jamesonwilliams, line 153: To self document it, you could express this as `int expectedModelCount = Arrays.asList(EventObject1.class, EventObject2.class).size()`.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -400,6 +509,60 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
...
+                return forEvent(DataStoreChannelEventName.MODEL_SYNCED).filter(hubEvent) &&
+                    modelNames.contains(hubEventData.getModel());
+            }
+        };
+    }
+
+    private static HubEventFilter syncQueryStartedForModels(int modelCount) {
+        return new HubEventFilter() {
+            @Override
+            public boolean filter(@NonNull HubEvent<?> hubEvent) {
```
jamesonwilliams, line 557: Lambda-fy this?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -400,6 +509,60 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
...
+    @SuppressWarnings("unchecked")
+    private HubEventFilter syncMetricsEmittedFor(Class<? extends Model>... models) {
+        List<String> modelNames = Observable.fromIterable(Arrays.asList(models))
+            .map(model -> model.getSimpleName())
+            .toList()
+            .blockingGet();
+
+        return new HubEventFilter() {
+            @Override
+            public boolean filter(@NonNull HubEvent<?> hubEvent) {
```
jamesonwilliams, line 543: Lambda-fy?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -400,6 +509,60 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
...
+    }
+
+    private static HubAccumulator createAccumulator(HubEventFilter eventFilter, int times) {
+        return HubAccumulator.create(HubChannel.DATASTORE, eventFilter, times);
+    }
+
+    private static HubEventFilter forEvent(DataStoreChannelEventName eventName) {
+        return new HubEventFilter() {
+            @Override
+            public boolean filter(@NonNull HubEvent<?> hubEvent) {
```
jamesonwilliams, line 528: Lambda-fy?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -116,6 +140,91 @@ public void setup() throws AmplifyException {
...
+                    if (!SYSTEM_MODEL_NAMES.contains(eventModel)) {
+                        // Verify there were no unexpected events for any other models.
+                        assertOperationTypeCounts(eventData, 0, 0, 0);
+                    }
+            }
+        }
+        // Check - END
+    }
+
+
```
jamesonwilliams, line 226:
```suggestion
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: Yeah...initially, I was going to filter based on isSyncable, but we don't have such a flag for Android.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: I refactored the logic to make it a bit lighter in terms of dependencies. I created an accumulator class that keeps track of the sync metrics for an individual model. Rather than depending on the local storage adapter and creating an observer, I'm just passing in a reference to this new object into the merge method.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
rjuliano: No longer exists after refactor.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
rjuliano: Made it final after renaming it to `ModelSyncMetricsAccumulator`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncMetricsObserver.java
rjuliano: No longer needed.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: Removed.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: Done.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: That's a good idea.

rjuliano: Done.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: Done

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: That will probably be a little trickier since we'd be checking for the absence of an event. It's probably a useful pattern to create in the `HubAccumulator`, but I'd prefer to do that as a separate PR if that's OK.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -66,6 +67,18 @@
      * @return A completable operation to merge the model
      */
     <T extends Model> Completable merge(ModelWithMetadata<T> modelWithMetadata) {
+        return merge(modelWithMetadata, value -> { });
```
jamesonwilliams, line 70: There is a `NoOpConsumer.create()`, I think? Which might be a little more explicit about the intention.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -87,8 +100,8 @@
             .filter(currentVersion -> currentVersion == -1 || incomingVersion > currentVersion)
             // If we should merge, then do so now, starting with the model data.
             .flatMapCompletable(shouldMerge ->
-                (isDelete ? delete(model) : save(model))
-                    .andThen(save(metadata))
+                (isDelete ? delete(model, storageItemChangeConsumer) : save(model, storageItemChangeConsumer))
+                    .andThen(save(metadata, value -> { }))
```
jamesonwilliams, line 104: Ditto here

core/src/main/java/com/amplifyframework/datastore/events/SyncQueriesStartedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+            return true;
+        }
+        if (thatObject == null || getClass() != thatObject.getClass()) {
+            return false;
+        }
+
+        SyncQueriesStartedEvent that = (SyncQueriesStartedEvent) thatObject;
+        if (models.length != that.models.length) {
+            return false;
+        }
```
jamesonwilliams, line 70: You don't need this, if you use the `Arrays.equals`, right?

core/src/main/java/com/amplifyframework/datastore/events/SyncQueriesStartedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * Event payload emitted when the sync process starts.
+ */
+public final class SyncQueriesStartedEvent {
+    private String[] models;
```
jamesonwilliams, line 27:
```suggestion
    private final String[] models;
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -116,6 +130,88 @@ public void setup() throws AmplifyException {
...
+        syncProcessor.hydrate().subscribe(hydrationObserver);
+
+        // Check - BEGIN
+        // Verify that sync completes.
+        assertTrue(hydrationObserver.awaitTerminalEvent(OP_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+        hydrationObserver.assertNoErrors();
+        hydrationObserver.assertComplete();
+
+        // Verify that syncQueriesStarted was emitted once.
+        assertEquals(1, syncStartAccumulator.await((int) OP_TIMEOUT_MS, TimeUnit.MILLISECONDS).size());
```
jamesonwilliams, line 180: Ah crumbs. I guess `HubAccumulator` should have taken `long`, in hindsight, huh?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -116,6 +130,88 @@ public void setup() throws AmplifyException {
...
+        assertEquals(expectedModelCount, hubEvents.size());
+
+        // For each event (excluding system models), verify the desired count.
+        for (HubEvent<?> event : hubEvents) {
+            ModelSyncedEvent eventData = (ModelSyncedEvent) event.getData();
+            assertTrue(eventData.isFullSync());
+            assertFalse(eventData.isDeltaSync());
+            String eventModel = eventData.getModel();
+            switch (eventModel) {
+                case "BlogOwner":
```
jamesonwilliams, line 196: Can we do `BlogOwner.class.getSimpleName()` for these case statements, instead of string literal?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -400,6 +496,48 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
...
+    private static HubEventFilter forEvent(DataStoreChannelEventName eventName) {
+        return hubEvent -> eventName.toString().equals(hubEvent.getName());
+    }
+
+    @SuppressWarnings("unchecked")
+    private HubEventFilter syncMetricsEmittedFor(Class<? extends Model>... models) {
+        List<String> modelNames = Observable.fromIterable(Arrays.asList(models))
+            .map(model -> model.getSimpleName())
+            .toList()
+            .blockingGet();
```
jamesonwilliams, line 521: We have this pattern pretty frequently, by now.

We could wrap it into a utility:
```java
List<String> modelNames = ForEach.inList(models, Model::getSimpleName);
```
```java
// In testutils, probskis.
public final class ForEach {
     private ForEach() {}

     <I, O> List<O> inList(List<I> in, Mapping<I, O> mapping) {
         final List<O> out = new ArrayList<>();
         for (I item : in) {
             out.add(mapping.apply(item));
         }
         return Immutable.of(out);
     }

     @FunctionalInterface
     interface Mapping<I, O> {
         O apply(I in);
     }
}
```
Software is so cool. Sigh.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -400,6 +496,48 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
         assertEquals(1, errorHandlerCallCount);
     }
 
+    private void assertOperationTypeCounts(ModelSyncedEvent event,
+                                           int expectedAdd,
+                                           int expectedUpdate,
+                                           int expectedDelete) {
```
jamesonwilliams, line 502: When you call this, it is hard to remember which int is which. While potentially overkill, one solution is like:
```java
assertOperationTypeCounts(event, TypeCounts.builder().added(5).updated(1).deleted(0).build());
```
With a little inner data class to store those vals.

Or hey, I mean, it's just a test class. Life doesn't always need to be complicated, we could even just have:
```
class TypeCounts {
    int added;
    int updated;
    int deleted;
}
```
And then prepare them:
```java
TypeCounts counts = new TypeCounts();
counts.added = 4;
counts.updated = 9;
counts.deleted = 1;
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -69,13 +78,17 @@
 public final class SyncProcessorTest {
     private static final long OP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(2);
     private static final long BASE_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
+    private static final List<String> SYSTEM_MODEL_NAMES =
+        Observable.fromIterable(SystemModelsProviderFactory.create().models())
+            .map(Class::getSimpleName).toList().blockingGet();
```
jamesonwilliams, line 83: I discuss a `ForEach `utility below that may make sense, if we have to pull this same trick more than a time or two.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
jamesonwilliams: Oh duh, or wait -- update. Can you just reuse the `ModelSyncMetrics` as a pojo to store these immediate values?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -340,4 +380,25 @@ public SyncProcessor build() {
             return modelSchemaRegistry.getModelSchemaForModelInstance(modelWithMetadata.getModel());
         }
     }
+
+    private static final class ModelSyncMetrics {
```
jamesonwilliams, line 384: Does this get sent back to user? Does it need to be public, with equals/hashCode/toString and field accessors?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -113,13 +144,21 @@ private Completable createHydrationTask(
...
             .flatMapCompletable(syncType -> {
-                return SyncType.DELTA.equals(syncType) ?
+                Completable syncTimeSaveCompletable = SyncType.DELTA.equals(syncType) ?
                     syncTimeRegistry.saveLastDeltaSyncTime(modelClass, SyncTime.now()) :
                     syncTimeRegistry.saveLastBaseSyncTime(modelClass, SyncTime.now());
+                return syncTimeSaveCompletable.andThen(Completable.fromAction(() -> {
+                    metricsAccumulator
+                        .toModelSyncedEvent(syncType)
+                        .toHubEvent()
+                        .publish(HubChannel.DATASTORE, Amplify.Hub);
```
jamesonwilliams, line 160: What implications does this have on testability, if you inline `Amplify.Hub`? Should we instead snake a `hub` instance in through the constructor? That way, you could provide a mock in test, if you needed to?

I forget why we put this in the publish arg... it seemed like a good idea at the time?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -74,6 +84,11 @@ private SyncProcessor(
         this.appSync = Objects.requireNonNull(appSync);
         this.merger = Objects.requireNonNull(merger);
         this.dataStoreConfigurationProvider = dataStoreConfigurationProvider;
+        this.modelNames = Observable.fromIterable(modelProvider.models())
+                                        .map(m -> m.getSimpleName())
+                                        .toList()
+                                        .blockingGet()
+                                        .toArray(new String[0]);
```
jamesonwilliams, line 91: See other comments about `ForEach`. This one returns `String[]` though instead of `List<String>`. I guess we'd have to standardize. Oh, or like `ForEach.inArray`, `ForEach.inList`, etc. Could do multiple collection types

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -97,6 +101,9 @@ public Orchestrator(
         Merger merger = new Merger(mutationOutbox, versionRepository, localStorageAdapter);
         SyncTimeRegistry syncTimeRegistry = new SyncTimeRegistry(localStorageAdapter);
 
+        this.localStorageAdapter = localStorageAdapter;
+        syncableModels = modelProvider.models();
```
jamesonwilliams, line 105: Is this a dead store variable? Doesn't look like `syncableModels` gets used.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -97,6 +101,9 @@ public Orchestrator(
         Merger merger = new Merger(mutationOutbox, versionRepository, localStorageAdapter);
         SyncTimeRegistry syncTimeRegistry = new SyncTimeRegistry(localStorageAdapter);
 
+        this.localStorageAdapter = localStorageAdapter;
```
jamesonwilliams, line 104: Is this dead store, too?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ModelSyncMetricsAccumulator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Class that encapsulates the logic of keeping track of sync metrics
+ * by operation type for a given model.
+ */
+final class ModelSyncMetricsAccumulator {
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
+    private final Map<String, AtomicInteger> syncMetrics;
```
jamesonwilliams, line 36: Does it make sense for the key to be the constrained type, `DataStoreItemChange.TYPE`, instead of just `String`?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -66,6 +67,18 @@
...
+
+    /**
+     * Merge an item back into the local store, using a default strategy.
+     * @param modelWithMetadata A model, combined with metadata about it
+     * @param storageItemChangeConsumer A callback invoked when the merge method saves or deletes the model.
+     * @param <T> Type of model
+     * @return A completable operation to merge the model
+     */
+    <T extends Model> Completable merge(ModelWithMetadata<T> modelWithMetadata,
+                                        Consumer<StorageItemChange<T>> storageItemChangeConsumer) {
```
jamesonwilliams, line 81: Is there anyway we can avoid having these callbacks?

The `Completable` is already an asynchronous signal. Now, this method has two different async signals that have different lifecycles. It's a bit to handle, conceptually.

One possible option, if you need to observe save/deletes on the datastore, is the `observe()` method on the `LocalStorageAdapter`? Hopefully that can be setup internally, without effecting the Merger's contract?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
rjuliano: I thought we had something like that. I was looking for `EmptyConsumer` 🤦

core/src/main/java/com/amplifyframework/datastore/events/SyncQueriesStartedEvent.java
rjuliano: Ah...yes I can.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: Yeah...we can always change that. I'll make a note to do that in a follow up PR

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: Interesting...Evidently `BlogOwner.class.getSimpleName()` is not considered a compile-time constant. Even if I create a final static field and assign it the value of `BlogOwner.class.getSimpleName()`, the compiler still flags it with a `Constant expression required` error

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: I like it :)

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
rjuliano: Oh yeah...I don't even have to have a separate method. I just created "expected" objects and use `assertEquals`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: Actually...this can be removed entirely.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: The original intent was to not have `Amplify.Hub` references inside the `HubEvent` class. Let me ponder over this one for a bit.

jamesonwilliams: I may have simply been wrong on this. It does feel like encapsulating it HubEvent is cleaner than stating it at random call sites like this, doesn't it?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
jamesonwilliams: Sounds good, yes -- for a separate PR!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: I changed the input parameter to be `Collection<T>` which covers a majority of cases and also added a `inArray` method in case someone needs it.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: Yup...can't remember why, but I think I was using it somewhere other than the constructor in one of my previous interations.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ModelSyncMetricsAccumulator.java
rjuliano: You know...I tried to do this and couldn't get it to work. Finally just realized why: there is a `StorageItemChange.Type` and a `DataStoreItemChange.Type`. When I initially tried this, I was using `StorageItemChange.Type` in the key declaration, and then trying to retrieve by passing in a variable of `DataStoreItemChange.Type`. Just brilliant 🤣

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
rjuliano: That was my previous implementation. However, it proved to be cumbersome from the standpoint of determining the lifecycle of the sync attempt for a given model. 

I was creating an observer before kicking off the sync. That observer kept an in-memory map of counters for each model and applied the following logic to each event received.:

```
// pseudo-code
onChangeEvent (changeEvent) {
 if (changeEvent.initiator == SYNC_ENGINE && isUserModel(changeEvent.model)) {
   incrementCounter(changeEvent.model, changeEvent.changeType)
 } else if (changeEvent.initiator == SYNC_ENGINE && changeEvent.model == LastSyncMetadata) {
  // That means the sync attempt finished for one of the models
  let syncMetadata = changeEvent.eventPayload
  //end of sync attempt for model specified in the 
  emitMetricsFor(syncMetadata.modelName);
 } else {
   NoOp
 }
}
```
While this approach worked and it was nicely decoupled, it felt kind of brittle. It assumes that the initial sync  is the only thing emitting events with initiator set to SYNC_ENGINE. While this is true right now, it's one of those things that could change and break this logic. It also assumes that an update to the LastSyncMetadata table made by SYNC_ENGINE was the end of the sync attempt for the model.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: I've gone back and changed this a bit. I see `HubEvent` more as a POJO rather than something that has behavior. I changed the call sites back to:
```java 
Amplify.Hub.publish(HubChannel.DATASTORE, <eventDataPojo>.toHubEvent());
```

core/src/main/java/com/amplifyframework/datastore/events/SyncQueriesStartedEvent.java
```diff
@@ -0,0 +1,70 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * Event payload emitted when the sync process starts.
+ */
+public final class SyncQueriesStartedEvent {
+    private final String[] models;
```
raphkim, line 27: Perhaps we can use `List<String>` or `Set<String>` here?

Most de/serializers should be able to easily convert these collections into a json array.

core/src/main/java/com/amplifyframework/util/ForEach.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Utility class that reduces boilerplate code in tests
+ * where an oft-used pattern of mapping a list of values to another by
+ * applying a function.
+ */
+public final class ForEach {
```
raphkim, line 27: cool!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
rjuliano: If I have bandwidth tomorrow, I'll revisit this one to see what I can come up with. Maybe instead of returning a completable, I can return a Single or a Maybe with the StorageItemChange<T>

core/src/main/java/com/amplifyframework/datastore/events/SyncQueriesStartedEvent.java
rjuliano: I'll probably do this as a follow up PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
rjuliano: I'll do this as a follow-up.

Issue #707, opened 2020-08-07
SQL commands fail if models have words that correspond to sqlite reserved words


saltonmassally:
### Describe the bug

 Datastore SQL commands fail if your model has SQLite reserved words in them. take for example this create table query generated by the library:

```
CREATE TABLE IF NOT EXISTS Request (
  id TEXT PRIMARY KEY NOT NULL,
  state TEXT NOT NULL,
  groupId TEXT,
  FOREIGN KEY (groupId) REFERENCES Group(id) ON DELETE CASCADE,
);
```

or

```
CREATE TABLE IF NOT EXISTS Group (
  id TEXT PRIMARY KEY NOT NULL,
);
```

Both will fail because of the presence of `Group` which is a reserved word.

### Expected behavior
Model names and fields should be quoted

raphkim:
Hi @saltonmassally ,

Thank you for reporting this issue. We will add this item to the backlog and provide an update here when a fix is implemented.

PR #712, opened 2020-08-08
Fix (datastore) quote sql commands

saltonmassally:
*Issue #, if available:*
Datastore SQL commands fail if your model has SQLite reserved words in them. see #707

*Description of changes:*

This quotes table and column names to avoid that.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageReservedWordTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageReservedWordTest.java
@@ -1,0 +1,559 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.storage.sqlite;
+
+import androidx.core.util.ObjectsCompat;
+
+import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.ModelProvider;
+import com.amplifyframework.core.model.annotations.BelongsTo;
+import com.amplifyframework.core.model.annotations.HasMany;
+import com.amplifyframework.core.model.annotations.Index;
+import com.amplifyframework.core.model.annotations.ModelConfig;
+import com.amplifyframework.core.model.annotations.ModelField;
+import com.amplifyframework.core.model.query.Where;
+import com.amplifyframework.core.model.query.predicate.QueryField;
+import com.amplifyframework.datastore.DataStoreException;
+import com.amplifyframework.datastore.StrictMode;
+import com.amplifyframework.datastore.storage.SynchronousStorageAdapter;
+import com.amplifyframework.util.Immutable;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+
+import static com.amplifyframework.core.model.query.predicate.QueryField.field;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test that reserved SQLite words can be safely used in Model names and fields.
+ */
+public final class SQLiteStorageReservedWordTest {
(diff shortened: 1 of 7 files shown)


saltonmassally:
> Cool @saltonmassally! Is the backtick a SQL _keyword_ (`SqlKeyword`)?
> 
> If not -- we also have a utility `Wrap` in the `core` module. You could consider to extend this with a `Wrap.inBackticks(...)` utility method.
> 
> It'd be great to add a test case for this, against a real live SQL database. May need to build an instrumentation test under `aws-datastore/src/androidTest`, for that purpose.
> 
> Would love to see this come back as a PR to `amplify-android`. Thanks @saltonmassally.

Let me do that in a further commit

saltonmassally:
@jamesonwilliams I replaced quote in SqlKeyword with Wrap.inBackticks.

On the subject of having a test case for this, what exactly are we targeting not covered by existing cases? Additionally, I am still finding my way through the code so a pointer to a similar case will be helpful.

jamesonwilliams:
> @jamesonwilliams I replaced quote in SqlKeyword with Wrap.inBackticks.
> 
> On the subject of having a test case for this, what exactly are we targeting not covered by existing cases? Additionally, I am still finding my way through the code so a pointer to a similar case will be helpful.

The existing tests are passing, so that's a great sign.

We don't currently have any tests that check what you documented in #707. In other words: what happens when a model includes a field that is a SQL reserved word?

The instrumentation tests for the SQL storage layer are in [`aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite`](https://github.com/aws-amplify/amplify-android/tree/main/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite).

The test we want to run is:
1. Model contains a reserted SQL keyword.
2. A `SQLiteStorageAdapter` is initialized, using that model.
3. Assert: Table creation succeeds
4. An item is saved into the storage adapter.
5. Assert: save succeeds
6. Query for that item.
7. Assert: query returns original item that you saved in step (4).

There are two strategies to gain this coverage.

1. Add your own new test file, that does exactly this, referencing some `@Model` class you creat.
2. Recycle an existing test, by modifying an existing model class. For example, if you look through the existing SQL storage adapter tests, many of them make reference to the [`BlogOwner`](https://github.com/aws-amplify/amplify-android/blob/ba4218e47b055f3b36eede7209b8ba40db81ad62/testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/BlogOwner.java#L20) in the `testmodels` module. You could add your SQL-reserved-word-named fields into _this_ model. Then, _all_ of the existing tests would serve as code coverage for your current changes in this PR.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SqlCommandTest.java
```diff
@@ -55,7 +55,7 @@
 
     private static final String PERSON_BASE_QUERY =
             "SELECT Person.id AS Person_id, Person.age AS Person_age, Person.firstName AS Person_firstName, " +
-                    "Person.lastName AS Person_lastName FROM Person";
+                    "Person.lastName AS Person_lastName FROM `Person`";
```
raphkim, line 58: shouldn't these be
```
"SELECT `Person`.`id` AS `Person_id`, `Person`.`age` AS `Person_age`, `Person`.`firstName` AS `Person_firstName`, " +
    "`Person`.`lastName` AS `Person_lastName` FROM `Person`";
```
?

saltonmassally: fixed in the latest commit

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageReservedWordTest.java
```diff
@@ -0,0 +1,559 @@
...
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test that reserved SQLite words can be safely used in Model names and fields.
+ */
+public final class SQLiteStorageReservedWordTest {
+    private SynchronousStorageAdapter adapter;
+
+    /**
+     * Enables Android Strict Mode, to help catch common errors while using.
```
raphkim, line 57:
```suggestion
     * Enables Android Strict Mode, to help catch common errors while using
```

core/src/main/java/com/amplifyframework/util/Wrap.java
```diff
@@ -24,6 +24,20 @@
 public final class Wrap {
     private Wrap() {}
 
+    /**
+     * Returns original string wrapped with in backticks.
+     * @param original Original string to modify.
+     * @return Original string wrapped with backtick.
+     *         If original string is null or empty, it just returns the original.
```
raphkim, line 31: this description is not true for empty strings right? you can use [`Empty#check(String)`](https://github.com/aws-amplify/amplify-android/blob/b0980c7e4f8e3cb0729c8136a4e186fd628433f6/core/src/main/java/com/amplifyframework/util/Empty.java#L58-L60) utility method to return the original string if true.

saltonmassally: Good catch. Fixed in next commit

jamesonwilliams:
LGTM, thanks for the contribution @saltonmassally. You get a "Contributors" badge on Discord, now, as well as cool green lettering for your name. :-D

PR #713, opened 2020-08-10
Adding SyncType to LastSyncMetadata

rjuliano:
*Issue #, if available:*

*Description of changes:*
As part of implementing the hub events for DataStore sync (#710), we need to save sync type (BASE or DELTA) as part of the `LastSyncMetadata` system model. This will allow us to decouple the process of tracking the sync metrics from the actual sync process. 

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
@@ -38,12 +38,14 @@ public final class LastSyncMetadata implements Model {
     private final @ModelField(targetType = "ID", isRequired = true) String id;
     private final @ModelField(targetType = "String", isRequired = true) String modelClassName;
     private final @ModelField(targetType = "AWSTimestamp", isRequired = true) Long lastSyncTime;
+    private final @ModelField(targetType = "String", isRequired = true) String lastSyncType;
 
     @SuppressWarnings("checkstyle:ParameterName") // The field is named "id" in the model; keep it consistent
-    private LastSyncMetadata(String id, String modelClassName, Long lastSyncTime) {
+    private LastSyncMetadata(String id, String modelClassName, Long lastSyncTime, SyncType syncType) {
         this.id = id;
         this.modelClassName = modelClassName;
         this.lastSyncTime = lastSyncTime;
+        this.lastSyncType = syncType.name();
     }
 
     /**
@@ -53,9 +55,11 @@ public final class LastSyncMetadata implements Model {
      * @param lastSyncTime Last time it was synced
      * @return {@link LastSyncMetadata} for the model class
      */
-    static <T extends Model> LastSyncMetadata lastSyncedAt(@NonNull Class<T> modelClass, long lastSyncTime) {
+    static <T extends Model> LastSyncMetadata lastSyncedAt(@NonNull Class<T> modelClass,
+                                                           @Nullable long lastSyncTime,
+                                                           @NonNull SyncType syncType) {
         Objects.requireNonNull(modelClass);
-        return create(modelClass, lastSyncTime);
+        return create(modelClass, lastSyncTime, syncType);
     }
 
     /**
@@ -67,22 +71,23 @@ public final class LastSyncMetadata implements Model {
      */
     static <T extends Model> LastSyncMetadata neverSynced(@NonNull Class<T> modelClass) {
         Objects.requireNonNull(modelClass);
-        return create(modelClass, null);
+        return create(modelClass, null, SyncType.BASE);
     }
 
     /**
      * Creates an {@link LastSyncMetadata} for the provided model class.
      * @param modelClass Class of model for which metadata pertains
      * @param lastSyncTime Time of last sync; null, if never.
+     * @param syncType The type of sync (FULL or DELTA).
      * @param <T> Type of model
      * @return {@link LastSyncMetadata}
      */
     @SuppressWarnings("WeakerAccess")
     static <T extends Model> LastSyncMetadata create(
-            @NonNull Class<T> modelClass, @Nullable Long lastSyncTime) {
+            @NonNull Class<T> modelClass, @Nullable Long lastSyncTime, @NonNull SyncType syncType) {
         Objects.requireNonNull(modelClass);
         String modelClassName = modelClass.getSimpleName();
-        return new LastSyncMetadata(hash(modelClassName), modelClassName, lastSyncTime);
+        return new LastSyncMetadata(hash(modelClassName), modelClassName, lastSyncTime, syncType);
     }
 
     @NonNull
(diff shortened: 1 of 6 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -53,9 +55,11 @@ private LastSyncMetadata(String id, String modelClassName, Long lastSyncTime) {
      * @param lastSyncTime Last time it was synced
      * @return {@link LastSyncMetadata} for the model class
      */
-    static <T extends Model> LastSyncMetadata lastSyncedAt(@NonNull Class<T> modelClass, long lastSyncTime) {
+    static <T extends Model> LastSyncMetadata lastSyncedAt(@NonNull Class<T> modelClass,
+                                                           @Nullable long lastSyncTime,
+                                                           @NonNull SyncType syncType) {
```
jamesonwilliams, line 60: To cut down on the number of arguments, and since there is already a `create(...)`, how about we replace `lastSyncedAt` with:

```java
static <T extends Model> LastSyncMetadata baseSyncedAt(@NonNull Class<T> modelClass, long lastBaseSyncTime) {
    ....
}

static <T extends Model> LastSyncMetadata deltaSyncedAt(@NonNull Class<T> modelClass, long lastDeltaSyncTime) {
    ....
}

static <T extends Model> LastSyncMetadata neverSynced(@NonNull Class<T> modelClass) {
    ...
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -38,12 +38,14 @@
     private final @ModelField(targetType = "ID", isRequired = true) String id;
     private final @ModelField(targetType = "String", isRequired = true) String modelClassName;
     private final @ModelField(targetType = "AWSTimestamp", isRequired = true) Long lastSyncTime;
+    private final @ModelField(targetType = "String", isRequired = true) String lastSyncType;
```
jamesonwilliams, line 41: Make sure to update `equals()`, `hashCode()`, and `toString()`, as well.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -102,8 +103,17 @@ Completable hydrate() {
...
+        // Default this to BASE since if there is no LastSyncMetadata record for this
+        // model, a BASE sync will be performed.
+        AtomicReference<SyncType> syncType = new AtomicReference<>(SyncType.BASE);
+
         return syncTimeRegistry.lookupLastSyncTime(modelClass)
             .map(this::filterOutOldSyncTimes)
+            .map(lastSyncTime -> {
+                long syncIntervalMs = dataStoreConfigurationProvider.getConfiguration().getSyncIntervalMs();
+                syncType.set(SyncType.fromSyncTimeAndThreshold(lastSyncTime, syncIntervalMs));
+                return lastSyncTime;
```
jamesonwilliams, line 115: Seems like we should just keep the return type from this `.map(...)` as `LastSyncMetadata`, here. If you do that, you won't need the `AtomicReference`. And, you can access the component data in the `syncModel(...)` call, below, by pulling out `getModelClass()` and `getLastSyncTime()`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
```diff
@@ -55,9 +56,11 @@
         });
     }
 
-    <T extends Model> Completable saveLastSyncTime(@NonNull Class<T> modelClazz, SyncTime syncTime) {
+    <T extends Model> Completable saveLastSyncTime(@NonNull Class<T> modelClazz,
```
jamesonwilliams, line 59: Possibly, consider the same `saveBaseSyncTime(...)` and `saveDeltaSyncTime(...)` to keep the arg count down? (As with the factories on `LastSyncMetadata`.)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
rjuliano: I refactored it a bit so we don't need that AtomicReference. Let me know if it makes sense. I did keep the return type of the map as `SyncTime` because we don't have the storage adapter available to get the `LastSyncMetadata` record.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
rjuliano: Followed the same pattern here as well.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -67,22 +83,23 @@ private LastSyncMetadata(String id, String modelClassName, Long lastSyncTime) {
      */
     static <T extends Model> LastSyncMetadata neverSynced(@NonNull Class<T> modelClass) {
         Objects.requireNonNull(modelClass);
-        return create(modelClass, null);
+        return create(modelClass, null, SyncType.BASE);
```
jamesonwilliams, line 86: Are these two `LastSyncMetadata` `equals()`? Should they have the same `hashCode()`?

1. Person.class, null, SyncType.BASE
2. Person.class, null, SyncType.DELTA

Might be an edge case / not super relevant.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
```diff
@@ -55,9 +56,26 @@
         });
     }
 
-    <T extends Model> Completable saveLastSyncTime(@NonNull Class<T> modelClazz, SyncTime syncTime) {
+    <T extends Model> Completable saveLastDeltaSyncTime(@NonNull Class<T> modelClazz,
+                                                   @Nullable SyncTime syncTime) {
```
jamesonwilliams, line 60: nit: alignment -- i'd either align arg, or wrap and +8 indent from method head

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
rjuliano: Interesting...Right now they're treated as not equals, which I think is OK. I ran the following test to verify that:

```
        LastSyncMetadata data1 = LastSyncMetadata.create(BlogOwner.class, null, SyncType.BASE);
        LastSyncMetadata data2 = LastSyncMetadata.create(BlogOwner.class, null, SyncType.DELTA);
        assertNotEquals(data1, data2);
        assertNotEquals(data1.hashCode(), data2.hashCode());
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
rjuliano: Ugh...that's ugly. Just fixed it.

PR #716, opened 2020-08-11
feature(datastore) Trigger network-related hub events

rjuliano:
*Description of changes:*
Trigger the following Hub events:

- **subscriptionsEstablished**: triggered from the `SubscriptionProcessor` after all subscriptions start.
- **ready**: triggered from the `Orchestrator` when it reaches its target state. This seemed to be the most appropriate place for it because it will trigger the event after `StorageObserver` starts (when no API is configured) or after all the remote sync processes start (if API is configured)
- **networkStatus**: Triggered when the DataStore plugin receives a Hub event on the API channel indicating network connectivity changes (more below).

In order to trigger a meaningful **networkStatus** event from DataStore, I'm leveraging [connectivity events](https://square.github.io/okhttp/events/#events-with-failures) triggered by the `OkHttpClient`. This is done by implementing an [EventListener](https://square.github.io/okhttp/events/#eventlistener) and triggering a `HubEvent` on the API channel when the listener changes status.

It might make sense to look at the changes by module in the following order:
- core
- api
- datastore


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -732,4 +748,43 @@ public int hashCode() {
             return result;
         }
     }
+
+    /**
+     * This class implements OkHttp's {@link EventListener}. Its main purpose
+     * is to listen to network-related events reported by the http client and trigger
+     * a Hub event if necessary.
+     */
+    private final class ApiConnectionEventListener extends EventListener {
```
jamesonwilliams, line 757:
```suggestion
    private static final class ApiConnectionEventListener extends EventListener {
```

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -732,4 +748,43 @@ public int hashCode() {
...
+        private AtomicReference<ApiEndpointStatus> currentNetworkStatus;
```
jamesonwilliams, line 758:
```suggestion
        private final AtomicReference<ApiEndpointStatus> currentNetworkStatus;
```

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -732,4 +748,43 @@ public int hashCode() {
...
+        @Override
+        public void connectionAcquired(@NotNull Call call, @NotNull Connection connection) {
+            super.connectionAcquired(call, connection);
+            triggerEventIfStatusChanged(currentNetworkStatus.getAndSet(ApiEndpointStatus.REACHABLE));
+        }
+
+        private void triggerEventIfStatusChanged(@NonNull ApiEndpointStatus previousStatus) {
+            if (!previousStatus.equals(currentNetworkStatus.get())) {
+                ApiEndpointStatusChangeEvent eventData =
+                    new ApiEndpointStatusChangeEvent(currentNetworkStatus.get(), previousStatus);
```
jamesonwilliams, line 783: You call `currentNetworkStatus.get()` twice here, without any synchronization guards, it appears. What happens if the value changes between the two calls?

1. Do you need to add syncrhonization primitives?
2. You can defend by making one `.get()`, caching the result, and then using the cached result in both locations.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -732,4 +748,43 @@ public int hashCode() {
             return result;
         }
     }
+
+    /**
```
jamesonwilliams, line 752: In a microservice I did a couple years back, I found a pattern where you can add _behaviors_ to an `enum`. In that case, I would do like `Metrics.SOME_SERVICE_CRAP_HAPPENED.report()`.

But, you could use a similar mechanism here, potentially. Maybe something like:
```java
ApiEndpointStatus.UNKNOWN.announce();
```
And then have the hub broadcast logic contained inside the enum itself.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -246,6 +255,13 @@ public void graphQlMutationGetsResponse() throws JSONException, ApiException {
 
         // Assert that the expected response was received
         assertEquals(expectedName, actualResponse.getData().getName());
+
+        // Verify that the expected hub event fired.
+        List<HubEvent<?>> events = networkStatusObserver.await();
```
jamesonwilliams, line 260: I used to have a convenient method on the `HubAccumulator` that allowed you to just get the first one. like `takeOne()`, it was called. Should we include an `awaitFirst()` to simplify the `List` return type? Maybe not. Just sayin'.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -246,6 +255,13 @@ public void graphQlMutationGetsResponse() throws JSONException, ApiException {
...
+        assertTrue(events.size() > 0);
+        assertTrue(events.get(0).getData() instanceof ApiEndpointStatusChangeEvent);
+        ApiEndpointStatusChangeEvent eventData = (ApiEndpointStatusChangeEvent) events.get(0).getData();
```
jamesonwilliams, line 263: **Aha!**

We have this unfortunate situation with Hub right now, where we have to cast the data type.

I wonder if instead we can include a factory on the event structure, like:
```java
ApiEndpointStatusChangedEvent event = ApiEndpointStatusChangedEvent.fromEvent(events.get(0));
```
Or, very similar:
```java
ApiEndpointStatuChangedEvent event = ApiEndpointStatusChangedEvent.from(events.get(0).getData());
```

This would at least pull the cast into our own code where we can do it safely, without the customer needing to worry about it.

Impl might start off looking like an `equals(....)` almost:
```
public static ApiEndpointStatusChangedEvent from(Object someBlob) {
    if (!(someBlob instanceof ApiEndpointStatuChangedEvent)) {
        throw BadNewsBearsError("Shucks!");
    }
    ... etc. validate fields,
    return new ApiEndpointStatusChangeEvent(..., ...);
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -98,6 +103,20 @@ private AWSDataStorePlugin(
...
+            return ApiChannelEventName.API_ENDPOINT_STATUS_CHANGED.name().equals(hubEvent.getName());
+        }, hubEvent -> {
+                if (hubEvent.getData() instanceof ApiEndpointStatusChangeEvent) {
+                    ApiEndpointStatusChangeEvent eventData = (ApiEndpointStatusChangeEvent) hubEvent.getData();
+                    boolean isActive = ApiEndpointStatus.REACHABLE.equals(eventData.getCurrentStatus());
+                    Amplify.Hub.publish(HubChannel.DATASTORE,
+                        HubEvent.create(DataStoreChannelEventName.NETWORK_STATUS, new NetworkStatusEvent(isActive))
+                    );
+                }
+            });
```
jamesonwilliams, line 119: Can you shuffle this off to a helper class, and then do a thin integration to the plugin with like 
```
NetworkStatusMonitor.enable();
```
Or something like that?

Also, does this _need_ to be int the constructor? Or could it go down into the `configure(...)` or `initialize(...)` hooks? Which is the most appropriate lifecycle hook for this?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -132,14 +139,27 @@ public void configureAndInitializeInApiMode() throws JSONException, AmplifyExcep
...
                 .start();
+        HubAccumulator dataStoreReadyObserver =
+            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.READY, 1)
+                .start();
+        HubAccumulator subscriptionsEstablishedObserver =
+            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.SUBSCRIPTIONS_ESTABLISHED, 1)
+                .start();
+        HubAccumulator networkStatusObserver =
+            HubAccumulator.create(HubChannel.DATASTORE, DataStoreChannelEventName.NETWORK_STATUS, 1)
+                .start();
```
jamesonwilliams, line 150: You've got a mechanism repeated four times -- it's a signal to make it a data problem, instead:

```java
List<DataStoreChannelEventName> expectedEventNames = Arrays.asList(
    DataStoreChannelEventName.REMOTE_SYNC_STARTED,
    DataStoreChannelEventName.READY,
    DataStoreChannelEventName.SUBSCRIPTIONS_ESTABLISHED,
    DataStoreChannelEventName.NETWORK_STATUS
);
Map<DataStoreChannelEventName, HubAccumulator> accums = new HashMap<>();
for (DataStoreChannelEventName eventName : eventNames) {
    HubAccumulator accum = HubAccumulator.create(HubChannel.STORAGE, eventName, 1);
    accums.add(accum);
    accum.start();
}
```

Or actually -- for extra credit, we could make a `MultipleHubAccumulator` utility, to get some of this boilerplate out of the test class itself.

It could accept a ` LinkedHashMap` (to mandate and ordering in which to call `await()`. And have a usage pattenr like this:
```java
// It's like a list of events, in order. Allows duplicates by Integer value > 1
// HubEvent1 --> 1 time
// HubEvent2 --> 1 time
// HubEvent1 --> 5 times
// etc.
LinkedHashMap<HubEvent<?>, Integer> tape = new LinkedHashMap<>();
MultiHubAccumulator accum = MultiHubAccumulator.create(tape);

accum.start();

accum.await(); // Waits for the entire "tape" to play.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -306,8 +326,16 @@ private ApiCategory mockApiCategoryWithGraphQlApi() throws AmplifyException {
         when(mockApiPlugin.getPluginKey()).thenReturn(MOCK_API_PLUGIN_NAME);
         when(mockApiPlugin.getCategoryType()).thenReturn(CategoryType.API);
 
+        ApiEndpointStatusChangeEvent eventData =
+            new ApiEndpointStatusChangeEvent(ApiEndpointStatusChangeEvent.ApiEndpointStatus.REACHABLE,
```
jamesonwilliams, line 330: You could play a game here like:
```java
ApiEndpointStatuChangeEvent eventData =
    ApiEndpointStatus.REACHABLE.changeTo(ApiEndpiontStatus.UNKNOWN);
```

Basically, add afactory method to the enum, that has the ability to generate the change events. It knows the "from" based on what is being invoked, and it knows the "to", if we pass that as an argument.

core/src/main/java/com/amplifyframework/api/ApiChannelEventName.java
```diff
@@ -0,0 +1,32 @@
...
+import com.amplifyframework.hub.HubEvent;
+
+/**
+ * An enumeration of the names of events relating the the {@link ApiCategory},
+ * that are published via {@link HubCategory#publish(HubChannel, HubEvent)} on the
+ * {@link HubChannel#API} channel.
+ */
+public enum ApiChannelEventName {
+    /**
+     * The status of the remote API endpoint has changed.
```
jamesonwilliams, line 29: What kinds of status changes are these? Like, we are able to reach it or not?

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -52,7 +52,23 @@
     /**
      * The remote synchonization processes stopped.
      */
-    REMOTE_SYNC_STOPPED("remote_sync_stopped");
+    REMOTE_SYNC_STOPPED("remote_sync_stopped"),
+
+    /**
+     * Notifies if the network is active or not.
```
jamesonwilliams, line 58:
```suggestion
     * Indicates that the network is active or not.
```

core/src/main/java/com/amplifyframework/datastore/events/NetworkStatusEvent.java
```diff
@@ -0,0 +1,68 @@
...
+     * @return The value of the active field.
+     */
+    public boolean getActive() {
+        return active;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Boolean.valueOf(active).hashCode();
+        return result;
```
jamesonwilliams, line 47:
```suggestion
        return Boolean.valueOf(active).hashCode();
```

core/src/test/java/com/amplifyframework/hub/HubEventDataObjectsTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        assertNotEquals(status1, status2);
+        assertEquals(status1, status3);
+
+        assertNotEquals(status1.hashCode(), status2.hashCode());
+        assertEquals(status1.hashCode(), status3.hashCode());
+        assertFalse(status1 == status3);
+
+        assertNotEquals(status1.toString(), status2.toString());
+        assertEquals(status1.toString(), status3.toString());
```
jamesonwilliams, line 53: Personally I agree with you that these are useful types of tests. However, I recently changed _how_ I do this test, of the `hashCode()`. Rarely will anyone ever call `hashCode()`. Perhaps they will if they are writing _their own_, higher level `hashCode()`, and include one of these types in it. But more often, they'll probably try to use it in a collection. So like:
```java
PossibleSaneHashObject one = new PossiblySaneHashObject();
PossibleSaneHashObject two = new PossiblySaneHashObject();
Set<PossiblySaneHashObject> set = new HashSet<>();
set.add(one);
set.add(two);
assertEquals(1, set.size());
```

This is straight-up pedantry at this point, feel free to ignore 😆

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
rjuliano: Good call. I'm going to implement the latter.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
rjuliano: That's a good idea. Just added a commit with that change.

core/src/test/java/com/amplifyframework/hub/HubEventDataObjectsTest.java
rjuliano: It makes sense...especially as we start adding more events down the road. That being said, the test itself seems pretty generic, so I created a class under tesutils called `ObjectValidatorUtils`. Right now it has one method that takes 3 objects: object1, object2, object3. The objects must be constructed such that:

object1.equals(object3), but not ==.
object2 is different from object1 and object3.

Let me know if this makes sense. Naming stuff is definitely not my superpower 🤣

core/src/main/java/com/amplifyframework/api/ApiChannelEventName.java
rjuliano: I did a little word smithing there. Let me know if it looks OK.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -44,8 +45,12 @@
 import com.amplifyframework.core.model.ModelOperation;
 import com.amplifyframework.util.UserAgent;
 
+import org.jetbrains.annotations.NotNull;
```
jamesonwilliams, line 48: Oops, we want the `androidx` one here, I believe.

OkHttp uses the JetBrains annotations on its APIs, and so this probably got added when you implemented an OkHttp interface, right?

For better or worse, I have been _swapping_ such annotations, as I encounter them, into the AndroidX ones.

OkHttp is itself a vanilla Java library, so they _can't_ depend on the AndroidX annotations.

I feel emotionally that these annotations ought to have better support for Android Studio / Android Lint / Kotlin interop. However, I don't know it any scientifically backed objective way. 😆

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
```diff
@@ -44,12 +44,14 @@
...
     private final CopyOnWriteArrayList<HubEvent<?>> events;
     private final AtomicReference<SubscriptionToken> token;
+    private final CountDownLatch firstItemLatch;
 
     private HubAccumulator(
             @NonNull HubChannel channel, @NonNull HubEventFilter filter, int quantity) {
         this.channel = channel;
         this.filter = filter;
         this.latch = new CountDownLatch(quantity);
+        this.firstItemLatch = new CountDownLatch(1);
```
jamesonwilliams, line 54: I was hoping we could just reuse the `latch` -- is it possible? I don't think it makes sense to have multiple ones.

testutils/src/main/java/com/amplifyframework/testutils/ObjectValidatorUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Utility class that can be used to test core
+ * Object behavior of a class that implements its own
+ * equals, hashCode and toString methods.
+ */
+public final class ObjectValidatorUtils {
```
jamesonwilliams, line 28: Okay, cool. This looks like a good general purpose testutil.

Gripes about naming (classic jhwill@):

- Let's drop the `Utils`
- From the method it contains, maybe `EqualsToStringHashValidator` is a reasonable name? It's kinda long, but it's very clear and narrow in its utility:

testutils/src/main/java/com/amplifyframework/testutils/ObjectValidatorUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Function to make some basic assertions as it related to the
+     * proper implementation of equals, hashCode and toString.
+     * @param subject1 An instance of T
+     * @param suject2 An instance of T that is not equal to subject1.
+     * @param subject3ThatEqualsSubject1 An instance of T such that when passed as
+     *                                   a parameter to subject1.equals, will return true.
+     * @param <T> The type being tested.
+     */
+    public static <T> void assertCoreObjectBehavior(T subject1, T suject2, T subject3ThatEqualsSubject1) {
```
jamesonwilliams, line 44: Assuming this is contained `EqualsToStringHashValidator`, then maybe:

```suggestion
    public static <T> void validate(T reference, T equalsReference, T notEqualsReference) {
```

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
rjuliano: Oh...you know what? We don't need that. Just need to check if `latch.getCount() == quantity - 1`. I'll change that.

testutils/src/main/java/com/amplifyframework/testutils/ObjectValidatorUtils.java
rjuliano: Hahaha...I knew "Utils" was going to get some attention. I'm good with that and the method name as well. LIke I said before, I would have never made it as a marketing person.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
rjuliano: I created a couple of factory methods in each of the classes used as payloads for hub events. There's definitely some work to do in the Hub in general to make the approach cleaner. As it stands, the `subscribe` method definition takes a `HubSubscriber` instance which is currently just taking `HubEvent<?>` as a parameter which kind of prevents a cleaner approach to the problem. 

TLDR: a cleaner approach would probably require changes to the Hub category definition.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: I tried to clean up a bit, by moving the casting logic into the event types. That being said, those methods throw exceptions so I ended up with an ugly try/catch here :( 

Maybe I'll just encapsulate the whole subscribe block somewhere. 

I think I can probably move this out of the constructor. It just has to happen before we trigger the sync.

rjuliano: Ok...I move it out to configure and did a variation of what you suggested with regards to encapsulating that block of code to listen and trigger events.

testutils/src/main/java/com/amplifyframework/testutils/ObjectValidatorUtils.java
rjuliano: Refactored this logic into its own class in testutils.

testutils/src/main/java/com/amplifyframework/testutils/HubAccumulator.java
rjuliano: This is done.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
rjuliano: Fixed.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -119,6 +122,8 @@ synchronized void startSubscriptions() {
             return;
         }
         if (subscriptionsStarted) {
+            Amplify.Hub.publish(HubChannel.DATASTORE,
```
jamesonwilliams, line 125: We have full liberty to add a new `publish(...)` method, if you would like to, as well!

Maybe like:
```java
<E extends Enum<E>> void publish(HubChannel channel, Enum<E> eventName) {
    publish(channel, HubEvent.create(eventName));
}
```
A small optimization to reduce some boiler plate. I've already baked in some `Enum<E>` type mumbo-jumbo elsewhere around the Hub APIs.

core/src/main/java/com/amplifyframework/api/ApiChannelEventName.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api;
```
jamesonwilliams, line 16: Should this go into `api.events`?

core/src/main/java/com/amplifyframework/api/events/ApiEndpointStatusChangeEvent.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public String toString() {
+        return "ApiEndpointStatusChangeEvent{" +
+            "currentStatus=" + currentStatus +
+            ", previousStatus=" + previousStatus +
+            "}";
+    }
+
+    @Override
+    public HubEvent<ApiEndpointStatusChangeEvent> toHubEvent() {
```
jamesonwilliams, line 95: Ooooooh :-D

core/src/main/java/com/amplifyframework/api/events/ApiEndpointStatusChangeEvent.java
```diff
@@ -0,0 +1,144 @@
...
+     * {@link HubEvent} object as an instance of {@link ApiEndpointStatusChangeEvent}.
+     * @param hubEvent An instance of {@link HubEvent}
+     * @return An instance of {@link ApiEndpointStatusChangeEvent}.
+     * @throws AmplifyException If unable to cast to the target type.
+     */
+    public static ApiEndpointStatusChangeEvent from(HubEvent<?> hubEvent) throws AmplifyException {
+        if (hubEvent.getData() instanceof ApiEndpointStatusChangeEvent) {
+            return (ApiEndpointStatusChangeEvent) hubEvent.getData();
+        }
+        String expectedClassName = ApiEndpointStatusChangeEvent.class.getName();
```
jamesonwilliams, line 110: To keep the noise down for the reader, maybe:
```suggestion
        String expectedClassName = ApiEndpointStatusChangeEvent.class.getSimpleName();
```
?

core/src/main/java/com/amplifyframework/datastore/events/NetworkStatusEvent.java
```diff
@@ -0,0 +1,127 @@
...
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.datastore.DataStoreChannelEventName;
+import com.amplifyframework.hub.HubChannel;
+import com.amplifyframework.hub.HubEvent;
+import com.amplifyframework.logging.Logger;
+
+/**
+ * Event payload for the {@link com.amplifyframework.datastore.DataStoreChannelEventName#NETWORK_STATUS} event.
+ */
+public final class NetworkStatusEvent implements HubEvent.Data<NetworkStatusEvent> {
```
jamesonwilliams, line 35: ooooh :-D

core/src/main/java/com/amplifyframework/datastore/events/NetworkStatusEvent.java
```diff
@@ -0,0 +1,127 @@
...
+    public static NetworkStatusEvent from(ApiEndpointStatusChangeEvent apiEvent) {
+        boolean isActive = ApiEndpointStatus.REACHABLE.equals(apiEvent.getCurrentStatus());
+        return new NetworkStatusEvent(isActive);
+    }
+
+    /**
+     * Start listening for API_ENDPOINT_STATUS_CHANGED to be emitted by the API and trigger
+     * the DataStore NETWORK_STATUS event accordingly.
+     */
+    public static void monitor() {
```
jamesonwilliams, line 115: Does this belong in the `NetworkStatusEvent`, or in some external `NetworkStatusMontior.start()` utility class? I don't feel strongly either way I guess. If push came to shove I think I probably would have written it the second way, but this is also cool, and it is well coupled.

core/src/main/java/com/amplifyframework/hub/HubEvent.java
```diff
@@ -125,6 +127,14 @@ public UUID getId() {
         return uuid;
     }
 
+    /**
+     * Publish the instance of the {@link HubEvent} to Amplify Hub.
+     * @param channel The channel to publish the event to.
+     */
+    public void publish(@NonNull HubChannel channel) {
+        Amplify.Hub.publish(channel, this);
```
jamesonwilliams, line 135: Eh, I feel iffy about binding `Amplifly`, here. I like the intention of this utility method though. Could you instead have it like:
```java
public void publish(@NonNull HubChannel toChannel, @NonNull HubCategoryBehavior usignHubInstance) {
    usingHubInstance.publish(channel, this);
}
```
?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
rjuliano: Oh...I will do that then. I think that will go a long way in cleaning some of this stuff up.

core/src/main/java/com/amplifyframework/api/ApiChannelEventName.java
rjuliano: Good call. Moved it. Was going to do the same for DataStore, but it affects too many files so I'll do it as a follow up PR.

core/src/main/java/com/amplifyframework/hub/HubEvent.java
rjuliano: That does feel kind of out of place here. I'm going to move this.

core/src/main/java/com/amplifyframework/datastore/events/NetworkStatusEvent.java
rjuliano: Yeah...you're right. Triggering the event from core seems weird. I'm going to move it to its own utility class in the datastore module.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
rjuliano: I did some refactoring for this to make the calling code a bit cleaner by adding a factory method to `HubSubscriber`. Calling code then looks like this: 

```java
HubSubscriber subscriber = HubSubscriber.<ApiEndpointStatusChangeEvent>create(eventData -> {
	NetworkStatusEvent.from(eventData).toHubEvent().publish(HubChannel.DATASTORE, Amplify.Hub);
};

HubFilter filter = hubEvent -> {
	return ApiChannelEventName.API_ENDPOINT_STATUS_CHANGED.name().equals(hubEvent.getName());
};

Amplify.Hub.subscribe(HubChannel.API, filter, subscriber);
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: I'll handle this in a separate PR :)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
rjuliano: Well, turns out it's not as much on the publishing side as it is on the subscriber side. The `HubSubscriber` interface currently doesn't have a method that takes in a generic type. Modifying the existing one would probably make this a breaking change. Adding a new method to that interface would then force us to change every calling site from lambda-style to an anonymous object declaration. Neither of those sounded like a good plan.

Instead, I went with adding a factory method in `HubSubscriber` which makes the calling code a little cleaner.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
rjuliano: I think this one ended up getting addressed by some of the refactoring I did.

Issue #595, opened 2020-06-24
"E/SQLiteLog: (1) no such table: " occurs when I use "Amplify.DataStore.save()"


Shumpei0706:
Hi, I am facing an error again and again when I change my GraphQL schema very slightly from tutorial:

>  "E/SQLiteLog: (1) no such table: "

I only changed the name of the model from `Todo` to `Tod`. I tried to check many cases, but even this small change can cause this error.

Am I supposed to register the schema when it's changed, in addition to running `modelgen`?

jamesonwilliams:
> Do I supposed to register the schema when it's changed in addition to run "modelgen"?

Hi @Shumpei0706 -- yes, if you change the structure/names of the models, you should re-run code generation, and push your changes to the backend, as well.

woutermilants:
I'm having the same issue. 
So first run modelgen and then ? 

My schema is created in AWS, but locally it still fails

AntonPolkanov:
I am getting the same error (_"E/SQLiteLog: (1) no such table"_), but it slightly defferent use case.

I did the getting started tutorial https://docs.amplify.aws/start/q/integration/android and everything worked fine. Then I 
1. added a new type Comment to the gql schema
2. ran modelgen
3. ran the application 
and got "E/SQLiteLog: (1) no such table: Comment" error.

Current schema: 
```
enum Priority {
  LOW
  NORMAL
  HIGH
}

type Comment @model {
  id: ID!
  content: String
}

type Todo @model {
  id: ID!
  name: String!
  priority: Priority
  description: String
}
```
If I delete the Comment type from the schema and do steps (2) and (3) the error goes away.

jamesonwilliams:
The intended behavior of the system is as follows.

__On the development machine, at build time:__ Every set of model schema gets a unique hash. If any of the model schema change, the hash changes.

__On the Android device, at run time:__ If there is no hash in the database, the current model schema hash is stored. If there is a hash in the database, it is compared against the current model schema hash. If they are different, the old database is dropped, and a new one is created, to accommodate the new schema. If the hashes are the same, nothing happens.

It sounds to me like there is a bug, and the system is not correctly re-establishing the database, when the model hash changes.

[...]

PR #718, opened 2020-08-11
fix(aws-datastore): remove insert statement cache

raphkim:
*Issue #, if available:*
#595 

*Description of changes:*
Removed caching logic for insert statements:

When the `SQLiteStorageAdapter` was first written, small optimization logic was added to cache insert statements during initialization. These statements were to be re-used whenever new entries were being entered in a table. Currently, no other types of sqlite statements are cached like this, and are generated on the spot whenever they are called. 

When the database was updated with new tables, it caused a chicken-and-egg problem between generating insert statements and updating the database. To prevent this problem, the original optimization logic is removed in this PR so that insert statements are prepared on the spot like the other types of sqlite operations. 

No explicit tests were written, but I was able to confirm the fix with @rjuliano 's suggestion for [reproducing the bug](https://github.com/aws-amplify/amplify-android/issues/595#issuecomment-662651419).

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
@@ -605,21 +605,6 @@ public final class SQLiteStorageAdapter implements LocalStorageAdapter {
         return new CreateSqlCommands(createTableCommands, createIndexCommands);
     }
 
-    private Map<String, SqlCommand> getInsertSqlPreparedStatements() {
-        final Map<String, SqlCommand> modifiableMap = new HashMap<>();
-        final Set<Map.Entry<String, ModelSchema>> modelSchemaEntrySet =
-                modelSchemaRegistry.getModelSchemaMap().entrySet();
-        for (final Map.Entry<String, ModelSchema> entry : modelSchemaEntrySet) {
-            final String tableName = entry.getKey();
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -217,17 +212,6 @@ public synchronized void initialize(
...
-                /*
-                 * Create INSERT INTO TABLE_NAME statements for all SQL tables
-                 * and compile them and store in an in-memory map. Later, when a
-                 * {@link #save(T, Consumer, Consumer)} operation needs to insert
-                 * an object (sql rows) into the database, it can bind the input
-                 * values with the prepared insert statement.
-                 *
-                 * This is done to improve performance of database write operations.
-                 */
-                this.insertSqlPreparedStatements = getInsertSqlPreparedStatements();
```
richardmcclellan, line 229: Can you delete the `getInsertSqlPreparedStatements()` method itself now too?

raphkim: Ah yes good call!

raphkim:
> Well that's a nice and simple fix :)
> 
> Would it make sense to cherry pick @rjuliano's [unit test change](https://github.com/aws-amplify/amplify-android/commit/959dc270006404f95a0a6f0d9ab6629def1d2abb) into this PR as well to verify your fix?

Included it!

PR #719, opened 2020-08-11
Display environment info on developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Displays environment information (Amplify plugin information and developer environment information) on the developer menu. Note that the developer environment information currently contains only simulated values. Also makes the following changes:

- Increases the height of the developer menu.
- Changes the message displayed on the file issue screen and increases the font size of this message.
- Reduces the minimum character count requirement for the issue description on the file issue screen.
- Adds environment information to the issue body generated when the "Copy to Clipboard" or "File Issue" button is pressed on the file issue screen.

Environment information screen (note that the PersistentLogStoragePlugin is part of core and inside my sample app's build.gradle, the version for the dependency on the core package is set to main):
Dev Menu Env Info Screen

File issue screen after pressing the "File Issue" button on the main screen of the developer menu:
Dev Menu File Issue Modified

When the user clicks the "File Issue" button on the file issue screen, they are taken to the new issue page on GitHub (after logging in, if necessary). The issue body on GitHub looks as follows when clicking the "Preview" button on the GitHub new issue page:
Dev Menu GitHub File New Issue

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -71,10 +71,7 @@ public final class Amplify {
     @SuppressWarnings("checkstyle:all") public static final DataStoreCategory DataStore = new DataStoreCategory();
     @SuppressWarnings("checkstyle:all") public static final PredictionsCategory Predictions = new PredictionsCategory();
 
-    /**
-     * A map from each type of category to an entry point for that category.
-     */
-    public static final Map<CategoryType, Category<? extends Plugin<?>>> CATEGORIES = buildCategoriesMap();
+    private static final LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> CATEGORIES = buildCategoriesMap();
 
     // Used as a synchronization locking object. Set to true once configure() is complete.
     private static final AtomicBoolean CONFIGURATION_LOCK = new AtomicBoolean(false);
@@ -89,9 +86,9 @@ public final class Amplify {
         throw new UnsupportedOperationException("No instances allowed.");
     }
 
-    // The fact that this is a LinkedHashMap is important.
+    // The fact that this is a LinkedHashMap, and not a Map, is important.
     // We are relying on the ordering of this data-structure, for configuration.
-    private static Map<CategoryType, Category<? extends Plugin<?>>> buildCategoriesMap() {
+    private static LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> buildCategoriesMap() {
         final LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> categories = new LinkedHashMap<>();
         categories.put(CategoryType.ANALYTICS, Analytics);
         categories.put(CategoryType.API, API);
@@ -101,7 +98,15 @@ public final class Amplify {
         categories.put(CategoryType.HUB, Hub);
         categories.put(CategoryType.DATASTORE, DataStore);
         categories.put(CategoryType.PREDICTIONS, Predictions);
-        return Immutable.of(categories);
+        return categories;
+    }
+
+    /**
+     * Returns an unordered map from each type of category to an entry point for that category.
+     * @return a Map from CategoryType to Category.
+     */
+    public static Map<CategoryType, Category<? extends Plugin<?>>> getCategoriesMap() {
+        return Immutable.of(CATEGORIES);
     }
 
     /**
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -33,8 +33,10 @@ import com.amplifyframework.logging.LoggingCategory;
 import com.amplifyframework.predictions.PredictionsCategory;
 import com.amplifyframework.storage.StorageCategory;
 import com.amplifyframework.util.Empty;
+import com.amplifyframework.util.Immutable;
 
 import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -69,7 +71,10 @@ public final class Amplify {
     @SuppressWarnings("checkstyle:all") public static final DataStoreCategory DataStore = new DataStoreCategory();
     @SuppressWarnings("checkstyle:all") public static final PredictionsCategory Predictions = new PredictionsCategory();
 
-    private static final LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> CATEGORIES = buildCategoriesMap();
+    /**
+     * A map from each type of category to an entry point for that category.
+     */
+    public static final Map<CategoryType, Category<? extends Plugin<?>>> CATEGORIES = buildCategoriesMap();
 
     // Used as a synchronization locking object. Set to true once configure() is complete.
     private static final AtomicBoolean CONFIGURATION_LOCK = new AtomicBoolean(false);
@@ -84,9 +89,9 @@ public final class Amplify {
         throw new UnsupportedOperationException("No instances allowed.");
     }
 
-    // The fact that this is a LinkedHashMap, and not a Map, is important.
+    // The fact that this is a LinkedHashMap is important.
     // We are relying on the ordering of this data-structure, for configuration.
-    private static LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> buildCategoriesMap() {
+    private static Map<CategoryType, Category<? extends Plugin<?>>> buildCategoriesMap() {
         final LinkedHashMap<CategoryType, Category<? extends Plugin<?>>> categories = new LinkedHashMap<>();
         categories.put(CategoryType.ANALYTICS, Analytics);
         categories.put(CategoryType.API, API);
@@ -96,7 +101,7 @@ public final class Amplify {
         categories.put(CategoryType.HUB, Hub);
         categories.put(CategoryType.DATASTORE, DataStore);
         categories.put(CategoryType.PREDICTIONS, Predictions);
-        return categories;
+        return Immutable.of(categories);
     }
 
     /**
--- a/core/src/main/java/com/amplifyframework/core/Resources.java
+++ b/core/src/main/java/com/amplifyframework/core/Resources.java
@@ -16,6 +16,7 @@
 package com.amplifyframework.core;
 
 import android.content.Context;
+import android.content.res.Resources.NotFoundException;
 import androidx.annotation.RawRes;
 
 import com.amplifyframework.AmplifyException;
@@ -75,7 +76,7 @@ public final class Resources {
 
         try {
             inputStream = context.getResources().openRawResource(resourceId);
-        } catch (android.content.res.Resources.NotFoundException exception) {
+        } catch (NotFoundException exception) {
             throw new AmplifyException(
                     "Failed to find the resource with ID " + resourceId + ".",
                     exception, "Please check that it has been created."
--- a/core/src/main/java/com/amplifyframework/core/Resources.java
+++ b/core/src/main/java/com/amplifyframework/core/Resources.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core;
+
+import android.content.Context;
+import androidx.annotation.RawRes;
+
+import com.amplifyframework.AmplifyException;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.InputStream;
+import java.util.Scanner;
+
+/**
+ * A utility to read resource files.
+ */
+public final class Resources {
+
+    private Resources() { }
+
+    /**
+     * Obtains the raw resource ID for the given resource identifier.
+     * @param context An Android Context
+     * @param identifier Name of a raw resource
+     * @return ID of the raw resource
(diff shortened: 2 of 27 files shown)


jamesonwilliams:
Can you break this PR into (at least) git commits, and then use the `rebase and merge` option, when it goes in?
- One commit to add the `getVersion()` across the code base (itself, a broad change)
- Another commit to begin using it from our feature

This way, the two will be distinct in the git log, and from git blame.

But, we don't have to review them separately. A single PR is fine.

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
```diff
@@ -0,0 +1,121 @@
...
+        devEnvironmentItems.put("osVersion", "OS Version");
+    }
+
+    /**
+     * Returns a String representation of the version of each plugin added, or
+     * "No plugins added." if no plugins were added.
+     * @return version information for each plugin added
+     */
+    public String getPluginVersions() {
+        String pluginVersions = getCategoryPluginVersions(Amplify.Analytics) +
```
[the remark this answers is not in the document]
eeatonaws: I was originally hoping to use Amplify.CATEGORIES, but I cannot access it since it's a private field.

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Objects;
+import java.util.Scanner;
+
+/**
+ * Contains information about the plugin versions added to the application
+ * and the developer's environment.
+ */
+public final class EnvironmentInfo {
+    // Name of the file that contains the developer environment information.
+    private static final String DEV_ENV_INFO_FILE_NAME = "localenvinfo";
```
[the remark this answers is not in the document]
eeatonaws: The file local-env-info.json currently resides in the amplify/.config directory in a customer's app project. Developer environment information will eventually be stored in this file and a copy of the file (named localenvinfo.json due to Android resource file naming restrictions) will be added to the app/src/main/res/raw directory in order to retrieve the information for the developer menu.

core/src/main/java/com/amplifyframework/core/Resources.java
```diff
@@ -0,0 +1,101 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.InputStream;
+import java.util.Scanner;
+
+/**
+ * A utility to read resource files.
+ */
+public final class Resources {
```
jamesonwilliams, line 32: Oh, huh. Nice cleanup of the `AmplifyConfiguration`, with this change!

core/src/main/java/com/amplifyframework/core/Resources.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Reads the content of the resource with the given ID, as a {@link JSONObject}.
+     * @param context An Android Context
+     * @param resourceId ID of a raw resource
+     * @return JSON Object equivalent of the raw resource
+     * @throws AmplifyException If resource with given ID does not exist or cannot be read
+     */
+    public static JSONObject readJsonResourceFromId(Context context, int resourceId) throws AmplifyException {
```
jamesonwilliams, line 73: `@ResId` annotation on this one?
```suggestion
    public static JSONObject readJsonResourceFromId(Context context, @ResId int resourceId) throws AmplifyException {
```

core/src/main/java/com/amplifyframework/devmenu/DeveloperMenu.java
```diff
@@ -33,6 +34,10 @@
  * developer menu.
  */
 public final class DeveloperMenu implements ShakeDetector.Listener {
+    /**
+     * Logger for the developer menu.
+     */
+    protected static final Logger LOG = Amplify.Logging.forNamespace("amplify:devmenu");
```
jamesonwilliams, line 40: Can it be just `private` (and thus, not need documentation?)
```suggestion
    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:devmenu");
```

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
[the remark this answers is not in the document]
jamesonwilliams: ^ I think it would be preferable to just make a new config block inside of our existing `amplifyconfiguration.json`. We _can't_ read the host-side file, and we probably won't benefit from having two separate `.json` files littered about the `src/main/res/raw`. This one should get design reviewed, ultimately. lizeyu@ is actively working on this so let's loop him in for review.

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public String getPluginVersions() {
+        String pluginVersions = getCategoryPluginVersions(Amplify.Analytics) +
+                getCategoryPluginVersions(Amplify.API) +
+                getCategoryPluginVersions(Amplify.Auth) +
+                getCategoryPluginVersions(Amplify.Logging) +
+                getCategoryPluginVersions(Amplify.Storage) +
+                getCategoryPluginVersions(Amplify.Hub) +
+                getCategoryPluginVersions(Amplify.DataStore) +
+                getCategoryPluginVersions(Amplify.Predictions);
```
jamesonwilliams, line 70: In situations like this, it's best to separate out a collection, and then use control flow. e.g.,

```java
// TODO: can we centralize these, somewhere? We want ~1 file in the code base that lists all the categories.
List<Category<?>> categories = Arrays.asList(Amplify.Whatever, ...);
for (Category<?> c : categories) {
    getCategoryPluginVersions(c);
}
```

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Returns a String representation of information about the developer's environment.
+     * @param context an Android Context
+     * @return developer environment information
+     * @throws AmplifyException if the developer environment information could not be read
+     */
+    public String getDeveloperEnvironmentInfo(@NonNull Context context) throws AmplifyException {
+        Context appContext = Objects.requireNonNull(context).getApplicationContext();
+        JSONObject envInfo = Resources.readJsonResource(appContext, DEV_ENV_INFO_FILE_NAME);
+        StringBuilder formattedEnvInfo = new StringBuilder();
```
jamesonwilliams, line 83: Let's assume this info will be in `amplifyconfiguraiton.json`, pending some support from the CLI team. If that's so, you can write this in a forward-compatible way by checking:
```java
if (amplifyConfigJson.has("keyName")) {
    // use it!
} else {
    // env info not available
}
```

core/src/main/java/com/amplifyframework/devmenu/DeveloperMenu.java
eeatonaws: I made the Logger protected in order to access it in DevMenuEnvironmentFragment. I was thinking it would be helpful to have a single logger instance for the developer menu that can be shared across the developer menu code.

jamesonwilliams: @eeatonaws That is probably a more efficient way to do it, yes. The rest of the code base right now uses multiple logger impls, even for the same tag, though. It would be sticking with consistency, anyway, to do the same.

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
```diff
@@ -60,14 +62,14 @@ public EnvironmentInfo() {
...
-        String pluginVersions = getCategoryPluginVersions(Amplify.Analytics) +
-                getCategoryPluginVersions(Amplify.API) +
-                getCategoryPluginVersions(Amplify.Auth) +
-                getCategoryPluginVersions(Amplify.Logging) +
-                getCategoryPluginVersions(Amplify.Storage) +
-                getCategoryPluginVersions(Amplify.Hub) +
-                getCategoryPluginVersions(Amplify.DataStore) +
-                getCategoryPluginVersions(Amplify.Predictions);
+        // TODO: can we centralize these somewhere? We want ~1 file in the code base that lists all the categories.
+        List<Category<?>> categories = Arrays.asList(Amplify.Analytics, Amplify.API, Amplify.Auth, Amplify.Logging,
```
jamesonwilliams, line 66: I'm trying to think: what would be the _harm_ of making this `CATEGORIES` public? If it were, then you could just do `Amplify.CATEGORIES.values()`, here. We'd want to change [this one to `Immutable.of(...)`](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/core/Amplify.java#L99), so that nobody could inject/modify categories or play games like that, I guess.

richardmcclellan: > I'm trying to think: what would be the harm of making this CATEGORIES public? 

One downside is that `CATEGORIES` would then show up in autocomplete after typing `Amplify.`, even though there's really no utility for it, which could cause confusion for customers.   Not a huge deal, but something to consider.

> We'd want to change [this one to Immutable.of(...)](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/core/Amplify.java#L99)

When doing this, you'd want to make sure the `Immutable` utility class can handle a `LinkedHashMap`.  It would need to maintain the order, and I'm not sure if it would or not.

core/src/main/java/com/amplifyframework/core/Resources.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param resourceId ID of a raw resource
+     * @return JSON Object equivalent of the raw resource
+     * @throws AmplifyException If resource with given ID does not exist or cannot be read
+     */
+    public static JSONObject readJsonResourceFromId(Context context, @RawRes int resourceId) throws AmplifyException {
+        InputStream inputStream;
+
+        try {
+            inputStream = context.getResources().openRawResource(resourceId);
+        } catch (android.content.res.Resources.NotFoundException exception) {
```
richardmcclellan, line 78: Can you import this class instead of using the fully qualified form inline?

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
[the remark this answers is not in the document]
TrekSoft: @jamesonwilliams: Erica, Daniel, Abhash, and I already did a design review with Jake Litwicki where it was decided the local env info file the CLI team is already generating is the correct one for this purpose and this is what iOS has implemented and has the user importing as part of the setup process. The problem with using amplifyconfiguration.json is that it would lead to conflicts when developers make commits from different machines.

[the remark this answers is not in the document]
richardmcclellan: > The problem with using amplifyconfiguration.json is that it would lead to conflicts when developers make commits from different machines with different environment info.

Very good point. 

Is there a design doc available to understand this a little more?    I'm curious to know how `localenvinfo.json` is generated.  Is it at compile time?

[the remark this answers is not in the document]
eeatonaws: There is a design doc, I just shared it with you.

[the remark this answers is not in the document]
richardmcclellan: From the design doc, I can see that `local-env-info.json` already exists, and is generated on `amplify configure`.    However, there's still several questions that don't seem to be addressed in the doc.  

1. How/when is `local-env-info.json` copied to `localenvinfo.json` in the res/raw folder?
2. How is it kept up to date when node version or amplify-cli version on my machine changes?  I would say that incorrect environment info is probably worse than no info at all.
3. Since`local-env-info.json` is not intended to be under source control, how it is added to `.gitignore`?  Or is this something developers need to do manually?

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -84,9 +89,9 @@ private Amplify() {
         throw new UnsupportedOperationException("No instances allowed.");
     }
 
-    // The fact that this is a LinkedHashMap, and not a Map, is important.
+    // The fact that this is a LinkedHashMap is important.
```
jamesonwilliams, line 92: !!!!!

core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -96,7 +101,7 @@ private Amplify() {
         categories.put(CategoryType.HUB, Hub);
         categories.put(CategoryType.DATASTORE, DataStore);
         categories.put(CategoryType.PREDICTIONS, Predictions);
-        return categories;
+        return Immutable.of(categories);
```
jamesonwilliams, line 104: Ah crap.

https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableMap(java.util.Map)

Not sure how you'll actually return a unmodifiable `LinkedHashMap`, though. You probably don't want to implement your own, that sounds like overkill. 

Hmmm.

core/src/main/java/com/amplifyframework/devmenu/EnvironmentInfo.java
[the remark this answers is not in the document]
richardmcclellan: Questions answered via Chime:

> 1. How/when is local-env-info.json copied to localenvinfo.json in the res/raw folder?

The CLI will be updated to copy this file.  TBD on which command will trigger this.  Possibly `amplify push`.

> 2. How is it kept up to date when node version or amplify-cli version on my machine changes? I would say that incorrect environment info is probably worse than no info at all.

The goal is not to keep it up to date, but to make it match the environment info that was used to generate the Amplify project, since that is most useful for debugging.  

> 3. Since `local-env-info.json` is not intended to be under source control, how it is added to .gitignore? Or is this something developers need to do manually?

The CLI will be updated to add it to .gitignore when it is copied to the `res/raw` folder.

core/src/main/java/com/amplifyframework/core/Amplify.java
richardmcclellan: Agreed!  How about this:

 - Keep `CATEGORIES` as a private `LinkedHashMap`, and the `Amplify` class can continue using it as is.
 - Create a public getter method `getCategoriesMap()`, which returns an immutable, unordered categories `Map`, which is used by the DevMenu.

PR #727, opened 2020-08-14
Search the logs on the developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Adds the ability to search the logs displayed on the developer menu.

Logs screen when all logs are displayed:
Dev Menu Logs New

Logs screen when there is a log that contains the search query:
Dev Menu Search Logs New

Logs screen when no logs contain the search query:
Dev Menu No Search Results New


core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
```diff
@@ -39,10 +46,23 @@ public DevMenuLogsFragment() { }
...
+        // Search the logs when the search button is pressed.
+        logsView.findViewById(R.id.search_logs_button).setOnClickListener(view -> searchLogs());
         return logsView;
     }
+
+    /**
+     * Display the logs (if any) that contain the search query.
+     */
+    private void searchLogs() {
+        EditText searchText = logsView.findViewById(R.id.search_logs_field);
```
richardmcclellan, line 63: Could you use a [SearchView](https://developer.android.com/reference/android/widget/SearchView) instead of an `EditText` here?  I believe it is basically just a wrapper around an `EditText`, but you get a lot of additional search functionality for free, like:
 - a search icon on the left side
 - a search button on the keyboard, which I believe is the more standard way for search to work, instead of having a button on the right side.
 - If your Android device has a hardware Search button, it'll actually work.

core/src/main/res/layout/dev_menu_fragment_logs.xml
```diff
@@ -55,14 +58,24 @@
             android:requiresFadingEdge="horizontal"
             android:fadingEdgeLength="70dp">
 
-            <TextView
+            <TextSwitcher
```
richardmcclellan, line 61: Interesting, I haven't seen this component before.  Is the purpose here just to add the fade animation when the text changes?

core/src/main/java/com/amplifyframework/devmenu/DeveloperMenu.java
```diff
@@ -129,6 +129,32 @@ public String getLogs() {
...
     }
 
+    /**
+     * Returns a String representation of the stored logs that contain the given
+     * String, or all logs if the given String is null or the empty string.
+     * @param searchText the text to search for in the logs
+     * @return the stored logs as a String.
+     */
+    public String getFilteredLogs(String searchText) {
+        if (searchText == null || searchText.isEmpty()) {
```
richardmcclellan, line 139:
```suggestion
        if (Empty.check(searchText)) {
```

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
```diff
@@ -39,10 +46,23 @@ public DevMenuLogsFragment() { }
...
-        View logsView = inflater.inflate(R.layout.dev_menu_fragment_logs, container, false);
+        logsView = inflater.inflate(R.layout.dev_menu_fragment_logs, container, false);
         // Display the logs (if any).
-        TextView logsText = logsView.findViewById(R.id.logs_text);
-        logsText.setText(DeveloperMenu.singletonInstance(getContext()).getLogs());
+        developerMenu = DeveloperMenu.singletonInstance(getContext());
+        logsText = logsView.findViewById(R.id.logs_text);
+        logsText.setText(developerMenu.getLogs());
+        // Search the logs when the search button is pressed.
+        logsView.findViewById(R.id.search_logs_button).setOnClickListener(view -> searchLogs());
```
richardmcclellan, line 55: Instead of having an explicit button, could you just display results in real time, updating them each time a new character is entered in the search bar?    If you use a SearchView, you could implement this with the onQueryTextChanged() method in [OnQueryTextListener](https://developer.android.com/reference/android/widget/SearchView.OnQueryTextListener).

core/src/main/res/layout/dev_menu_fragment_logs.xml
eeatonaws: Yes, the TextSwitcher switches between displaying the text on its two child TextViews and animates the action of displaying new text.

eeatonaws:
> What does the UI look like when the software keyboard is visible? Does it cover the search bar, or push it up?
> 
> That scroll area for the logs themselves is awfully small - I'd suggest making this view fill a larger portion of the screen if possible.

When the software keyboard is visible, the developer menu is pushed up to prevent the keyboard from blocking the logs (shown in the picture below). I agree that the scroll area for the logs is somewhat small. In https://github.com/aws-amplify/amplify-android/pull/719, I increase the height of the developer menu, which will give more space for displaying the logs.

Screen Shot 2020-08-14 at 11 38 14 AM

richardmcclellan:
> When the software keyboard is visible, the developer menu is pushed up to prevent the keyboard from blocking the logs (shown in the picture below). I agree that the scroll area for the logs is somewhat small. In #719, I increase the height of the developer menu, which will give more space for displaying the logs.

Given that the developer menu height is defined as a constant, that might push the search bar up off the for devices with certain resolutions.  For example, what would happen in landscape view?    Instead of defining the height of the developer menu it might be better to constrain the view to its parent (`layout_marginTop=XXdp`), so that it never gets pushed off the top of the screen.

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
eeatonaws: I initially tried to use a SearchView instead of EditText, but the search icon that should be displayed to the left of the SearchView was not visible. The XML attribute used to set this icon is only available in API level 21+, yet Android still allocates space for the icon to the left of where the user would enter a query. Setting margin and padding to 0 does not remove the extra space for the icon. The result is shown in the picture below. I think having the extra space to the left of the SearchView looks strange and I could not find a way to either align the SearchView to the left or have an equal amount of space on either side of the SearchView.

Screen Shot 2020-08-14 at 12 21 47 PM

richardmcclellan: Can you fix this by using `androidx.appcompat.widget.SearchView` instead of `android.widget.SearchView`?

eeatonaws: Yes, although it will require making Theme.AppCompat the parent style for the activity, which changes the styling of many components on the developer menu, but I should be able to fix the styling issues.

eeatonaws:
> > When the software keyboard is visible, the developer menu is pushed up to prevent the keyboard from blocking the logs (shown in the picture below). I agree that the scroll area for the logs is somewhat small. In #719, I increase the height of the developer menu, which will give more space for displaying the logs.
> 
> Given that the developer menu height is defined as a constant, that might push the search bar up off the for devices with certain resolutions. For example, what would happen in landscape view? Instead of defining the height of the developer menu it might be better to constrain the view to its parent (`layout_marginTop=XXdp`), so that it never gets pushed off the top of the screen.

When the device is in landscape view and the user presses the search box, the screen looks as follows (the "Search" button is added automatically and closes the soft keyboard when pressed):
Dev Menu Search Logs Landscape

If I set layout_marginTop on the outermost layout for the developer menu activity instead of setting the height, the developer menu disappears when typing in the search box.

richardmcclellan:
> When the device is in landscape view and the user presses the search box, the screen looks as follows (the "Search" button is added automatically and closes the soft keyboard when pressed):

Interesting!  I guess this a feature of `SearchView`.  My first thought was that it looked a little odd, as I don't think you would need such a large text area to enter a query, but given that this is the `platform behavior`, I think we should go with it.

PR #732, opened 2020-08-18
feat(auth): Adds user info to sign up result

TrekSoft:
After a successful sign up, the sign up result will now contain the user info.

Note that the user will be null for `resendSignUpCode()` and `confirmSignUp()` operations.

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -802,7 +802,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     }
 
     private AuthSignUpResult convertSignUpResult(@NonNull SignUpResult result, @NonNull String username) {
-        UserCodeDeliveryDetails details = result.getUserCodeDeliveryDetails();
+        UserCodeDeliveryDetails details = Objects.requireNonNull(result).getUserCodeDeliveryDetails();
         AuthCodeDeliveryDetails newDetails = details != null
                 ? new AuthCodeDeliveryDetails(
                     details.getDestination(),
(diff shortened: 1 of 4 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -801,7 +801,7 @@ private String getUserIdFromToken(String token) {
         }
     }
 
-    private AuthSignUpResult convertSignUpResult(SignUpResult result) {
+    private AuthSignUpResult convertSignUpResult(@NonNull SignUpResult result, @NonNull String username) {
         UserCodeDeliveryDetails details = result.getUserCodeDeliveryDetails();
```
eeatonaws, line 805:
```suggestion
        UserCodeDeliveryDetails details = Objects.requireNonNull(result).getUserCodeDeliveryDetails();
```

rxbindings/src/test/java/com/amplifyframework/rx/RxAuthBindingTest.java
```diff
@@ -86,7 +87,7 @@ public void testSignUpSucceeds() {
         AuthCodeDeliveryDetails details = new AuthCodeDeliveryDetails(RandomString.string(), DeliveryMedium.SMS);
         AuthSignUpStep step = AuthSignUpStep.CONFIRM_SIGN_UP_STEP;
         AuthNextSignUpStep nextStep = new AuthNextSignUpStep(step, Collections.emptyMap(), details);
-        AuthSignUpResult result = new AuthSignUpResult(false, nextStep);
+        AuthSignUpResult result = new AuthSignUpResult(false, nextStep, new AuthUser(userId, username));
```
eeatonaws, line 90: In the JavaDoc comment for the AuthSignUpResult constructor, it says the user will be null if isSignUpComplete is false, but here you're passing in a non-null AuthUser and false for isSignUpComplete. When creating this new AuthSignUpResult here, should you either pass in true for isSignUpComplete or null for the user?

TrekSoft: Good point - I'll add null for the user

PR #737, opened 2020-08-18
feat(datastore) add pagination, respecting syncPageSize and syncMaxRecords

richardmcclellan:
Currently, DataStore only syncs up to `1_000` items for each model.  The sync query response includes a "nextToken" if there are more pages of results available, but this is currently ignored.

This PR adds support for fetching additional pages of results.
 - Customers can set `syncPageSize` on the `DataStoreConfiguration` to control the number of results per page.   This defaults to `1_000`. Customers may need to lower this number if the model is large, because the result set from a DDB Scan is limited to 1 MB per call.
 - Customers can set `syncMaxRecords` on the `DataStoreConfiguration to control the total number of items to sync.  This defaults to `10_000`.


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -166,28 +168,41 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
...
+     *                response.getData().getRequestForNextResult() for subsequent requests.
+     * @param singleEmitter A SingleEmitter which emits an Iterable&lt;ModelWithMetadata&gt;, a concatenation of the
+     *                      results from all pages.
+     * @param emittedValue a Set&lt;ModelWithMetadata&lt;T&gt;&gt; containing the concatenated results of all requests.
+     * @param <T> The type of model to sync.
+     */
+    private <T extends Model> void syncPage(
+        GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>> request,
+        SingleEmitter<Iterable<ModelWithMetadata<T>>> singleEmitter,
+        Set<ModelWithMetadata<T>> emittedValue) {
```
jamesonwilliams, line 204: I like to use +8 for method params, so that the body is clearly distinguished at +4
```suggestion
            GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>> request,
            SingleEmitter<Iterable<ModelWithMetadata<T>>> singleEmitter,
            Set<ModelWithMetadata<T>> emittedValue) {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -198,13 +213,26 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
...
+        singleEmitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable));
+    }
+
+    private Integer syncMaxRecords() {
+        try {
+            return dataStoreConfigurationProvider.getConfiguration().getSyncMaxRecords();
+        } catch (DataStoreException exception) {
+            LOG.warn("Failed to retrieve datastore configuration, using default syncMaxRecords value.", exception);
+            return DEFAULT_SYNC_MAX_RECORDS;
+        }
```
jamesonwilliams, line 235: Is this a realistic scenario?

If so, can we handle this in the code which build the configuration object, instead? S.t. the configuration always has some sane value, by the time we get here to access it?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncClientTest.java
```diff
@@ -68,7 +69,8 @@ public void setup() {
 
         // We need it to response with **something** by default.
         // Use this same method to send more interesting test values back...
-        mockApiResponse(new GraphQLResponse<>(new ArrayList<>(), new ArrayList<>()));
+        PaginatedResult<ModelWithMetadata<BlogOwner>> data = new PaginatedResult<>(new ArrayList<>(), null);
+        mockApiResponse(new GraphQLResponse<>(data, new ArrayList<>()));
```
jamesonwilliams, line 73:
```suggestion
        PaginatedResult<ModelWithMetadata<BlogOwner>> data = new PaginatedResult<>(Collections.emptyList(), null);
        mockApiResponse(new GraphQLResponse<>(data, Collections.emptyList()));
```

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
```diff
@@ -213,91 +224,159 @@ public static OnDeleteConfigurator onDelete(AppSync mock) {
...
-                any(), // Consumer<Iterable<ModelWithMetadata<T>>>
-                any() // Consumer<DataStoreException>
+                argThat(requestMatcher),
+                any(), // Consumer<GraphQLResponse<PaginatedResult<ModelWithMetadata<M>>>>
+                any()  // Consumer<DataStoreException>
             );
             return SyncConfigurator.this;
         }
 
+        private <M extends Model> ArgumentMatcher<GraphQLRequest<PaginatedResult<ModelWithMetadata<M>>>> matcherFor(
```
jamesonwilliams, line 351: Method name: `matchesRequest`? It'll look cool above when you write `argThat(matchesRequest(`!

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMocking.java
```diff
@@ -213,91 +224,159 @@ public static OnDeleteConfigurator onDelete(AppSync mock) {
...
+         */
+        public <M extends Model> SyncConfigurator mockSuccessResponse(
+                ArgumentMatcher<GraphQLRequest<PaginatedResult<ModelWithMetadata<M>>>> requestMatcher,
+                GraphQLResponse<PaginatedResult<ModelWithMetadata<M>>> mockResponse) {
             doAnswer(invocation -> {
                 // Get a handle to the response consumer that is passed into the sync() method
-                // Response consumer is the third param, at index 2 (@0, @1, @2, @3).
-                final int argumentPositionForResponseConsumer = 2;
-                final Consumer<GraphQLResponse<Iterable<ModelWithMetadata<T>>>> consumer =
+                // Response consumer is the second param, at index 1 (@0, @1, @2).
```
jamesonwilliams, line 333: I really appreciate you taking the time and consideration to update some of this ancillary stuff. This would be easy to overlook, and often is.

richardmcclellan:
> It might be that you don't have to solve the `StackOverflowException` in this PR. Your current PR is already a significant upgrade in functionality. We could "unwrap" the recursion in a follow-up PR.

Hmm, it would be an improvement for customers with ~10,000 items, since all of their items would be synced, but for a customer with ~1,000,000 items, it would cause their app to crash every time.  Given that, I don't think it's a great idea to ship it as is. 

> Do you have an understanding for which platform limit you're running into, that triggers the SOE?

When running the unit tests on my Mac, the tests pass for 500 pages, but result in a stack overflow for 1000 pages.  I believe the memory available for the call stack is different per device though, probably lower on an Android phone.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: Agreed on the indentation preference!  This method was refactored in my latest push though, so no longer relevant.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: `DataStoreConfigurationProvider::getConfiguration()` does throw a `DataStoreException`, if errors are found building the configuration, but I don't think it is actually thrown in practice.  If the configuration.json is not found, then it uses falls back to defaults.  

I've updated this so that if a `DataStoreException` is actually thrown, it will cause the whole sync to fail, instead of falling back to some default here.  I think this is reasonable, since, as I said, it shouldn't ever be thrown in practice.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -402,6 +423,138 @@ public void userProvidedErrorCallbackInvokedOnFailure() {
...
+     * @throws AmplifyException on error building sync request for next page.
+     * @throws InterruptedException If interrupted while awaiting terminal result in test observer
+     */
+    @Test
+    public void syncStopsAfterMaxRecords() throws AmplifyException, InterruptedException {
+        syncAndExpect(10, 5);
+    }
+
+    /**
+     * Validate the sync can handle 500 of pages.  Even with a recursive, functional algorithm, this should pass.
```
richardmcclellan, line 447:
```suggestion
     * Validate the sync can handle 100 of pages.  Even with a recursive, functional algorithm, this should pass.
```

richardmcclellan:
This is ready for review now!  I was able to solve the StackOverflowError with a [BehaviorProcessor](http://reactivex.io/RxJava/javadoc/io/reactivex/processors/BehaviorProcessor.html).  Quite elegant once it worked, but RxJava really makes my head spin!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -172,45 +169,62 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
...
-        ).doOnError(failureToSync ->
-            LOG.warn("Failed to sync down cloud state for model type = " + modelClass.getSimpleName(), failureToSync)
-        );
+        final Integer syncPageSize = dataStoreConfigurationProvider.getConfiguration().getSyncPageSize();
+
+        // Create a BehaviorProcessor, and set the default value to a GraphQLRequest that fetches the first page.
+        BehaviorProcessor<GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>>> processor =
+                BehaviorProcessor.createDefault(appSync.buildSyncRequest(modelClass, lastSyncTimeAsLong, syncPageSize));
+
+        return processor.concatMap(request -> syncPage(request).toFlowable())
```
jamesonwilliams, line 189: 189-200 -- some advanced stuff going on here, very cool!

richardmcclellan:
> Nice!
> 
> Curious how you arrived to `Flowable` & `BehaviorProcessor`? (I don't doubt their use! I as an individual just have not used them yet.)

It was adapted from [this StackOverflow solution](https://stackoverflow.com/a/48580794/282279).   There's actually a [whole list of processors](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/processors/package-summary.html) besides BehaviorProcessor.  All of them are basically for handling more complex routing between producers and subscribers.  My understanding is that `Flowable` is similar to `Observable`, except that it supports back pressure, which ensures that if the producer is emitting items faster than a subscriber can consume them, it can tell the producer to slow down.   That's probably not necessary here since I imagine we can save items to SQLite much faster than we are receiving them over the network.  I don't think there's any downside though.     I think this problem could be solved with an `Observable` stream by using `Subjects`, but I haven't looked too deeply into that.

I'm thinking we should also update our [API docs](https://docs.amplify.aws/lib/graphqlapi/query-data/q/platform/android#list-multiple-pages-of-items) to suggest this approach to customers as well, because the current sample code is susceptible to a stack overflow, since it is recursive.

PR #738, opened 2020-08-19
feat(aws-auth-cognito): implement device tracking

raphkim:
*Issue #, if available:*

*Description of changes:*
Add `rememberDevice()`, `forgetDevice()`, and `fetchDevices()` in the Auth category

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -25,6 +25,7 @@ import androidx.annotation.VisibleForTesting;
 import com.amplifyframework.AmplifyException;
 import com.amplifyframework.auth.AuthChannelEventName;
 import com.amplifyframework.auth.AuthCodeDeliveryDetails;
+import com.amplifyframework.auth.AuthDevice;
 import com.amplifyframework.auth.AuthException;
 import com.amplifyframework.auth.AuthPlugin;
 import com.amplifyframework.auth.AuthProvider;
@@ -65,8 +66,10 @@ import com.amazonaws.mobile.client.SignInUIOptions;
 import com.amazonaws.mobile.client.SignOutOptions;
 import com.amazonaws.mobile.client.UserState;
 import com.amazonaws.mobile.client.UserStateDetails;
+import com.amazonaws.mobile.client.results.Device;
 import com.amazonaws.mobile.client.results.ForgotPasswordResult;
 import com.amazonaws.mobile.client.results.ForgotPasswordState;
+import com.amazonaws.mobile.client.results.ListDevicesResult;
 import com.amazonaws.mobile.client.results.SignInResult;
 import com.amazonaws.mobile.client.results.SignInState;
 import com.amazonaws.mobile.client.results.SignUpResult;
@@ -78,8 +81,10 @@ import com.amazonaws.services.cognitoidentity.model.NotAuthorizedException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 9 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -510,6 +515,99 @@ public void onError(Exception exception) {
...
+    public void fetchDevices(
+            @NonNull Consumer<List<AuthDevice>> onSuccess,
+            @NonNull Consumer<AuthException> onException
+    ) {
+        awsMobileClient.getDeviceOperations().list(new Callback<ListDevicesResult>() {
+            @Override
+            public void onResult(ListDevicesResult result) {
+                List<AuthDevice> devices = new ArrayList<>();
+                for (Device device : result.getDevices()) {
+                    devices.add(AuthDevice.fromId(device.getDeviceKey()));
```
TrekSoft, line 595: Did you verify we can't get the device names from AWSMobileClient?

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
```diff
@@ -252,11 +254,52 @@ public Object confirmResetPassword(
...
+     * Forgets the current device synchronously.
+     * @param device Auth device to forget
+     * @return Dummy object - just indicates it completed successfully
+     * @throws AuthException exception
+     */
+    public Object forgetDevice(@NonNull AuthDevice device) throws AuthException {
+        return Await.<Object, AuthException>result(AUTH_OPERATION_TIMEOUT_MS, (onResult, onError) ->
+                asyncDelegate.forgetDevice(device, () -> onResult.accept(new Object()), onError)
         );
     }
```
jamesonwilliams, line 292: Passing the dummy `new Object()` into the `Await` util is okay. But you don't need to return it.

Success can be == doesn't throw an exception.

I'd update this for every method that uses this pattern.

```suggestion
    /**
     * Forgets the current device.
     * @param device Auth device to forget
     * @throws AuthException On failure to forget the provided device
     */
    public void forgetDevice(@NonNull AuthDevice device) throws AuthException {
        Await.<Object, AuthException>result(AUTH_OPERATION_TIMEOUT_MS, (onResult, onError) ->
                asyncDelegate.forgetDevice(device, () -> onResult.accept(new Object()), onError)
        );
    }
```

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
```diff
@@ -252,11 +254,52 @@ public Object confirmResetPassword(
...
+                asyncDelegate.rememberDevice(() -> onResult.accept(new Object()), onError)
+        );
+    }
+
+    /**
+     * Forgets the current device synchronously.
+     * @return Dummy object - just indicates it completed successfully
+     * @throws AuthException exception
+     */
+    public Object forgetDevice() throws AuthException {
```
jamesonwilliams, line 276: Naming: this should be `forgetCurrentDevice()`, to clearly distinguish it from `forgetDevice(AuthDevice specificDevice)`. I would not compromise the naming just to match iOS. We should ask the iOS team to update their name if they're not using this, too.

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
```diff
@@ -252,11 +254,52 @@ public Object confirmResetPassword(
...
-                asyncDelegate.fetchAuthSession(onResult, onError)
+        return Await.result(AUTH_OPERATION_TIMEOUT_MS, asyncDelegate::fetchAuthSession);
+    }
+
+    /**
+     * Remembers current device synchronously.
+     * @return Dummy object - just indicates it completed successfully
+     * @throws AuthException exception
+     */
+    public Object rememberDevice() throws AuthException {
```
jamesonwilliams, line 265: Since we have a `forceDevice(AuthDevice specificDevice)`, I suggest that we call this `rememberCurrentDevice()`. (Same for Rx bindings, of course, as well.)

rxbindings/src/main/java/com/amplifyframework/rx/RxAuthBinding.java
```diff
@@ -126,6 +128,26 @@ public void handleWebUISignInResponse(Intent intent) {
...
+    }
+
+    @Override
+    public Completable forgetDevice(@NonNull AuthDevice device) {
+        return toCompletable((onComplete, onError) -> delegate.forgetDevice(device, onComplete, onError));
+    }
+
+    @Override
+    public Single<List<AuthDevice>> fetchDevices() {
+        return toSingle((delegate::fetchDevices));
```
jamesonwilliams, line 148: Extra parens
```suggestion
        return toSingle(delegate::fetchDevices);
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
raphkim: Will verify with jithin or someone else before I push this!

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
raphkim: This is the pattern that I usually follow, but I was trying to match the other methods in this class.

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
raphkim: They have it set up such that `AuthDevice` parameter is nullable, and if none is passed, the method exhibits above behavior. It was in my discretion to overload the method and make `AuthDevice` a `@NonNull` parameter. I believe it is okay to rename this method as you suggest.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
raphkim: https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/Device%2BExtension.swift
Swift appears to be ignoring name when creating a device also.

Issue #724, opened 2020-08-13
[DataStore] initial sync fails when using schema with belongs-to connection


a-melnyk:
Hi everyone,

I have a problem when initial sync fails on a schema with [belongs-to connection](https://docs.amplify.aws/cli/graphql-transformer/directives#belongs-to), it seems that DataStore doesn't support this type of connection. The problem reproduces with any number of items in the table (also with 1 Post and 1 Comment)

supporting info:
* [logs](https://gist.github.com/a-melnyk/7b02b55d5f88df246c7bfb7a4942f455)
* [code](https://github.com/a-melnyk/simple-datastore/tree/initial_sync_fails_SQLITE_CONSTRAINT_FOREIGNKEY)
* [schema](https://github.com/a-melnyk/simple-datastore/blob/initial_sync_fails_SQLITE_CONSTRAINT_FOREIGNKEY/schema.graphql)

steps to reproduce:
1. start the app and sign in by clicking the corresponding button
2. click the "Post + COMMENT" button, a post and a comment should be created, logged and synchronised to DynamoDB
3. remove the app (and data) from your android device
4. start the app and sign in
5. click the "LIST ALL POSTS" to trigger initial sync, observe application logs -> the initial sync fails

My assumption is that the synchronisation order causes the issue, as we can see from logs and the sqlite state, DataStore tries to synchronise and persist a `Comment` locally without having persisted the corresponding `Post` yet.
Renaming `Comment` entity to `Xomment` seems to change and synchronisation order and fix the issue.

Issue #487, opened 2020-05-19
DataStore multiple one has many relation error


michalisAsp:
With below schema, latest master branch, DataStore sync disabled:
```graphql
type Blog @model
{
    id: ID!
    posts: [Post] @connection(name: "BlogPosts")
}
type Post @model
{
    id: ID!
    blog: Blog @connection(name: "BlogPosts")
    comments: [Comment!] @connection(name: "PostComments")
    text: String
}
type Comment @model
{
    id: ID!
    post: Post @connection(name: "PostComments")
}
```

I am seeing the exception below when I am trying to query for comments
`Amplify.DataStore.query(Comment::class.java)`

It seems that defining relations A has many B & B has many C is what causing the issue.
Keep in mind that I am able to store Comments, and query/store Posts.

 ```console
AmplifyException {message=Error in querying the model., cause=AmplifyException {message=Error converting field "post" from model "com.amplifyframework.datastore.generated.model.Comment", cause=AmplifyException {message=Error converting field "blog" from model "com.amplifyframework.datastore.generated.model.Post", cause=AmplifyException {message=Error converting field "id" from model "com.amplifyframework.datastore.generated.model.Blog", cause=java.lang.IllegalArgumentException: column 'Blog_id' does not exist. Available columns: [Comment_id, Comment_commentPostId, Post_id, Post_text, Post_postBlogId], recoverySuggestion=There is a possibility that there is a bug if this error persists. Please take a look at 
https://github.com/aws-amplify/amplify-android/issues to see if there are any existing issues that 
match your scenario, and file an issue with the details of the bug if there isn't.}, recoverySuggestion=There is a possibility that there is a bug if this error persists. Please take a look at 
https://github.com/aws-amplify/amplify-android/issues to see if there are any existing issues that 
match your scenario, and file an issue with the details of the bug if there isn't.}, recoverySuggestion=There is a possibility that there is a bug if this error persists. Please take a look at 
https://github.com/aws-amplify/amplify-android/issues to see if there are any existing issues that 
match your scenario, and file an issue with the details of the bug if there isn't.}, recoverySuggestion=See attached exception for details.}
```

jamesonwilliams:
This looks like the "atom" of the issue:

[...]

Issue #464, opened 2020-05-14
Synchronization Error with relations


Andrew-Dickinson:
Might be a duplicate of #461 but I don't want to crowd that thread if it's different. The main difference is that I can cause this crash to occur with a small amount of data.

**Schema**
```grahphql
type GroceryItem @model {
  id: ID!
  checked: Boolean!
  name: String
  position: Int!
  aisle: Aisle! @connection(name: "AisleGroceries")
}

type Aisle @model {
  id: ID!
  name: String!
  position: Int!
  groceries: [GroceryItem!]! @connection(name: "AisleGroceries")
}
```

**Steps to Reproduce**
If while my app is running, I run the following two GraphQL queries against the AppSync API directly:

```graphql
mutation CreateAisle {
  createAisle(input: {
    id: "d49e54cd-83d6-4fe5-0197-00afa72ac41c",
    name: "Aisle9",
    position: 0
  }) {
    id
    name
    position
    _lastChangedAt
    _version
    _deleted
  }
}

mutation CreateGroceryItem {
  createGroceryItem(input: {
    id: "b326a9e7-b225-4d20-9a51-6470f75f6361",
    name: "GroceryItem1",
    checked: false,
    position: 0,
    groceryItemAisleId: "d49e54cd-83d6-4fe5-0197-00afa72ac41c"
  }) {
    id
    name
    position
    _lastChangedAt
    _version
    _deleted
  }
}

```

The first query (creating the aisle), passes without a hitch and I see my app's UI update to reflect it. However, the second query causes a crash.

**Full stacktrace**
```console
W/amplify:aws-datastore: Error enqueuing mutation from subscription.
    com.amplifyframework.datastore.DataStoreException: Bad subscription data for GroceryItem: [Cannot return null for non-nullable type: 'Aisle' within parent 'GroceryItem' (/onCreateGroceryItem/aisle), Cannot return null for non-nullable type: 'Boolean' within parent 'GroceryItem' (/onCreateGroceryItem/checked)]
        at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$subscription$2$AppSyncClient(AppSyncClient.java:3)
        at com.amplifyframework.datastore.appsync.-$$Lambda$AppSyncClient$KK1x1_fo5sffWREukMLi8z0goy4.accept(Unknown Source:10)
        at com.amplifyframework.api.aws.SubscriptionEndpoint.notifySubscriptionData(SubscriptionEndpoint.java:2)
[...]

PR #741, opened 2020-08-20
fix(datastore): order hydration tasks topologically

saltonmassally:
*Issue #, if available:*

topological ordering was being applied to the individual hydration stream, i.e. given classes of model A and Model B, it appears that the original implementation was being applied on the model A hydration observable and Model B hydration observable. As such model A will always be compared to model A and B to B. Which I don't believe we want. What we want is to have models that depend on another hydrate first and so sort the hydration task by that.

*Description of changes:*

Proposing a different implementation that first sorts the hydration tasks by the model's typological sort and then continue with Compete.concat which then handles the completion of each stream sequentially.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
@@ -93,8 +93,6 @@ final class SyncProcessor {
      * @return An Rx {@link Completable} which can be used to perform the operation.
      */
     Completable hydrate() {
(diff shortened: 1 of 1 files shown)


rjuliano:
> This looks great @saltonmassally, thanks so much! This one was my bug. :-D
> 
> I believe this will fix _at least one of_:
> 
> * #464
> * #724
> * #487
> 
> Can you confirm?

@jamesonwilliams For sure #724. I created a sample app with the schema from that issue and this change definitely fixes it. 

I think #464 and #487 are different issues though. I'll try to confirm that though.

@saltonmassally Thanks so much for this contribution

PR #744, opened 2020-08-21
chore(core) Enable logging during unit tests

richardmcclellan:
This enables logging to the console window when running unit tests.  The AndroidLogger, which relies on `android.util.Log` does not work during unit tests, since it relies on the Android library, which is not available during unit tests.    There is now a new JavaLogger which is used instead, if run from a unit test.


richardmcclellan:
@jamesonwilliams currently:
 - `core` depends on `testImplementation(project(path: ':testutils'))`
 - `testutils` depends on `implementation project(path: ':core')`

If we move `JavaLoggingPlugin` to `testutils`, then we need:
 - `core` to depend on `implementation(project(path: ':testutils'))`

However, this results in a circular dependency, so we can't do that.   

Not really sure of a way to move these to `src/test` either.  I suppose the alternative is to mock `AndroidLogger` in test, and have it call `System.out.println()` instead of `android.util.Log`.  That approach requires adding a @Before method to every test, or having a common superclass for all test classes though, which I was hoping to avoid.

core/src/main/java/com/amplifyframework/logging/JavaLoggingPlugin.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Implementation of the {@link LoggingCategoryBehavior} designed for use by unit tests,
+ * which emits logs via {@code System.out.println()}.
+ */
+public final class JavaLoggingPlugin extends LoggingPlugin<Void> {
+    private static final String AMPLIFY_NAMESPACE = "amplify";
+    private final LogLevel defaultLoggerThreshold;
+
+    /**
+     * Creates a logging plugin using {@link LogLevel#INFO} as the default
```
eeatonaws, line 33: The comment says `LogLevel.INFO` is used as the default logging threshold, but the constructor itself uses `LogLevel.VERBOSE`.

jamesonwilliams:
> @jamesonwilliams currently:
> 
> * `core` depends on `testImplementation(project(path: ':testutils'))`
> * `testutils` depends on `implementation project(path: ':core')`
> 
> If we move `JavaLoggingPlugin` to `testutils`, then we need:
> 
> * `core` to depend on `implementation(project(path: ':testutils'))`
> 
> However, this results in a circular dependency, so we can't do that.
> 
> Not really sure of a way to move these to `src/test` either. I suppose the alternative is to mock `AndroidLogger` in test, and have it call `System.out.println()` instead of `android.util.Log`. That approach requires adding a @before method to every test, or having a common superclass for all test classes though, which I was hoping to avoid.

Thanks, @richardmcclellan. I agree, your solution is the cleanest among the various strategies, and provides some immediate value. Please proceed!

core/src/main/java/com/amplifyframework/logging/JavaLogger.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public void verbose(@Nullable String message) {
+        if (threshold.above(LogLevel.VERBOSE)) {
+            return;
+        }
+        log(LogLevel.VERBOSE, message);
+    }
+
+    private void log(@NonNull LogLevel error, @Nullable String message) {
```
eeatonaws, line 100: Since the `LogLevel` passed in is not always `LogLevel.ERROR`, I think it would be more clear to rename the `LogLevel error` parameter to something like `logLevel` or just `level`.

richardmcclellan: Good catch!  Naming it `error` was not intentional :)

PR #746, opened 2020-08-21
Remove logs from issue body on developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* On the developer menu, removes the option to include logs in the issue body generated on the "File Issue" screen. Logs will not be included in the issue body whether the "Copy to Clipboard" or "File Issue" button is pressed.

Updated file issue screen on the developer menu:
Dev Menu New File Issue Screen

Updated GitHub issue body:
Dev Menu File Issue No Logs Included


core/src/main/java/com/amplifyframework/devmenu/DeveloperMenu.java
```diff
@@ -160,12 +160,11 @@ public String getFilteredLogs(String searchText) {
...
     /**
      * Creates and returns the issue body containing the issue description,
-     * environment information, device information, and (optionally) logs.
+     * environment information, and device information.
      * @param description the description of the issue
-     * @param includeLogs whether to include logs in the issue body
      * @return the issue body
      */
-    public String createIssueBody(String description, boolean includeLogs) {
+    public String createIssueBody(String description) {
```
TrekSoft, line 167: Rather than remove this good work you already did, I would just modify the code which calls the function to always send false for now with a ToDo comment that for now logs aren't being included but if that changes later they can be sent.

jpignata: That seems good to me! I advised Erica to remove code versus comment out, but I think the solution above is reasonable and ensures we can easily pick this back up.

TrekSoft: Yeah it makes sense to me to remove the UI code but as far as this method goes it's optional, additional functionality so felt like it made sense to make an exception here.

PR #757, opened 2020-08-25
Categorize sign in and sign up errors

changxu0306:
Specify error categories under Sign in and Sign up scenarios.
Add `AWSCognitoAuthException` class and its subclasses under `AuthException.java`
Check error categories in `AWSCognitoAuthPlugin.java`

Note that there are still some error categories not checked within SignUp and SignIn methods.

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -57,6 +57,7 @@ import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.hub.HubEvent;
 import com.amplifyframework.util.UserAgent;
 
+import com.amazonaws.AmazonClientException;
 import com.amazonaws.logging.LogFactory;
 import com.amazonaws.mobile.client.AWSMobileClient;
 import com.amazonaws.mobile.client.Callback;
@@ -75,7 +76,15 @@ import com.amazonaws.mobile.client.results.UserCodeDeliveryDetails;
 import com.amazonaws.mobile.config.AWSConfiguration;
 import com.amazonaws.mobileconnectors.cognitoauth.AuthClient;
 import com.amazonaws.mobileconnectors.cognitoidentityprovider.util.CognitoJWTParser;
-import com.amazonaws.services.cognitoidentity.model.NotAuthorizedException;
+import com.amazonaws.services.cognitoidentityprovider.model.AliasExistsException;
+import com.amazonaws.services.cognitoidentityprovider.model.CodeDeliveryFailureException;
+import com.amazonaws.services.cognitoidentityprovider.model.CodeMismatchException;
+import com.amazonaws.services.cognitoidentityprovider.model.ExpiredCodeException;
+import com.amazonaws.services.cognitoidentityprovider.model.InvalidPasswordException;
+import com.amazonaws.services.cognitoidentityprovider.model.NotAuthorizedException;
+import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
+import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
+import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -279,9 +288,24 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
                 @Override
                 public void onError(Exception error) {
-                    onException.accept(
-                        new AuthException("Sign up failed", error, "See attached exception for more details")
-                    );
+                    if (error instanceof UsernameExistsException) {
+                        onException.accept(
+                                new AuthException.AWSCognitoAuthException.UsernameExistsException(error.getCause())
+                        );
+                    } else if (error instanceof AliasExistsException) {
+                        onException.accept(
+                                new AuthException.AWSCognitoAuthException.AliasExistsException(error.getCause())
+                        );
+                    } else if (error instanceof AmazonClientException) {
+                        onException.accept(
+                                new AuthException.AWSCognitoAuthException.NetworkException(error.getCause())
+                        );
+                    } else {
+                        onException.accept(
+                                new AuthException("Sign up failed", error, "See attached exception for more details")
+                        );
+                    }
+
                 }
             }
         );
@@ -302,9 +326,36 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
             @Override
             public void onError(Exception error) {
-                onException.accept(
-                    new AuthException("Confirm sign up failed", error, "See attached exception for more details")
-                );
+                if (error instanceof UserNotFoundException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.UserNotFoundException(error.getCause())
+                    );
+                } else if (error instanceof CodeMismatchException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.CodeMismatchException(error.getCause())
+                    );
+                } else if (error instanceof ExpiredCodeException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.CodeExpiredException(error.getCause())
+                    );
+                } else if (error instanceof CodeDeliveryFailureException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.CodeDeliveryFailureException(error.getCause())
+                    );
+                } else if (error instanceof UserNotConfirmedException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.UserNotConfirmedException(error.getCause())
+                    );
+                } else if (error instanceof AmazonClientException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.NetworkException(error.getCause())
+                    );
+                } else {
+                    onException.accept(
+                            new AuthException("Confirm sign up failed", error,
+                                            "See attached exception for more details")
+                    );
+                }
             }
         });
     }
@@ -361,9 +412,23 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
             @Override
             public void onError(Exception error) {
-                onException.accept(
-                    new AuthException("Sign in failed", error, "See attached exception for more details")
-                );
+                if (error instanceof InvalidPasswordException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.InvalidPasswordException(error.getCause())
+                    );
+                } else if (error instanceof UserNotFoundException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.UserNotFoundException(error.getCause())
+                    );
+                } else if (error instanceof AmazonClientException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.NetworkException(error.getCause())
+                    );
+                } else {
+                    onException.accept(
+                            new AuthException("Sign in failed", error, "See attached exception for more details")
+                    );
+                }
             }
         });
     }
@@ -579,10 +644,10 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                 }
 
                 @Override
-                public void onError(Exception exception) {
+                public void onError(Exception error) {
                     onException.accept(new AuthException(
                             "An error occurred confirming password recovery code",
-                            exception,
+                            error,
                             "See attached exception for more details"
                     ));
                 }
--- a/core/src/main/java/com/amplifyframework/auth/AuthException.java
+++ b/core/src/main/java/com/amplifyframework/auth/AuthException.java
@@ -191,4 +191,384 @@ public class AuthException extends AmplifyException {
             );
         }
     }
+
+    /**
+     * Exception thrown by AWS Cognito Auth.
+     */
+    public static class AWSCognitoAuthException extends AuthException {
+
+        private static final long serialVersionUID = 1L;
+
+        /**
+         * Creates a new AWS cognito auth related exception with a message, root cause, and recovery suggestion.
+         *
+         * @param message            An error message describing why this exception was thrown
+         * @param cause              The underlying cause of this exception
+         * @param recoverySuggestion Text suggesting a way to recover from the error being described
+         */
+        public AWSCognitoAuthException(
+                @NonNull final String message,
+                @NonNull final Throwable cause,
+                @NonNull final String recoverySuggestion
+        ) {
+            super(message, cause, recoverySuggestion);
+        }
+
+        /**
+         * Constructs a new AWS cognito auth related exception using a provided message and an associated error.
+         *
+         * @param message            Explains the reason for the exception
+         * @param recoverySuggestion Text suggesting a way to recover from the error being described
+         */
+        public AWSCognitoAuthException(
+                @NonNull final String message,
+                @NonNull final String recoverySuggestion
+        ) {
+            super(message, recoverySuggestion);
+        }
+
+        /**
+         * Could not perform the action because user not found in the system.
+         */
+        public static class UserNotFoundException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "User not found in the system.";
+            private static final String RECOVERY_SUGGESTION = "Please enter correct username.";
+
+            /**
+             * Default message/recovery suggestion with a cause.
+             *
+             * @param cause The original error.
+             */
+            public UserNotFoundException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because user not confirmed in the system.
+         */
+        public static class UserNotConfirmedException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "User not confirmed in the system.";
+            private static final String RECOVERY_SUGGESTION = "Retry operation and make a confirmation.";
+
+            /**
+             * Default message/recovery suggestion with a cause.
+             * @param cause The original error.
+             */
+            public UserNotConfirmedException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because username already exists in the system.
+         */
+        public static class UsernameExistsException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "Username already exists in the system.";
+            private static final String RECOVERY_SUGGESTION = "Retry operation and enter another username.";
+
+            /**
+             * Default message/recovery suggestion with a cause.
+             * @param cause The original error.
+             */
+            public UsernameExistsException(Throwable cause) {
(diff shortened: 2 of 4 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -279,9 +288,24 @@ public void onResult(SignUpResult result) {
 
                 @Override
                 public void onError(Exception error) {
-                    onException.accept(
-                        new AuthException("Sign up failed", error, "See attached exception for more details")
-                    );
+                    if (error instanceof UsernameExistsException) {
+                        onException.accept(
+                                new AuthException.AWSCognitoAuthException.UsernameExistsException(error.getCause())
```
TrekSoft, line 293: Hey, I think as we talked about on Slack, errors like these (UsernameExists, NetworkException, etc.) aren't specific to Cognito so it'd be better just to have them up at the Global level of AuthException instead of making the subclass of AWSCognitoAuthException for them

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -279,9 +288,24 @@ public void onResult(SignUpResult result) {
...
+                        );
+                    } else if (error instanceof AliasExistsException) {
+                        onException.accept(
+                                new AuthException.AWSCognitoAuthException.AliasExistsException(error.getCause())
```
TrekSoft, line 297: What's the difference between AliasExistsException and UsernameExistsException? Might be better to check for either AliasExists or UsernameExists and return either as our UsernameExists exception.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -302,9 +326,36 @@ public void onResult(SignUpResult result) {
...
+                    );
+                } else if (error instanceof ExpiredCodeException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.CodeExpiredException(error.getCause())
+                    );
+                } else if (error instanceof CodeDeliveryFailureException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.CodeDeliveryFailureException(error.getCause())
+                    );
+                } else if (error instanceof UserNotConfirmedException) {
```
TrekSoft, line 345: Same with all these errors - none of these would be specific to Cognito.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            }
+        }
+
+        /**
+         * Could not perform the action because user not confirmed in the system.
+         */
+        public static class UserNotConfirmedException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "User not confirmed in the system.";
+            private static final String RECOVERY_SUGGESTION = "Retry operation and make a confirmation.";
```
TrekSoft, line 254: "Please confirm user first and then retry operation"

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -302,9 +326,36 @@ public void onResult(SignUpResult result) {
 
             @Override
             public void onError(Exception error) {
-                onException.accept(
-                    new AuthException("Confirm sign up failed", error, "See attached exception for more details")
-                );
+                if (error instanceof UserNotFoundException) {
+                    onException.accept(
+                            new AuthException.AWSCognitoAuthException.UserNotFoundException(error.getCause())
```
richardmcclellan, line 331: Hmm, any way to shorten these class names?  Looks like you are using `AuthException` and `AWSCognitoAuthException` for two purposes - as a namespace AND as a parent class for the inner class exceptions.  Do you need to do both?  
  
Could you instead name it something like `AWSCognitoAuthUserNotFoundException`, and define it as a top level class?

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+             */
+            public UsernameExistsException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because alias already exists in the system.
+         */
+        public static class AliasExistsException extends AWSCognitoAuthException {
```
TrekSoft, line 285: Per earlier comment - this may be redundant with UsernameExists.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            }
+        }
+
+        /**
+         * Could not perform the action because confirmation code has expired.
+         */
+        public static class CodeExpiredException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "Confirmation code has expired.";
+            private static final String RECOVERY_SUGGESTION = "Retry operation and send another confirmation code.";
```
TrekSoft, line 339: "Resend a new confirmation code and then retry operation."

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            }
+        }
+
+        /**
+         * Could not perform the action because number of allowed operation has exceeded.
+         */
+        public class LimitExceededException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "Number of allowed operation has exceeded.";
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 390: Hmm - this recovery suggestion seems wrong. I'd expect it'd be something along the lines of increasing their service limit or waiting awhile before re-attempting.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because Amazon Cognito service cannot find the requested resource.
+         */
+        public static class ResourceNotFoundException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "Amazon Cognito service cannot find the requested resource.";
```
TrekSoft, line 423: I think this could also be brought up to the parent AuthException level and change the message to, "Could not find the requested online resource."

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 424: "Retry with exponential back-off"

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            }
+        }
+
+        /**
+         * Could not perform the action because user made too many failed attempts for a given action.
+         */
+        public static class FailedAttemptsLimitExceededException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "User has made too many failed attempts for a given action.";
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 441: This recovery suggestion is wrong - for this type of error we may want to let the developer specify the recovery suggestion since it depends on the way the service is configured. Is the account now locked until a password reset? Is there a cool off period? If so what is the cool off period?

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+             */
+            public FailedAttemptsLimitExceededException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because user made too many requests for a given operation.
+         */
+        public static class RequestLimitExceededException extends AWSCognitoAuthException {
```
TrekSoft, line 455: This seems like a duplicate of LimitExceeded.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            public RequestLimitExceededException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because Amazon Cognito service encounters an invalid AWS Lambda response
+         * or encounters an unexpected exception with the AWS Lambda service.
+         */
+        public static class LambdaException extends AWSCognitoAuthException {
```
TrekSoft, line 473: This one makes sense to include in the AWSCognito specific exception type.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE =
+                    "Amazon Cognito service encounters an invalid AWS Lambda response " +
```
TrekSoft, line 476: *encountered

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+                            "or encounters an unexpected exception with the AWS Lambda service.";
```
TrekSoft, line 477: *or an unexpected exception...

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 478: "Retry with exponential back-off"

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            }
+        }
+
+        /**
+         * Could not perform the action because device is not tracked.
+         */
+        public static class DeviceNotTrackedException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "Device is not tracked.";
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 495: Not a good recovery suggestion - not sure how this exception would get called but should give a recovery suggestion that would help them avoid getting this.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+             */
+            public DeviceNotTrackedException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because there exists error in loading web UI.
+         */
+        public static class ErrorLoadingUIException extends AWSCognitoAuthException {
```
TrekSoft, line 509: I don't think we'll use this error and can remove it.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because user cancelled the step.
+         */
+        public static class UserCancelledException extends AWSCognitoAuthException {
+            private static final long serialVersionUID = 1L;
+            private static final String MESSAGE = "User cancelled the step.";
```
TrekSoft, line 528: *the flow

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+            private static final String RECOVERY_SUGGESTION = "Turn off and re-attempt this operation.";
```
TrekSoft, line 529: "You can provide an option to retry the operation to the user but should not automatically do it as the user has expressed they wanted to cancel it."

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+             */
+            public UserCancelledException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action because requested resource is not available with the current account setup.
+         */
+        public static class InvalidAccountTypeException extends AWSCognitoAuthException {
```
TrekSoft, line 543: We already have this exception type in the parent class - this can be removed.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
...
+             */
+            public InvalidAccountTypeException(Throwable cause) {
+                super(MESSAGE, cause, RECOVERY_SUGGESTION);
+            }
+        }
+
+        /**
+         * Could not perform the action since request was not completed because of any network related issue.
+         */
+        public static class NetworkException extends AWSCognitoAuthException {
```
TrekSoft, line 560: Same for this one - remove.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -279,9 +288,24 @@ public void onResult(SignUpResult result) {
 
                 @Override
                 public void onError(Exception error) {
```
raphkim, line 290: I think it will be useful to create an `AuthExceptionUtil` class or something along that line to help declutter the plugin code.

See [how iOS handles them](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AuthErrorHelper.swift) for example.

Although, it will end up looking quite different compared to iOS. I imagine that the class will end up looking like:
```
public final class AuthExceptionUtil {
    private AuthExceptionUtil() {}
    public static AuthException resolveSignInError() {
        //...your code here
    }
    public static AuthException resolveSignOutError() {
        //...your code here
    }
    //...more methods to resolve each auth operation
}
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,384 @@ public UnknownException(Throwable cause) {
             );
         }
     }
+
+    /**
+     * Exception thrown by AWS Cognito Auth.
+     */
+    public static class AWSCognitoAuthException extends AuthException {
```
jamesonwilliams, line 198: We shouldn't put this in the `core` module, since it has "Cognito" in the name. Instead, this should live in the `aws-auth-cognito` module, somewhere.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -361,9 +412,23 @@ public void onResult(SignInResult result) {
 
             @Override
             public void onError(Exception error) {
-                onException.accept(
-                    new AuthException("Sign in failed", error, "See attached exception for more details")
-                );
+                if (error instanceof InvalidPasswordException) {
```
jamesonwilliams, line 415: A few thoughts on the error mapping strategy:

You could make a:
```java
final class CognitoErrorMapping {
    private CognitoErrorMapping() {}

    static AWSCognitoAuthException lookup(Exception exception) {
        if (exception instanceof ...) {
            return ...
        }
    }
}
```
and then centralize the mapping in a separate class. This would help keep the length of `AWSCognitoAuthPlugin` down, too.

Secondly, can you include the _entire_ `error` in the new exception's "cause" field, instead of just extracting the `getCause()`? Right now, I think you're removing/masking one layer in the exception trace. But better would be to simply _wrap_ the entire thing, into our types.

core/src/main/java/com/amplifyframework/auth/AuthException.java
jamesonwilliams, line 198: Since you define its subtypes inline in the `AWSCognitoAuthException`, I believe you can mark this exception (and all of its subtypes) as `final` classes.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
TrekSoft: No we don't want to change the top level class - this is one of the main Amplify error tenets that we keep exceptions with the category top level classes so as not to introduce breaking changes.

core/src/main/java/com/amplifyframework/auth/AuthException.java
changxu0306: Got it, a real big lesson😂

jamesonwilliams: Haha :-D

core/src/main/java/com/amplifyframework/auth/AuthException.java
changxu0306: So in this case I'll change the recovery suggestion to "Please check out the service configuration to see the condition of locking."

core/src/main/java/com/amplifyframework/auth/AuthException.java
changxu0306: Will change to "Please check out the device configuration."

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
changxu0306: AliasExistsException is thrown when a user tries to confirm the account with an email or phone number that has already been supplied as an alias from a different account. This exception tells user that an account with this email or phone already exists.

UsernameExistsException is thrown when Amazon Cognito encounters a user name that already exists in the user pool.

So, both of them should be kept.

core/src/main/java/com/amplifyframework/auth/AuthException.java
changxu0306: They are a little bit different. For example, if user sign up a bunch of different accounts and exceed the limit, an RequestLimitExceededException will be thrown. But these operations are not failed.

PR #758, opened 2020-08-26
Developer menu code

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* This PR is to merge all of the code for the developer menu into main.


core/src/main/java/com/amplifyframework/devmenu/PersistentLogStoragePlugin.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.devmenu;
```
TrekSoft, line 16: Don't know if it's worth changing now but was it discussed whether this would make more sense in the logging package vs. under dev menu?

eeatonaws: In the PR to store logs for the developer menu, Jameson left a comment about this: "One could argue either way. I'd say devmenu, since its the only consumer we currently expect for the plugin. If we find another use for it, we can always move it back to the more general location, later."

jamesonwilliams:
> That is _a lot_ of impressive work @eeatonaws !!
> 
> Cat21u1

This guy looks like my first ever pet, Friendly. 😢 Rest in Power, dear friend, Friendly.

PR #759, opened 2020-08-26
feature(rxbindings) Added progress-aware Rx constructs

rjuliano:
*Issue #, if available:*

*Description of changes:*
Created some constructs to support the work necessary to provide a more idiomatic experience for consumers wishing to use RxJava. As a first step, I modified the download operations in `RxStorageCategoryBehavior` to reflect these changes.

### Common
- **RxSingleOperation\<T\>(Interface)** where T is the type expected by the onResult Consumer in the callback-style API.

### Storage-specific
- **RxStorageTransferCallbackMapper\<T>(Interface)** where T is the type expected by the onResult Consumer in the callback-style API.
- **RxProgressAwareSingleOperation<T>(Class)** implements **RxSingleOperation\<T>**  and provides a constructor that accepts **RxStorageTransferCallbackMapper<T, P, E>**. It serves as a representation of an operation that returns a Single and emits progress events.

Internally, **RxProgressAwareSingleOperation** uses a couple of subjects (aka Rx objects you can publish and subscribe) to enqueue results, errors and progress updates received from the callback-style API. The constructor for this class maps the callbacks defined in **RxStorageTransferCallbackMapper** to methods that publish results to the subjects I just mentioned. It also gives us a reference to the `AmplifyOperation` returned by the callback-style API. This is needed in case we need to cancel the operation.

### Updated download APIs.
To validate this approach, I have updated the downloadFile APIs for the storage category in rxbindings.

I created a couple of interfaces to serve almost as type aliases:

Type Alias | Actual Type
-----------|-------------
RxStorageDownloadCallbackMapper|RxProgressAwareCallbackMapper<StorageDownloadFileResult,StorageTransferProgress,StorageException>
RxStorageDownloadOperation|RxProgressAwareSingle<StorageDownloadFileResult, StorageTransferProgress>

Consumers of the API will look something like this.

```java
RxStorageBinding.RxProgressAwareSingleOperation<StorageDownloadFileResult> rxOperation = rxStorage.downloadFile(remoteKey, localFile, StorageDownloadFileOptions.defaultInstance());
rxOperation.observeProgress().subscribe(storageStatus -> ...);
Single<StorageDownloadFileResult> resultSingle = rxOperation.observeResult();
// Use resultSingle however they wish
```


rxbindings/src/main/java/com/amplifyframework/rx/RxAdapters.java
```diff
@@ -100,4 +103,85 @@ private RxAdapters() {}
...
+         */
+        Single<T> observeResult();
+    }
+
+    /**
+     * Interface that should be implemented by reactive-style operations
+     * that emit progress information during processing.
+     * @param <P> The type that represents the progress of a given operation.
+     */
+    interface RxProgressAwareOperation<P> {
```
jamesonwilliams, line 137: Since progress is currently only defined for _Storage_, should we move these new `RxProgressAwareSingle` and `RxProgressAwareOperation` out of `RxAdapters`, and into `RxStorageBinding.java`? (Well, the interface might need to go in `RxStorageCategoryBehavior.java`?)

rxbindings/src/main/java/com/amplifyframework/rx/RxAdapters.java
```diff
@@ -100,4 +103,85 @@ private RxAdapters() {}
...
+        @Override
+        public void cancel() {
+            amplifyOperation.cancel();
+            resultSubject.onComplete();
+            progressSubject.onComplete();
+        }
+
+        @Override
+        public Single<T> observeResult() {
+            Single<T> objectSingle = Single.create(emitter -> {
```
jamesonwilliams, line 176: Just `return Single` without caching to variable?

rxbindings/src/main/java/com/amplifyframework/rx/RxStorageBinding.java
```diff
@@ -113,4 +116,25 @@
...
+            super(callbacks);
+        }
+    }
+
+    /**
+     * Type alias that defines the generic parameters for a download operation.
+     * @see RxAdapters.RxProgressAwareCallbackMapper
+     */
+    interface RxStorageDownloadCallbackMapper
+        extends RxAdapters.RxProgressAwareCallbackMapper<StorageDownloadFileResult,
```
jamesonwilliams, line 136: If we only have this one thing that's progress aware, should we flatten these two interfaces into just one?

rxbindings/src/main/java/com/amplifyframework/rx/RxAdapters.java
rjuliano: That makes sense. I was thinking we'd run across some progress aware API in one of the other categories, but now that I think about it, I can't think of an example.

rxbindings/src/main/java/com/amplifyframework/rx/RxAdapters.java
rjuliano: Doh 🤦

rxbindings/src/main/java/com/amplifyframework/rx/RxStorageBinding.java
rjuliano: Oh yeah...After moving some of this stuff around, I was able to reduce the footprint quite a bit.

rjuliano:
> Nice, @rjuliano! Looks pretty on-point to me.
> 
> If we have a few PRs coming for this body of work, does it make sense to compile it against a feature branch? `origin/rxbindings`, or something?

I think that makes sense. I'll create a branch for it and will make that a target for the PRs related to the RX stuff

PR #763, opened 2020-08-27
Include wrapper APIs for signing in without password

harvsu:
*Issue #, if available:*

*Description of changes:*
Include Auth wrapper APIs corresponding to changes in https://github.com/aws-amplify/aws-sdk-android/pull/2095

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -376,53 +376,11 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     @Override
     public void signIn(
             @Nullable String username,
-            @NonNull AuthSignInOptions options,
-            @NonNull final Consumer<AuthSignInResult> onSuccess,
-            @NonNull final Consumer<AuthException> onException
-    ) {
-        Map<String, String> metadata = null;
-
-        if (options != null && options instanceof AWSCognitoAuthSignInOptions) {
-            metadata = ((AWSCognitoAuthSignInOptions) options).getMetadata();
-        }
-
-        awsMobileClient.signIn(username, metadata, new Callback<SignInResult>() {
-            @Override
-            public void onResult(SignInResult result) {
-                try {
-                    AuthSignInResult newResult = convertSignInResult(result);
-                    fetchAndSetUserId(() -> onSuccess.accept(newResult));
-                } catch (AuthException exception) {
-                    onException.accept(exception);
-                }
-            }
-
-            @Override
-            public void onError(Exception error) {
-                onException.accept(
-                    new AuthException("Sign in failed", error, "See attached exception for more details")
-                );
-            }
-        });
-    }
-
-    @Override
-    public void signIn(
-            @Nullable String username,
             @Nullable String password,
             @NonNull final Consumer<AuthSignInResult> onSuccess,
             @NonNull final Consumer<AuthException> onException
     ) {
         signIn(username, password, AWSCognitoAuthSignInOptions.builder().build(), onSuccess, onException);
-    }
-
-    @Override
-    public void signIn(
-            @Nullable String username,
-            @NonNull final Consumer<AuthSignInResult> onSuccess,
-            @NonNull final Consumer<AuthException> onException
-    ) {
-        signIn(username, AWSCognitoAuthSignInOptions.builder().build(), onSuccess, onException);
     }
 
     @Override
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -375,12 +375,54 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
 
     @Override
     public void signIn(
-            @Nullable String username,
-            @Nullable String password,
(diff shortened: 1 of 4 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -375,14 +375,56 @@ public void onError(Exception error) {
...
+        @NonNull final Consumer<AuthSignInResult> onSuccess,
+        @NonNull final Consumer<AuthException> onException
+    ) {
+        Map<String, String> metadata = null;
+
+        if (options != null && options instanceof AWSCognitoAuthSignInOptions) {
+            metadata = ((AWSCognitoAuthSignInOptions) options).getMetadata();
+        }
+
+        awsMobileClient.signIn(username, metadata, new Callback<SignInResult>() {
```
richardmcclellan, line 389: According to the build log, this line is failing to compile with this error: 
```
incompatible types: Map<String,String> cannot be converted to String
        awsMobileClient.signIn(username, metadata, new Callback<SignInResult>() {
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -375,14 +375,56 @@ public void onError(Exception error) {
 
     @Override
     public void signIn(
-            @Nullable String username,
-            @Nullable String password,
-            @NonNull final Consumer<AuthSignInResult> onSuccess,
-            @NonNull final Consumer<AuthException> onException
+        @Nullable String username,
```
richardmcclellan, line 378: Could you revert this indentation?   As a convention throughout the codebase, we generally indent 8 spaces, when method signatures are wrapped, to differentiate it from the method implementation, which is indented 4 spaces.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -375,14 +375,56 @@ public void onError(Exception error) {
...
+                onException.accept(
+                    new AuthException("Sign in failed", error, "See attached exception for more details")
+                );
+            }
+        });
+    }
+
+    @Override
+    public void signIn(
+        @Nullable String username,
```
richardmcclellan, line 411: Same for these two `signIn` methods (indent 8 spaces instead of 4).

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
harvsu: > According to the build log, this line is failing to compile with this error:
> 
> ```
> incompatible types: Map<String,String> cannot be converted to String
>         awsMobileClient.signIn(username, metadata, new Callback<SignInResult>() {
> ```

This is dependent on https://github.com/aws-amplify/aws-sdk-android. 
Actually, the underlying aws sdk(https://github.com/aws-amplify/aws-sdk-android) also has the same. It also assumes password by default. 
I also created a PR to add the APIs without password in the underlying SDK. Here is the PR https://github.com/aws-amplify/aws-sdk-android. 
This issue will be automatically resolved after the aws-sdk-android merges the PR mentioned above.

jamesonwilliams:
Thanks for the proposal, @harvsu! Our team discussed this PR today. We are curious if you can simply provide `null` for the password, into the existing API? We don't think we need to expose another category API for this.

harvsu:
Thanks @jamesonwilliams for discussing and getting back to me! My proposal comes from the fact that `SRP Password` is a special/additional case in custom authentication per documentation. 
Screen Shot 2020-08-31 at 9 06 11 PM
Screen Shot 2020-08-31 at 9 06 32 PM

https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-authentication-flow.html

Yes, I tried providing `null` simply to the API, but the underling AWSMobileClient still considers it as a password and invokes constructor ```
AuthenticationDetails(String userId, String password, Map<String, String> authenticationParameters,
                                 Map<String, String> validationData) 
```(which sets SRP_A as the `CHALLENGE_NAME` in authentication parameters)instead of ```
AuthenticationDetails(String userId, Map<String, String> authenticationParameters,
            Map<String, String> validationData)
```
More details are in this PR https://github.com/aws-amplify/aws-sdk-android/pull/2095

Please review this PR https://github.com/aws-amplify/aws-sdk-android/pull/2095, which gives better context to this PR.

Having said that, it is still possible to pass `null` for password instead of exposing another API, but the underlying AWSMobileClient.java should check for `null` password inside ```
_signIn(final String username,
                             final String password,
                             final Map<String, String> validationData,
                             final Map<String, String> clientMetadata,
                             final Callback<SignInResult> callback) 
``` method and invoke the respective `AuthenticationDetails` constructor.

TrekSoft:
Wow! Thanks for all your contribution here @harvsu ! As Jameson mentioned, given the low usage of the passwordless case, we want the supported interface to be passing null instead of having a method override for it but it's good to know you've tried this and it's not working properly.

If you could revise your aws-amplify/aws-sdk-android#2095 PR to just make the existing _signIn method handle a null password as indicating the user wants to try passwordless login, that would be great and I'd be happy to review and merge that.

harvsu:
@TrekSoft Awesome! Thanks for the quick response! Will push that change!

harvsu:
@TrekSoft aws-amplify/aws-sdk-android#2095 now includes the requested changes and is ready for review! Thanks!

TrekSoft:
Sweet, I'll close this one and take a look at that one.

PR #768, opened 2020-08-27
feat(rxbindings) Create operation for subscription

rjuliano:
*Issue #, if available:*

*Description of changes:*
Created an operation class to address the subscription use case. The callback-style API invokes an onStart consumer with the subscriptionId. To align with iOS that aspect of the functionality will look a bit different the rxbindings for the API category.

The `RxSubscriptionOperation` will have two methods that return observables. 
- observeSubscriptionData: emits one event for each item received via the subscription.
- observeConnectionState: when the onStart callback is triggered by the API category, we will publish a message to the observable and will save the subscriptionId as a property of the `RxSubscriptionOperation`. This will enable customer to listen for a start event if they wish, and also aligns our async functionality with the  iOS implementation. 

To still provide access to the subscriptionId, the `RxSubscriptionOperation` also has a `getSubscriptionId()`, which consumers can use if they need that data.

### DIFF
--- a/rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
+++ b/rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
@@ -38,7 +38,7 @@ import io.reactivex.rxjava3.subjects.BehaviorSubject;
  * An implementation of the RxApiCategoryBehavior which satisfies the API contract by wrapping
  * {@link ApiCategoryBehavior} in Rx primitives.
  */
-final class RxApiBinding implements RxApiCategoryBehavior {
+public final class RxApiBinding implements RxApiCategoryBehavior {
     private final ApiCategoryBehavior api;
 
     RxApiBinding() {
@@ -51,6 +51,7 @@ final class RxApiBinding implements RxApiCategoryBehavior {
     }
 
     @NonNull
+    @Override
     public <T> Single<GraphQLResponse<T>> query(@NonNull GraphQLRequest<T> graphQlRequest) {
         return toSingle((onResult, onError) -> api.query(graphQlRequest, onResult, onError));
     }
@@ -173,7 +174,13 @@ final class RxApiBinding implements RxApiCategoryBehavior {
         return CancelableBehaviors.toObservable(method);
     }
 
-    static class RxSubscriptionOperation<T> implements Cancelable {
+    /**
+     * A class that represents a subscription operation and exposes
+     * observables for consumers to listen to subscription data and
+     * status events.
+     * @param <T> The type representing the subscription data.
+     */
+    public static final class RxSubscriptionOperation<T> implements Cancelable {
         private BehaviorSubject<ConnectionState> connectionStateSubject;
         private Observable<T> subscriptionData;
         private Cancelable amplifyOperation;
@@ -197,14 +204,30 @@ final class RxApiBinding implements RxApiCategoryBehavior {
             });
         }
 
+        /**
+         * Returns an {@link Observable} which consumers can use to
+         * retrieve data received by the subscription operation.
+         * @return Reference to the {@link Observable} with subscription data.
+         */
         public Observable<T> observeSubscriptionData() {
             return subscriptionData;
         }
 
+        /**
+         * Once the subscription starts, this method returns
+         * the value of the subscriptionId.
+         * @return The value of the subscriptionId.
+         */
         public String getSubscriptionId() {
             return subscriptionId;
         }
 
+        /**
+         * Returns an {@link Observable} which consumers can use to
+         * receive notfication about the status of the subscription connection. Currently,
+         * only {@link ConnectionState#CONNECTED} is emitted.
+         * @return Reference to the {@link Observable} that receives connection events.
+         */
         public Observable<ConnectionState> observeConnectionState() {
             return connectionStateSubject;
         }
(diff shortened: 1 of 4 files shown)


rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -26,16 +26,19 @@
...
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.subjects.BehaviorSubject;
 
 /**
  * An implementation of the RxApiCategoryBehavior which satisfies the API contract by wrapping
  * {@link ApiCategoryBehavior} in Rx primitives.
  */
-final class RxApiBinding implements RxApiCategoryBehavior {
+public final class RxApiBinding implements RxApiCategoryBehavior {
```
jamesonwilliams, line 41: Is this access level escalation required, for the _implementation class_? (The public methods will still be public.)

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -168,4 +173,83 @@
...
         return CancelableBehaviors.toObservable(method);
     }
+
+    /**
+     * A class that represents a subscription operation and exposes
+     * observables for consumers to listen to subscription data and
+     * status events.
+     * @param <T> The type representing the subscription data.
+     */
+    public static final class RxSubscriptionOperation<T> implements Cancelable {
```
jamesonwilliams, line 183: does it need to be public?

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -168,4 +173,83 @@
...
+        private BehaviorSubject<ConnectionState> connectionStateSubject;
+        private Observable<T> subscriptionData;
+        private Cancelable amplifyOperation;
+        private String subscriptionId;
+
+        private Consumer<String> onConnected = new Consumer<String>() {
```
jamesonwilliams, line 189: Maybe do this as an inner class (FooConsumer extends Consumer), and initialize it inside the constructor body? S.t. `private final Consumer<String> onConnected;` is up with the other ordinary declarations?

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -168,4 +173,83 @@
...
+        public Observable<T> observeSubscriptionData() {
+            return subscriptionData;
+        }
+
+        /**
+         * Once the subscription starts, this method returns
+         * the value of the subscriptionId.
+         * @return The value of the subscriptionId.
+         */
+        public String getSubscriptionId() {
```
jamesonwilliams, line 221: Is there any way to bundle this into the `ConnectionState` object?

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
```diff
@@ -168,4 +173,83 @@
...
+        /**
+         * Enum representing connection states of a
+         * subscription operation.
+         */
+        public enum ConnectionState {
+            /**
+             * The subscription successfully established a connection and is
+             * ready to receive data.
+             */
+            CONNECTED
```
jamesonwilliams, line 252: @palpatim notes:
> FWIW, we also notify on "connecting" and "disconnected" states. "Connecting" isn't terribly meaningful, and "disconnected" degrades to a completion on the subscription data publisher, so that may not be necessary in Android

Not sure if we need to do this today -- or can defer.

rxbindings/src/test/java/com/amplifyframework/rx/RxApiBindingTest.java
```diff
@@ -225,12 +228,20 @@ public void subscribeStartsEmitsValuesAndCompletes() throws InterruptedException
         );
 
         // Act: subscribe via binding
-        TestObserver<GraphQLResponse<Model>> observer =
-            rxApi.subscribe(request).test();
+        RxApiBinding.RxSubscriptionOperation<GraphQLResponse<Model>> rxOperation = rxApi.subscribe(request);
```
jamesonwilliams, line 231: I'd add an `import ... ` for `RxSubscriptionOperation..` directly just to cut down on some of this boiler plate a tiny bit.

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
rjuliano: Yeah...that's because anything outside of the package wouldn't have access to the RxSubscriptionOperation object. That being said, rather than making this public, I created a top level class called RxOperations where we can put of the Rx-operation related objects. That, we can make public.

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
rjuliano: This one does, but now lives under RxOperations.

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
rjuliano: Good call on this one. Rather that doing it this way, I created a ConnectionStateEvent that has a connectionState and subscriptionId properties. We can expand on that down the road.

rxbindings/src/main/java/com/amplifyframework/rx/RxApiBinding.java
rjuliano: I looked into this, but it will require some additional plumbing to surface those events from the SubscriptionEndpoint class. I will make note to create a task for this.

PR #770, opened 2020-08-28
Categorize auth errors

changxu0306:
Add `CognitoAuthExceptionConverter` as a utility to centralize error mapping.
Resolve incorrect error messages and recovery suggestions.
Get rid of improper subclass `AWSCognitoAuthException` of `AuthException`. Now all the exception class are children of `AuthException`.

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.cognito.util;
+
+import com.amplifyframework.auth.AuthException;
+
+import com.amazonaws.services.cognitoidentityprovider.model.AliasExistsException;
+import com.amazonaws.services.cognitoidentityprovider.model.CodeDeliveryFailureException;
+import com.amazonaws.services.cognitoidentityprovider.model.CodeMismatchException;
+import com.amazonaws.services.cognitoidentityprovider.model.ExpiredCodeException;
+import com.amazonaws.services.cognitoidentityprovider.model.InvalidParameterException;
+import com.amazonaws.services.cognitoidentityprovider.model.InvalidPasswordException;
+import com.amazonaws.services.cognitoidentityprovider.model.LimitExceededException;
+import com.amazonaws.services.cognitoidentityprovider.model.PasswordResetRequiredException;
+import com.amazonaws.services.cognitoidentityprovider.model.ResourceNotFoundException;
+import com.amazonaws.services.cognitoidentityprovider.model.TooManyFailedAttemptsException;
+import com.amazonaws.services.cognitoidentityprovider.model.UnexpectedLambdaException;
+import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
+import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
+import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
+
+/**
+ * Convert AWS Cognito Exceptions to AuthExceptions.
+ */
+public final class CognitoAuthExceptionConverter {
+
+    /**
+     * Dis-allows instantiation of this class.
+     */
+    private CognitoAuthExceptionConverter() {}
+
+    /**
+     * Lookup method to convert AWS Cognito Exception to AuthException.
+     * @param error Exception thrown by AWSMobileClient
+     * @return AuthException
+     */
+    public static AuthException lookup(Exception error) {
+        if (error instanceof UserNotFoundException) {
+            return new AuthException.UserNotFoundException(error);
+        }
+
+        if (error instanceof UserNotConfirmedException) {
+            return new AuthException.UserNotConfirmedException(error);
+        }
+
+        if (error instanceof UsernameExistsException) {
+            return new AuthException.UsernameExistsException(error);
+        }
+
+        if (error instanceof AliasExistsException) {
+            return new AuthException.AliasExistsException(error);
+        }
+
+        if (error instanceof InvalidPasswordException) {
+            return new AuthException.InvalidPasswordException(error);
+        }
+
+        if (error instanceof InvalidParameterException) {
+            return new AuthException.InvalidParameterException(error);
+        }
+
+        if (error instanceof ExpiredCodeException) {
+            return new AuthException.CodeExpiredException(error);
+        }
+
+        if (error instanceof CodeMismatchException) {
+            return new AuthException.CodeMismatchException(error);
+        }
+
+        if (error instanceof CodeDeliveryFailureException) {
+            return new AuthException.CodeDeliveryFailureException(error);
+        }
+
+        if (error instanceof LimitExceededException) {
+            return new AuthException.LimitExceededException(error);
+        }
+
+        if (error instanceof ResourceNotFoundException) {
+            return new AuthException.ResourceNotFoundException(error);
+        }
+
+        if (error instanceof TooManyFailedAttemptsException) {
+            return new AuthException.FailedAttemptsLimitExceededException(error);
+        }
+
+        if (error instanceof UnexpectedLambdaException) {
+            return new AuthException.LambdaException(error);
+        }
+
+        if (error instanceof PasswordResetRequiredException) {
+            return new AuthException.PasswordResetRequiredException(error);
+        }
+
+        return null;
+    }
+}
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
@@ -15,6 +15,8 @@
 
 package com.amplifyframework.auth.cognito.util;
 
+import androidx.annotation.NonNull;
+
 import com.amplifyframework.auth.AuthException;
 
 import com.amazonaws.services.cognitoidentityprovider.model.AliasExistsException;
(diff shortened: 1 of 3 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -0,0 +1,109 @@
...
+
+        if (error instanceof UnexpectedLambdaException) {
+            return new AuthException.LambdaException(error);
+        }
+
+        if (error instanceof PasswordResetRequiredException) {
+            return new AuthException.PasswordResetRequiredException(error);
+        }
+
+        return null;
```
jamesonwilliams, line 107: You could further simplify your code in the plugin, by providing a default exception, here:
```suggestion
        return new AuthException(fallbackMessage, error, "See attached exception for more details.");
```
Then, you'd only need to pass in an additional `fallbackMessage` when you call the lookup:
```java
AuthException mapped =
    CognitoAuthExceptionConverter.lookup("Fallback message for some situation, here.", error);
```

Note: this is just a suggestion. You can choose not to do this.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,286 @@ public UnknownException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public PasswordResetRequiredException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because Amazon Cognito service cannot find the requested resource.
```
jamesonwilliams, line 387:
```suggestion
     * Could not perform the action because Amplify cannot find the requested resource.
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,286 @@ public UnknownException(Throwable cause) {
...
+        public FailedAttemptsLimitExceededException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because Amazon Cognito service encounters an invalid AWS Lambda response
+     * or encounters an unexpected exception with the AWS Lambda service.
+     */
+    public static class LambdaException extends AuthException {
```
jamesonwilliams, line 425: This type of exception is also implementation specific. (The Auth _category_ doesn't need to involved AWS Lambda, necessarily.) I'd recommend to remove this `LambdaException` entirely.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,286 @@ public UnknownException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public FailedAttemptsLimitExceededException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because Amazon Cognito service encounters an invalid AWS Lambda response
```
jamesonwilliams, line 422: Mentions two details of the implementation -- Cognito, Lambda. That's fine for the _plugin_, but not the category, which is abstract.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
changxu0306: This is a very good suggestion!

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private CognitoAuthExceptionConverter() {}
+
+    /**
+     * Lookup method to convert AWS Cognito Exception to AuthException.
+     * @param fallbackMessage Fallback message to inform failure
+     * @param error Exception thrown by AWSMobileClient
+     * @return AuthException
+     */
+    public static AuthException lookup(String fallbackMessage, Exception error) {
```
TrekSoft, line 50: A bit of nit picking but it was counter intuitive to me seeing this used with the string as the first parameter. It makes it seem like the string is what you're looking up. I think it'd be more intuitive to change the order of these parameters to make clear you're looking up the error and then specifying what message to use. You can use Android Studio to easily refactor the parameter order without having to worry about changing all the places this is used manually.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -0,0 +1,106 @@
...
+        if (error instanceof UserNotConfirmedException) {
+            return new AuthException.UserNotConfirmedException(error);
+        }
+
+        if (error instanceof UsernameExistsException) {
+            return new AuthException.UsernameExistsException(error);
+        }
+
+        if (error instanceof AliasExistsException) {
+            return new AuthException.AliasExistsException(error);
```
TrekSoft, line 64: I missed if you replied to this on the earlier PR but why do we need Alias Exists in addition to UsernameExists?

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,230 @@ public UnknownException(Throwable cause) {
...
+        }
+    }
+
+    /**
+     * Could not perform the action because Amplify cannot find the requested resource.
+     */
+    public static class ResourceNotFoundException extends AuthException {
+        private static final long serialVersionUID = 1L;
+        private static final String MESSAGE = "Could not find the requested online resource.";
+        private static final String RECOVERY_SUGGESTION = "Retry with exponential back-off";
```
TrekSoft, line 392: "... or check your config file to be sure the endpoint is valid."

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
changxu0306: I really like this detailed suggestion😄 Yeah it's just like reading a book and the order of parameters matters a lot.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Dis-allows instantiation of this class.
+     */
+    private CognitoAuthExceptionConverter() {}
+
+    /**
+     * Lookup method to convert AWS Cognito Exception to AuthException.
+     * @param error Exception thrown by AWSMobileClient
+     * @param fallbackMessage Fallback message to inform failure
+     * @return AuthException
```
raphkim, line 48: quick description? e.g.
```
@return AuthException Specific exception for Amplify Auth
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,231 @@ public UnknownException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public CodeDeliveryFailureException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because user enters incorrect confirmation code.
```
raphkim, line 283:
```suggestion
     * Could not perform the action because user entered incorrect confirmation code.
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,231 @@ public UnknownException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public AliasExistsException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because error occurs when delivering the confirmation code.
```
raphkim, line 266:
```suggestion
     * Could not perform the action because error occurred when delivering the confirmation code.
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,231 @@ public UnknownException(Throwable cause) {
...
+         *
+         * @param cause The original error.
+         */
+        public UserNotFoundException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because user not confirmed in the system.
```
raphkim, line 214:
```suggestion
     * Could not perform the action because user is not confirmed in the system.
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,231 @@ public UnknownException(Throwable cause) {
             );
         }
     }
+
+    /**
+     * Could not perform the action because user not found in the system.
```
raphkim, line 196:
```suggestion
     * Could not perform the action because user was not found in the system.
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -191,4 +191,231 @@ public UnknownException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public CodeExpiredException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     * Could not perform the action because there exists incorrect parameters.
```
raphkim, line 318:
```suggestion
     * Could not perform the action because there are incorrect parameters.
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public static AuthException lookup(Exception error, String fallbackMessage) {
```
raphkim, line 50: Prefix with `@NonNull` to let developers know that this operation always returns a non-null exception.
The parameters should also have `@NonNull` and/or `@Nullable` tags.

The above practice should be used for every publicly exposed methods throughout the repo :)

raphkim, line 50:
```suggestion
    @NonNull
    public static AuthException lookup(@NonNull Exception error, @NonNull String fallbackMessage) {
```

Issue #215, opened 2020-01-02
Amplify Initialization and Teardown


jamesonwilliams:
Currently, the [`Amplify` entry-point](https://github.com/aws-amplify/amplify-android/blob/a213c0e2965aaccbe8da2560424e1efd89006fd4/core/src/main/java/com/amplifyframework/core/Amplify.java#L57) is a process-wide singleton. It is initialized via [a single, blocking configure(...) call](https://github.com/aws-amplify/amplify-android/blob/a213c0e2965aaccbe8da2560424e1efd89006fd4/core/src/main/java/com/amplifyframework/core/Amplify.java#L90-L99).

1. Amplify cannot be re-initialized

This is desirable in high-level tests, where there is a need to exercise some various configurations/arrangements, for test purposes. But, the only way to re-initialize the framework is to start with a new heap, i.e., a new process.

2. The user has no control over the resources we allocate.

> "What goes up must come down."

The user may not be integrating Amplify to satisfy a primary business case in their application. If so, they may wish to lazy-load the framework, when it is needed, and save memory, otherwise.

3. The configuration process may involve costly disk/network operations

The DataStore, for example, starts GraphQL subscriptions, sets up SQL databases - a bunch of gnarly stuff. But, since Amplify uses a blocking, synchronous configuration, the caller must block on these, too. If the application’s startup is gated upon the completion of this method, then the application's cold start time will be directly impacted.

## Proposal

Three changes are needed:

1. Make `configure(...)` into an asynchronous call, since it may be expensive. As an added benefit, it will match the contract of our existing category behaviors, which are async.
2. Rename `configure(...)` to `initialize(...)`, to clarify that categories/plugins are actually doing more than setting configuration values at this time.
3. Provide a `terminate()` which should undo any work done by `initialize(...)`.

```
/**
 * Initialize the Amplify framework. This initializes all categories and their
 * registered plugins.
 * @param context An Android Context
 * @param onSuccess Called back when initialization succeeds
 * @param onError Called when there is an error initializing the framework
 */
void initialize(
        Context context,
        Action onSuccess,
        Consumer<AmplifyException> onError);

/**
 * Terminates use of the Amplify framework, freeing any resources
 * used by its categories and plugins. After this is called,
 * Amplify must be re-initialized via
 * {@link #initialize(Context, Action, Consumer<AmplifyException>)}.
 */
void terminate(
        Action onSuccess,
        Consumer<AmplifyException> onError);
```

Example usage:
```
[...]

PR #776, opened 2020-08-31
[core] Add Amplify.unconfigure method that is opposite to configure

budlex:
Amplify.configure configures: plugins, AWSMobileClient, UserAgent, etc. There is no way to reset configuration. Our application is an Android Service that consumes awspinpointconfiguration.json in runtime. A client passes configuration when binded and analytics events as user interacts with the client. Client might change configuration (AWS account) or another client might bind with a different configuration. To handle this case and taking into account heavily used singleton pattern within amplify-android and underlying libraries we need a way to de-configure the singleton and set new parameters.

[aws-analytics-pinpoint] expose startSession and stopSession

Currently sessions are tracked automatically and the logic is bound to Android Activity lifecycle. However, if the code is running in Android Service this path is not invoked. The exposure of methods allows to manually track sessions for library consumers.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AWSPinpointAnalyticsPlugin.java
+++ b/aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/AWSPinpointAnalyticsPlugin.java
@@ -22,6 +22,7 @@ import androidx.annotation.Nullable;
 import androidx.annotation.StringDef;
 import androidx.annotation.VisibleForTesting;
 
+import com.amazonaws.mobileconnectors.pinpoint.analytics.SessionClient;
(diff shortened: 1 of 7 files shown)


jamesonwilliams:
Hi @budlex -- Re-configuring Amplify brings about a range of complexities. Unfortunately, this PR does not address most of them. Our team had decided not to build reconfiguration. Please provide your thoughts on the original design artifact: https://github.com/aws-amplify/amplify-android/discussions/215.

raphkim:
> Hi @budlex -- Re-configuring Amplify brings about a range of complexities. Unfortunately, this PR does not address most of them. Our team had decided not to build reconfiguration. Please provide your thoughts on the original design artifact: #215.

Closing until the above concern is addressed.

Issue #778, opened 2020-08-31
Seems out of place that we are checking ModelOperation in isOwnerArgumentRequired


saltonmassally:
Looking at com.amplifyframework.api.aws.AWSApiPlugin#isOwnerArgumentRequired it seems out of place that we are checking ModelOperation  before passing ownerField in our subscription request.

1) Correct me if wrong but  i don't think a user needs CREATE, UPDATE, DELETE, perms to be able to receive ON_CREATE, ON_UPDATE, ON_DELETE subscriptions. I don't have to give a user UPDATE permission to my api for them to be notified of an update made to a record they observe.

2) amplify cli still build the subscriptions with the owner field... (will send examples shortly)

take this model definition as an example:
```
type Share
  @model
  @auth(
    rules: [
      { allow: groups, groups: ["ForbiddenGroup"] }
      { allow: owner, ownerField: "accessUsers", operations: [read] }
      { allow: private, provider: iam }
    ]
  )
 {
  id: ID!
  accessUsers: [String!]! # users that have access to the record
}
```

this is what cli outputs
```
  onCreateShare(accessUsers: String): Share @aws_subscribe(mutations: ["createShare"]) @aws_iam @aws_cognito_user_pools
  onUpdateShare(accessUsers: String): Share @aws_subscribe(mutations: ["updateShare"]) @aws_iam @aws_cognito_user_pools
  onDeleteShare(accessUsers: String): Share @aws_subscribe(mutations: ["deleteShare"]) @aws_iam @aws_cognito_user_pools
```

Trying to sub from the android app will fail as owner field will not be passed in given the conditions set in the method in question.

I would think that all we need to check in this case is ModelOperations.READ

richardmcclellan:
Thanks for reporting this @saltonmassally !  I believe your understanding is accurate.  I'm going to resolve this issue, as I think it is a duplicate of https://github.com/aws-amplify/amplify-android/issues/778, but I will take a look at your PR shortly as well.   Before we finalize this change, we need to make sure iOS and JS are in sync with this logic as well.

PR #779, opened 2020-08-31
fix[datastore] owner with readonly access should be able to sub to on_delete, on_create, on_update

saltonmassally:
solves https://github.com/aws-amplify/amplify-android/issues/778

*Issue #, if available:*
#778 

*Description of changes:*
If user has read access to a model they should be able to observe on_delete, on_create, and on_update events

Units tests are failing now some of which will be unnecessary if we are proceeding with this. WIll fix if we are at a consensus on this change


richardmcclellan:
Closing in favor of https://github.com/aws-amplify/amplify-android/pull/807

PR #782, opened 2020-09-01
Fix/duplicating subscriptions

saltonmassally:
*Issue #, if available:*

SubscriptionProcessor.startSubscriptions can be called multiple times as datastore transitions through
STOPPED, LOCAL_ONLY, SYNC_VIA_API states. Guard ensures that we are not duplicating subscriptions.

At worse this can lead to appsync's maximum allowed subscription threshold of 100 being exceeded which throws an error

*Description of changes:*
Check if subscriptions already started before running SubscriptionProcessor.startSubscriptions

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
@@ -86,6 +86,15 @@ final class SubscriptionProcessor {
      * Start subscribing to model mutations.
      */
     synchronized void startSubscriptions() {
+        // let's ensure that we are not duplicating subscriptions here
+        if (ongoingOperationsDisposable.size() > 0) {
+            if  (!ongoingOperationsDisposable.isDisposed()){
+                LOG.info("Subscription already started");
+                return;
+            } else {
+                ongoingOperationsDisposable.clear();
+            }
+        }
         int subscriptionCount = modelProvider.models().size() * SubscriptionType.values().length;
         // Create a latch with the number of subscriptions are requesting. Each of these will be
         // counted down when each subscription's onStarted event is called.
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
@@ -88,7 +88,7 @@ final class SubscriptionProcessor {
     synchronized void startSubscriptions() {
         // let's ensure that we are not duplicating subscriptions here
         if (ongoingOperationsDisposable.size() > 0) {
-            if  (!ongoingOperationsDisposable.isDisposed()){
+            if (!ongoingOperationsDisposable.isDisposed()) {
                 LOG.info("Subscription already started");
                 return;
             } else {


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -86,6 +86,15 @@
      * Start subscribing to model mutations.
      */
     synchronized void startSubscriptions() {
+        // let's ensure that we are not duplicating subscriptions here
+        if (ongoingOperationsDisposable.size() > 0) {
```
rjuliano, line 90: This `ongoingOperationsDisposable` is also used to track the disposable created in the `startDrainingMutationBuffer` method (line 207). It's possible (although not likely) that the disposable created when initializing subscriptions is disposed, but the other one (from line 207) is not, so you'd still have 1 item in the `ongoingOperationsDisposable` variable.

In that case, subscriptions wouldn't be restarted. Maybe one thing we could do here is track those disposables separately rather than lumping them together. If we do that, we would also need to update the `stopAllSubscriptionActivity` method to dispose of both.

rjuliano (changes requested):
Thanks again for the contribution @saltonmassally.

jamesonwilliams:
> SubscriptionProcessor.startSubscriptions can be called multiple times as datastore transitions through
> STOPPED, LOCAL_ONLY, SYNC_VIA_API states. Guard ensures that we are not duplicating subscriptions.

Perhaps the best way to address this is with synchronization on the state transitions in the `Orchestrator`. This way, none of the child components need special intelligence, we can just implement safe transition logic in one place.

This would probably mean adding some "transitioning" states into the state machine.

saltonmassally:
> This would probably mean adding some "transitioning" states into the state machine.
I like the approach you propose. Because CRUD operations are wrapped with `com.amplifyframework.datastore.AWSDataStorePlugin#beforeOperation`  `orchestrator.start()` gets called many times if you fire a bunch of CRUD operations before `orchestrator` is started. 

How do we proceed, close this, and open an issue?

jamesonwilliams:
> How do we proceed, close this, and open an issue?

That sounds like a good plan. I think we have a little design work remaining. We can work on that together, with @rjuliano.

The last big round of effort on the Orchestrator was in https://github.com/aws-amplify/amplify-android/pull/662. As of that PR, the Orchestrator now uses a state machine, to transition between a "current state" and a "target state." The currently permissible states are `STOPPED`, `LOCAL_ONLY`, and `SYNC_VIA_API`.

The work needed is to add 1-2 more "_currently in-transition_" states. While the orchestrator is in a transition state, calls to start/stop it should block until arriving at a terminal state (`STOPPED`, `LOCAL_ONLY`, `SYNC_VIA_API`.) There are going to be a number of multi-threaded concerns to consider. We can simplify some of that by consistently running the main "what's the state? transition to..." logic always on the same thread, to create a strong guarantee of sequential execution.

saltonmassally:
@jamesonwilliams doing some work on this (actually already have and it works in my case) will send a PR but wanted your thoughts on the approach.

Thread calling `Orchestrator.start` acquires a lock only if it is not held by another thread at the time of invocation. I don't see the reason why we need other threads to queue waiting for this to be released so If the lock is held by another thread then this method will return immediately

Thread calling `Orchestrator.stop` acquires a lock. If another thread holds that lock we wait.

Using java.util.concurrent.locks.ReentrantLock we guarantee that transitions are sequential and `Orchestrator.start` also does not block other threads if it is "in transition"

Your thoughts

rjuliano:
@saltonmassally I created an issue with some background info on the problem. https://github.com/aws-amplify/amplify-android/issues/853

PR #730, opened 2020-08-16
fix[aws-datastore] selection set for nested custom types

saltonmassally:
This PR fixes a bug in the selection set involving fields of custom types.
*Issue #, if available:*

#702 

Custom types were not being transversed to build selection sets.  

*Description of changes:*

This PR recursively introspects a custom type and adds its properties to the selection set.

*TODO* 

- [x] Tests to ensure that mutation selection sets with inputs are being properly built. 
- [x] Tests to ensure serialization and deserialization to/from SQLite works as expected

*Not Handled**
-  there are instances that might lead to infinite recursion, e.g. circular relationships so we may want to set a hard max_depth unrelated to datastore's model association max_depth .

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #783, opened 2020-09-01
fix(datastore) Add missing converters for AWSTimestamp used by SQLiteStorageAdapter

richardmcclellan:
After this gets merged, we should be able to re-merge @saltonmassally 's PR (https://github.com/aws-amplify/amplify-android/pull/730) which adds support for non-model types, as well as https://github.com/aws-amplify/amplify-android/pull/670, which updates the representation of `lastChangedAt` from `long` to `AWSTimestamp`.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/TypeConverter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/TypeConverter.java
@@ -64,6 +64,7 @@ public final class TypeConverter {
         JAVA_TO_SQL.put(JavaFieldType.DATE, SQLiteDataType.TEXT);
         JAVA_TO_SQL.put(JavaFieldType.DATE_TIME, SQLiteDataType.TEXT);
         JAVA_TO_SQL.put(JavaFieldType.TIME, SQLiteDataType.TEXT);
+        JAVA_TO_SQL.put(JavaFieldType.TIMESTAMP, SQLiteDataType.INTEGER);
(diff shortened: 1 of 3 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/TypeConverter.java
```diff
@@ -64,6 +64,7 @@ private TypeConverter() {
         JAVA_TO_SQL.put(JavaFieldType.DATE, SQLiteDataType.TEXT);
         JAVA_TO_SQL.put(JavaFieldType.DATE_TIME, SQLiteDataType.TEXT);
         JAVA_TO_SQL.put(JavaFieldType.TIME, SQLiteDataType.TEXT);
+        JAVA_TO_SQL.put(JavaFieldType.TIMESTAMP, SQLiteDataType.INTEGER);
```
TrekSoft, line 67: Just to double check, Integer is correct vs. Long or Float?

TrekSoft (approved):
Cat20u2

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/TypeConverter.java
richardmcclellan: Yep, this is correct.  SQLite doesn't have different types for an int vs a long, there is only `INTEGER`. It can hold up to 64 bits, which is the same as a Java `long`, so it works for either.

Issue #591, opened 2020-06-22
ConflictUnhandled error when using API and Datastore in the same project


KayIlory:
On 17th June, I was able to do update via Amplify with no issues. From Sunday, the 21st June 2020, I am no longer able to do this. The values do not get updated and I get errors.

Steps to reproduce.. 
Set api conflict resolution to auto merge

```graphql
type UserTest @model
{
    id: ID!
    gender: Gender!
}
```

### Android class
```java
 public void createUser() {
        UserTest userTest = UserTest.builder().gender(Gender.Male).build();
        Amplify.API.mutate(ModelMutation.create(userTest),
                response -> {
                    if(response.getData() != null) {
                        updateUser(response.getData());
                    }
                }, error -> {
                    Log.e("MyAmplifyApp", "Create failed", error);
                });
    }

    public void updateUser(UserTest userTest)
    {
        UserTest _userTest = userTest.copyOfBuilder().gender(Gender.Female).build();
        Amplify.API.mutate(ModelMutation.update(_userTest),
                response -> {
                    if(response.getData() != null) {
                        Log.i("Output", response.getData().toString());
                    }
                }, error -> {
                    Log.e("MyAmplifyApp", "Create failed", error);
                });

    }
```
It doesn't update the sex from Male to Female. 

![image](https://user-images.githubusercontent.com/51476926/85336767-1dd9c200-b4d7-11ea-8925-9e4980c42ccd.png)

KayIlory:
Now I found a new BUG, i insert one record but two records get inserted instead of 1.

KayIlory:
waitForSignIn: userState:SIGNED_IN
I/Output: UserTest {id=a236299d-efd1-43ac-aaac-508c2e24be9c, gender=Male} 

After an update to try and change Gender to Female and doesnt work..

KayIlory:
E/Delete Error:: GraphQLResponse.Error{message='Conflict resolver rejects mutation.', locations='[GraphQLLocation{line='1', column='56'}]', path='[GraphQLPathSegment{value='deleteUserTest'}]', extensions='{errorInfo=null, data={gender=Male, id=76253e92-b18d-4f1e-b364-a766e8e32ab0}, errorType=ConflictUnhandled}'}

When trying to deleted a newly created user..

KayIlory:
public void createUser() {
        UserTest userTest = UserTest.builder().gender(Gender.Male).build();
[...]

PR #786, opened 2020-09-02
feat(api): allow pinning of api name in datastore

jamesonwilliams:
If you have multiple APIs configured in your
`amplifyconfiguration.json`, you won't be able to use the DataStore.
There's no deep reason for this, it is just a simple shortcoming of the
configuration.

Before this change, the DataStore would call API methods which
auto-select a first-and-only configured API.

After this change, the user can optionally "pin" the DataStore to a
*specific* API (other than the automatically detected one), by adding
the following `"datastore"` block to `amplifyconfiguration.json`:

```json
{
    ...
    "api": {
        "plugins": {
            "awsApiPlugin": {
                "[API NAME 1]": {
                    ...
                },
                "[API NAME 2]": {
                    ...
                }
            }
        }
    },
    ...
    "datastore": {
        "plugins": {
            "awsDataStorePlugin": {
                "apiName": "[API NAME 1]"
            }
        }
    }
    ...
}
```

If this value is *not* specified, there is no change in behavior.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -94,7 +94,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
             modelProvider,
             modelSchemaRegistry,
             sqliteStorageAdapter,
-            AppSyncClient.via(api),
+            AppSyncClient.via(api, () -> pluginConfiguration.getApiName()),
             () -> pluginConfiguration,
             () -> api.getPlugins().isEmpty() ? Orchestrator.Mode.LOCAL_ONLY : Orchestrator.Mode.SYNC_VIA_API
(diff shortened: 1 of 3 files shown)


richardmcclellan:
A few thoughts to consider:
 - Have we had any customers requests for this yet?  Doesn't mean it's not useful, though I haven't seen any yet.  It would be useful to me as a library developer, since I would be able to test the API and DataStore category functionalities in the same sample app.  I don't think it really addresses #591 at all - that issue is about customers who have _one_ AppSync API, and want the ability to use both `Amplify.DataStore`  and `Amplify.API` to interact with it.
 - How would this work with multi-auth?  For example, if I have more than one API, I might also have more than one Cognito user pool.  Whatever solution we come up with for supporting multiple plugin instances for a single category should ideally work for all categories.  @rjuliano any thoughts on this?

jamesonwilliams:
Closing for now until we can come back and evaluate some of the additional DX topics raised in the reviews.

PR #659, opened 2020-07-21
Create logs UI for developer menu

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* Created the developer menu logs UI shown in the picture below.

Screen Shot 2020-07-23 at 9 54 38 AM

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #788, opened 2020-09-02
feat(core): filter developer menu logs by log level

eeatonaws:
*Issue #, if available:* N/A

*Description of changes:* On the developer menu, adds the ability to filter the logs by log level (in addition to searching the logs).

Dev Menu New Logs Screen

Dev Menu Filter Logs Menu


core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
```diff
@@ -54,11 +73,55 @@ public boolean onQueryTextSubmit(String searchText) {
...
+                                    @Nullable ContextMenu.ContextMenuInfo menuInfo) {
+        super.onCreateContextMenu(menu, view, menuInfo);
+        requireActivity().getMenuInflater().inflate(R.menu.dev_menu_logs_menu, menu);
+        menu.findItem(logLevelMenuId).setChecked(true);
+    }
+
+    @Override
+    public boolean onContextItemSelected(MenuItem item) {
+        int itemId = item.getItemId();
+        if (itemId == R.id.all_logs) {
```
TrekSoft, line 99: This is a perfect place to use a Switch statement instead of the if else chain.

TrekSoft: Also, I wonder if there's a way you can simply include the LogLevel value in each MenuItem.

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
```diff
@@ -54,11 +73,55 @@ public boolean onQueryTextSubmit(String searchText) {
...
+        return true;
+    }
+
+    /**
+     * Display the logs that contain the provided search query and
+     * whose log level matches the selected log level.
+     */
+    private void displayFilteredLogs() {
+        logsText.setText(R.string.placeholder_logs);
+        logsText.setText(developerMenu.getFilteredLogs(searchQuery, selectedLogLevel));
```
TrekSoft, line 125: What's the purpose for setting the placeholder text first before setting it to getFilteredLogs?

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
eeatonaws: I did try to use a Switch statement instead of all the if else's, but resource IDs cannot be used in switch statements.

core/src/main/java/com/amplifyframework/devmenu/DevMenuLogsFragment.java
eeatonaws: In a previous PR Jameson commented that it would be helpful to display some placeholder text (i.e. "Search results will appear here...") in case it takes time to display the logs (https://github.com/aws-amplify/amplify-android/pull/659#discussion_r459085658), although that comment was more related to when you initially navigate to the logs view. If it does take time to display the logs, I thought it could be helpful to display the placeholder text indicating that search results are being retrieved, otherwise it may seem like the search functionality is broken.

TrekSoft (approved):
Cat27u1

PR #798, opened 2020-09-04
feat(Auth): Add user attribute operations (Fetch/Update/Verify/Confirm user attributes)

changxu0306:
- Add `AuthUpdateAttributeResult` under `core/java/com.amplifyframework.auth/result`

- Add `AuthNextUpdateAttributeStep`, `AuthUpdateAttributeStep` under `core/java/com.amplifyframework.auth/result/step`

- Add `AWSCognitoAuthUpdateAttributeOptions` under `aws-auth-cognito/java/com.amplifyframework.auth.cognito/options`

- Update `AuthCategory` and `AuthCategoryBehavior`(add signatures of user attribute operations)

- Implemented `fetchUserAttributes`, `updateUserAttribute`, `updateUserAttributes`, `confirmUserAttribute`, `resendUserAttributeConfirmationCode` methods in `AWSCognitoAuthPlugin`

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -692,6 +692,30 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     }
 
     @Override
+    public void updatePassword(
+            @NonNull String oldPassword,
+            @NonNull String newPassword,
+            @Nullable Action onSuccess,
+            @Nullable Consumer<AuthException> onException
+    ) {
+        awsMobileClient.changePassword(oldPassword, newPassword, new Callback<Void>() {
+            @Override
+            public void onResult(Void result) {
+                onSuccess.call();
+            }
+
+            @Override
+            public void onError(Exception error) {
+                onException.accept(new AuthException(
+                        "Failed to change password",
+                        error,
+                        "See attached exception for more details"
+                ));
+            }
+        });
+    }
+
+    @Override
     public void fetchUserAttributes(
             @NonNull Consumer<List<AuthUserAttribute>> onSuccess,
             @NonNull Consumer<AuthException> onError
@@ -878,30 +902,6 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                         ));
                     }
                 });
-    }
-
-    @Override
-    public void updatePassword(
-            @NonNull String oldPassword,
-            @NonNull String newPassword,
-            @Nullable Action onSuccess,
-            @Nullable Consumer<AuthException> onException
-    ) {
-        awsMobileClient.changePassword(oldPassword, newPassword, new Callback<Void>() {
-            @Override
-            public void onResult(Void result) {
-                onSuccess.call();
-            }
-
-            @Override
-            public void onError(Exception error) {
-                onException.accept(new AuthException(
-                        "Failed to change password",
-                        error,
-                        "See attached exception for more details"
-                ));
-            }
-        });
     }
 
     @Override
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -88,6 +88,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -700,11 +701,9 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
             @Override
             public void onResult(Map<String, String> result) {
                 List<AuthUserAttribute> userAttributes = new ArrayList<>();
-                for (String userAttributeKey : result.keySet()) {
+                for (Map.Entry<String, String> entry : result.entrySet()) {
                     userAttributes.add(new AuthUserAttribute(
-                            new AuthUserAttributeKey(userAttributeKey),
-                            result.get(userAttributeKey)
-                    ));
+                            new AuthUserAttributeKey(entry.getKey()), entry.getValue()));
                 }
                 onSuccess.accept(userAttributes);
             }
@@ -712,7 +711,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
             @Override
             public void onError(Exception error) {
                 onError.accept(new AuthException(
-                        "Failed to fetch user attributes",
+                        "Failed to fetch user attributes since user is not authenticated",
                         error,
                         "See attached exception for details"
                 ));
@@ -737,36 +736,61 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                 new Callback<List<UserCodeDeliveryDetails>>() {
                     @Override
                     public void onResult(List<UserCodeDeliveryDetails> result) {
-                        if (result != null && isUserAttributeVerifiedByCode(userAttributeKeyString)) {
-                            for (UserCodeDeliveryDetails details : result) {
-                                if (details.getAttributeName().equals(userAttributeKeyString)) {
+                        if (result != null) {
+                            if (isValidUserAttribute(attribute.getKey())) {
+                                if (isUserAttributeVerifiedByCode(attribute.getKey())) {
+                                    for (UserCodeDeliveryDetails details : result) {
+                                        if (details.getAttributeName().equals(userAttributeKeyString)) {
+                                            onSuccess.accept(new AuthUpdateAttributeResult(
+                                                    true,
+                                                    new AuthNextUpdateAttributeStep(
+                                                            AuthUpdateAttributeStep.CONFIRM_ATTRIBUTE_WITH_CODE,
+                                                            Collections.emptyMap(),
+                                                            convertCodeDeliveryDetails(details))
+                                            ));
+                                        } else {
+                                            onError.accept(new AuthException(
+                                                    "Returned an incorrect attribute name",
+                                                    "See attached exception for more details"
+                                            ));
+                                        }
+                                    }
+                                } else {
+                                    onError.accept(new AuthException(
+                                            "Code was returned with an attribute not verified by code",
+                                            "See attached exception for more details"
+                                    ));
+                                }
+                            } else {
+                                onError.accept(new AuthException(
+                                        "Invalid attribute name",
+                                        "See attached exception for more details"
+                                ));
+                            }
+                        }
+
+                        if (result == null) {
+                            if (isValidUserAttribute(attribute.getKey())) {
+                                if (!isUserAttributeVerifiedByCode(attribute.getKey())) {
                                     onSuccess.accept(new AuthUpdateAttributeResult(
                                             true,
                                             new AuthNextUpdateAttributeStep(
-                                                AuthUpdateAttributeStep.CONFIRM_ATTRIBUTE_WITH_CODE,
-                                                Collections.emptyMap(),
-                                                convertCodeDeliveryDetails(details))
+                                                    AuthUpdateAttributeStep.DONE,
+                                                    Collections.emptyMap(),
+                                                    null)
                                     ));
                                 } else {
                                     onError.accept(new AuthException(
-                                            "Returned a wrong attribute name",
-                                            "See attached exception for more details"));
+                                            "Code was not returned with an attribute verified by code",
+                                            "See attached exception for more details"
+                                    ));
                                 }
+                            } else {
+                                onError.accept(new AuthException(
+                                        "Invalid attribute name",
+                                        "See attached exception for more details"
+                                ));
                             }
-                        } else if (result == null &&
-                               !isUserAttributeVerifiedByCode(userAttributeKeyString)) {
-                            onSuccess.accept(new AuthUpdateAttributeResult(
-                                    true,
-                                    new AuthNextUpdateAttributeStep(
-                                            AuthUpdateAttributeStep.DONE,
-                                            Collections.emptyMap(),
-                                            null)
-                            ));
-                        } else {
-                            onError.accept(new AuthException(
-                                    "Code delivery failed",
-                                    "See attached exception for more details"
-                            ));
                         }
                     }
 
@@ -788,10 +812,10 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
             @NonNull Consumer<Map<AuthUserAttributeKey, AuthUpdateAttributeResult>> onSuccess,
             @NonNull Consumer<AuthException> onError
     ) {
-        for (AuthUserAttribute userAttribute : attributes) {
+        for (AuthUserAttribute attribute : attributes) {
             Map<String, String> userAttributeMap = new HashMap<>();
-            String userAttributeKeyString = userAttribute.getKey().getKeyString();
-            String userAttributeValue = userAttribute.getValue();
+            String userAttributeKeyString = attribute.getKey().getKeyString();
+            String userAttributeValue = attribute.getValue();
             userAttributeMap.put(userAttributeKeyString, userAttributeValue);
 
             awsMobileClient.updateUserAttributes(
@@ -799,43 +823,69 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                     new Callback<List<UserCodeDeliveryDetails>>() {
                         @Override
                         public void onResult(List<UserCodeDeliveryDetails> result) {
-                            if (result != null && isUserAttributeVerifiedByCode(userAttributeKeyString)) {
-                                for (UserCodeDeliveryDetails details : result) {
-                                    if (details.getAttributeName().equals(userAttributeKeyString)) {
+                            if (result != null) {
+                                if (isValidUserAttribute(attribute.getKey())) {
+                                    if (isUserAttributeVerifiedByCode(attribute.getKey())) {
+                                        for (UserCodeDeliveryDetails details : result) {
+                                            if (details.getAttributeName().equals(userAttributeKeyString)) {
+                                                Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap =
+                                                        new HashMap<>();
+                                                resultMap.put(new AuthUserAttributeKey(userAttributeKeyString),
+                                                        new AuthUpdateAttributeResult(
+                                                                true,
+                                                                new AuthNextUpdateAttributeStep(
+                                                                        AuthUpdateAttributeStep.
+                                                                                CONFIRM_ATTRIBUTE_WITH_CODE,
+                                                                        Collections.emptyMap(),
+                                                                        convertCodeDeliveryDetails(details))
+                                                        ));
+                                                onSuccess.accept(resultMap);
+                                            } else {
+                                                onError.accept(new AuthException(
+                                                        "Returned an incorrect attribute name",
+                                                        "See attached exception for more details"));
+                                            }
+                                        }
+                                    } else {
+                                        onError.accept(new AuthException(
+                                                "Code was returned with an attribute not verified by code",
+                                                "See attached exception for more details"
+                                        ));
+                                    }
+                                } else {
+                                    onError.accept(new AuthException(
+                                            "Invalid attribute name",
+                                            "See attached exception for more details"
+                                    ));
+                                }
+                            }
+
+                            if (result == null) {
+                                if (isValidUserAttribute(attribute.getKey())) {
+                                    if (!isUserAttributeVerifiedByCode(attribute.getKey())) {
                                         Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap =
                                                 new HashMap<>();
                                         resultMap.put(new AuthUserAttributeKey(userAttributeKeyString),
                                                 new AuthUpdateAttributeResult(
                                                         true,
                                                         new AuthNextUpdateAttributeStep(
-                                                                AuthUpdateAttributeStep.CONFIRM_ATTRIBUTE_WITH_CODE,
+                                                                AuthUpdateAttributeStep.DONE,
                                                                 Collections.emptyMap(),
-                                                                convertCodeDeliveryDetails(details))
+                                                                null)
                                                 ));
                                         onSuccess.accept(resultMap);
                                     } else {
                                         onError.accept(new AuthException(
-                                                "Returned a wrong attribute name",
-                                                "See attached exception for more details"));
-                                    }
-                                }
-                            } else if (result == null && !isUserAttributeVerifiedByCode(userAttributeKeyString)) {
-                                Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap =
-                                        new HashMap<>();
-                                resultMap.put(new AuthUserAttributeKey(userAttributeKeyString),
-                                        new AuthUpdateAttributeResult(
-                                                true,
-                                                new AuthNextUpdateAttributeStep(
-                                                        AuthUpdateAttributeStep.DONE,
-                                                        Collections.emptyMap(),
-                                                        null)
+                                                "Code was not returned with an attribute verified by code",
+                                                "See attached exception for more details"
                                         ));
-                                onSuccess.accept(resultMap);
-                            } else {
-                                onError.accept(new AuthException(
-                                        "Code delivery failed",
-                                        "See attached exception for more details"
-                                ));
+                                    }
+                                } else {
+                                    onError.accept(new AuthException(
+                                            "Invalid attribute name",
+                                            "See attached exception for more details"
+                                    ));
+                                }
                             }
                         }
 
@@ -1179,15 +1229,34 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
             : null;
     }
 
-    private boolean isUserAttributeVerifiedByCode(String userAttributeKey) {
-        if (userAttributeKey.equals(AuthUserAttributeKey.email().getKeyString())) {
-            return true;
-        }
+    private boolean isValidUserAttribute(AuthUserAttributeKey userAttributeKey) {
+        List<AuthUserAttributeKey> validUserAttributeKeys = Arrays.asList(
+                AuthUserAttributeKey.phoneNumber(),
+                AuthUserAttributeKey.email(),
+                AuthUserAttributeKey.address(),
+                AuthUserAttributeKey.birthdate(),
+                AuthUserAttributeKey.familyName(),
+                AuthUserAttributeKey.gender(),
+                AuthUserAttributeKey.middleName(),
+                AuthUserAttributeKey.givenName(),
+                AuthUserAttributeKey.locale(),
+                AuthUserAttributeKey.nickname(),
+                AuthUserAttributeKey.picture(),
+                AuthUserAttributeKey.preferredUsername(),
+                AuthUserAttributeKey.profile(),
+                AuthUserAttributeKey.updatedAt(),
+                AuthUserAttributeKey.website(),
+                AuthUserAttributeKey.zoneInfo(),
+                AuthUserAttributeKey.name()
+        );
+        return validUserAttributeKeys.contains(userAttributeKey);
+    }
 
-        if (userAttributeKey.equals(AuthUserAttributeKey.phoneNumber().getKeyString())) {
-            return true;
-        }
-
-        return false;
+    private boolean isUserAttributeVerifiedByCode(AuthUserAttributeKey userAttributeKey) {
+        List<AuthUserAttributeKey> codeVerifiedAttributeKeys = Arrays.asList(
+                AuthUserAttributeKey.phoneNumber(),
+                AuthUserAttributeKey.email()
+        );
+        return codeVerifiedAttributeKeys.contains(userAttributeKey);
     }
 }
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -779,36 +779,36 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
                 new Callback<List<UserCodeDeliveryDetails>>() {
                     @Override
                     public void onResult(List<UserCodeDeliveryDetails> result) {
-                        Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap =
-                                new HashMap<>();
+                        Map<String, UserCodeDeliveryDetails> codeDetailsMap = new HashMap<>();
+                        Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap = new HashMap<>();
 
                         for (UserCodeDeliveryDetails details : result) {
-                            Map<String, UserCodeDeliveryDetails> codeDetailsMap =
-                                    Collections.singletonMap(details.getAttributeName(), details);
-                            for (String attributeKey : attributesMap.keySet()) {
-                                if (codeDetailsMap.containsKey(attributeKey)) {
-                                    resultMap.put(new AuthUserAttributeKey(attributeKey),
-                                            new AuthUpdateAttributeResult(
-                                                    true,
-                                                    new AuthNextUpdateAttributeStep(
-                                                            AuthUpdateAttributeStep.
-                                                                    CONFIRM_ATTRIBUTE_WITH_CODE,
-                                                            Collections.emptyMap(),
-                                                            convertCodeDeliveryDetails(details))
-                                            ));
-                                } else {
-                                    resultMap.put(new AuthUserAttributeKey(attributeKey),
-                                            new AuthUpdateAttributeResult(
-                                                    true,
-                                                    new AuthNextUpdateAttributeStep(
-                                                            AuthUpdateAttributeStep.
-                                                                    DONE,
-                                                            Collections.emptyMap(),
-                                                            null)
-                                            ));
-                                }
-                                onSuccess.accept(resultMap);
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -260,6 +262,66 @@ void confirmResetPassword(
...
+     * @param onError Error callback
+     */
+    void updateUserAttribute(
+            @NonNull AuthUserAttribute attribute,
+            @NonNull Consumer<AuthUpdateAttributeResult> onSuccess,
+            @NonNull Consumer<AuthException> onError);
+
+    /**
+     * Update multiple user attributes.
+     * @param attributes Attribute to be updated
```
TrekSoft, line 288: *Attributes

core/src/main/java/com/amplifyframework/auth/result/AuthUpdateAttributeResult.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Wraps the result of a user attribute operation.
+ */
+public class AuthUpdateAttributeResult {
+    private final boolean isUpdated;
+    private final AuthNextUpdateAttributeStep nextStep;
+
+    /**
+     * Wraps the result of a user attribute operation.
```
TrekSoft, line 33: *of an update user attribute operation

core/src/main/java/com/amplifyframework/auth/result/AuthUpdateAttributeResult.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param isUpdated True if the user attribute has now been updated, False otherwise.
+     *                  Check {@link #getNextStep()} for details on additional steps that may be required.
+     * @param nextStep Details about the next step in the sign up process (or whether the flow is now done).
```
TrekSoft, line 36: *in the update attribute process

core/src/main/java/com/amplifyframework/auth/result/step/AuthNextUpdateAttributeStep.java
```diff
@@ -0,0 +1,127 @@
...
+ * additional details which can optionally accompany it. If there is no next step, {@link #getUpdateAttributeStep()}
+ * will have a value of DONE.
+ */
+public final class AuthNextUpdateAttributeStep {
+    private final AuthUpdateAttributeStep updateAttributeStep;
+    private final Map<String, String> additionalInfo;
+    private final AuthCodeDeliveryDetails codeDeliveryDetails;
+
+    /**
+     * Gives details on the next step, if there is one, in the sign in flow.
```
TrekSoft, line 40: *in the update attribute flow

core/src/main/java/com/amplifyframework/auth/result/step/AuthNextUpdateAttributeStep.java
```diff
@@ -0,0 +1,127 @@
...
+     * Details about how a code was sent, if relevant to the current step.
+     * @return Details about how a code was sent, if relevant to the current step - null otherwise
+     */
+    @Nullable
+    public AuthCodeDeliveryDetails getCodeDeliveryDetails() {
+        return codeDeliveryDetails;
+    }
+
+    /**
+     * When overriding, be sure to include signInStep, additionalInfo, and codeDeliveryDetails in the hash.
```
TrekSoft, line 83: *include updateAttributeStep

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+        );
+
+        awsMobileClient.updateUserAttributes(
+                userAttribute,
+                new Callback<List<UserCodeDeliveryDetails>>() {
+                @Override
+                public void onResult(List<UserCodeDeliveryDetails> result) {
+                    if (result != null) {
+                        for (UserCodeDeliveryDetails details : result) {
+                            if (details.getAttributeName().equals(attribute.getKey().getKeyString())) {
```
TrekSoft, line 743: There's a problem here - if this condition isn't satisfied, neither success nor error will be called. Will a UserCodeDeliveryDetails be returned by AWSMobileClient for every kind of attribute, including those that don't need to be confirmed?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+                                onSuccess.accept(new AuthUpdateAttributeResult(
+                                        true,
+                                        new AuthNextUpdateAttributeStep(
+                                                AuthUpdateAttributeStep.CONFIRM_ATTRIBUTE_WITH_CODE,
```
TrekSoft, line 747: This is correct if there's a UserCodeDeliveryDetails object returned only for attributes that are being confirmed. If it is returned in an empty state or something like that for attributes that don't need confirmation this is wrong. Most attributes will simply be done after you call update attribute without needing a confirm state.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+                                                Collections.emptyMap(),
+                                                convertCodeDeliveryDetails(details))
+                                ));
+                            }
+                        }
+                    } else {
+                        onError.accept(new AuthException(
+                                "Code delivery failed",
```
TrekSoft, line 755: Again, this assumes all attributes are going to have codes sent which they don't - in fact most don't.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+                            error,
+                            "See attached exception for more details"
+                            )
+                    );
+                }
+            });
+    }
+
+    @Override
+    public void updateUserAttributes(
```
TrekSoft, line 774: Same comments as above apply to this method

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+    public void resendUserAttributeConfirmationCode(
+            @NonNull AuthUserAttributeKey attributeKey,
+            @NonNull Consumer<AuthCodeDeliveryDetails> onSuccess,
+            @NonNull Consumer<AuthException> onError
+    ) {
+        String attributeName = attributeKey.getKeyString();
+        awsMobileClient.verifyUserAttribute(attributeName, new Callback<UserCodeDeliveryDetails>() {
+            @Override
+            public void onResult(UserCodeDeliveryDetails result) {
+                if (result.getAttributeName().equals(attributeName)) {
```
TrekSoft, line 840: Again, you need to handle the case where it doesn't equal it - perhaps returning an error if that's proper. But currently it'll just hang and never return a success or error callback if this condition is false.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+                new Callback<Void>() {
+                    @Override
+                    public void onResult(Void result) {
+                        onSuccess.call();
+                    }
+
+                    @Override
+                    public void onError(Exception error) {
+                        onError.accept(new AuthException(
+                                "An error occurred confirming update user attribute",
```
TrekSoft, line 877: Remove the word "update"

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -1227,4 +1196,16 @@ private AuthCodeDeliveryDetails convertCodeDeliveryDetails(UserCodeDeliveryDetai
                     details.getAttributeName())
             : null;
     }
+
+    private boolean isUserAttributeVerifiedByCode(String userAttributeKey) {
+        if (userAttributeKey.equals(AuthUserAttributeKey.email().getKeyString())) {
+            return true;
+        }
+
+        if (userAttributeKey.equals(AuthUserAttributeKey.phoneNumber().getKeyString())) {
```
jamesonwilliams, line 1187: It is better practice to write:
```java
if (definitelyNotNull.equals(possiblyNull)) {
    // No NullPointerException!
}
```
Than:
```java
if (possiblyNull.equals(definitelyNotNul)) {
    // Uh oh, might have NPE
}
```

Here, I think `AuthUserAttributeKey.phoneNumber().getKeyString()` is always non null -- but what about `userAttributeKey`? 

(Same comment for `if` check above)

------

Other thought -- you could do like:
```java
List<String> codeVerifiedAttributeKeys = Arrays.asList(
    AuthUserAttributeKey.phoneNumber().getKeyString(),
    AuthUserAttributeKey.email().getKeyString()
);
return codeVerifiedAttributeKeys.contains(/* provided */ userAttributeKey);
```

4 lines instead of 10 🤷‍♂️

changxu0306: I got your idea. Yeah I should always avoid any possible NPE. 😭

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,274 @@ public void onError(Exception error) {
...
+                }
+                onSuccess.accept(userAttributes);
+            }
+
+            @Override
+            public void onError(Exception error) {
+                onError.accept(new AuthException(
+                        "Failed to fetch user attributes",
+                        error,
+                        "See attached exception for details"
```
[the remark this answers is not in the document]
changxu0306: Yes, exactly due to `CognitoNotAuthorizedException`

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,274 @@ public void onError(Exception error) {
...
+        awsMobileClient.updateUserAttributes(
+                userAttributeMap,
+                new Callback<List<UserCodeDeliveryDetails>>() {
+                    @Override
+                    public void onResult(List<UserCodeDeliveryDetails> result) {
+                        if (result != null) {
+                            if (isValidUserAttribute(attribute.getKey())) {
+                                if (isUserAttributeVerifiedByCode(attribute.getKey())) {
+                                    for (UserCodeDeliveryDetails details : result) {
+                                        if (details.getAttributeName().equals(userAttributeKeyString)) {
```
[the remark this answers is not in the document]
changxu0306: Adopted solution 1! :-D

TrekSoft:
Running this sample, I'm only seeing the non-verifiable attributes (name and familyName) showing up in the results - not email (even though I see from fetching the attributes that it did get updated):

```java
 List<AuthUserAttribute> attributes = new ArrayList<>();
            attributes.add(new AuthUserAttribute(AuthUserAttributeKey.email(), "ddaudeli@amazon.com"));
            attributes.add(new AuthUserAttribute(AuthUserAttributeKey.name(), "David"));
            attributes.add(new AuthUserAttribute(AuthUserAttributeKey.familyName(), "Daudelin"));
            Amplify.Auth.updateUserAttributes(attributes, result -> Log.i(TAG, result.toString()), error -> Log.e(TAG, error.toString()));
```

changxu0306:
> Running this sample, I'm only seeing the non-verifiable attributes (name and familyName) showing up in the results - not email (even though I see from fetching the attributes that it did get updated):
> 
> ```java
>  List<AuthUserAttribute> attributes = new ArrayList<>();
>             attributes.add(new AuthUserAttribute(AuthUserAttributeKey.email(), "ddaudeli@amazon.com"));
>             attributes.add(new AuthUserAttribute(AuthUserAttributeKey.name(), "David"));
>             attributes.add(new AuthUserAttribute(AuthUserAttributeKey.familyName(), "Daudelin"));
>             Amplify.Auth.updateUserAttributes(attributes, result -> Log.i(TAG, result.toString()), error -> Log.e(TAG, error.toString()));
> ```

If your email attribute is already set as "ddaudeli@amazon.com", or you have updated email attribute with this email before, you'll get

TrekSoft:
Hey, we should leave the pull request open and just commit more changes to it. Per our call, we found that what I'd discovered was an edge case where it currently doesn't return a result object if it's a verifiable attribute which has already updated but it should to match the non-verifiable attribute behavior (and what the customer expects). 

Also, we discussed that we should only be making one call to AWSMobileCients updateAttributes method with our list of attributes and only call the onSuccess and onError callbacks once with the results of what we get back

changxu0306:
> Hey, we should leave the pull request open and just commit more changes to it. Per our call, we found that what I'd discovered was an edge case where it currently doesn't return a result object if it's a verifiable attribute with the same value as before but it should be updated to return a result object with a DONE next step like it does for a non-verifiable attribute behavior (and what the customer expects).
> 
> Also, we discussed that we should only be making one call to AWSMobileCients updateAttributes method with our list of attributes and only call the onSuccess and onError callbacks once with the results of what we get back

Sorry I clicked close PR by mistake... I just want to close the comment I was typing. 😭

jamesonwilliams (changes requested):
Since you are adding new functionalities to the `AuthCategoryBehavior`, we will to surface the same functionality in the Rx version of the API, the `RxAuthCategoryBehavior`.

-----

Also need unit/component tests in the `AuthComponentTest`, and the `RxAuthBindingTest`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,250 @@ public void onError(Exception error) {
...
+                    userAttributes.add(new AuthUserAttribute(
+                            new AuthUserAttributeKey(entry.getKey()), entry.getValue()));
+                }
+                onSuccess.accept(userAttributes);
+            }
+
+            @Override
+            public void onError(Exception error) {
+                onError.accept(new AuthException(
+                        "Failed to fetch user attributes since user is not authenticated",
```
TrekSoft, line 714: Hmm - this message is not always accurate. I would remove the "since user is not authenticated" piece since there could be other reasons it fails (e.g. no internet) and then per Jameson's suggestion set the suggestion parameter below to "Ensure that you are logged in and online." instead of "See attached exception..."

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,250 @@ public void onError(Exception error) {
...
+            @NonNull Consumer<AuthUpdateAttributeResult> onSuccess,
+            @NonNull Consumer<AuthException> onError
+    ) {
+
+        awsMobileClient.updateUserAttributes(
+                Collections.singletonMap(attribute.getKey().getKeyString(), attribute.getValue()),
+                new Callback<List<UserCodeDeliveryDetails>>() {
+                    @Override
+                    public void onResult(List<UserCodeDeliveryDetails> result) {
+                        if (!isValidUserAttribute(attribute.getKey().getKeyString())) {
```
TrekSoft, line 734: If the user reached this success callback, that means they sent an attribute to update which Cognito accepted and successfully updated. We certainly don't want to throw an error in that case. 

Also, Cognito supports custom attributes in addition to the standard ones we have listed so we don't know whether an attribute is valid or not - hence you can't have a working isValidUserAttribute method.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
changxu0306: Make sense!

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,250 @@ public void onError(Exception error) {
...
+                            onError.accept(new AuthException(
+                                    "Invalid attribute name",
+                                    "See attached exception for more details"
+                            ));
+                        }
+
+                        if (!isUserAttributeVerifiedByCode(attribute.getKey().getKeyString())) {
```
TrekSoft, line 741: This isn't the best way to handle this because what if Cognito adds verification methods for other attributes in the future? What we know is that the onResult method only gets called if all attributes (in this case just the one) are updated successfully. The list type result object contains an entry for any attribute(s) which now need to be confirmed with a sent code and contains no entry for the others. So the logic here for updating one attribute should be as simple as:

onResult:
   if(result.size() == 0) return result with DONE status
   else return result with CONFIRM_ATTRIBUTE_WITH_CODE status

onError:
   return error

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,250 @@ public void onError(Exception error) {
...
+        for (AuthUserAttribute attribute : attributes) {
+            attributesMap.put(attribute.getKey().getKeyString(), attribute.getValue());
+        }
+
+        awsMobileClient.updateUserAttributes(
+                attributesMap,
+                new Callback<List<UserCodeDeliveryDetails>>() {
+                    @Override
+                    public void onResult(List<UserCodeDeliveryDetails> result) {
+                        for (String attributeKey : attributesMap.keySet()) {
```
TrekSoft, line 808: Similar to above, the logic inside the onResult method should go something like this:

Create a resultMap<AuthUserAttributeKey, AuthUpdateAttributeResult> object upfront.

Loop through the objects inside the `result` object and add them to a codeDetailsMap<String, UserCodeDeliveryDetails> keyed by the attribute string name so you can easily check whether a given attribute exists in it. 

Loop through each attribute in the `attributes` list and for each one:

If it exists in the `codeDetailsMap` add an entry for it to `resultMap` with a status of CONFIRM_ATTRIBUTE_WITH_CODE
else add an entry to `resultMap` for it with a status of DONE.

There's no way there can be an error if we're inside the onResult callback.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +692,250 @@ public void onError(Exception error) {
...
+    public void resendUserAttributeConfirmationCode(
+            @NonNull AuthUserAttributeKey attributeKey,
+            @NonNull Consumer<AuthCodeDeliveryDetails> onSuccess,
+            @NonNull Consumer<AuthException> onError
+    ) {
+        String attributeName = attributeKey.getKeyString();
+        awsMobileClient.verifyUserAttribute(attributeName, new Callback<UserCodeDeliveryDetails>() {
+            @Override
+            public void onResult(UserCodeDeliveryDetails result) {
+                if (result.getAttributeName().equals(attributeName)) {
```
TrekSoft, line 890: So I actually think a better option here is to remove this check. If it comes back with a successful response, it means it recognized the attribute you sent and resent the code for it. You can simply return the converted response without checking this.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -955,4 +1204,16 @@ private AuthCodeDeliveryDetails convertCodeDeliveryDetails(UserCodeDeliveryDetai
                     details.getAttributeName())
             : null;
     }
+
+    private boolean isValidUserAttribute(String userAttributeKey) {
+        return AuthUserAttributeKey.authUserAttributeKeyList().contains(userAttributeKey);
+    }
+
+    private boolean isUserAttributeVerifiedByCode(String userAttributeKey) {
```
TrekSoft, line 1212: Both these two methods can be removed given my comments above.

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
```diff
@@ -195,6 +198,32 @@ public static AuthUserAttributeKey custom(String attributeKey) {
         return new AuthUserAttributeKey(attributeKey);
     }
 
+    /**
+     * Method for getting a list of all valid user attribute keys.
+     * @return a list of user attribute keys
+     */
+    public static List<String> authUserAttributeKeyList() {
```
TrekSoft, line 205: This also can be removed now

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -757,22 +738,15 @@ public void onResult(List<UserCodeDeliveryDetails> result) {
                                             Collections.emptyMap(),
                                             null)
                             ));
-                        }
-
-                        for (UserCodeDeliveryDetails details : result) {
-                            if (details.getAttributeName().equals(attribute.getKey().getKeyString())) {
+                        } else {
+                            for (UserCodeDeliveryDetails details : result) {
```
TrekSoft, line 742: What is this loop for?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -687,6 +691,197 @@ public void onError(Exception error) {
...
+        awsMobileClient.updateUserAttributes(
+                attributesMap,
+                new Callback<List<UserCodeDeliveryDetails>>() {
+                    @Override
+                    public void onResult(List<UserCodeDeliveryDetails> result) {
+                        Map<AuthUserAttributeKey, AuthUpdateAttributeResult> resultMap =
+                                new HashMap<>();
+
+                        for (UserCodeDeliveryDetails details : result) {
+                            Map<String, UserCodeDeliveryDetails> codeDetailsMap =
```
TrekSoft, line 786: Defining a singleton map here defeats the purpose of having a map. My suggested logic was to do a loop through the `UserCodeDeliveryDetails` in `result` and add them all to a map before looping through the `attributesMap` to avoid having a nested loop like this one.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
changxu0306: Since result is a list, we must go through the it to get the detail (even though in this case there is only one result)

TrekSoft: Since there's only one result you get it by calling `result.get(0)`

changxu0306: Make sense. I've already changed that

TrekSoft:
Code looks good now - only thing left at this point is to add the Rx methods and unit tests. Feel free to ping either @jamesonwilliams or I if you have questions on this and we'd be happy to give you an intro to how these are done in the codebase.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -831,14 +855,7 @@ public void resendUserAttributeConfirmationCode(
...
             public void onResult(UserCodeDeliveryDetails result) {
-                if (result.getAttributeName().equals(attributeName)) {
-                    onSuccess.accept(convertCodeDeliveryDetails(result));
-                } else {
-                    onError.accept(new AuthException(
-                            "Returned an incorrect attribute name",
-                            "See attached exception for more details"
-                    ));
-                }
+                onSuccess.accept(convertCodeDeliveryDetails(result));
```
TrekSoft, line 858: Good catch - thanks for fixing this.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -553,6 +557,156 @@ public void confirmResetPassword() throws AuthException {
         verify(mobileClient).confirmForgotPassword(eq(NEW_PASSWORD), eq(CONFIRMATION_CODE), any());
     }
 
+    /**
+     * Tests that fetchUserAttributes method of the Auth wrapper of AWSMobileClient (AMC) calls
+     * AMC.getUserAttributes and get to fetch the user attributes.
```
TrekSoft, line 562: "and get to fetch the user attributes" ?? wording

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -553,6 +557,156 @@ public void confirmResetPassword() throws AuthException {
...
+     * Also ensures that in the onResult case, the success callback receives a valid Map which maps
+     * AuthUserAttributeKey into AuthUpdateAttributeResult, and in the onError case, the error call
+     * back receives an AuthException with the root cause attached.
+     * @throws AuthException test fails if this gets thrown since method should succeed
+     */
+    @Test
+    public void updateUserAttributes() throws AuthException {
+        List<AuthUserAttribute> attributes = new ArrayList<>();
+        AuthUserAttributeKey attributeKey = new AuthUserAttributeKey(ATTRIBUTE_KEY);
+        attributes.add(new AuthUserAttribute(attributeKey, ATTRIBUTE_VAL));
```
TrekSoft, line 628: To properly test this method as contrasted with the previous one, you should test passing in another attribute as well which doesn't get UserCodeDeliveryDetails back and verify that the correct result object gets created for it with a DONE status.

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
```diff
@@ -322,20 +336,6 @@ void confirmUserAttribute(
...
-     * @param onSuccess Success callback
-     * @param onError Error callback
-     */
-    void updatePassword(
-            @NonNull String oldPassword,
-            @NonNull String newPassword,
-            @NonNull Action onSuccess,
-            @NonNull Consumer<AuthException> onError
-    );
-
```
TrekSoft, line 276: Just out of curiosity what was the reason behind changing the order of this method?

rxbindings/src/main/java/com/amplifyframework/rx/RxAuthBinding.java
```diff
@@ -166,6 +171,36 @@ public Completable updatePassword(@NonNull String oldPassword, @NonNull String n
             delegate.updatePassword(oldPassword, newPassword, onComplete, onError));
     }
 
+    @Override
+    public Single<List<AuthUserAttribute>> fetchUserAttributes() {
+        return toSingle((delegate::fetchUserAttributes));
```
TrekSoft, line 176: Remove double parantheses

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
changxu0306: Yeah it should be "*to obtain a list of user attributes."😅

core/src/main/java/com/amplifyframework/auth/AuthCategoryBehavior.java
changxu0306: Previously, the series of user attribute methods were located between `confirmResetPassword()` and `updatePassword()`. Now I want to put the methods related to password operations together.

TrekSoft (approved):
Approved pending the one comment

Cat5u4

rxbindings/src/main/java/com/amplifyframework/rx/RxAuthBinding.java
```diff
@@ -166,6 +171,36 @@ public Completable updatePassword(@NonNull String oldPassword, @NonNull String n
...
+        return toSingle(((onResult, onError) ->
+                delegate.resendUserAttributeConfirmationCode(attributeKey, onResult, onError)));
+    }
+
+    @Override
+    public Completable confirmUserAttribute(@NonNull AuthUserAttributeKey attributeKey,
+                                            @NonNull String confirmationCode) {
+        return toCompletable(((onComplete, onError) ->
+                delegate.confirmUserAttribute(attributeKey, confirmationCode, onComplete, onError)));
+    }
```
raphkim, line 202: all the extra parentheses should be removed

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
```diff
@@ -310,6 +315,71 @@ public void updatePassword(
...
+
+    /**
+     * Resend user attribute confirmation code to verify user attribute synchronously.
+     * @param attributeKey The user attribute key
+     * @return result object
+     * @throws AuthException exception
+     */
+    public AuthCodeDeliveryDetails resendUserAttributeConfirmationCode(
+            @NonNull AuthUserAttributeKey attributeKey) throws AuthException {
+        return Await.<AuthCodeDeliveryDetails, AuthException>result(((onResult, onError) -> {
```
raphkim, line 363: extra parentheses

testutils/src/main/java/com/amplifyframework/testutils/sync/SynchronousAuth.java
```diff
@@ -310,6 +315,71 @@ public void updatePassword(
         );
     }
 
+    /**
+     * Fetch a list of user attributes synchronously.
+     * @return List of user attributes upon successful fetch
+     * @throws AuthException exception
+     */
+    public List<AuthUserAttribute> fetchUserAttribute() throws AuthException {
+        return Await.<List<AuthUserAttribute>, AuthException>result(((onResult, onError) -> {
```
raphkim, line 324: extra parentheses

raphkim (approved):
Noticed a few occasions where you used redundant parentheses that you should clean up, but LGTM otherwise!

TrekSoft (approved):
Looks good

Cat7u4

core/src/main/java/com/amplifyframework/auth/AuthUserAttributeKey.java
```diff
@@ -46,7 +46,7 @@
      * Construct a new instance of AuthUserAttributeKey.
      * @param attributeKey the attribute key to use in the new object.
      */
-    protected AuthUserAttributeKey(String attributeKey) {
+    public AuthUserAttributeKey(String attributeKey) {
```
jamesonwilliams, line 49: You don't have to make this public. You can use `AuthUserAttributeKey.custom(CUSTOM_ATTR)`, or one of the existing `AuthUserAttributeKey.address()`, etc.

core/src/main/java/com/amplifyframework/auth/result/AuthUpdateAttributeResult.java
```diff
@@ -0,0 +1,101 @@
...
+import androidx.core.util.ObjectsCompat;
+
+import com.amplifyframework.auth.result.step.AuthNextUpdateAttributeStep;
+
+import java.util.Objects;
+
+/**
+ * Wraps the result of an update user attribute operation.
+ */
+public class AuthUpdateAttributeResult {
```
jamesonwilliams, line 28: If you don't extend this class anywhere, make this `public final class`.

rxbindings/src/main/java/com/amplifyframework/rx/RxAuthCategoryBehavior.java
```diff
@@ -237,6 +242,48 @@
...
+     *         {@link AuthException} on failure
+     */
+    Single<AuthCodeDeliveryDetails> resendUserAttributeConfirmationCode(@NonNull AuthUserAttributeKey attributeKey);
+
+    /**
+     * Confirm user attribute.
+     * @param attributeKey The user attribute key
+     * @param confirmationCode The confirmation code the user received after starting
+     *                         the resendUserAttributeConfirmationCode process
+     * @return An Rx {@link Completable} which completes successfully if userAttributeKey is confirmed,
```
jamesonwilliams, line 282:
```suggestion
     * @return An Rx {@link Completable} which completes successfully if attributeKey is confirmed,
```

rxbindings/src/test/java/com/amplifyframework/rx/RxAuthBindingTest.java
```diff
@@ -796,6 +811,178 @@ public void testFetchDevices() throws InterruptedException {
...
+    /**
+     * Tests that a successful request to fetch user attributes will propagate a completion
+     * back through the binding.
+     * @throws InterruptedException  If test observer is interrupted while awaiting terminal event
+     */
+    @Test
+    public void testFetchUserAttributes() throws InterruptedException {
+        // Arrange an invocation of the success Action
+        List<AuthUserAttribute> expected = Collections.singletonList(
+                new AuthUserAttribute(new AuthUserAttributeKey(ATTRIBUTE_KEY), ATTRIBUTE_VAL)
```
jamesonwilliams, line 823: Can this just be `AuthUserAttributeKey.email()`?

jamesonwilliams (approved):

rxbindings/src/test/java/com/amplifyframework/rx/RxAuthBindingTest.java
changxu0306: Yes. But I was thinking about using the existing fields defined at the beginning.

PR #810, opened 2020-09-09
feature: user can provide a datastore config

jamesonwilliams:
Previous work created a `DataStoreConfiguration`, and created a
two-phase configuration approach, wherein a file-based configuration can
be selectively overridden by a Java/Kotlin-provided configuration.

However, this work did not surface a public API for the user to actually
input their configuration.

As of this commit, it can be done by:
```kotlin
AWSDataStorePlugin(DataStoreConfiguration.builder()
    .syncMaxRecords(1_000_000)
    .build())
```


rjuliano (approved):
Ship it!!! 🚢 

Yeah...the only existing constructor to accept that parameter is private. When I added this, I tested it by passing it to the sub-components that used it, but not via the plugin 🤦  doh

richardmcclellan (approved):

Issue #800, opened 2020-09-04
Developer Menu can't be disabled in Debug Mode


estiukamobi:
When integrating '**_com.amplifyframework:aws-auth-cognito:1.3.1_**' I have noticed that a Developer Menu Screen has been introduced when the application is in Debug mode. It appears if the device is shaken.

However, the developer Menu is enabled by default and cannot be disabled, also the shake detector is way too sensitive
https://github.com/aws-amplify/amplify-android/blob/f345bda690eec84d5e83fecb4efba7e50bcf8c11/core/src/main/java/com/amplifyframework/core/Amplify.java#L145

KayIlory:
I am also facing the same issues. Anyway to disable this 
![image](https://user-images.githubusercontent.com/51476926/92325755-f3fb5a80-f044-11ea-9350-b4bfc35bb90f.png)

jpignata:
Howdy! That will only show up in debug mode. We'll circle back next week on an option to disable and tweaks to the shake sensitivity.

24Lathiya:
@estiukamobi Android application generating two app shortcuts with same icon and name. One is opening app were another open menu like [KayIlory's Issue](https://github.com/aws-amplify/amplify-android/issues/800#issuecomment-687777020)

Please help @jamesonwilliams

jpignata:
Heya @24Lathiya -- are you seeing this behavior with our 1.3.1 release?

24Lathiya:
@jpignata OMG this issue gone with 1.3.1 thanks.

jamesonwilliams:
Resolving as per https://github.com/aws-amplify/amplify-android/issues/800#issuecomment-688847165

jpignata:
I think we should discuss two items:

1) Do we have an option to disable this functionality in debug?
2) Should we tweak the sensitivity?

richardmcclellan:
Here is the [ReactNative ShakeDetector implementation](https://github.com/facebook/react-native/blob/aee88b6843cea63d6aa0b5879ad6ef9da4701846/ReactAndroid/src/main/java/com/facebook/react/common/ShakeDetector.java), that is used for showing the ReactNative debug menu.  Copying the sensitivity values would probably make sense.  The comments indicate they only trigger on detecting a "rage shake". 😂

kwatkins:
This is still in 1.3.1... And bad form adding without a disable flag, it's super annoying.

jpignata:
[...]

PR #821, opened 2020-09-11
feat(core): Adds option to disable dev menu

TrekSoft:
Addresses #800 by decreasing default sensitivity of dev menu and adding the ability to disable it altogether even when in debug mode (it's always disabled in prod mode). To disable it, replace your current command, `Amplify.configure(getApplicationContext())` with the following:

```java
AmplifyConfiguration config = AmplifyConfiguration.builder(getApplicationContext()).devMenuEnabled(false).build();
Amplify.configure(config, getApplicationContext());
```


core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -142,7 +144,24 @@ public static void configure(@NonNull final AmplifyConfiguration configuration,
             // Configure User-Agent utility
             UserAgent.configure(configuration.getPlatformVersions());
 
-            DeveloperMenu.singletonInstance(context).enableDeveloperMenu();
+            boolean disableDevMenu = false;
+
+            try {
+                ApplicationInfo ai = context.getPackageManager().getApplicationInfo(
```
richardmcclellan, line 150: I believe our checkstyle prevents variable names less than 3 characters, so. maybe renaming this to `applicationInfo`

TrekSoft: It passed checkstyle but I can change it

richardmcclellan (commented):
Could we change the name of the flag from `disable_dev_menu` to `dev_menu_enabled`?   Then we would define the default value (`true`) in `core/src/main/AndroidManifest.xml`, and customers could override it to be `false` if desired.  A couple reasons:

 - Negating the meaning (`enable` instead of `disable`) makes it easier to understand - one less negative to think about.
 - Having the default defined in the library itself is more self documenting, and will make it easier for customers to figure out how to disable it if desired.
 - `enabled` at the end vs `enable` at the beginning aligns better with Android naming conventions

I realize we should try and get this out to customers quickly, but before we ship it, can we ask iOS for a quick review on this flag, and verify sure they could implement something similar?

core/src/main/java/com/amplifyframework/core/Amplify.java
jamesonwilliams: It is currently configured do dis-allow them on public APIs. But, I don't think we have a check for internal implementation. We could add it? (I personally prefer to use >=3 everywhere FWIW)

TrekSoft:
> Could we change the name of the flag from `disable_dev_menu` to `dev_menu_enabled`? Then we would define the default value (`true`) in `core/src/main/AndroidManifest.xml`, and customers could override it to be `false` if desired.

The issue with this is it make it more complicated and more invasive for the user to override it because then they have to add the android tools namespace to their manifest as well. The way it is now, it's a simple one line add.

richardmcclellan:
> The issue with this is it make it more complicated and more invasive for the user to override it because then they have to add the android tools namespace to their manifest as well. The way it is now, it's a simple one line add.

Hmm, sort of a valid concern I suppose.  Adding the android-tools namespace is fairly common though, and would be pretty easy for an Android developer to do.

I'm actually not too sure about using the AndroidManifest at all.  We don't currently put flags here - we either read them from amplifyconfiguration.json, or pass them in when creating plugins, or on Amplify.configure.  Asking customers to put a flag in the AndroidManifest is adding yet another place for configuration to go.  The AndroidManifest used to be where all the configuration went, many years ago, but a lot of it has moved to Gradle now.

So, two other options would be:
1. Ask customers to add a flag to their build.gradle (https://stackoverflow.com/a/17201265/282279).  This would still be yet another place for configuration to go.
2. Add a boolean to AmplifyConfiguration.java, `devMenuEnabled`, so that customers can set it to false, and pass it in via Amplify.configure();

I actually kinda like option 2 the best, because we're not asking them to make change in some new place.  Thoughts?

jamesonwilliams:
> 2. Add a boolean to `AmplifyConfiguration.java`, `devMenuEnabled`, so that customers can set it to false, and pass it in via `Amplify.configure();`
> 
> I actually kinda like option 2 the best, because we're not asking them to make change in some new place. Thoughts?

I agree. We already have two configuration mechanisms -- `amplifyconfiguration.json`, and the `AmplifyConfiguration.builder()`. Can't we just use these?

TrekSoft:
Talked to @richardmcclellan some more and found a solution using Amplify Configuration that makes sense. Also talked with Jithin and they actually have the dev menu as opt in since iOS wasn't able to make it automatic so they're good with the change too.

richardmcclellan (approved):
🚢

jamesonwilliams (commented):
I see two uses of a `true` as a default for "is it on." However, I don't see that these consider the app's debuggable configuraiton? @eeatonaws has a utility method to inspect and determine the app's debuggability. I think you want to use it's output as the default value, instead of a literal `true` or `false.`

richardmcclellan:
> I see two uses of a `true` as a default for "is it on." However, I don't see that these consider the app's debuggable configuraiton? @eeatonaws has a utility method to inspect and determine the app's debuggability. I think you want to use it's output as the default value, instead of a literal `true` or `false.`

Hmm, if we did this, then customers would have the ability to enable the debug menu in production, which we definitely don't 
want them to do accidentally.  Right now, that can't happen because the check for whether the app is debuggable happens further down the stack.

jamesonwilliams:
> > I see two uses of a `true` as a default for "is it on." However, I don't see that these consider the app's debuggable configuraiton? @eeatonaws has a utility method to inspect and determine the app's debuggability. I think you want to use it's output as the default value, instead of a literal `true` or `false.`
> 
> Hmm, if we did this, then customers would have the ability to enable the debug menu in production, which we definitely don't
> want them to do accidentally. Right now, that can't happen because the check for whether the app is debuggable happens further down the stack.

Ah okay, I got it. So this option means "Turn it on, if you can, please." And the default for the request is "yes, please." But later on, we do a hard check and can still say "Nope, no thanks." That's how it is setup @richardmcclellan ?

richardmcclellan:
> > > I see two uses of a `true` as a default for "is it on." However, I don't see that these consider the app's debuggable configuraiton? @eeatonaws has a utility method to inspect and determine the app's debuggability. I think you want to use it's output as the default value, instead of a literal `true` or `false.`
> > 
> > 
> > Hmm, if we did this, then customers would have the ability to enable the debug menu in production, which we definitely don't
> > want them to do accidentally. Right now, that can't happen because the check for whether the app is debuggable happens further down the stack.
> 
> Ah okay, I got it. So this option means "Turn it on, if you can, please." And the default for the request is "yes, please." But later on, we do a hard check and can still say "Nope, no thanks." That's how it is setup @richardmcclellan ?

Exactly!

jamesonwilliams (approved):

PR #824, opened 2020-09-14
fix(aws-api):  use classloader to determine whether a class is a cust…

richardissuperman:
Any jdk preloaded classes like integer, string will be using bootstrap classloader (native) so it will be null, using it to determine whether a class is a custom type or not is faster than going through the JavaFieldType list and try catch a run time exception

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java
@@ -469,6 +469,14 @@ public final class Temporal {
         }
 
         /**
+         * Returns a new Timestamp instance that represents the current system time.
+         * @return a new Timestamp instance that represents the current system time.
+         */
+        public static Timestamp now() {
+            return new Timestamp();
+        }
+
+        /**
          * Gets the number of seconds that have elapsed since the UNIX epoch.
          *
          * @return Seconds since UNIX epoch
--- a/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelMetadata.java
@@ -22,6 +22,7 @@ import androidx.core.util.ObjectsCompat;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.annotations.ModelConfig;
 import com.amplifyframework.core.model.annotations.ModelField;
+import com.amplifyframework.core.model.temporal.Temporal;
 
 import java.util.Objects;
 
@@ -37,7 +38,7 @@ public final class ModelMetadata implements Model {
     private final @ModelField(targetType = "ID", isRequired = true) String id;
     private final @ModelField(targetType = "Boolean") Boolean _deleted;
     private final @ModelField(targetType = "Int") Integer _version;
-    private final @ModelField(targetType = "AWSTimestamp") Long _lastChangedAt;
+    private final @ModelField(targetType = "AWSTimestamp") Temporal.Timestamp _lastChangedAt;
 
     /**
      * Constructor for this metadata model.
@@ -50,7 +51,7 @@ public final class ModelMetadata implements Model {
             @NonNull String id,
             @Nullable Boolean deleted,
             @Nullable Integer version,
-            @Nullable Long lastChangedAt) {
+            @Nullable Temporal.Timestamp lastChangedAt) {
         this.id = Objects.requireNonNull(id);
         this._deleted = deleted;
         this._version = version;
@@ -89,7 +90,7 @@ public final class ModelMetadata implements Model {
      * @return last changed at time
      */
     @Nullable
-    public Long getLastChangedAt() {
+    public Temporal.Timestamp getLastChangedAt() {
         return _lastChangedAt;
     }
 
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
@@ -32,6 +32,7 @@ import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.core.model.query.predicate.QueryPredicates;
+import com.amplifyframework.core.model.temporal.Temporal;
 import com.amplifyframework.datastore.appsync.AppSync;
 import com.amplifyframework.datastore.appsync.AppSyncClient;
 import com.amplifyframework.datastore.appsync.ModelWithMetadata;
@@ -88,7 +89,7 @@ public final class AppSyncClientInstrumentationTest {
     @Test
     @SuppressWarnings("MethodLength")
     public void testAllOperations() throws DataStoreException {
-        Long startTime = new Date().getTime();
+        Long startTimeSeconds = TimeUnit.MILLISECONDS.toSeconds(new Date().getTime());
 
         // Create simple model with no relationship
         BlogOwner owner = BlogOwner.builder()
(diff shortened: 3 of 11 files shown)


aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -276,16 +276,20 @@ private SelectionSet wrapPagination(SelectionSet node) {
          */
         private static boolean isCustomType(@NonNull Field field) {
             Class<?> cls = getClassForField(field);
-            if (Model.class.isAssignableFrom(cls) || Enum.class.isAssignableFrom(cls)) {
+            if (cls.getClassLoader() == null || (cls.getClassLoader() != null && isDateType(cls))) {
```
TrekSoft, line 279: Isn't including the cls.getClassLoader() != null check redundant since the cls.getClassLoader() == null check would have already resolved if true?

TrekSoft (approved):
Cat1u3

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardissuperman: @TrekSoft ah thats right, let me revise!

TrekSoft:
Thanks for the revision @richardissuperman - I've asked @richardmcclellan who's more familiar with the code to also take a quick look at it from a "any unintended side effects" standpoint.

richardissuperman:
> Thanks for the revision @richardissuperman - I've asked @richardmcclellan who's more familiar with the code to also take a quick look at it from a "any unintended side effects" standpoint.

sure thanks!

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -276,16 +276,20 @@ private SelectionSet wrapPagination(SelectionSet node) {
          */
         private static boolean isCustomType(@NonNull Field field) {
             Class<?> cls = getClassForField(field);
-            if (Model.class.isAssignableFrom(cls) || Enum.class.isAssignableFrom(cls)) {
+            if (cls.getClassLoader() == null || isDateType(cls)) {
```
richardmcclellan, line 279: Could you add a comment explaining how this works?  (essentially what you've put in PR description)

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -276,16 +276,20 @@ private SelectionSet wrapPagination(SelectionSet node) {
...
                 return false;
-            } catch (IllegalArgumentException exception) {
-                // if we get here then field is  a custom type
-                return true;
             }
+            return true;
+        }
+
+        private static boolean isDateType(Class<?> cls) {
+            return cls.getSimpleName().equals(JavaFieldType.DATE.stringValue()) ||
```
richardmcclellan, line 289: Maybe something like this to make it a little DRYer?
```
        Set<String> nonNativeTypeNames = new HashSet<>(Arrays.asList(
                DATE.stringValue(),
                DATE_TIME.stringValue(),
                TIME.stringValue(),
                TIMESTAMP.stringValue()));
        return nonNativeTypeNames.contains(clazz.getSimpleName());
```

richardmcclellan (changes requested):
I think this change will work technically, though a minor concern is maintaining the list of non native types.  If anyone adds a non native JavaFieldType in the future, this list will need to be updated.  Perhaps we could move the entire isCustomType method to `JavaFieldType`?  That would help reduce this concern.

Also, could you add a unit test to verify that `isCustomType` works as expected?

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardissuperman: @richardmcclellan  sure will add!

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -276,16 +276,20 @@ private SelectionSet wrapPagination(SelectionSet node) {
...
+            if (Model.class.isAssignableFrom(cls) || Enum.class.isAssignableFrom(cls)) {
                 return false;
-            } catch (IllegalArgumentException exception) {
-                // if we get here then field is  a custom type
-                return true;
             }
+            return true;
+        }
+
+        private static boolean isDateType(Class<?> cls) {
```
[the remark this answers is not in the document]
richardissuperman: > I think this change will work technically, though a minor concern is maintaining the list of non native types. If anyone adds a non native JavaFieldType in the future, this list will need to be updated. Perhaps we could move the entire isCustomType method to `JavaFieldType`? That would help reduce this concern.
> 
> Also, could you add a unit test to verify that `isCustomType` works as expected?

makes sense, i think i will put these logic into ```JavaFieldType.java``` !

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardissuperman: > Maybe something like this to make it a little DRYer?
> 
> ```
>         Set<String> nonNativeTypeNames = new HashSet<>(Arrays.asList(
>                 DATE.stringValue(),
>                 DATE_TIME.stringValue(),
>                 TIME.stringValue(),
>                 TIMESTAMP.stringValue()));
>         return nonNativeTypeNames.contains(clazz.getSimpleName());
> ```

sounds good, also i noticed that we are only comparing the class's name , shall we use package path + class name ? cause we may have classes that have duplicated names but different packages

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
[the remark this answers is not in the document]
richardissuperman: @jamesonwilliams in java any jdk preloaded classes like ```Integer```, ```String``` （anything inside **rt.jar**） will use bootstrap classloader which is a native one, so in java it will show as null, any application classes (custom type) will use a concrete application classloader.

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardmcclellan: That is a valid point.  I'm not really sure why `JavaFieldType` has a `String javaFieldType` property instead of a `Class javaFieldClass` property.  Maybe you could try making that change?  Then we could just compare the `Class` directly, which would always be accurate.

richardissuperman: @richardmcclellan sure will add it in the next revision

jamesonwilliams (changes requested):
> Any jdk preloaded classes like integer, string will be using bootstrap classloader (native) so it will be null,

I don't understand what you mean by this

> using it to determine whether a class is a custom type or not is faster than going through the JavaFieldType list and try catch a run time exception

Can you demonstrate this with an automated test?

I don't think this is the right optimization. The `aws-api` module is used to perform network operations. The cost of individual CPU operations will be minute compared to the network ops.

Instead, we should optimize for logical & maintainable code units, not CPU performance.

richardissuperman:
> I don't understand what you mean by this. Is there a functional defect in the `JavaFieldType`?
> 
> >

There is no functional defect on ```JavaFieldType```, just think that the purpose of ```JavaFieldType``` is determine whether a class/type is a JDK preloaded classes like Integer , String, Boolean, so instead of listing them all in a single class and compare them 1 by 1 ,  we can just use the class's classloader instance to determine whether this class is a custom type, or a primitive type from Java.

I agree that if the so called **optimisation** broke some code maintainability , its not a good trade-off to make. Do you think what richard suggestion above work as it keep the code maintainability as well as using classloader ? If not i will cancel this PR.

richardmcclellan:
> > I don't understand what you mean by this. Is there a functional defect in the `JavaFieldType`?
> > >
> 
> There is no functional defect on `JavaFieldType`, just think that the purpose of `JavaFieldType` is determine whether a class/type is a JDK preloaded classes like Integer , String, Boolean, so instead of listing them all in a single class and compare them 1 by 1 , we can just use the class's classloader instance to determine whether this class is a custom type, or a primitive type from Java.
> 
> I agree that if the so called **optimisation** broke some code maintainability , its not a good trade-off to make. Do you think what richard suggestion above work as it keep the code maintainability as well as using classloader ? If not i will cancel this PR.

I would agree that checking for a null classloader instead of iterating the `values()` is a negligible optimization, and not worth the new complexity - knowledge of how the classloader works should not be required for maintaining `JavaFieldType`. 

The other idea discussed of replace `String javaFieldType` with `Class javaFieldType` would still be worthwhile, so that we can compare the actual Class objects instead of their simpleNames, which is error prone if there are duplicate classes with the same name.

richardissuperman:
> I would agree with @jamesonwilliams that checking for a null classloader instead of iterating the `values()` is a negligible optimization, and not worth the new complexity. Knowledge of how the classloader works should not be required for maintaining `JavaFieldType`.

make sense

PR #825, opened 2020-09-14
Added Exceptions to Temporal

hellosagar:
*Issue #, if available:*

*Description of changes:*


aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
```diff
@@ -30,53 +34,73 @@
  * AppSync temporal scalars</a>.
  */
 public final class TemporalDeserializers {
+    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
```
richardmcclellan, line 37:
```suggestion
    private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-api");
```

I think I gave you the wrong namespace for Logging when chatting on Discord, this should be `aws-api`.

richardmcclellan (changes requested):
Good start!   Looks like the build failed because of some checkstyle issues, so please take a look at those.

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
```diff
@@ -30,53 +34,73 @@
...
                 throws JsonParseException {
-            return new Temporal.Date(json.getAsString());
+            try {
+                return new Temporal.Date(json.getAsString());
+            } catch (IllegalArgumentException exception) {
+                LOG.warn("Failed to deserialize " + json.getAsString() + " as a Temporal.Date due to " + exception);
+                return null;
+            }
+
+
```
jamesonwilliams, line 56:
```suggestion
```

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
```diff
@@ -30,53 +34,73 @@
...
     public static final class DateDeserializer implements JsonDeserializer<Temporal.Date> {
         @Override
         public Temporal.Date deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
                 throws JsonParseException {
-            return new Temporal.Date(json.getAsString());
+            try {
+                return new Temporal.Date(json.getAsString());
+            } catch (IllegalArgumentException exception) {
+                LOG.warn("Failed to deserialize " + json.getAsString() + " as a Temporal.Date due to " + exception);
+                return null;
```
jamesonwilliams, line 53: It would be better to throw exceptions early here instead of returning null. This "erases" information, going from a detailed failure, to a generic `null` state.

Same for other deserializers

aws-api-appsync/src/test/java/com/amplifyframework/core/model/temporal/TemporalTimeTest.java
```diff
@@ -53,6 +53,11 @@ public void parsesExpectedFormats() {
         assertEquals("01:22:00", new Temporal.Time("01:22").format());
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void parseInvalidFormat(){
```
jamesonwilliams, line 57:
```suggestion
    public void parseInvalidFormat() {
```

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
hellosagar: sure sir
so we need this right?
image

richardmcclellan: I would suggest rethrowing as a JsonParseException, since that is something that consumers of this class will be expecting.

hellosagar: ok understand as its an exception because of wrong parsing

hellosagar: that's we want to throw JsonParseException?
image

richardmcclellan: No, you'll want to catch the IllegalArgumentException (not JsonParseException, and then throw a JsonParseException.  And yes, let's do that for all three.

hellosagar: ok understand this right
``` 
           try {
                return new Temporal.Date(json.getAsString());
            } catch (IllegalArgumentException exception) {
                throw new JsonParseException("Failed to deserialize " + json.getAsString() + " as a Temporal.Date due to " + exception);
            } ```

File: aws-api-appsync/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java [hunk 5]
```diff
@@ -326,8 +335,7 @@ public Time(@NonNull String text) {
                 zoneOffset = ZoneOffset.from(offsetTime);
             } catch (DateTimeParseException exception) {
                 // Optional timezone offset not present
-                localTime = LocalTime.parse(text, DateTimeFormatter.ISO_LOCAL_TIME);
-                zoneOffset = null;
+                throw new IllegalArgumentException("Failed to create Temporal.Time object from " + text, exception);
```
richardmcclellan, line 338: For Temporal.Time, the offset is optional, so this implementation should be very similar to what you did in `Temporal.Date`.  So this method should look something like:

```
try {
    localTime = LocalTime.parse(text, DateTimeFormatter.ISO_LOCAL_TIME);
    zoneOffset = null;
} catch (DateTimeParseException dateTimeParseException) {
    throw new IllegalArgumentException("Failed to create Temporal.Time object from " + text, exception);
}

File: aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java [hunk 6]
```diff
@@ -30,53 +30,69 @@
...
      * Deserializer of Temporal.Date, an extended ISO-8601 Date string, with an optional timezone offset.
-     *
+     * <p>
      * Based on the <a href=https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html> AWS AppSync AWSDate
      * scalar.</a>
      */
     public static final class DateDeserializer implements JsonDeserializer<Temporal.Date> {
         @Override
-        public Temporal.Date deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
-                throws JsonParseException {
```
richardmcclellan, line 42: I believe `throws JsonParseException` needs to stay here, since this method can still throw a JsonParseException?

richardmcclellan (commented):
Looking better!   It looks like there some of the existing tests are failing.  Now that the Temporal constructors can throw an `IllegalArgumentException`, you probably need to add `throws IllegalArgumentException` to any existing test method signatures that use them.

hellosagar:
Now I'm able to fix all issues

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
```diff
@@ -54,11 +54,12 @@
      * Deserializer of Temporal.DateTime, an extended ISO-8601 DateTime string.  Time zone offset is required.
      * <p>
      * Based on the <a href=https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html>AWS AppSync AWSDateTime
-     * scalar.</a>
+     * scalar.</a>git fetch upstream
```
richardmcclellan, line 57:
```suggestion
     * scalar.</a>
```

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/TemporalDeserializers.java
```diff
@@ -30,53 +30,71 @@
  * AppSync temporal scalars</a>.
  */
 public final class TemporalDeserializers {
+
```
richardmcclellan, line 33:
```suggestion
```

hellosagar:
Thanks for helping a lot

jpignata:
> Thanks for helping a lot

Thanks so much for contributing!

hellosagar:
> > Thanks for helping a lot
> 
> Thanks so much for contributing!

My Pleasure

aws-api-appsync/src/main/java/com/amplifyframework/core/model/temporal/Temporal.java
```diff
@@ -211,9 +215,14 @@ public DateTime(@NonNull java.util.Date date, int offsetInSeconds) {
          * Constructs an {@link Temporal.DateTime} from a valid extended ISO-8601 DateTime string.
          *
          * @param text a valid extended ISO-8601 DateTime string
+         * @throws IllegalArgumentException if no zoneOffset is set.
```
richardmcclellan, line 218:
```suggestion
         * @throws IllegalArgumentException when text is not a valid ISO-8601 DateTime string.
```

richardmcclellan: Can you add something similar to the Date and Time constructor javadoc?

richardmcclellan (approved):
I'm going to merge this now, so we can get it into our next release.  Thanks again @hellosagar!

hellosagar:
Give me a second sir let me made those doc changes

hellosagar:
Push done for doc changes

PR #832, opened 2020-09-16
chore(build): Add buildspec.yml file for CodeBuild

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR adds a buildspec.yml file that:
- runs `./gradlew build`
- Consolidates unit test reports in one directory
- Publishes those reports to CodeBuild

It has no impact on the current CircleCI setup.


TrekSoft (approved):
Cat4u3

buildspec.yml
```diff
@@ -0,0 +1,39 @@
+version: 0.2
```
jamesonwilliams, line 1: Can we add a
```yml
# Paragraph up here explaining
# what this file is for
```
(It's clear right now during PR review, but may not be in other contexts / for other readers.)

Does the `buildspec.yml` work generically accross multiple systems, or is this specific to CodeBuild? Can you add in the comment?

rjuliano: I've added some information about the file, including a couple links to the AWS docs in case the reader is not familiar with the product. One other thing we can do is to move this file out of here and make it part of the Cloudformation template generated in the `amplify-ci-support` repo.

jamesonwilliams (approved):

PR #839, opened 2020-09-17
chore(aws-api-appsync) Use class as value instead of simpleName for JavaFieldTy…

richardissuperman:
Use class as value instead of simpleName for JavaFieldType in case of duplicated Class simple name.

We were using class's simple name as identifier , this will cause issues when classes have same name (different package) , In fact, there are already two duplicated classes:

java.lang.Date vs Temporal.Date , former is used in our model class.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
@@ -20,6 +20,8 @@ import androidx.annotation.NonNull;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.temporal.Temporal;
 
+import java.util.Date;
+
 /**
  * Enumerate the types used in the fields
  * of {@link com.amplifyframework.core.model.Model} classes.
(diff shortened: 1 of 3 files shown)


aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
```diff
@@ -28,86 +30,83 @@
...
-    JavaFieldType(@NonNull String javaFieldType) {
+    JavaFieldType(@NonNull Class<?> javaFieldType) {
         this.javaFieldType = javaFieldType;
     }
 
-    /**
-     * Return the string that represents the value of the enumeration constant.
-     * @return the string that represents the value of the enumeration constant.
-     */
-    public String stringValue() {
```
jamesonwilliams, line 98: This one wasn't being used anywhere, huh?

jamesonwilliams (approved):
Thanks @richardissuperman. Does this fix any of the [open issues](https://github.com/aws-amplify/amplify-android/issues)? If so, can you tag in the PR description? If not, can you change the title to `chore(aws-api-appsync)` or similar?

Overall the change here looks reasonable to me.

aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
richardissuperman: @jamesonwilliams right this one was only used internally inside the ```JavaFieldType.java```

richardissuperman:
> Thanks @richardissuperman. Does this fix any of the [open issues](https://github.com/aws-amplify/amplify-android/issues)? If so, can you tag in the PR description? If not, can you change the title to `chore(aws-api-appsync)` or similar?
> 
> Overall the change here looks reasonable to me.

right, this is not tracking any open issues, but just a suggestion from richard https://github.com/aws-amplify/amplify-android/pull/824#issuecomment-693454524 , in fact i already see duplicated classes issues. java.lang.Date vs Temporal.Date , former is used in our model class.

jamesonwilliams (approved):

PR #850, opened 2020-09-21
fix(Auth): Fixes guest / anonymous user access

TrekSoft:
Changes the order of operations to the same one that iOS uses to avoid the bug in AWSMobileClient where it doesn't return identity id until getAWSCredentials is called.


rjuliano (approved):
Ship it! 🚢

Issue #854, opened 2020-09-23
Derive subscription startup timeout from the number of models


rjuliano:
Currently, subscription startup timeout is set to 10 seconds. We need to change this so we can take into account the number of models in the model provider

jamesonwilliams:
Carrying over [some discussion from Discord, 01SEP2020](https://discordapp.com/channels/705853757799399426/735987596462653461/750261715287146606):

@saltonmassally:
> For projects with models > 10, timing out at 10 second is problematic and subscriptions startup consistently times out. Can we increase the timeout threshold?

@jamesonwilliams:
> The current code establishes the connections serially, not in parallel, right? Yes I think we could make it dynamic (k * modelCount), or even just user-configurable?

@saltonmassally:
> yes it is serial, i thought about parallelising it as i see no reason why we can't do that.. also making it dynamic might be a good short-term fix while we throw user-configurability in the backlog
Let me know your thoughts, I can open a PR accordingly

@jamesonwilliams 
> I think I originally had it in parallel; I believe I made it serial only to simplify the error reporting from Rx. Otherwise, if multiple things fail in parallel, you have to sort out which ones to report -- or wait for all the errors and report multiple errors, etc. However, if we can get that implemented cleanly, it would be a great optimization.
> I think either of those solutions could be an improvement. User-configurable might be more explicit, and less confusing? @richardmcclellan @rjuliano What do you guys think of the options here?

@rjuliano 
> I think it's OK in the short-term to derive that value based on the number of models. Adding a configurable setting would be good long term, but it probably needs to be uniform across all platforms.

@richardmcclellan 
> Maybe I'm not understanding the code. Isn't it already in parallel, because we use `Observable.merge` [here](https://github.com/aws-amplify/amplify-android/blob/4d426937c2e0a66345f7c63ca02d5c6f3f675cc5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L103).   Serial would be `Observable.concat` I think, right?

@rjuliano
> Regarding parallelizing it, I believe we currently do that. The list of observables is constructed serially, however, the [Observable.merge](https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L103) call is set on run on the IO scheduler

@saltonmassally 
> So for now let's just derive that value based on the number of models. any recommendation of an equation for that? or rather a formula
[...]

PR #858, opened 2020-09-25
fix(aws-datastore): make subscription timeout model count dependent

raphkim:
*Issue #, if available:*
#854 

*Description of changes:*
* Orchestrator operations now have timeout that depends on the number of models.
* Orchestrator now throws `TimeoutException` if `blockingAwait()` times out.
* Subscription processor operations also have timeout that depends on the number of models.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -322,11 +322,11 @@ public final class Orchestrator {
 
     private void stopApiSyncBlocking() {
         try {
-            boolean subscribed = stopApiSync()
+            boolean stopped = stopApiSync()
                 .subscribeOn(startStopScheduler)
-                .blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
-            if (!subscribed) {
-                throw new TimeoutException("Subscription timed out.");
+                .blockingAwait(LOCAL_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            if (!stopped) {
+                throw new TimeoutException("Operation timed out.");
             }
         } catch (Throwable failure) {
             LOG.warn("Failed to stop API sync.", failure);
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -39,6 +39,7 @@ import org.json.JSONObject;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.rxjava3.core.Completable;
@@ -51,7 +52,8 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
  */
 public final class Orchestrator {
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
-    private static final long OP_TIMEOUT_SECONDS = 10;
+    private static final long TIMEOUT_SECONDS_PER_MODEL = 2;
+    private static final long MINIMUM_OP_TIMEOUT_SECONDS = 10;
 
     private final SubscriptionProcessor subscriptionProcessor;
     private final SyncProcessor syncProcessor;
@@ -62,6 +64,7 @@ public final class Orchestrator {
     private final MutationOutbox mutationOutbox;
     private final CompositeDisposable disposables;
     private final Scheduler startStopScheduler;
+    private final long adjustedTimeoutSeconds;
 
     /**
      * Constructs a new Orchestrator.
@@ -112,6 +115,13 @@ public final class Orchestrator {
         this.targetMode = targetMode;
         this.disposables = new CompositeDisposable();
         this.startStopScheduler = Schedulers.single();
+
+        // Operation times out after 10 seconds. If there are more than 5 models,
+        // then 2 seconds are added to the timer per additional model count.
+        this.adjustedTimeoutSeconds = Math.max(
+            MINIMUM_OP_TIMEOUT_SECONDS,
+            TIMEOUT_SECONDS_PER_MODEL * modelProvider.models().size()
+        );
     }
 
     /**
@@ -241,11 +251,14 @@ public final class Orchestrator {
     private void startObservingStorageChanges() {
         LOG.info("Starting to observe local storage changes.");
         try {
-            mutationOutbox.load()
+            boolean subscribed = mutationOutbox.load()
                 .andThen(Completable.create(emitter -> {
                     storageObserver.startObservingStorageChanges(emitter::onComplete);
                     currentMode.set(Mode.LOCAL_ONLY);
-                })).blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                })).blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+            if (!subscribed) {
+                throw new TimeoutException("Subscription timed out.");
+            }
         } catch (Throwable throwable) {
             LOG.warn("Failed to start observing storage changes.", throwable);
         }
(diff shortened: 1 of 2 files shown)


rjuliano (commented):
I know it's in draft so you may not be done with it, but we also need to take into account the [timeout in SubscriptionProcessor](https://github.com/aws-amplify/amplify-android/blob/0159843d9f3ef52e72c8c3e55f9f5299595a70af/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L119)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -305,9 +321,12 @@ private Completable startApiSync() {
 
     private void stopApiSyncBlocking() {
         try {
-            stopApiSync()
+            boolean subscribed = stopApiSync()
                 .subscribeOn(startStopScheduler)
-                .blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                .blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
```
jamesonwilliams, line 326: Here, the timeout is being applied "outside of" `stopApiSync()`. For `startApiSync()`, though, the timeout is applied _inside_, on the `syncProcessor.hydrate()`. Is there any way to change this, so that the blocking timeout happens at the same abstraction level, on start/stop APIs, as opposed to on stop, and on a detail of start?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -241,11 +251,14 @@ private Completable transitionToApiSync(Mode current) {
...
-            mutationOutbox.load()
+            boolean subscribed = mutationOutbox.load()
                 .andThen(Completable.create(emitter -> {
                     storageObserver.startObservingStorageChanges(emitter::onComplete);
                     currentMode.set(Mode.LOCAL_ONLY);
-                })).blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                })).blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+            if (!subscribed) {
+                throw new TimeoutException("Subscription timed out.");
+            }
```
jamesonwilliams, line 261: I could be wrong, here, but I _believe_ this timeout should be _way_ shorter. Starting the subscription to the local storage should be O(ms), to complete a CPU-bound operation.

The sync and subscription timeouts are different. Both involve k-many _network_ operations. They're like the opposite, basically.

Should the storage observer timeout be more like ~2 seconds?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: For the start scenario, we also have to take into account that we block inside the `SubscriptionProcessor` as well as on the call to `syncProcessor.hydrate()` as you pointed out. So if we move the timeout higher in the call hierarchy (for start), we'll have to make sure that timeout is doubled since the worst case scenario would be for both SubscriptionProcessor and SyncProcessor to take `numberOfModels* TIMEOUT_SECONDS_PER_MODEL` seconds

rjuliano (commented):
Left a couple of comments...let me know what you think

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
jamesonwilliams: ^ Ah good call out, though. We should find out whether or not this triggers loading of the queue from SQL. A disk-bound timeout will be different than a CPU-bound timeout.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -305,9 +321,12 @@ private Completable startApiSync() {
 
     private void stopApiSyncBlocking() {
         try {
-            stopApiSync()
+            boolean subscribed = stopApiSync()
                 .subscribeOn(startStopScheduler)
-                .blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                .blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+            if (!subscribed) {
+                throw new TimeoutException("Subscription timed out.");
```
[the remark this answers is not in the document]
jamesonwilliams: The "do online stuff" orchestrator mode is currently called `SYNC_VIA_API`. If we change this language here, we should keep both in sync. Maybe it would make more sense to simply call this `ONLINE` mode? `startOnlineBehaviors()` and `stopOnlineBehaviors()`?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
raphkim: I think this effort is unnecessary. Correct me if i'm wrong, but shouldn't `blockingAwait()` be applied to any operation where you want to ensure that it is completed before moving onto the next instruction? What is the purpose of putting a blocking await on the entire `startApiSync()` method instead of on its components that actually need it?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -241,11 +252,14 @@ private Completable transitionToApiSync(Mode current) {
...
         try {
-            mutationOutbox.load()
+            boolean subscribed = mutationOutbox.load()
                 .andThen(Completable.create(emitter -> {
                     storageObserver.startObservingStorageChanges(emitter::onComplete);
                     currentMode.set(Mode.LOCAL_ONLY);
-                })).blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                })).blockingAwait(LOCAL_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            if (!subscribed) {
+                throw new TimeoutException("Subscription timed out.");
```
jamesonwilliams, line 261:
```suggestion
                throw new TimeoutException("Timed out while preparing local-only mode.");
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -272,8 +286,11 @@ private Completable startApiSync() {
 
             LOG.debug("About to hydrate...");
             try {
-                syncProcessor.hydrate()
-                    .blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                boolean subscribed = syncProcessor.hydrate()
+                    .blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+                if (!subscribed) {
+                    throw new TimeoutException("Subscription timed out.");
```
jamesonwilliams, line 292:
```suggestion
                    throw new TimeoutException("Timed out while performing initial model sync.");
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -305,9 +322,12 @@ private Completable startApiSync() {
 
     private void stopApiSyncBlocking() {
         try {
-            stopApiSync()
+            boolean stopped = stopApiSync()
                 .subscribeOn(startStopScheduler)
-                .blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                .blockingAwait(LOCAL_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            if (!stopped) {
+                throw new TimeoutException("Operation timed out.");
```
jamesonwilliams, line 329:
```suggestion
                throw new TimeoutException("Timed out while waiting for API synchronization to end.");
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -305,9 +322,12 @@ private Completable startApiSync() {
 
     private void stopApiSyncBlocking() {
         try {
-            stopApiSync()
+            boolean stopped = stopApiSync()
                 .subscribeOn(startStopScheduler)
-                .blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                .blockingAwait(LOCAL_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
jamesonwilliams, line 327: This is probably actually be a _network_ op, since it has to send the `stop` messages on the WebSocket subscription. May want to use the long timeout for this one.

jamesonwilliams (approved):

PR #860, opened 2020-09-25
Adds support for mixed owner and group based auth rules

TrekSoft:
Currently subscriptions do not work if you have both an owner auth directive with a READ operation and a group auth directive with a READ operation. This fixes that.


rjuliano (approved):
Ship it ~~~🚢 ~~~~

PR #867, opened 2020-09-28
chore(api) remove JAVA_DATE since codegen does not generate java.lang.Date

richardmcclellan:
Follow up to https://github.com/aws-amplify/amplify-android/pull/839.    

 - Removes unused code in `JavaFieldType` and `TypeConverter`
 - Updates related tests so that our test models use `Temporal.Date` instead of `java.lang.Date`, since that is what codegen uses.

### DIFF
--- a/core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
+++ b/core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
@@ -45,10 +45,9 @@ public final class ModelSchemaTest {
      * The factory {@link ModelSchema#fromModelClass(Class)} will produce
      * an {@link ModelSchema} that meets our expectations for the {@link Person} model.
      * @throws AmplifyException from model schema parsing
-     * @throws ClassNotFoundException from Class.forName("com.amplifyframework.core.model.temporal.Temporal$Date")
      */
     @Test
-    public void modelSchemaIsGeneratedForPersonModel() throws AmplifyException, ClassNotFoundException {
+    public void modelSchemaIsGeneratedForPersonModel() throws AmplifyException {
         Map<String, ModelField> expectedFields = new HashMap<>();
         expectedFields.put("id", ModelField.builder()
             .targetType("ID")
--- a/core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
+++ b/core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
@@ -20,6 +20,7 @@ import androidx.annotation.NonNull;
 import com.amplifyframework.AmplifyException;
 import com.amplifyframework.core.model.annotations.AuthRule;
 import com.amplifyframework.core.model.annotations.ModelConfig;
+import com.amplifyframework.core.model.temporal.Temporal;
 import com.amplifyframework.testmodels.personcar.MaritalStatus;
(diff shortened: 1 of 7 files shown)


core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
```diff
@@ -70,7 +70,7 @@ public void modelSchemaIsGeneratedForPersonModel() throws AmplifyException {
         expectedFields.put("dob", ModelField.builder()
             .targetType("AWSDate")
             .name("dob")
-            .type(Date.class)
+            .type(Class.forName("com.amplifyframework.core.model.temporal.Temporal$Date"))
```
richardmcclellan, line 73: This is a bit of a hack because I cannot reference `Temporal.Date.class` from here, since it is in the `aws-api-appsync` package, so I have to get it using reflection.   A better solution for another day could be to move more of the `Model` classes from `core` to `aws-api-appsync`.   

Previously, this test was only passing by coincidence - it was comparing `java.lang.Date.class.getSimpleName()` with the `Temporal.Date.class.getSimpleName()`, and it passed since they are equal.  Now that we are comparing the actual class, this test started failing (as it should), so this needs to be updated to actually reference the correct `Class`.

jamesonwilliams: > aws-api-appsync module, which I can't reference from core

Certainly not from _src_ code. But what about as a `testImplementation`?

You could also access it through the `testutils` module, perhaps -- an additional level of indirection?

jamesonwilliams (commented):
Have a suggestion to avoid the reflection. Otherwise, looks good.

core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
```diff
@@ -45,9 +45,10 @@
      * The factory {@link ModelSchema#fromModelClass(Class)} will produce
      * an {@link ModelSchema} that meets our expectations for the {@link Person} model.
      * @throws AmplifyException from model schema parsing
+     * @throws ClassNotFoundException from Class.forName("com.amplifyframework.core.model.temporal.Temporal$Date")
      */
     @Test
-    public void modelSchemaIsGeneratedForPersonModel() throws AmplifyException {
+    public void modelSchemaIsGeneratedForPersonModel() throws AmplifyException, ClassNotFoundException {
```
jamesonwilliams, line 51: The `ClassNotFoundException` has gone away now, too.

jamesonwilliams (approved):
Looks like a nice improvement! The integration tests have a red 'x'. Not sure if that's legitimate, or an issue with our tests (haven't looked yet).

richardmcclellan:
> Certainly not from src code. But what about as a testImplementation?

Duh.  Thanks for the tip.  This worked :)

> Looks like a nice improvement! The integration tests have a red 'x'. Not sure if that's legitimate, or an issue with our tests (haven't looked yet).

It's the devmenu tests.  I just merged from main, which should fix them.

core/src/test/java/com/amplifyframework/core/model/ModelSchemaTest.java
richardmcclellan: Just fixed!

jamesonwilliams (approved):

Issue #848, opened 2020-09-21
NullPointerException thrown on calling toString or hashCode of Temporal.DateTime


saltonmassally:
A model with Temporal.DateTime (AWSDatetime) field with null values results in the following:

```
  java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.time.LocalDateTime.toString()' on a null object reference
        at java.time.OffsetDateTime.toString(OffsetDateTime.java:1906)
        at java.lang.String.valueOf(String.java:2924)
        at java.lang.StringBuilder.append(StringBuilder.java:132)
        at com.amplifyframework.core.model.temporal.Temporal$DateTime.toString(Temporal.java:277)
        at java.lang.String.valueOf(String.java:2924)
        at java.lang.StringBuilder.append(StringBuilder.java:132)
        at com.amplifyframework.datastore.generated.model.Group.hashCode(Group.java:340)
```

I think `this.offsetDateTime` below evaluates to null
```
return "Temporal.DateTime{offsetDateTime='" + this.offsetDateTime + "'" + '}';
```

saltonmassally:
Got a few more details on this. Looks like either the datetime format i am using is wrong or some issue with the conversion to Temporal.Datetime.

Datetime stored in sqllite `2020-09-21T16:57:03.132Z`

<img width="1199" alt="Screenshot 2020-09-22 at 01 38 54" src="https://user-images.githubusercontent.com/2057234/93815962-77878f00-fc74-11ea-967e-c6b09bacd813.png">

saltonmassally:
Furthering this. I suspect the issue might be the zone handling as if I use java.time.format.DateTimeFormatter#ISO_ZONED_DATE_TIME is works fine for my case above in which a zone rather than an offset is supplied. I haven;t tested for offset case though

jamesonwilliams:
> Got a few more details on this. Looks like either the datetime format i am using is wrong or some issue with the conversion to Temporal.Datetime.

https://github.com/aws-amplify/amplify-android/pull/825 will make the `Temporal.DateTime` "fail fast" if it sees an input it doesn't like. If you are getting into this state due to a bad format, this might help, @saltonmassally?

[...]

Issue #815, opened 2020-09-10
How can I serialize the auto-generated model classes myself?


hellosagar:
I try to send **auto-generated model class**   from one activity to another activity using intent through Gson as currently there is no way to parse this auto-generated model class as `Serializable` or as `Parcelable` so I have to convert it into String first using Gson
```kotlin
val intent = Intent(this,TestResultAddOrEditActivity::class.java)
intent.putExtra("testResult", Gson().toJson(item))
startActivity(intent)
```

which looks like this
![image](https://user-images.githubusercontent.com/50016799/92685339-24007180-f355-11ea-8ff6-cd8c8892a530.png)

but in another activity what I get is the null value of `createDate` and `date` which is an object of `Temporal.DateTime`
only this object is causing issue and I get other variables without any problem

Screenshot of object TestResult before serializing
![image](https://user-images.githubusercontent.com/50016799/92686023-7aba7b00-f356-11ea-9627-593140c93b69.png)
screenshot of TestResult object from the second activity after second after deserializing this serialized object using this code snippet
```kotlin
val intent = intent
testResult = Gson().fromJson(intent.getStringExtra("testResult"), TestResult::class.java)
```
and that's what the object of TestResult I get from this above code snippet
![image](https://user-images.githubusercontent.com/50016799/92686440-485d4d80-f357-11ea-9e07-2869a10e9efa.png)
This is quite surprising why I get `createDate` and `date` as null
I'll try yo give my best to explain this bug
Please tell me what's the problem
I'm waiting eagerly to get the solution to this problem
Thanks

KayIlory:
I agree, this is a major BUG that got introduced in the version >1.3. @jamesonwilliams  any ideas would help

hellosagar:
I think the problem is in the toString() method of Temporal.java line 266
![image](https://user-images.githubusercontent.com/50016799/92692358-a8f18800-f361-11ea-823a-a737e12d734f.png)

richardmcclellan:
Hi @hellosagar and @Kayllory, a couple thoughts here:

- I think what's happening is that a `Temporal` object was constructed from a String that was an invalid format.  An improvement that I will try to make is to throw an exception instead of creating the `Temporal` object, so that you get a more helpful stacktrace that shows you what that invalid String is.  
[...]

PR #875, opened 2020-10-01
chore(aws-api-appsync): consolidate serde logic

jamesonwilliams:
Around the code base, Gson is used with various different
configurations. However, most uses require the same basic set of
serializers and deserializers, to handle custom Amplify and AppSync
types.

To address this, the serializers and deserializesr are consolidated into
the aws-api-appsync module. Construction of Gson instances is aligned to
use the common GsonFactory, now shared between the API and DataStore
categories.

The GsonFactory itself is augmented to provided different factory
methods, suitable for augmenting the set of adapters in a more
configurable way.

Resolves: https://github.com/aws-amplify/amplify-android/issues/848


aws-api-appsync/src/main/java/com/amplifyframework/api/aws/GsonFactory.java
```diff
@@ -26,28 +26,62 @@
...
+     *          adapters
+     */
+    public static Gson create(@NonNull Map<Class<?>, Object> additionalAdapters) {
+        GsonBuilder builder = new GsonBuilder();
+        withAdditionalAdapters(builder, additionalAdapters);
+        withDeserializers(builder);
+        return create(builder);
+    }
+
     private static void withDeserializers(GsonBuilder builder) {
```
richardmcclellan, line 78:
```suggestion
    private static void withAdapters(GsonBuilder builder) {
```

raphkim (approved):

richardmcclellan (approved):
This is great!  Nice cleanup!

richardmcclellan:
Now that `GsonFactory` is public, can we resolve [this feature request](https://github.com/aws-amplify/amplify-android/issues/815), which asks for an easy way to (de)serialize generated models?

jamesonwilliams:
> Now that `GsonFactory` is public, will this also resolve [this feature request](https://github.com/aws-amplify/amplify-android/issues/815), which asks for an easy way to (de)serialize generated models?

It's a step in the right direction for https://github.com/aws-amplify/amplify-android/issues/815. But as of this PR, we still have hardcorded calls to create and use `GsonFactory.create()`, etc.

To complete #815, we'll need to abstract all of that and expose a single place to set the used Gson instance (perhaps, as a constructor/builder argument to the plugin(s)).

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreGsonFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import com.amplifyframework.datastore.appsync.ModelWithMetadataDeserializer;
+import com.amplifyframework.datastore.syncengine.TimeBasedUuidTypeAdapter;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+/**
+ * A utility to create Gson instances for the DataStore plugin.
+ */
+public final class DataStoreGsonFactory {
```
richardmcclellan, line 31: I'm a little confused by having two separate factories, `ApiGsonFactory`, and `DataStoreGsonFactory`.  From what I can tell, this is how you are using them: 
 - `ApiGsonFactory`
   - API request serialization
   - API response deserialization
   - DataStore response deserialization - this doesn't feel right!
 - `DataStoreApiFactory`
   - DataStore request serialization
   - SQLite (de)serialization of custom types


The DataStore `AppSyncClient` is using the API category under the hood, and hence is using the `ApiGsonFactory` for  all of the network operations - sync, mutate, subscribe.  


I would suggest either:
1. Just have one GsonFactory with everything
2. Have an `ApiGsonFactory` which is used for all API related (de)serialization in both API & DataStore, and a `CustomTypeGsonFactory`, which is used for (de)serialization of custom types to SQLite

jamesonwilliams: I should be able to achieve (1). It is the last change I was toying with, prior to putting up the latest version. Let me go back and see that option to its natural conclusion.

richardmcclellan: Awesome.  Other than that, the PR looks great.  Nice cleanup!

Issue #853, opened 2020-09-23
Prevent Orchestrator's startup sequence from being re-triggered while still in-flight


rjuliano:
## Background
The startup sequence for the orchestrator performs the following tasks:
- Start observing changes to local storage and push changes to the mutation outbox (Always)
- Set up subscriptions for all models (Only if API is present)
- Execute initial sync for all models (Only if API is present)
- Start processing mutation outbox items (Only if API is present)
- Start processing changes received via remote subscriptions (Only if API is present)

Currently we attempt to trigger this sequence from a couple of different places:
- During [DataStore startup](https://github.com/aws-amplify/amplify-android/blob/c612d56181ba4a686ff2da4fc8a8913050f9863f/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L226)
- When [DataStore.clear is called](https://github.com/aws-amplify/amplify-android/blob/c612d56181ba4a686ff2da4fc8a8913050f9863f/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L482)
- Before [any interaction with the DataStore](https://github.com/aws-amplify/amplify-android/blob/c612d56181ba4a686ff2da4fc8a8913050f9863f/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L495)(other than clear)

The 3rd scenario was implemented as a way for us to ensure that the sync engine is always running (or recovers from failure, specifically in offline scenarios). 99% of the time, this will be a no-op because the [orchestrator is already started](https://github.com/aws-amplify/amplify-android/blob/df70332ab7f6197fe5756a85196f8505fa05e95d/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L157). However...

## The problem
As it stands, there's a period of time between when the Orchestrator's start method is called and the internal state of the Orchestrator is updated, that leaves room for that method to be invoked multiple times. This happens typically when the app calls Amplify.configure and immediately tries to query or save something to the DataStore.

## The solution
We need to add some safeguards to ensure that there can only be one invocation of the start/stop methods in flight at any given point. 

Whatever solution is implemented needs to ensure it does not impact DataStore interactions when the device is offline. The problem we had in the past is that if we block for the entirety of the startup sequence (those 5 steps above), it introduces delays on every interaction with the DataStore (for offline cases). We truly only need to block until the storage observer starts.

[...]

PR #876, opened 2020-10-02
fix(datastore):Prevent concurrent start/stop on orchestrator

rjuliano:
*Issue #, if available: #853

*Description of changes:*
Added a semaphore around start/stop operations for the orchestrator to prevent concurrent calls to these methods.

Modified the tests to attempt to start the orchestrator twice.

Also, instead of mocking the AppSyncClient class which is part of the DataStore package, I changed the test to mock the API category instead. If nothing else, it should provide increased coverage since we're now actually testing the code in AppSyncClient.


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -180,9 +189,16 @@ private Completable transitionCompletable() {
      */
     public Completable stop() {
         LOG.info("DataStore orchestrator stopping. Current mode = " + currentMode.get().name());
+        if (!startStopSemaphore.tryAcquire()) {
```
saltonmassally, line 192: Given that you generally want calls to stop to be terminal maybe use acquire so calling thread blocks or tryAcquire with timeout for a softer approach?

rjuliano: That's a good point. The only place where stop is called in a meaningful way is during a call to DataStore.clear. We'd want to block in that case, so I just went with the acquire option.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/ApiMocking.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.appsync;
```
jamesonwilliams, line 16: `AppSyncMocking` lives in `com.amplifyframework.datastore.appsync` since the `AppSync` interface does. Your new `ApiMocking` might be a `testutil`ity, since it mocks a `core` spec (the `ApiCategoryBehavior`.) In this view, it could live somewhere under [`testutils/src/main/java/com/amplifyframework/testutils`](https://github.com/aws-amplify/amplify-android/tree/main/testutils/src/main/java/com/amplifyframework/testutils). I don't feel 100% resolved in this, though. What do you think?

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/ApiMocking.java
```diff
@@ -0,0 +1,150 @@
...
+import static org.mockito.Mockito.doAnswer;
+
+/**
+ * Utility to mock behaviors of the API category.
+ */
+public final class ApiMocking {
+
+    private ApiMocking(){
+
+    }
```
jamesonwilliams, line 48:
```suggestion
    private ApiMocking() {}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/ApiMocking.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Mock the results of successful query operation triggered by the DataStore.
+     * @param mockApi Mock object of type {@link GraphQLBehavior}
+     * @param models A list of models to be returned in the result.
+     * @param <M> The model type.
+     */
+    @SuppressWarnings("unchecked")
+    public static <M extends Model> void mockSuccessfulSyncQuery(GraphQLBehavior mockApi, M... models) {
```
jamesonwilliams, line 106: This method name is leaking an abstraction, I think. The API category doesn't know about "sync"ing, that's a datastore thing. However, the class is called `ApiMocking`. To stay pure to its charter, this should probably not mention "sync" anywhere, just "query."

Also, since `ModelWithMetadata` is essentially a DataStore detail, I would pull that out of the implementation as well. Instead, you could provide them at the call site:
```java
public static <T> void mockSuccessfulSyncQuery(GraphQLBehavior mockApi, T... results) {
    ...
    // Implementation no longer mentions ModelWithMetadata either
    onResults.accept(new GraphQLResponse<PaginatedResult<T>>(
                new PaginatedResult<T>(syncQueryResults, null));
    ...
}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/ApiMocking.java
```diff
@@ -0,0 +1,150 @@
...
+            ));
+            return null;
+        }).when(mockApi).query(
+            any(), // Request
+            any(), // onResponse
+            any()  // onFailure
+        );
+    }
+
+    private static <M extends Model> ArgumentMatcher<GraphQLRequest<M>> getMatcherFor(M model) {
```
jamesonwilliams, line 138: Maybe just take the model ID as an arg, and rename it to:
```suggestion
    private static <M extends Model> ArgumentMatcher<GraphQLRequest<M>> requestContainsModelId(String modelId) {
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
```diff
@@ -70,19 +73,18 @@ public void itemsPlacedInStorageArePublishedToNetwork() throws AmplifyException
...
-        ModelWithMetadata<BlogOwner> modelWithMetadata = new ModelWithMetadata<>(susan, metadata);
-        AppSyncMocking.create(appSync).mockResponse(susan);
-        AppSyncMocking.sync(appSync).mockSuccessResponse(BlogOwner.class, modelWithMetadata);
-        AppSyncMocking.onCreate(appSync).callOnStart();
-        AppSyncMocking.onDelete(appSync).callOnStart();
-        AppSyncMocking.onUpdate(appSync).callOnStart();
+        // Mock behaviors from for the API category
+        GraphQLBehavior mockApi = mock(GraphQLBehavior.class);
+
+        ApiMocking.mockSubscriptionStart(mockApi);
```
jamesonwilliams, line 83: I forget exactly why I went with the static methods in the `AppSyncMocking` util. I think my goal was to get something that looked like Mockito:
```java
onCreate(appSync).callOnStart();
```

In hind-sight, it might have been a better design to create an object instance that encapsulated the `appSync` instance.

Or, in other words, what if your utility worked like:
```java
ApiMocking mocking = new ApiMocking(mock(GraphQLBehavior.class));
```
Or even  just:
```java
// internally, encapsulates a `Mockito.mock(GraphQLBehavior.class)`
MockGraphQLApi api = new MockGraphQLApi();
```

jamesonwilliams (commented):
The approach looks sound!

Some ideas on the mechanics.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -135,10 +138,16 @@ public boolean isStopped() {
      * Start performing sync operations between the local storage adapter
      * and the remote GraphQL endpoint.
      */
-    public void start() {
+    public synchronized void start() {
+        LOG.debug("Available permits = " + startStopSemaphore.availablePermits());
+        if (!startStopSemaphore.tryAcquire()) {
```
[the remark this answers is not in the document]
saltonmassally: The problem with using acquire is that CRUD operations will then block in `beforeOperation` as threads wait which can have a cascading effect.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -135,10 +138,16 @@ public boolean isStopped() {
...
+            LOG.warn("Unable to acquire orchestrator lock. Transition currently in progress.");
+            return;
+        }
         disposables.add(transitionCompletable()
             .subscribeOn(startStopScheduler)
             .doOnDispose(() -> LOG.debug("Orchestrator disposed a transition."))
+            .doFinally(startStopSemaphore::release)
```
richardissuperman, line 150: in the problem description here: https://github.com/aws-amplify/amplify-android/issues/853 you mentioned that we should only **We truly only need to block until the storage observer starts**, in here it looks like we are still blocking everything , shall we release the ```startStopSemaphore``` as soon as ```startObservingStorageChanges()``` is done? something like:

```
  mutationOutbox.load()
                    currentMode.set(Mode.LOCAL_ONLY);
                }))
.foFinally(startStopSemaphore::release)
.blockingAwait(OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);

```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: Thanks again for taking a look. I have refactored this to include a timeout.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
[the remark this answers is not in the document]
rjuliano: To build on my last comment, I ended up changing the start method(s) on the orchestrator to return a Completable. This gives the calling more control over timeout and blocking vs non-blocking.

As a result of this change, the logic to trigger the DataStore's READY event is now triggered from the plugin instead of the orchestrator, which seems appropriate.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
rjuliano: Let me circle back on this one. Let me know what you think of the other changes.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -142,24 +142,53 @@ public boolean isStopped() {
...
+    public synchronized Completable start() {
+        return start(NETWORK_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    }
+
     /**
      * Start performing sync operations between the local storage adapter
-     * and the remote GraphQL endpoint.
+     * and the remote GraphQL endpoint. The timeout parameters provided will be
+     * used to block execution of the current thread until lock acquisition has
+     * either failed or succeded, or the timeout is elapsed.
```
raphkim, line 157: succeeded*

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
```diff
@@ -100,8 +113,17 @@ public void itemsPlacedInStorageArePublishedToNetwork() throws AmplifyException
             );
 
         // Arrange: orchestrator is running
-        orchestrator.start();
+        orchestrator.start().subscribe();
+
+        // Try to start it in a new thread.
+        new Thread(() -> orchestrator.start().subscribe()).start();
+        // Try to start it again on a current thread.
+        orchestrator.start().subscribe();
```
raphkim, line 121: I understand that you are testing the semaphore, but can we move this to a separate test case? Someone without context checking out this test might get confused why this step is necessary in testing that the "items placed in storage are published to network".

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -142,24 +142,53 @@ public boolean isStopped() {
...
+                .subscribeOn(startStopScheduler);
+        } else {
+            return Completable.error(new DataStoreException("Unable to acquire orchestrator lock. " +
+                                                        "Transition currently in progress.",
+                                                        "Retry your operation."));
+        }
+    }
+
+    private boolean tryAcquireStartStopLock(long opTimeout, TimeUnit timeUnit) {
+        LOG.debug("Attempting to acquire lock. Permits available = " + startStopSemaphore.availablePermits());
```
raphkim, line 181: Considering that we are using this lock for mutual exclusion, isn't this log redundant?

raphkim (approved):
Looks really solid! Minor nit-picks

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -223,7 +225,22 @@ public void initialize(@NonNull Context context) throws AmplifyException {
                 initError, AmplifyException.TODO_RECOVERY_SUGGESTION
             );
         }
-        orchestrator.start();
+        startOrchestratorAsync();
+    }
+
+    private void startOrchestratorAsync() {
```
jamesonwilliams, line 231: To the fullest extent possible, we should keep this logic within the Orchestrator itself. The plugin is (theoretically) intended to be a thin veneer, and the Orchestrator itself has the job of managing start/stop lifecycle.

Is there a way to get the actions of the subscribe callbacks back into the Orchestrator, here?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -489,11 +506,8 @@ public void clear(@NonNull Action onComplete,
 
     private void beforeOperation(@NonNull final Runnable runnable) {
         try {
-            Completable.fromAction(
-                () -> {
-                    categoryInitializationsPending.await();
-                    orchestrator.start();
-                })
+            categoryInitializationsPending.await();
+            orchestrator.start(BEFORE_OP_TIMEOUT_MS, TimeUnit.MILLISECONDS)
```
jamesonwilliams, line 510: I'm a little confused by this. What does it mean to have a 0 second timeout? Would it be better to just fork the `start()` API, and have `startWithBehaviorFoo()` and `startWithBehaviorBar()`?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
jamesonwilliams: @raphkim I think your point is that this is a "yes/no" log, not a "0...k" log, right? Or in other words, the observation is that `availablePermits()` is either 0 or 1?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
rjuliano: That's a good point. I'm breaking this out.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
raphkim: Yep! It's either available or not haha

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
raphkim: Man I was confused when I first saw this, then saw his explanation and understood it, but now I'm confused again. If it's zero second timeout, does it not _ALWAYS_ fail? and return `false` regardless of whether transition is already happening or not? Maybe it should be a small value like 200ms instead?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
rjuliano: It's easy enough to make that true or false instead.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Yes...I was able to move this. The original thought was to have the orchestrator be controlled by the plugin, but it doesn't really buy us anything.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Timeout of 0 means that the caller only wants the lock if one is available immediately. Basically a fail fast. That being said, I can achieve similar behavior after refactoring the way the completable is built inside the orchestrator.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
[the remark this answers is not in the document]
rjuliano: I removed the overloads that had timeouts since I was able to achieve the same result by re-working the construction of the completable.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: I removed the new overload with the timeout parameter. Just had to re-work the completable a bit.

jamesonwilliams (approved):
Looks promising, thanks for diving into this @rjuliano! I think by now both of us are probably tiring of working in this orchestrator. 😄

Issue #841, opened 2020-09-17
Incomplete Support for User-Provided Conflict Handler


jamesonwilliams:
When the DataStore's sync engine publishes content up to the AppSync endpoint, it's possible that a conflict arrises between the local version of the data, and the remote version of the data. [The Amplify Docs discuss this topic in greater depth, here](https://docs.amplify.aws/lib/datastore/conflict/q/platform/android).

The JavaScript and iOS libraries provide a mechanism for the user to add their own custom, client-side handling logic, to be used whenever such a conflict may occur:

- [iOS - Custom Conflict Resolution](https://docs.amplify.aws/lib/datastore/conflict/q/platform/ios#custom-conflict-resolution)
- [JS - Custom Conflict Resolution](https://docs.amplify.aws/lib/datastore/conflict/q/platform/js#custom-configuration)

The Android code base includes a [`DataStoreConfiguration` ](https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java#L35) object. As of https://github.com/aws-amplify/amplify-android/pull/810 (released in Amplify Android [1.3.2](https://github.com/aws-amplify/amplify-android/releases/tag/release_v1.3.2)), the user can provide that config when initializing the DataStore plugin:
```java
Amplify.addPlugin(new AWSDataStorePlugin(DataStoreConfiguration.builder()
    .dataStoreConflictHandler((conflictData, decision) -> {
        decision.accept(DataStoreConflictHandlerResult.APPLE_REMOTE);
    })
    .build()));
```

___However___.

Android's [`MutationProcessor.java`](https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java) currently does not include any logic to handle [the various AppSync conflict errors](https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html#errors) (`ConflictUnhandled`, etc.)

In other words, you can use `decision.accept(DataStoreConflictHandlerResult.APPLE_REMOTE` in your handler, but the DataStore won't actually _honor_ that request, right now.

***Work needed:***
1. Wire the handler in to `MutationProcessor.java`
2. Code logic to act on the different handling approaches (`RETRY_REMOTE`, `RETRY_LOCAL`, `RETRY` (with user-provided model))
[...]

PR #883, opened 2020-10-07
feature(aws-datastore): handle mutation conflicts

jamesonwilliams:
When the DataStore's sync engine attempts to publish a local change to the AppSync backend, it is possible that the AppSync backend already has a copy of the model being mutated. In such cases, AppSync may send back a GraphQL response that contains a `ConflictUnhandled` error.

The client can now recognize a `ConflictUnhandled` error, and will respond to it by invoking a conflict handling function. The user may supply their own conflict handling strategy, by passing it in a`DataStoreConfiguration` provided to the `AWSDataStorePlugin`:
```kotlin
val config = DataStoreConfiguration.builder()
    .dataStoreConflictHandler((conflictData, onResult) -> {
        onResult.accept(DataStoreConflictHandlerResult.APPLY_REMOTE);
    })
    .build();
Amplify.addPlugin(AWSDataStorePlugin(config))
```

This current work updates the DataStore to invoke the handler. Subsequent work will implement default logic for the various handling strategies (retry local change, apply remote change, and retry with a custom model.) Currently, the only strategy that has been implemented is "apply remote."

Refer: https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html#errors
Refer: https://github.com/aws-amplify/amplify-android/issues/841

--------
## Status
To keep the size down, I'm planning to do a followup PR which implements handlers for the three different conflict handling strategies.


aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ConflictUnhandledError.java
```diff
@@ -0,0 +1,125 @@
...
+        Gson gson = GsonFactory.instance();
+
+        for (GraphQLResponse.Error error : errors) {
+            if (Empty.check(error.getExtensions())) {
+                continue;
+            }
+
+            AppSyncExtensions appSyncExtensions =
+                new AppSyncExtensions(error.getExtensions());
+            if (!"ConflictUnhandled".equals(appSyncExtensions.getErrorType())) {
```
richardmcclellan, line 79: Could we create an enum with all the error types [like iOS](https://github.com/aws-amplify/amplify-ios/blob/eb5b7ff98ae9ac090dc8c71e374b8f3bcb0ee472/AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift)?   I had intended to do it when I created `AppSyncExtensions`, but didn't get that far.

testutils/src/main/java/com/amplifyframework/testutils/Varargs.java
```diff
@@ -0,0 +1,48 @@
...
+    private Varargs() {}
+
+    /**
+     * Converts a (possibly null) array into a non-null (possibly empty) list.
+     * @param items A possibly null array of items
+     * @param <T> Type of items in the array
+     * @return A non-null, possibly empty collection
+     */
+    @NonNull
+    public static <T> List<T> toList(@Nullable T[] items) {
```
richardmcclellan, line 41: Neat! 🎉

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -284,4 +371,114 @@ void publish(
             Consumer<DataStoreException> onFailure
         );
     }
+
+    static final class Builder implements
```
richardmcclellan, line 375: Much needed, looks good.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -47,24 +57,40 @@
...
-    private final AppSync appSync;
+    private final VersionRepository versionRepository;
+    private final SyncTimeRegistry syncTimeRegistry;
     private final MutationOutbox mutationOutbox;
+    private final AppSync appSync;
+    private final DataStoreConfigurationProvider configurationProvider;
     private final CompositeDisposable ongoingOperationsDisposable;
 
-    MutationProcessor(
+    private MutationProcessor(
```
richardmcclellan, line 68: Can you change this to take a `Builder` instead of a bunch of parameters?

aws-datastore/src/test/resources/conflict-unhandled-response.json
```diff
@@ -0,0 +1,29 @@
...
+    {
+      "path": [
+        "updateNote"
+      ],
+      "data": {
+        "id": "KoolId22",
+        "content": "Resurecting the dataz",
+        "_version": 7,
+        "_deleted": true,
+        "_lastChangedAt": 1601499066604
```
richardmcclellan, line 15: `Temporal.Timestamp` should always be in seconds, not milliseconds.

richardmcclellan (commented):
Looks good so far!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
jamesonwilliams: Done!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
jamesonwilliams: Done!

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ConflictUnhandledError.java
jamesonwilliams: Done!

aws-datastore/src/test/resources/conflict-unhandled-response.json
jamesonwilliams: I believe this value was observed in the wild, in a real response from AppSync. Will confirm.

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMockingTest.java
```diff
@@ -0,0 +1,376 @@
...
+
+import io.reactivex.rxjava3.core.Completable;
+import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.functions.Predicate;
+
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests the {@link AppSyncMocking} test utility.
```
rjuliano, line 49: 🤯

rjuliano (approved):
Ship it! 👍

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ConflictUnhandledError.java
```diff
@@ -0,0 +1,125 @@
...
+
+        Gson gson = GsonFactory.instance();
+
+        for (GraphQLResponse.Error error : errors) {
+            if (Empty.check(error.getExtensions())) {
+                continue;
+            }
+
+            AppSyncExtensions appSyncExtensions = new AppSyncExtensions(error.getExtensions());
+            AppSyncErrorType errorType = AppSyncErrorType.fromErrorType(appSyncExtensions.getErrorType());
```
richardmcclellan, line 78: How about we move this inside of `AppSyncExtensions` and just make the getter return the enum? 

Instead of this current code:
```
public String getErrorType() {
    return errorType;
}
```

do something like:
```
public AppSyncErrorType getErrorType() {
    return AppSyncErrorType.fromErrorType(errorType);
}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMockingTest.java
```diff
@@ -0,0 +1,376 @@
...
+            if (!ObjectsCompat.equals(expectedResponse.getErrors(), response.getErrors())) {
+                return false;
+            }
+            ModelWithMetadata<BlogOwner> expectedMwm = expectedResponse.getData();
+            ModelWithMetadata<BlogOwner> mwm = response.getData();
+            return equalsRelaxingTimestamp(expectedMwm).test(mwm);
+        };
+    }
+
+    private static Predicate<ModelWithMetadata<BlogOwner>> equalsRelaxingTimestamp(
```
richardmcclellan, line 339: Ick... so what's the issue exactly for needing this workaround?

richardmcclellan (approved):

aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncMockingTest.java
jamesonwilliams: In `AppSyncMocking`, I have a couple of flavors of the `mockSuccessResponse(...)` which internally create their own `Temporal.Timestamp.now()`. (So, I can't know what it is, outside, in this unit test.)

However, I just realized that I could test the other flavors of `mockSuccessResponse(...)`, instead -- the ones that _will_ allow me to supply a known timestamp. That would get rid of these kludges and clean things up quite a bit. I'll do that.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/ConflictUnhandledError.java
jamesonwilliams: Good call. I will do this.

aws-datastore/src/test/resources/conflict-unhandled-response.json
jamesonwilliams: We discussed this offline. Our findings:

1. [AppSync returns this field in milliseconds](https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html), in the conflict handler.
2. But, in a generated schema, `_lastChangedAt`  is represented as an `AWSTimestamp`.  And an `AWSTimestamp` [is defined as number of seconds since 1970](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html).

Issue #885, opened 2020-10-07
Amplify S3 StorageUploadFileOptions.metadata is appended differently than AmazonS3Client.putObject metadata


edgar-zigis:
Hello,

In our S3 bucket policy we have a requirement for AES256 encryption. Our bucket
permission policy does look like this:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "DenyIncorrectEncryptionHeader",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::staging-customer-assets-XXXXXXXXXXXXXXXXX/*",
            "Condition": {
                "StringNotEquals": {
                    "s3:x-amz-server-side-encryption": "AES256"
                }
            }
        },
        {
            "Sid": "DenyUnEncryptedObjectUploads",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::staging-customer-assets-XXXXXXXXXXXXXXXXX/*",
            "Condition": {
                "Null": {
                    "s3:x-amz-server-side-encryption": "true"
                }
            }
        }
    ]
}
```

I did create a sample file like this:
```kotlin
val exampleFile = File(applicationContext.cacheDir, "ExampleKey")
exampleFile.writeText("Example file contents")
```

When using deprecated `AmazonS3Client` upload options I am able to upload this
file successfully:
```kotlin
val clientStorage = AmazonS3Client(credentialsProvider)
clientStorage.putObject(
    "staging-customer-assets-XXXXXXXXXXXXXXXXX",
    "ExampleKey",
     FileInputStream(exampleFile),
     ObjectMetadata().also {
         it.sseAlgorithm = ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION
     }
)
```
But when trying to upload it via Amplify, it fails to upload it:

```kotlin
val metadata = mutableMapOf<String, String>()
metadata[Headers.SERVER_SIDE_ENCRYPTION] = ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION

val options = StorageUploadFileOptions.builder()
    .metadata(metadata)
    .build()
Amplify.Storage.uploadFile("ExampleKey", exampleFile, options, {
    // success
}, {
    // failure
})
```

Now I would like to clarify, that if I disable encryption requirement, both
[...]

PR #886, opened 2020-10-07
Enable setting of the server side encryption algorithm in StorageUploadFileOptions

edgar-zigis:
*Issue #, if available:* https://github.com/aws-amplify/amplify-android/issues/885

*Description of changes:* Now server side encryption can be set via StorageUploadFileOptions


TrekSoft:
Hey @edgar-zigis - thanks so much for this PR! Other than the one minor checkstyle error it looks good!

```
[ant:checkstyle] [ERROR] /home/circleci/amplify-android-home/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/request/AWSS3StorageUploadFileRequest.java:56:30: Expected @param tag for 'sseAlgorithm'. [JavadocMethod]
```

jamesonwilliams (changes requested):
Hey @edgar-zigis, thanks so much for the initiative on this!! Looks pretty good, but I have a couple of suggestions.

-------

When I read SSD, I think of Intel's Streaming SIMD Extensions. When the S3 service uses this acronym, they mean Server-Side Extensions.
1. To simplify this comprehension for the Amplify customer, can we call this option `serverSideEncryption()` throughout the PR?
2. Alternately, we could call it `bucketEncryption()` through-out the PR, if that is more clearly phrased, even?

-----------
There are only [three valid options for the value of this field](https://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html). So, we can do better than asking the user to supply a raw `String` in the `Builder`.

Instead, we could:
1. Use an enum, instead of a String.
```java
.bucketEncryption(EncryptionMethod.MANAGED_KEYS)
```
with:
```java
enum EncryptionMethod {
    MANAGED_KEYS("AES256"),
    KMS_KEYS("aws:kms");
    // Technically also there is a way for customer to provide keys
}
```
2. Even simpler, we could just only support the S3 managed keys, for now. Then, the builder option might be:
```java
.enableBucketEncryption()
```
Internally, this would add the `"AES256"` option to the S3 call.

edgar-zigis:
@TrekSoft thank you for noticing!
@jamesonwilliams very good points. I have decided to stick with `server side encryption`, because this is the term which is usually looked for when people try to find how to set bucket encryption :)

jamesonwilliams (changes requested):
Looking good, thanks for making the change. One last thought:

Instead of accepting the `@Nullable StorageServerSideEncryption` arg in the builder, can you make it `@NonNull StorageServerSideEncryption` and enforce `Objects.requireNonNull(...)` on the supplied value? We can also add a `NONE` to the enum, so that the default assignment in the Builder can be `StorageServerSideEncryption.NONE`.

Then:
```java
if (!StorageServerSideEncryption.NONE.equals(storageServerSideEncryption)) {
    objectMetadata.setSSEAlgorithm(storageServerSideEncryption.getName());
}
```

Sorry didn't think of this in the last round. Appreciate your contribution!

edgar-zigis:
@jamesonwilliams thank you for dedicating your time! Now I will know the guidelines better :) I have adjusted the PR to match your observations!

richardmcclellan:
@edgar-zigis can you address the one outstanding checkstyle issue?
```
> Task :aws-storage-s3:checkstyle FAILED
[ant:checkstyle] [ERROR] /home/circleci/amplify-android-home/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageUploadFileOperation.java:27: Wrong order for 'com.amplifyframework.storage.StorageException' import. [ImportOrder]
```

edgar-zigis:
Import order has been changed!

core/src/main/java/com/amplifyframework/storage/StorageServerSideEncryption.java
```diff
@@ -0,0 +1,51 @@
...
+public enum StorageServerSideEncryption {
+    /**
+     * AES256 encryption.
+     */
+    MANAGED_KEYS("AES256"),
+
+    /**
+     * aws:kms encryption.
+     */
+    KMS_KEYS("aws:kms"),
```
raphkim, line 30: Is this too AWS specific to be in `core` module?

At Amplify layer, the categories should not adhere too much to a specific AWS service, and such details should be specified at the plugin layer. Exposing `AES256` and `KMS` as supported encryption methods at a category level doesn't seem ideal to me, although I guess one can still argue that these are generic enough encryptions to be applicable to storage?

raphkim (commented):
Looks good to me

core/src/main/java/com/amplifyframework/storage/StorageServerSideEncryption.java
edgar-zigis: Well, this is used across multiple modules, so I did not find a better place to put it.
Also, on the AWSMobileClient only these encryption methods are specified. I think this is a pretty required feature.

raphkim: I believe the ideal place for `StorageServerSideEncryption` enum to be housed in is inside `aws-storage-s3` module. However, the way that `StorageOptions` is structured makes it difficult to pass in plugin-specific options parameter to the category. This may require some refactoring from our side.

Meanwhile, this PR looks pretty solid otherwise. As you mentioned, the enum in question cannot be placed outside of `core` as it currently stands, so I think we will do a quick refactor of `StorageOptions` after merging in your changes.  How does that sound to you?

edgar-zigis: @raphkim sounds great! <3

jamesonwilliams (approved):

Issue #868, opened 2020-09-29
NullPointerException in AWSMobileClientStore.get(java.lang.String[])' on a null object reference


hellosagar:
I got this exception in the Amplify.Auth.fetchAuthSession error block
this is the code I used 
``` Amplify.Auth.fetchAuthSession(
            { result ->
                Timber.i("AmplifyQuickstart $result")
            },
            { error ->
                Timber.e("AmplifyQuickstart $error")
            }
        )```
and this is the error stacktrace
```AmplifyException {message=An error occurred while attempting to retrieve your user details, cause=java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Map com.amazonaws.mobile.client.AWSMobileClientStore.get(java.lang.String[])' on a null object reference, recoverySuggestion=See attached exception for more details}```
and this is my ```amplifyconfiguration.json```
```{
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "auth": {
        "plugins": {
            "awsCognitoAuthPlugin": {
                "UserAgent": "aws-amplify-cli/0.1.0",
                "Version": "0.1.0",
                "IdentityManager": {
                    "Default": {}
                },
                "CredentialsProvider": {
                    "CognitoIdentity": {
                        "Default": {
                            "PoolId": "ap-south-1:a7081471-eba6-41eb-a14e-ec7747ccfa06",
                            "Region": "ap-south-1"
                        }
                    }
                },
                "CognitoUserPool": {
                    "Default": {
                        "PoolId": "ap-south-1_gDZTd26eh",
                        "AppClientId": "b04qhon5u958o41jdmc8ov769",
                        "AppClientSecret": "v4ih0la56d9t1e0vtb0rhvjojc38sb6f0v6ear9u6n7mpb8e5gi",
                        "Region": "ap-south-1"
                    }
                },
                "Auth": {
                    "Default": {
                        "authenticationFlowType": "USER_SRP_AUTH"
                    }
                },
                "AppSync": {
                    "Default": {
                        "ApiUrl": "https://whoolaonwnhl5n2mjcg2jfgsnq.appsync-api.ap-south-1.amazonaws.com/graphql",
                        "Region": "ap-south-1",
                        "AuthMode": "AWS_IAM",
                        "ClientDatabasePrefix": "HelpingProjectApi_AWS_IAM"
                    }
                }
            }
        }
    },
    "api": {
        "plugins": {
            "awsAPIPlugin": {
[...]

PR #888, opened 2020-10-07
(fix) Fail early and throw useful exception if customer forgets to call Amplify.configure

richardmcclellan:
- Fixes https://github.com/aws-amplify/amplify-android/issues/868 by throwing a useful exception if customer tries to use a plugin before calling Amplify.configure(). 
 - One consequence of this new exception being thrown is that plugins need to be configured in a particular order now.  That is, Auth must be configured _before_ Analytics, API, Storage, and Predictions now.  I accomplished this by moving `Auth` to the top of the list of the `Amplify.buildCategoriesMap()` method.    Previously, the API category was actually being configured before the Auth category.  During configuration of API, it actually grabs a reference to the Auth plugin, and saves a reference to the Auth escape hatch, the `AWSMobileClient`.   At this point, the `AWSMobileClient` is instantiated, but not initialized.  The `AWSApiPlugin` only saves a reference to it during configuration, so the fact that initialization happens later is actually okay, for now.  That said, if the `AWSApiPlugin` configure method is modified in the future to call any methods on the `AWSMobileClient`, it could cause issues that would be hard to debug.  This PR adds some safety to ensure that no plugin is used before it is configured.

### DIFF
--- a/aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
+++ b/aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
@@ -66,7 +66,6 @@ import com.amazonaws.mobile.client.results.SignInState;
 import com.amazonaws.mobile.client.results.SignUpResult;
 import com.amazonaws.mobile.client.results.Tokens;
 import com.amazonaws.mobile.client.results.UserCodeDeliveryDetails;
-import com.amazonaws.mobile.config.AWSConfiguration;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Before;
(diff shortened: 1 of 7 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -323,6 +324,7 @@ private void assertSyncProcessorNotStarted(ApiCategory mockApi) {
     private ApiCategory mockApiCategoryWithGraphQlApi() throws AmplifyException {
         ApiCategory mockApiCategory = spy(ApiCategory.class);
         ApiPlugin<?> mockApiPlugin = mock(ApiPlugin.class);
+
```
richardmcclellan, line 327:
```suggestion
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthComponentTest.java
```diff
@@ -1124,6 +1081,7 @@ public void signedInSessionWithExpiredTokens() throws AuthException {
     @Test
     public void getCurrentUser() {
         doAnswer(invocation -> USERNAME).when(mobileClient).getUsername();
+
```
richardmcclellan, line 1084:
```suggestion
```

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -172,18 +174,21 @@ public final void removePlugin(@NonNull P plugin) {
...
     public final P getPlugin(@NonNull final String pluginKey) throws IllegalStateException {
         P plugin = plugins.get(pluginKey);
-        if (plugin != null) {
-            return plugin;
-        } else {
+        if (plugin == null) {
             throw new IllegalStateException(
-                "Tried to get a plugin but that plugin was not present. " +
-                "Check if the plugin was added originally or perhaps was already removed."
+                    "Tried to get a plugin but that plugin was not present. " +
```
richardmcclellan, line 184:
```suggestion
                "Tried to get a plugin but that plugin was not present. " +
```

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -172,18 +174,21 @@ public final void removePlugin(@NonNull P plugin) {
...
     public final P getPlugin(@NonNull final String pluginKey) throws IllegalStateException {
         P plugin = plugins.get(pluginKey);
-        if (plugin != null) {
-            return plugin;
-        } else {
+        if (plugin == null) {
             throw new IllegalStateException(
                 "Tried to get a plugin but that plugin was not present. " +
-                "Check if the plugin was added originally or perhaps was already removed."
+                            "Check if the plugin was added originally or perhaps was already removed."
```
richardmcclellan, line 185:
```suggestion
                    "Check if the plugin was added originally or perhaps was already removed."
```

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -172,18 +174,21 @@ public final void removePlugin(@NonNull P plugin) {
...
     public final P getPlugin(@NonNull final String pluginKey) throws IllegalStateException {
         P plugin = plugins.get(pluginKey);
-        if (plugin != null) {
-            return plugin;
-        } else {
+        if (plugin == null) {
             throw new IllegalStateException(
                 "Tried to get a plugin but that plugin was not present. " +
-                "Check if the plugin was added originally or perhaps was already removed."
+                    "Check if the plugin was added originally or perhaps was already removed."
```
richardmcclellan, line 185:
```suggestion
                "Check if the plugin was added originally or perhaps was already removed."
```

jamesonwilliams (approved):

PR #889, opened 2020-10-07
fix(API): Adds support for custom identity claim

TrekSoft:
Modifies our code to fetch whatever value from the access token specified by the user (or the default username if the user didn't specify one)

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -319,7 +319,11 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
                     )
                 ) {
                     request = appSyncRequest.newBuilder()
-                            .variable(ownerRuleWithReadRestriction.getOwnerFieldOrDefault(), "String!", getUsername())
+                            .variable(
+                                ownerRuleWithReadRestriction.getOwnerFieldOrDefault(),
+                                    "String!",
+                                getIdentityValue(ownerRuleWithReadRestriction.getIdentityClaimOrDefault())
+                            )
                             .build();
                 }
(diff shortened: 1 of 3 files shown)


richardmcclellan:
When I have the most basic owner auth rule on a schema like this: 

```
@auth(rules: [{ allow: owner }])
```

The generated annotation is:
```
@AuthRule(allow = AuthStrategy.OWNER, ownerField = "owner", identityClaim = "cognito:username", operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ })
```

This PR will change the behavior for this basic use case to look for `cognito:username` instead of `username` in the access token.  Is this correct?

TrekSoft:
> When I have the most basic owner auth rule on a schema like this:
> 
> ```
> @auth(rules: [{ allow: owner }])
> ```
> 
> The generated annotation is:
> 
> ```
> @AuthRule(allow = AuthStrategy.OWNER, ownerField = "owner", identityClaim = "cognito:username", operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ })
> ```
> 
> This PR will change the behavior for this basic use case to look for `cognito:username` instead of `username` in the access token. Is this correct?

Riiip - let me double check this. I was seeing something different on my schema.

TrekSoft:
Yup, I see that - looks like a bug in the CLI that was just never discovered before because we never used that value. Will double check with iOS here and then talk to CLI. So this PR is not ready to merge as is now.

richardmcclellan (commented):
The logic for backwards compatibility with codegen using `cognito:username` as the default, instead of `username` looks correct!  As discussed offline, let's just confirm with them that this approach makes sense.

richardmcclellan (approved):
Approving based on confirmation from CLI team on this approach!

PR #892, opened 2020-10-10
recursively build joins for multilevel nested models

saltonmassally:
*Issue #, if available:*

*Description of changes:*
Currently joins only transverse the first level for nested models. This PR attempts to make SQLiteCommandFactory build joins for models with more than one level of nesting.

A potential gotcha for deeply nested models is the performance hit so we may want to make how deep we follow nesting configurable in DatastoreConfiguration and set and document a default limit.

A related Important fix here is the order of the limit and order by clause "The ORDER BY clause should appear before the LIMIT clause" or it results in an error

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
@@ -24,6 +24,7 @@ import com.amplifyframework.datastore.storage.SynchronousStorageAdapter;
 import com.amplifyframework.testmodels.commentsblog.AmplifyModelProvider;
 import com.amplifyframework.testmodels.commentsblog.Blog;
 import com.amplifyframework.testmodels.commentsblog.BlogOwner;
+import com.amplifyframework.testmodels.commentsblog.Comment;
 import com.amplifyframework.testmodels.commentsblog.Post;
 import com.amplifyframework.testmodels.commentsblog.PostStatus;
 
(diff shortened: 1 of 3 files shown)


raphkim:
Hey @saltonmassally ,

Thanks for the PR (as always). I actually thought about recursing through the nested models as you have while initially working on the storage engine, and have decided against it at the time for the following reasons:

1. Recursing through deeply nested models increases the workload significantly for little reward. `join` operation doesn't scale very well with increasing database size, and the operation load will be unscalable past a few nested-layers for large databases. Meanwhile, it is unlikely that the customer is interested in a model that is several relationship away from the queried model most of the time.
2. What is the expected behavior upon encountering a loop? 

Meanwhile, I do agree that the current implementation is less than ideal. To throw some ideas around, a good compromise would be to go 2-ish layers deep by default, and allow customers to customize this value per query.

raphkim:
It appears that I was confusing this with a totally different case here. 

We definitely want to avoid eager-loading every descendant of a model and recurse all the way down for the reasons I mentioned above... but this PR is _NOT_ doing that! I am just realizing that this recursion goes **upward** to cover all of the ancestors, which is indeed necessary in order for `query` to behave correctly.

raphkim (approved):

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -146,6 +147,46 @@ public void querySavedDataWithForeignKey() throws DataStoreException {
...
+
+        final Comment comment = Comment.builder()
+            .content("Alan's first comment")
+            .post(post)
+            .build();
+
+        adapter.save(blogOwner);
+        adapter.save(blog);
+        adapter.save(post);
+        adapter.save(comment);
```
jamesonwilliams, line 181: If you like it, this should work, too:
```suggestion
        adapter.save(blogOwner, blog, post, comment);
```

jamesonwilliams (approved):

saltonmassally:
Been a bit swamped lately, will address these comments during the weekend.

PR #893, opened 2020-10-10
Have query operations not block if lock not acquired

saltonmassally:
*Issue #, if available:*

*Description of changes:*

Related to  #876 and tries to address the ff:

Have query operations not block if lock not acquired, currently, it blocks for 2 seconds.

Proposes removing synchronizing `com.amplifyframework.datastore.syncengine.Orchestrator#start` because of the ff concerns:
-  does synchronized make all CRUD operations effectively sequential given it a prerequisite for CRUD operations through the AWSDataStorePlugin? Is that a practical restriction?

- does it lead to a chain of suspended threads? Say we are in transition and at the same time make 5 queries on different threads concurrently... all are suspended while we wait for each in sequentially to try acquire to timeout after 2 seconds or acquire a thread... does that 5th call wait in the worse scenario for 5*2 seconds?

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -287,7 +287,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
             @NonNull QueryPredicate predicate,
             @NonNull Consumer<DataStoreItemChange<T>> onItemSaved,
             @NonNull Consumer<DataStoreException> onFailureToSave) {
-        beforeOperation(() -> sqliteStorageAdapter.save(
+        beforeOperationReadyStorageAndOrchestrator(() -> sqliteStorageAdapter.save(
             item,
             StorageItemChange.Initiator.DATA_STORE_API,
             predicate,
@@ -322,7 +322,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
             @NonNull QueryPredicate predicate,
             @NonNull Consumer<DataStoreItemChange<T>> onItemDeleted,
             @NonNull Consumer<DataStoreException> onFailureToDelete) {
-        beforeOperation(() -> sqliteStorageAdapter.delete(
+        beforeOperationReadyStorageAndOrchestrator(() -> sqliteStorageAdapter.delete(
             item,
             StorageItemChange.Initiator.DATA_STORE_API,
             itemDeletion -> {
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -505,6 +505,24 @@ private void beforeOperation(@NonNull final Runnable runnable) {
...
+                .blockingAwait(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        } catch (Throwable throwable) {
+            if (!orchestrator.isStarted()) {
+                LOG.warn("Failed to execute request because DataStore is not fully initialized.");
+            } else {
+                LOG.warn("Failed to execute request due to an unexpected error.", throwable);
+            }
+        }
+    }
+
```
rjuliano, line 525: Maybe instead of having two `beforeOperation*` and two `start*` methods, you could use an enum to define possible strategies for starting the orchestrator.

For example, you could create an enum inside the Orchestrator class called `StartWaitStrategy` with two items: `StorageAndSyncReady` and `StorageReady`.

Then in the `AWSDataStorePlugin` class, instead of having two `beforeOperation*` methods, you could have one with `Orchestrator.StartWaitStrategy` as a parameter: 

```java
private void beforeOperation(Orchestrator.StartWaitStrategy strategy, @NonNull final Runnable runnable) {
   ...
      orchestrator.start(strategy);
   ...
}
```
Likewise, you could apply a similar pattern to the start method in the Orchestrator. 

```java
public void start(StartWaitStrategy strategy) {
   if(StartWaitStrategy.StorageReady.equals(strategy)) {
     attemptStart(NO_WAIT_TIMEOUT);
   } else {
     attemptStart(LOCAL_OP_TIMEOUT_SECONDS);
   }
}
```

Let me know what you think.

rjuliano (commented):
@saltonmassally Thanks for another great contribution. I definitely agree with the intent of this PR. I made a suggestion with regards to using an enum to define potential start strategies, rather than having similar methods with different names.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
jamesonwilliams: I like your suggestion @rjuliano! What do you think, @saltonmassally?

saltonmassally: Sounds solid; I will make the changes

richardmcclellan:
I just implemented a change to release the semaphore as soon as the storage observer is initialized, rather than waiting for the entire sync to complete, which I believe will accomplish the same objective that this PR was going for.  Closing in favor of https://github.com/aws-amplify/amplify-android/pull/1027

PR #903, opened 2020-10-14
feature(core): auth rule provider specification

raphkim:
*Description of changes:*
* This PR sets foundation for supporting OIDC use-case with `@auth` directive by adding `AuthProvider` field in `AuthRule`. 

~~*__TODO:__* Talk to the codegen team to support generating models with updated `@AuthRule` annotation.~~
See https://github.com/aws-amplify/amplify-cli/issues/5580

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/model/AuthProvider.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthProvider.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.core.model;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+/**
+ * The type of provider for an @auth directive rule.
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth">
+ *     GraphQL Transformer @auth directive documentation.</a>
+ */
+public enum AuthProvider {
+    /**
+     * Uses API key to authenticate.
+     * The following strategies are permitted:
+     * <ul>
+     *     <li>{@link AuthStrategy#PUBLIC}</li>
+     * </ul>
+     */
+    API_KEY("apiKey"),
+
+    /**
+     * Uses AWS IAM to authenticate.
+     * The following strategies are permitted:
+     * <ul>
+     *     <li>{@link AuthStrategy#PUBLIC}</li>
+     *     <li>{@link AuthStrategy#PRIVATE}</li>
+     * </ul>
+     */
+    IAM("iam"),
+
+    /**
+     * Uses OpenID Connect to authenticate.
+     * The following strategies are permitted:
+     * <ul>
+     *     <li>{@link AuthStrategy#OWNER}</li>
+     *     <li>{@link AuthStrategy#GROUPS}</li>
+     * </ul>
+     */
+    OIDC("oidc"),
+
+    /**
+     * Uses Cognito User Pools to authenticate.
+     * This is the default mode of authorization.
+     * The following strategies are permitted:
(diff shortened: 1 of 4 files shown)


richardmcclellan:
When I first implemented the AuthRule annotation a few months ago, I considered specifying the provider, but did not, because iOS was not.  @lawmicha had mentioned at the time that there wasn't a use case where we would need to know the provider.

@raphkim can you share a basic summary of how you plan to use this new provider field?

Also, assuming this change is needed, I imagine it also needs to be added for iOS and JS codegen as well.

core/src/main/java/com/amplifyframework/core/model/annotations/AuthRule.java
```diff
@@ -40,28 +40,43 @@
...
+    /**
+     * Used for owner authorization.
      *
      * @return name of a {@link ModelField} of type String which specifies the user which should have access
      */
     String ownerField() default "";
 
     /**
-     * Used to specify a custom claim.  Defaults to "username" when using AuthStrategy.OWNER.
+     * Used to specify a custom claim.
```
richardmcclellan, line 64: Did you mean to remove "Defaults to "username" when using AuthStrategy.OWNER."?  I think that comment is still accurate.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -57,74 +66,111 @@ public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
...
-        if (Empty.check(this.operations)) {
-            return Arrays.asList(
-                    ModelOperation.CREATE,
-                    ModelOperation.UPDATE,
-                    ModelOperation.DELETE,
-                    ModelOperation.READ);
-        }
-        return this.operations;
+        return Immutable.of(Empty.check(operations)
+                ? Arrays.asList(ModelOperation.values())
```
richardmcclellan, line 172: While this changes makes the code more concise, I think it's a little more readable to explicitly enumerate the 4 values that make up the default.   That also ensures that if any values are added to the `ModelOperation` enum for some reason, this set of default operations won't be changed unexpectedly.

core/src/main/java/com/amplifyframework/core/model/annotations/AuthRule.java
raphkim: I think this statement is inaccurate inside `@AuthRule` annotation since default is "". We only decide the default value once it gets passed over to `AuthRule` object.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
raphkim: I think this change addresses your second concern better in that it ensures that every operation is included even if `ModelOperation` is modified in the future. However, I do agree that it was more readable initially, so I'll revert it back :)

raphkim:
> When I first implemented the AuthRule annotation a few months ago, I considered specifying the provider, but did not, because iOS was not. @lawmicha had mentioned at the time that there wasn't a use case where we would need to know the provider. @lawmicha can you confirm that?
> 
> @raphkim can you share a basic summary of how you plan to use this new provider field?
> 
> Also, assuming this change is needed, it will also need to be added for iOS and JS codegen as well.

The reason why we need this change (at least in the Android implementation) is that for owner based authentication we currently have it hardcoded to acquire username and groups from Cognito User Pools. This change will allow me to specify whether identity and groups should be obtained from user pools token vs OIDC token.

core/src/main/java/com/amplifyframework/core/model/AuthProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+/**
+ * The type of provider for an @auth directive rule.
+ * @see <a href="https://docs.amplify.aws/cli/graphql-transformer/directives#auth">
+ *     GraphQL Transformer @auth directive documentation.</a>
+ */
+public enum AuthProvider {
```
jamesonwilliams, line 26: Currently this is being proposed for the Core module, at `com.amplifyframework.core.model`. That doesn't seem right. How can we get this out of Core, and into a more purpose-built package?

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -139,29 +185,40 @@ public boolean equals(Object object) {
 
         AuthRule authRule = (AuthRule) object;
 
-        return authStrategy == authRule.authStrategy &&
+        return ObjectsCompat.equals(authStrategy, authRule.authStrategy) &&
```
richardmcclellan, line 188: This really doesn't matter too much since `==` and `equals` do behave the same way when comparing enum values - however, `==` has a slight benefit because it does type checking at _compile time_, whereas `equals` doesn't do it until _runtime_.

PR #904, opened 2020-10-14
feature: resolve conflicts according to strategies

jamesonwilliams:
The DataStore's sync engine will periodically try to publish local
changes up to AppSync. If AppSync already has a copy of the model
instance, a versioning conflict may occur. In this case, AppSync may
return a `ConflictUnhandledError` to the client, in the GraphQL response
error list.

As a customer, you can supply a handler to define custom logic when such
a conflict occurs. To do so, you provide a `DataStoreConflictHandler` to
the `DataStoreConfiguration` when constructing the `AWSDataStorePlugin`:

```java
DataStoreConfiguration config = DataStoreConfiguration.builder()
    .dataStoreConflictHandler(DataStoreConflictHandler.alwaysApplyRemote())
    .build();
Amplify.addPlugin(new AWSDataStorePlugin(config));
```

The ultimate goal of the user-provided handler will be to identify a
`ConflictResolutionDecision`, to be used to address the conflict:

1. `ConflictResolutionDecision.applyRemote()`:
   Just accept the remote copy of the data, and overwrite whatever we
   have locally
2. `ConflictResolutionDecision.retryLocal()`:
   Try to publish the local mutation again. If it fails a second time,
   surface an error, and do not rety additional times.
3. `ConflictResolutionDecision.retry(T userProvidedModel)`:
   Try to publish a new version of the model up to AppSync.
   If using this decision, please take care to ensure the model ID
   matches the ID of the data in conflict.

In the above configuration example, the
`DataStoreConflictHandler.alwaysApplyRemote()` configures a handler that
always chosen to accept the server's version of the data. A similar
`DataStoreConfigurationHandler.alwaysRetryLocal()` is also provided for
convenience. Fully-custom handlers may be provided as well, by
implementing the `DataSToreConfigurationHandler` interface's single
method, `onConflictDetected`.

Once you have elected a strategy in the handler, the DataStore's sync
engine will take over and fulfill the strategy you've requested. Prior
to the current commit, the conflict handler didn't actually do anything
on your behalf -- you were expected to code all resolution logic
directly in your handler. This is no longer necessary.

Resolves: https://github.com/aws-amplify/amplify-android/issues/841


aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelWithMetadataAdapter.java
```diff
@@ -58,4 +64,24 @@ public static void register(@NonNull GsonBuilder builder) {
...
+        // Flatten out the fields of the model and its metadata into a flat key-value map.
+        // To do this, serialize each individually, and then add the key/value pairs for each
+        // object into a new container.
+        JsonObject serializedMetadata = (JsonObject) context.serialize(src.getSyncMetadata());
+        for (Map.Entry<java.lang.String, JsonElement> entry : serializedMetadata.entrySet()) {
+            result.add(entry.getKey(), entry.getValue());
+        }
+        JsonObject serializedModel = (JsonObject) context.serialize(src.getModel());
+        for (Map.Entry<java.lang.String, JsonElement> entry : serializedModel.entrySet()) {
+            result.add(entry.getKey(), entry.getValue());
```
TrekSoft, line 82: There's the possibility here for a user key to conflict with a metadata key. Are we making it clear to the user somewhere what these reserved metadata keys are so they don't accidentally make a property with the same name?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandler.java
```diff
@@ -16,23 +16,288 @@
...
-     * Resolves a DataStore conflict.
-     * @param conflictData Data about the conflict.
-     * @param onResult A callback that should be invoked when the conflict handling is complete.
-     * @param <T> The type of model for which a conflict was observed
+     * Factory to obtain a handler that always applies the remote copy
+     * of the data.
+     * @return A DataStore conflict handler that always applies the remote
+     *         copy of the data.
+     */
+    static DataStoreConflictHandler alwaysApplyRemote() {
```
TrekSoft, line 41: Would it make sense to have the inverse of this, `alwaysApplyLocal()`?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ConflictResolver.java
```diff
@@ -0,0 +1,129 @@
...
+            });
+    }
+
+    @NonNull
+    private <T extends Model> Single<ModelWithMetadata<T>> resolveModelAndMetadata(
+            @NonNull ConflictData<T> conflictData,
+            @NonNull ModelMetadata metadata,
+            @NonNull ConflictResolutionDecision<T> decision) {
+        // If the strategy is apply remote or retry local, we already know the version we'll use.
+        // We simply extract it from the conflict data.
```
TrekSoft, line 98: It looks like this is done in all three cases below - not understanding the comment.

TrekSoft (approved):
Looks good - just a few minor comments/questions

Cat12u3

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelWithMetadataAdapter.java
jamesonwilliams: This is true! But, I think the conflict happens earlier in the system, than this point. By this point, we're already operating on a model that was generated by the CLI. So, the conflict would have been caught there.

I did a quick audit of where/how these fields are mentioned, and I agree it could use improvement.

AppSync & the CLI automatically include a range of fields. Some are prefixed with `_` (e.g., `_lastChangedAt`, `_deleted`, `_version`), while others are not (`updatedAt`, `createdAt`.)

We have a few pieces of existing documentation about this:

In the [GraphQL Directive usage](https://docs.amplify.aws/cli/graphql-transformer/directives#usage):

> Model directive automatically adds `createdAt` and `updatedAt` timestamps to each entities.

`_lastChangedAt`, `_version`, and `_deleted` are mentioned in a few places in the documentation, however, not explicitly as reserved words in the GraphQL schema that the user provides.
1. [DataStore "How it works" Guide to using the AppSync Console](https://docs.amplify.aws/lib/datastore/how-it-works/q/platform/android#writing-data-from-the-appsync-console)
2. [AppSync schema requirements for using versioned versioned data](https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html#versioned-data-source-metadata)

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConflictHandler.java
jamesonwilliams: There was a lot of deliberation on the naming back in April/May when iOS implemented this functionality. At this point, I'm just aligning to the other platforms.

The key difference is that "apply remote" just patches the local data, but doesn't talk to AppSync again. On the other hand, "retry local" will attempt another mutation, using the server's `_version`, but with the local data. If _that_ mutation fails, the overall transaction fails.

^ Maybe we can get this explanation into the docs site, somewhere.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ConflictResolver.java
jamesonwilliams: Good call! This comment looks like a stray. I'll remote it.

PR #906, opened 2020-10-14
feat(aws-api): support owner-based auth with oidc

raphkim:
*Description of changes:*
Owner-based auth currently works only for Cognito User Pools. This PR lets the plugin determine which auth provider to obtain the JWT token from instead of always assuming Cognito User Pools.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -27,6 +27,7 @@ import com.amplifyframework.api.ApiPlugin;
 import com.amplifyframework.api.aws.operation.AWSRestOperation;
 import com.amplifyframework.api.aws.sigv4.CognitoUserPoolsAuthProvider;
 import com.amplifyframework.api.aws.sigv4.DefaultCognitoUserPoolsAuthProvider;
+import com.amplifyframework.api.aws.sigv4.OidcAuthProvider;
 import com.amplifyframework.api.events.ApiEndpointStatusChangeEvent;
 import com.amplifyframework.api.events.ApiEndpointStatusChangeEvent.ApiEndpointStatus;
 import com.amplifyframework.api.graphql.GraphQLOperation;
@@ -315,15 +316,20 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
                 // them.
                 if (ownerRuleWithReadRestriction != null && (
                         readAuthorizedGroups.isEmpty() ||
-                        Collections.disjoint(readAuthorizedGroups, getUserGroups())
+                        Collections.disjoint(readAuthorizedGroups, getUserGroups(
+                                clientDetails.apiConfiguration.getAuthorizationType()
+                        ))
                     )
                 ) {
+                    String key = ownerRuleWithReadRestriction.getOwnerFieldOrDefault();
+                    String type = "String!";
+                    String value = getIdentityValue(
+                            ownerRuleWithReadRestriction.getIdentityClaimOrDefault(),
+                            clientDetails.apiConfiguration.getAuthorizationType()
+                    );
+
                     request = appSyncRequest.newBuilder()
-                            .variable(
-                                ownerRuleWithReadRestriction.getOwnerFieldOrDefault(),
-                                    "String!",
-                                getIdentityValue(ownerRuleWithReadRestriction.getIdentityClaimOrDefault())
-                            )
+                            .variable(key, type, value)
                             .build();
                 }
             } catch (AmplifyException exception) {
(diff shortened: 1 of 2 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -409,20 +414,34 @@ private String getIdentityValue(String identityClaim) throws ApiException {
...
+                return oidcProvider.getLatestAuthToken();
+            case API_KEY:
+            case AWS_IAM:
+            case NONE:
+            default:
                 throw new ApiException(
-                        "Attempted to subscribe to a model with owner based authorization without a Cognito provider",
-                        "Did you add the AWSCognitoAuthPlugin to Amplify before configuring it?"
+                    "Failed to obtain access token from the configured auth provider.",
+                    "Verify that the API is configured with either Cognito User Pools or OpenID Connect."
```
TrekSoft, line 442: I think this suggested recovery method could be clearer - in the most common cause of this error (forgetting to add the auth plugin) the API could be configured with the auth provider. Suggest leaving that suggestion in there and adding to it what would be a common error if it was OIDC.

TrekSoft (approved):
Cat10u3

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -383,13 +388,13 @@ private String getIdentityValue(String identityClaim) throws ApiException {
         return identityValue;
     }
 
-    private ArrayList<String> getUserGroups() throws ApiException {
-        CognitoUserPoolsAuthProvider cognitoProvider = getAuthProvider();
+    private ArrayList<String> getUserGroups(AuthorizationType authType) throws ApiException {
+        // TODO: this doesn't work with OIDC right now. Implement custom groups claim.
```
richardmcclellan, line 392: What needs to happen to support this?  

Instead of leaving this TODO here, can we either:

1. Add support for it
2. Or, throw an exception if a customer tries to use it?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -383,13 +388,13 @@ private String getIdentityValue(String identityClaim) throws ApiException {
...
         ArrayList<String> groups = new ArrayList<>();
         final String GROUPS_KEY = "cognito:groups";
 
         try {
-            JSONObject accessToken = CognitoJWTParser.getPayload(cognitoProvider.getLatestAuthToken());
+            JSONObject accessToken = CognitoJWTParser.getPayload(getAuthToken(authType));
```
richardmcclellan, line 397: `getPayload` can throw an `CognitoParameterInvalidException`.  Can we catch it and wrap it in an `APIException`?  You could throw the same one that is thrown for a `JsonException` below.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -358,17 +364,16 @@ private boolean isReadRestrictingStaticGroup(AuthRule authRule) {
...
-        CognitoUserPoolsAuthProvider cognitoProvider = getAuthProvider();
-
-        String identityValue;
+    private String getIdentityValue(String identityClaim, AuthorizationType authType) throws ApiException {
+        String identityValue = null;
 
         try {
             identityValue = CognitoJWTParser
-                    .getPayload(cognitoProvider.getLatestAuthToken())
+                    .getPayload(getAuthToken(authType))
```
richardmcclellan, line 372: Can we catch a `CognitoParameterInvalidException` here as well?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -409,20 +414,34 @@ private String getIdentityValue(String identityClaim) throws ApiException {
         return groups;
     }
 
-    private CognitoUserPoolsAuthProvider getAuthProvider() throws ApiException {
-        CognitoUserPoolsAuthProvider cognitoProvider = authProvider.getCognitoUserPoolsAuthProvider();
-        if (cognitoProvider == null) {
-            try {
-                cognitoProvider = new DefaultCognitoUserPoolsAuthProvider();
-            } catch (ApiException exception) {
+    private String getAuthToken(AuthorizationType authType) throws ApiException {
```
jamesonwilliams, line 417: The last few PRs for auth have gone right into the main body of this `AWSApiPlugin`. I'm worried about this. We should really be breaking the auth logic out into smaller classes.

Plugins, in a perfect world, are simply adapters between our business logic and the Amplify framework.

However, increasingly, this `AWSApiPlugin` is host to business logic.

@raphkim The model you used for your Predictions category -- where you had business logic services, and used the plugin as a thin-veneer into them -- was architecturally correct, for example.

jamesonwilliams (changes requested):
Let's get some unit tests, too!

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: to be addressed in a separate PR

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: Will be addressed in a separate PR to support custom group claims

richardmcclellan: Implementing in a separate PR is fine.  To merge this now though, could you just replace the TODO with something like this:
```
if(AuthorizationType.OPENID_CONNECT.equals(authType)) {
    throw new ApiException("Custom group claims with OIDC are not supported yet");
}
```

jamesonwilliams (approved):

PR #907, opened 2020-10-15
fix: align datastore config builder methods to ios

jamesonwilliams:
The following options are changed on the `DataStoreConfiguration.builder()`:

1. `dataStoreErrorHandler` -> `errorHandler`
2. `dataStoreConflictHandler` -> `conflictHandler`
3. `syncIntervalInMinutes` -> `syncInterval` (with required TimeUnit arg)

The corresponding accessor methods on the `DataStoreConfiguration` are also updated.

**Note:** This is technically a breaking change, but we don't have the `DataStoreConfiguration` usage documented anywhere, yet. So this is a good time to make the change, before we do.

Related: https://github.com/aws-amplify/docs/pull/2547


richardmcclellan (approved):
:shipit:

PR #909, opened 2020-10-17
feat(datastore) Add start and stop, and stop starting on configure and clear

richardmcclellan:
Summary
 - Added `void DataStore::start(Action onComplete, Consumer<DataStoreException> onError)` API 
 - Added `void DataStore::stop(Action onComplete, Consumer<DataStoreException> onError)` API
 - Added `Completable RxDataStore::start()` and `Completable RxDataStore::stop()` API
 - Modified `DataStore::configure()` so that it does not call start anymore
 - Modified `DataStore::clear()` so that it does not call start anymore
 - The following scenarios now bubble up an exception to the user provided `Consumer<DataStoreException>`, instead of just logging a warning:
   - [Timeout acquiring lock in Orchestrator::start()](https://github.com/aws-amplify/amplify-android/blob/89ed53d3f2b887c6b23439c776be133fa1ac0819/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L181)
   - [Timeout starting DataStore](https://github.com/aws-amplify/amplify-android/blob/c612d56181ba4a686ff2da4fc8a8913050f9863f/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L501)
   - [Any other unexpected error while starting DataStore](https://github.com/aws-amplify/amplify-android/blob/c612d56181ba4a686ff2da4fc8a8913050f9863f/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java#L503)

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -247,7 +247,7 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
                         () -> onComplete.call(),
                         throwable -> onError.accept(new DataStoreException("Request failed because DataStore is not " +
                                 "initialized.", throwable, "Retry your request."))
-                );
+            );
     }
 
     /**
@@ -274,9 +274,9 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
         waitForInitialization(() -> orchestrator.stop()
                 .subscribeOn(Schedulers.io())
                 .subscribe(
-                        onComplete::call,
-                        error -> onError.accept(new DataStoreException("Failed to stop DataStore.", error,
-                                "Retry your request."))), onError);
+                    onComplete::call,
+                    error -> onError.accept(new DataStoreException("Failed to stop DataStore.", error,
+                            "Retry your request."))), onError);
     }
 
     /**
@@ -293,8 +293,8 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
         stop(() -> Completable.create(emitter -> sqliteStorageAdapter.clear(emitter::onComplete, emitter::onError))
                         .subscribeOn(Schedulers.io())
                         .subscribe(onComplete::call,
-                                throwable -> onError.accept(new DataStoreException("Clear operation failed",
-                                        throwable, AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION))),
+                            throwable -> onError.accept(new DataStoreException("Clear operation failed",
+                                    throwable, AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION))),
                 onError);
     }
 
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -236,6 +236,68 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
         ));
     }
 
+    private void waitForInitialization(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        Completable.create(emitter -> {
+            categoryInitializationsPending.await();
+            emitter.onComplete();
+        })
+                .timeout(LIFECYCLE_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+                .subscribeOn(Schedulers.io())
+                .subscribe(
+                        () -> onComplete.call(),
+                        throwable -> onError.accept(new DataStoreException("Request failed because DataStore is not " +
+                                "initialized.", throwable, "Retry your request."))
+                );
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void start(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        waitForInitialization(() -> {
+            try {
+                orchestrator.start();
+            } catch (DataStoreException exception) {
+                onError.accept(exception);
+                return;
+            }
+            onComplete.call();
+        }, onError);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void stop(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        waitForInitialization(() -> orchestrator.stop()
+                .subscribeOn(Schedulers.io())
+                .subscribe(
+                        onComplete::call,
+                        error -> onError.accept(new DataStoreException("Failed to stop DataStore.", error,
+                                "Retry your request."))), onError);
+    }
+
+    /**
+     * Stops all synchronization processes and invokes the clear method of the underlying storage adapter. Any items
+     * pending synchronization in the outbound queue will be lost. Synchronization processes will be restarted on the
+     * next interaction with the DataStore.
+     *
+     * @param onComplete Invoked if the call is successful.
+     * @param onError Invoked if not successful
+     */
+    @SuppressWarnings("unused")
+    @Override
+    public void clear(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        stop(() -> Completable.create(emitter -> sqliteStorageAdapter.clear(emitter::onComplete, emitter::onError))
+                        .subscribeOn(Schedulers.io())
+                        .subscribe(onComplete::call,
+                                throwable -> onError.accept(new DataStoreException("Clear operation failed",
+                                        throwable, AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION))),
+                onError);
+    }
+
     /**
      * Terminate use of the plugin.
      */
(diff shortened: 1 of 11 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -102,25 +107,37 @@ public void setup() {
...
+        standAloneDataStorePlugin.initialize(context);
+    }
+
+    /**
+     * Starting the plugin in local mode (no API plugin) works without freezing or crashing the calling thread.
+     * @throws AmplifyException Not expected; on failure to configure of initialize plugin.
+     */
+    @Test
+    public void startInLocalMode() throws AmplifyException {
+//        //Configure DataStore with an empty config (All defaults)
```
jamesonwilliams, line 128:
```suggestion
        // Configure DataStore with an empty config (All defaults)
```

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -204,7 +225,7 @@ public void configureAndInitializeInApiModeWithoutApi() throws JSONException, Am
      * @throws AmplifyException on failure to arrange API plugin via Amplify facade
      */
     @Test
-    public void clearStopsSyncUntilNextInteraction() throws AmplifyException, JSONException {
+    public void clearStopsSyncAndDeletesDatabase() throws AmplifyException, JSONException {
```
jamesonwilliams, line 228: Line 223 still says "synchronization processes are restarted"; could you update the Javadoc?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
```diff
@@ -156,7 +158,13 @@ public void preventConcurrentStateTransitions() throws AmplifyException {
         orchestrator.start();
 
         // Try to start it in a new thread.
-        new Thread(() -> orchestrator.start()).start();
+        new Thread(() -> {
+            try {
+                orchestrator.start();
+            } catch (DataStoreException exception) {
+                fail("Error occurred starting the orchestrator." + exception);
```
jamesonwilliams, line 165: This won't actually fail the test. jUnit will only report a test failure if the assertion error is raised on the test runner's thread. This will crash the `new Thread`, but it won't get reported.

If you _don't_ rely on the `fail` for the test to be useful, I'd suggest changing this `fail` to a comment, like:
```java
// This doesn't matter, we'll catch the failure later when we validate (thing you'll validate later).
```

If you _are_ expecting this `fail` to catch a bad state, you'd have to coalesce it back into control flow on the test runner thread, somehow.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategory.java
```diff
@@ -183,6 +183,16 @@ public void observe(
...
+    @Override
+    public void start(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        getSelectedPlugin().start(onComplete, onError);
+    }
+
+    @Override
+    public void stop(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        getSelectedPlugin().stop(onComplete, onError);
+    }
+
```
jamesonwilliams, line 195: Maybe we should order these nearer to `configure(...)` and `initialize(...)` in the document, since they are more lifecycle-oriented?

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -227,7 +227,29 @@ void observe(
     );
 
     /**
-     * Resets the underlying DataStore to its pre-initialized state such that no data remains on the local
+     * Starts the DataStore.  This only needs to be called if you wish to start eagerly.  If you don't call it,
+     * it will be called automatically prior to executing any other operations (#query, #save, #delete, #observe).
```
jamesonwilliams, line 231: Suggest a few tweaks to language to:
1. Remove use of "it", which increases cognitive buffering in dense technical writing, and
2. Talk explicitly about synchronization with a remote system. Currently, there is not enough differentiation from `configure()` or `initialize()`. 
```suggestion
     * Starts the DataStore's synchronization with a remote system, if DataStore is configured to support
     * remote synchronization. This only needs to be called if you wish to start the synchronization eagerly.
     * If you don't call start(), the synchronization will start automatically, prior to executing any other
     * operations (#query, #save, #delete, #observe).
```

I might suggest going back and updating `configure()`, `initialize()` (well, it's on the category), and `clear()`, to more explicitly highlight their relationship to the new `start()` and `stop()`.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -227,7 +227,29 @@ void observe(
...
+     *
+     * @param onComplete Invoked after DataStore is initialized.  This does not block until subscriptions and
+     *                  sync are complete.  To block until sync and subscriptions are complete, use
+     *                   Amplify.Hub.subscribe to listen for the DataStoreChannelEventName.READY event on
+     *                   HubChannel.DATASTORE.
+     * @param onError Invoked if an exception occurs.
```
jamesonwilliams, line 237: Similar thoughts as above, about the `onComplete`, here. But also: what's the DX for this callback, why would someone use it? Shouldn't `start(...)` wait until model synchronization is fully started, to invoke this? Otherwise, why wouldn't I just always use the `READY` event? And if I should always prefer `READY`, then couldn't we remove this callback? Lastly, we now have:
1. Configured
2. Initialized
3. Started
4. Ready

All of which mean different things. That might be okay _internally_ to our code, but I wonder if we can eliminate ~2 of these entirely from the user's view.

testutils/src/main/java/com/amplifyframework/testutils/VoidResult.java
```diff
@@ -0,0 +1,37 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.testutils;
+
+/**
+ * Result placeholder to be used when there is no value emitted from a task, but we are interested in the completion of
+ * the task.
+ */
+public final class VoidResult {
```
jamesonwilliams, line 22: Nice!! I think there are a few places where I could have used this, around the code base. I can go back and patch them up, now that you've added this.

jamesonwilliams (commented):
Code changes look very clean, nice @richardmcclellan! Thoughts below are mostly about how we communicate these changes in the Javadoc.

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
richardmcclellan: The comment was actually wrong before - the previous implementation restarted synchronization immediately).   

The new behavior matches the comment now.  `clear` won't restart synchronization, but the next interaction with DataStore will.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
rjuliano: I've struggled with this one. If the intent is to align with JS (as much as possible), `start()` really means `startSyncEngine()`, but the naming is besides the point here. One of the things we discussed is that for the mobile platforms, blocking execution during the entirety of the initialization of sync engine could lead to poor UX.

In the past, I've advocated for blocking only until `StorageObserver` starts since that's the bare minimum we would need to ensure no customer data is lost. That being said, that would give the onComplete callback yet another meaning, so I'm not too sure about that.

Looking at the plugin code, it looks like the `onComplete` callback basically lets them know that we've received their request to start the sync engine. I commented under the start method of the plugin as well, but there may be a scenario which both onComplete and onError can fire for the same call, which could lead to confusion.

Maybe the "easy" answer is to have no callbacks on start and stop, and tell devs to listen to sync engine events on the Hub.

rjuliano (commented):
This looks good. Left a couple of comments for your consideration

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
richardmcclellan: I've been scratching my head for a while about this as well :).

Here's how `start()` is implemented in this PR:
 - `onComplete` is called after:
   - plugin initialization completes (if needed)
   - `storageObserver.startObservingStorageChanges` completes.
- `onError` is called:
  - if initialization doesn't complete within 2 seconds
  - if orchestrator start stop lock is not acquired within 2 seconds
  - if `storageObserver.startObservingStorageChanges` doesn't complete within 2 seconds

Realistically, if `onError` is called, the error is not going to be something clients can handle anyway.  It'll probably only be called because of a bug in the library.

A basic app built with DataStore should look something like this (`start()` is not even a critical component):

```
private void showListOfTodos() {
    runQuery(); // Show locally cached data (or local synced data if already synced)
    Amplify.Hub.subscribe(HubChannel.DATASTORE,
            hubEvent -> DataStoreChannelEventName.READY.toString().equals(hubEvent.getName()),
            hubEvent -> runQuery()); // Show synced data
}

private void runQuery() {
    Amplify.DataStore.query(Todo.class,
            result -> { /* Display results on UI */},
            error -> { }
    );
}
```

Customers only need to call `start()` if they want to pre-cache data before displaying it to the end user.  They would call start(), presumably on app launch, and then call `query` whenever the end user navigates to a view that requires data.   Based on this, I don't think onComplete / onError make much sense, so I will remove them.  

For `stop()`, there are two customer use cases:
1. Close subscription sockets to save battery when DataStore is not needed, such as on app close.
2. Force a re-evaluation of sync query predicates. A customer would immediately call `query(...)` after `stop()`. Since we have the startStopSemaphore locking the stop operation, this should work just fine.

Callbacks aren't particularly useful for either of those use cases.   I'll bring this up with iOS/JS in the DataStore sync meeting tomorrow, and if we all agree, we will remove the callbacks.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -238,6 +236,68 @@ private Completable initializeStorageAdapter(Context context) {
...
+    @Override
+    public void start(@NonNull Action onComplete, @NonNull Consumer<DataStoreException> onError) {
+        waitForInitialization(() -> {
+            try {
+                orchestrator.start();
+            } catch (DataStoreException exception) {
+                onError.accept(exception);
+                return;
+            }
+            onComplete.call();
```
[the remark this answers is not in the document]
richardmcclellan: > Could both onComplete and onError fire for the same call? I think onComplete will always fire, but if an exception is thrown by orchestrator.start() the onError(...) callback will also be invoked which could lead to devs assuming that if onComplete() fires, they're OK to move on with the rest of their call flow.

Hmm, did I mess this up?   I think `onComplete` will _not_ be called on error, because there's a `return` on line 263, right?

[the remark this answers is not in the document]
rjuliano: Nevermind. For whatever reason I thought the onComplete.call() was outside of the curly braces.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
richardmcclellan: I spoke with the iOS and JS teams about this a few days ago, and they opposed removing these callbacks. 

The JS version of `start()` returns a Promise, which will throw an exception if there are any exceptions thrown while evaluating the client provided selection sync query code block (not implemented on Android yet, but coming soon)

So, there are reasons for having callbacks, and reasons not to.  Given that adding this new API is a bit of a one-way door, I'd like to keep the callbacks, even if they don't report very meaningful errors, since we will likely need them later (probably for selective sync). 

If there are no objections to the above, this PR is ready to fly, and I'll merge as soon as I get an approval!

rjuliano (approved):
Ship it!!!🚀

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
```diff
@@ -146,17 +149,32 @@ public void itemsPlacedInStorageArePublishedToNetwork() throws AmplifyException
...
+                // No need to do anything here.  Test will fail down below when latch times out.
+            }
+        }).start();
+
         // Try to start it again on a current thread.
         orchestrator.start();
 
+        if (!startOrchestratorInNewThreadLatch.await(2, TimeUnit.SECONDS)) {
+            fail("Failed to start orchestrator on a background thread.");
+        }
```
jamesonwilliams, line 176: How about:
```suggestion
        Completable
            .create(subscriber -> {
                new Thread(() -> {
                   try {
                       orchestrator.start();
                       subscriber.onComplete();
                   } catch (DataStoreException failure) {
                       subscriber.onFailure(failure);
                   }
                }).start();
            })
            .blockingAwait(5, TimeUnit.SECONDS);
```
Or:
```suggestion
        Await
            .<VoidResult, DataStoreException>result((onResult, onError) -> {
                new Thread(() -> {
                   try {
                       orchestrator.start();
                       onResult.accept(VoidResult.instance());
                   } catch (DataStoreException failure) {
                       onError.accept(failure);
                   }
                }).start();
            });
```

jamesonwilliams (approved):

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/OrchestratorTest.java
richardmcclellan: I like both of these better, thanks for the suggestions.  I'll go with the RxJava approach :).   RxJava feels cleaner than using a `CountDownLatch`.    I'd like to start using the [RxJava timeout operator](http://reactivex.io/documentation/operators/timeout.html) more too, instead of CountDownLatch, where possible.

PR #924, opened 2020-10-23
chore(aws-api): refactor owner-based auth unit tests

raphkim:
*Description of changes:*
* Refactor `AuthRule` (remnant from #903)
* Isolated the owner-based auth related tests from `AWSApiPluginTest`
* Refactor test structure to allow testing for different types of auth modes
* Implement `FakeJWTToken` generator to replace potentially sensitive actual ID tokens
* Add unit tests to confirm that owner field is added correctly for OIDC usecase

### DIFF
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
@@ -286,21 +286,14 @@ public final class OwnerBasedAuthTest {
     }
 
     private static final class FakeCognitoAuthProvider implements CognitoUserPoolsAuthProvider {
-        private static final String USERNAME = "Facebook_100003287976754";
-        private static final String ID_TOKEN = "eyJraWQiOiJnMmtYXC8rSXRmNFwvcmwyODhBSTNCMk9kNDVsdEU4" +
-                "ZUtIZmF0RkNRWEVDMmM9IiwiYWxnIjoiUlMyNTYifQ.eyJzdWIiOiI2YTRjZjMxMi01MWM5LTQyNjAtYjRh" +
-                "ZC0wMDdjMDdkZDdmMzMiLCJjb2duaXRvOmdyb3VwcyI6WyJ1cy13ZXN0LTJfejVWM1ZQa1h5X0ZhY2Vib29" +
-                "rIiwiQWRtaW4iXSwidG9rZW5fdXNlIjoiYWNjZXNzIiwic2NvcGUiOiJhd3MuY29nbml0by5zaWduaW4udX" +
-                "Nlci5hZG1pbiBwaG9uZSBvcGVuaWQgcHJvZmlsZSBlbWFpbCIsImF1dGhfdGltZSI6MTU5OTY4MTk2Mywia" +
-                "XNzIjoiaHR0cHM6XC9cL2NvZ25pdG8taWRwLnVzLXdlc3QtMi5hbWF6b25hd3MuY29tXC91cy13ZXN0LTJf" +
-                "ejVWM1ZQa1h5IiwiZXhwIjoxNTk5NzU3OTE3LCJpYXQiOjE1OTk3NTQzMTcsInZlcnNpb24iOjIsImp0aSI" +
-                "6ImQyNzYxMDg3LTliNmYtNDYzMS1iYWJjLTY0ZWQzM2UyNGQzMiIsImNsaWVudF9pZCI6IjNjZDdjcGJ1N2" +
-                "huYzlka2xoMmZsamg3am0xIiwidXNlcm5hbWUiOiJGYWNlYm9va18xMDAwMDMyODc5NzY3NTQifQ.FAKE-S" +
-                "IGNATURE";
+        private static final String USERNAME = "facebook-test-user";
 
         @Override
         public String getLatestAuthToken() {
-            return ID_TOKEN;
+            return FakeJWTToken.builder()
+                    .putPayload("username", USERNAME)
+                    .build()
+                    .asString();
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -57,74 +64,89 @@ public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
...
     }
 
     /**
-     * Used to specify a custom claim.  Defaults to "username" when using AuthStrategy.OWNER.
-     * CLI has been incorrectly generating a value of "cognito:username" so we also check for this incorrect default
-     * value and convert it to the proper default of "username".
+     * Used to specify a custom claim.
+     * Defaults to "username" when using AuthStrategy.OWNER.
+     *
+     * Note: CLI has been incorrectly generating a value of "cognito:username"
```
richardmcclellan, line 89:
```suggestion
     * Note: An older version of the CLI incorrectly generated a value of "cognito:username"
```

richardmcclellan (approved):
Left some minor comments, looks good overall though!

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -139,18 +161,26 @@ public boolean equals(Object object) {
 
         AuthRule authRule = (AuthRule) object;
 
-        return authStrategy == authRule.authStrategy &&
+        return ObjectsCompat.equals(authStrategy, authRule.authStrategy) &&
```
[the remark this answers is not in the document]
raphkim: haha that's a subtlety that I didn't know about. thanks for that!

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -57,74 +64,89 @@ public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
...
     public List<ModelOperation> getOperationsOrDefault() {
-        if (Empty.check(this.operations)) {
-            return Arrays.asList(
-                    ModelOperation.CREATE,
-                    ModelOperation.UPDATE,
-                    ModelOperation.DELETE,
-                    ModelOperation.READ);
-        }
-        return this.operations;
+        return Immutable.of(Empty.check(operations)
```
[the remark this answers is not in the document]
raphkim: I still think that this change makes it clear that the intention is to add _every_ `ModelOperation` enums to the list, and also futureproofs for when a new value is added or removed from the enum. 

However, I added a comment with an explicit list of the values to improve readability as a compromise. Would that work?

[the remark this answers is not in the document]
richardmcclellan: Seems reasonable!

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    /**
+     * Test that owner argument fails to be appended to subscription request if
+     * the authorization mode is not OIDC compliant.
+     */
+    @Test
+    public void ownerArgumentNotAddedWithApiKey() {
+        // Set API to use API key auth mode
+        apiName = GRAPHQL_API_WITH_API_KEY;
```
jamesonwilliams, line 141: This is a little weird, where you set the `apiName`, but then `apiName` isn't used anywhere in this `@Test`. IF you need the `apiName` value elsewhere, can you pass it into a helper method, instance of relying on the instance state of the test class?

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
```diff
@@ -0,0 +1,351 @@
...
+
+        // Attempting to subscribe to a model with owner-based auth with API key auth mode.
+        GraphQLRequest<OwnerAuth> request = ModelSubscription.onCreate(OwnerAuth.class);
+        GraphQLOperation<OwnerAuth> operation = subscribe(request);
+
+        // Subscription should fail at pre-processing
+        assertNull(operation);
```
jamesonwilliams, line 148: Is returning `null` a reasonable failure mode? Why doesn't it throw a `BadAuthException` or something to that effect?

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
```diff
@@ -0,0 +1,351 @@
...
+        apiName = GRAPHQL_API_WITH_COGNITO;
+
+        assertTrue(isOwnerArgumentAdded(Owner.class, SubscriptionType.ON_UPDATE));
+        assertTrue(isOwnerArgumentAdded(OwnerRead.class, SubscriptionType.ON_UPDATE));
+
+        assertTrue(isOwnerArgumentAdded(Owner.class, SubscriptionType.ON_DELETE));
+        assertTrue(isOwnerArgumentAdded(OwnerRead.class, SubscriptionType.ON_DELETE));
+
+        assertTrue(isOwnerArgumentAdded(Owner.class, SubscriptionType.ON_CREATE));
+        assertTrue(isOwnerArgumentAdded(OwnerRead.class, SubscriptionType.ON_CREATE));
```
jamesonwilliams, line 186: Looks like it could be a `for` (or `for`-`for`) loop?

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
```diff
@@ -0,0 +1,351 @@
...
+    @Test
+    public void ownerArgumentNotAddedIfOperationNotRestrictedWithUserPools() throws AmplifyException {
+        // Set API to use Cognito User Pools auth mode
+        apiName = GRAPHQL_API_WITH_COGNITO;
+
+        assertFalse(isOwnerArgumentAdded(OwnerCreate.class, SubscriptionType.ON_UPDATE));
+        assertFalse(isOwnerArgumentAdded(OwnerUpdate.class, SubscriptionType.ON_UPDATE));
+        assertFalse(isOwnerArgumentAdded(OwnerDelete.class, SubscriptionType.ON_UPDATE));
+
+        assertFalse(isOwnerArgumentAdded(OwnerCreate.class, SubscriptionType.ON_DELETE));
```
jamesonwilliams, line 217: I think you need some helper methods to simplify all of these imperative checks.

Maybe varargs:
```java
assertOwnerArguments(SubscriptionType.ON_DELETE,
    OwnerCreate.class, OwnerUpdate.class, OwnerDelete.class));
```
or:
```java
assertOwnerArguments(OwnerUpdate.class, SubscriptionType.values());
```
^ Then you could use a single `for`-loop to go over a list of `OnwerUpdate.class`, `OwnerDelete.class`, etc.?

jamesonwilliams (approved):
Nice!

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
raphkim: I wanted to maintain the original syntax on how the tests were originally written, but I may revisit it another time to create static methods for assertion.

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
raphkim: This was not originally written by me, but I think this format improves readability.

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
raphkim: API plugin returns a `null` operation when encountering a failure even before sending in a network request. I can also do an assertion on the failure callback, but since the above behavior is already set in stone, I figured it would be redundant. Also the helper method `subscribe(GraphQLRequest)` was written to simplify a lot of the logic behind dealing with callbacks in this test (which mostly deals with request decoration rather than the actual result of the operation), so I would rather not add a layer of complexity here.

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
raphkim: I do agree that relying on the instance state is less than ideal, and I did give this approach a try. I decided against it because
1. most of these tests were not originally written by me, so I would rather maintain the syntax of the original tests when appending my own.
2. adding in another parameter to `isOwnerArgumentAdded()` method made the test suite look overall very verbose without adding too much value to the comprehensibility of the tests (since every assertion in a given test case will use the same API).

PR #926, opened 2020-10-23
fix(api): Add owner field to selection set if implicitly defined

richardmcclellan:
This PR fixes a bug where subscriptions will not be received at all for the schema below.    The fix is to modify the request for mutations to include the owner field to the selection set.  This ensures that when the mutation is broadcast out to the subscription listeners, the owner field will be hydrated, so it won't be filtered out, because it will be able to validate that the user is authorized to view this object.

```
type Post @model @auth(rules: [{ allow: owner }])
{
  id: ID!
  title: String!
}
```

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
@@ -22,6 +22,8 @@ import androidx.core.util.ObjectsCompat;
 import com.amplifyframework.AmplifyException;
 import com.amplifyframework.api.graphql.Operation;
 import com.amplifyframework.api.graphql.QueryType;
+import com.amplifyframework.core.model.AuthRule;
+import com.amplifyframework.core.model.AuthStrategy;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.types.JavaFieldType;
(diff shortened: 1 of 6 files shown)


changxu0306 (approved):
Looks good!

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -244,6 +246,12 @@ private SelectionSet wrapPagination(SelectionSet node) {
                 } else {
                     result.add(new SelectionSet(fieldName, null));
                 }
+                for (AuthRule authRule : schema.getAuthRules()) {
+                    if (AuthStrategy.OWNER.equals(authRule.getAuthStrategy())) {
+                        result.add(new SelectionSet(authRule.getOwnerFieldOrDefault(), null));
```
jamesonwilliams, line 251:
```suggestion
                        result.add(new SelectionSet(authRule.getOwnerFieldOrDefault(), Collections.emptySet()));
```

jamesonwilliams (approved):

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
richardmcclellan: I just pushed a change to require `SelectionSet.nodes` to be nonnull, let me know what you think!

jamesonwilliams: 'Tis a thing of exceptional beauty, @richardmcclellan, thanks!

PR #927, opened 2020-10-23
feature(build): Trigger unit and integration tests in CodeBuild

rjuliano:
*Issue #, if available:*

*Description of changes:*
Please ignore the amount of commits :) I was testing the webhooks

Here's a summary of the changes in the PR:
- Adding the @Ignore annotation at the method level for test suites that had the @Ignore at the class level. This was causing an issue when running the tests in DeviceFarm (working on figuring out why)
- Added buildspec files. One is used to run unit tests when a PR is created/updated against the main branch and the other is used in a CodeBuild project that builds, packages and triggers the integration tests in device farm.
- Created a gradle task that will allow us to pull backend configuration config files from inside CodeBuild. This is a work in progress. The current backend resources are spread across multiple accounts right now so it will take some time to consolidate them. For now, we're copying the files from an S3 bucket and using the same backend resources we leverage today. This will allow us to build equivalent functionality in Code* and then we can finish migrating backend resources.
- A couple of shell scripts to be used inside CodeBuild.
- A [shell script](https://docs.aws.amazon.com/codebuild/latest/userguide/use-codebuild-agent.html) to run the CodeBuild process locally. This is actually a copy of the script referenced in the docs. It's here for convenience but can be removed if we don't feel like it belongs here.


aws-api/build.gradle
```diff
@@ -17,6 +17,8 @@ apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
 
+ext.instrumentedTestAmplifyBackend = "ApiInstrumentedTests"
```
jamesonwilliams, line 20: Can this live in the root `build.gradle`, so that you don't have to add it to each module?

aws-api/src/androidTest/java/com/amplifyframework/api/aws/GraphQLInstrumentationTest.java
```diff
@@ -123,12 +120,12 @@ public void subscriptionReceivesMutationOverAwsIam() throws ApiException {
     @Ignore("Relies on an AWS account which is no longer active.  Resources need to be regenerated.")
     public void subscriptionReceivesMutationOverCognitoUserPools() throws
             ApiException, JSONException, SynchronousMobileClient.MobileClientException {
-        currentApiName = API_WITH_COGNITO_USER_POOLS;
+        /*currentApiName = API_WITH_COGNITO_USER_POOLS;
```
jamesonwilliams, line 123: The test is `@Ignore`'d anyway -- do you need to comment the code out, too?

build.gradle
```diff
@@ -110,11 +110,34 @@ ext {
 }
 
 subprojects { project ->
+    task pullInstrumentedTestAmplifyConfig() {
```
jamesonwilliams, line 113: Can you add these new task definitions (including `runCodeBuildScript`) under `configuration/instrumentation-tests.gradle` or something?

buildspec.yml
```diff
@@ -15,37 +15,44 @@ env:
   shell: /bin/sh
 phases:
   install:
+    runtime-versions:
+      java: corretto8
```
jamesonwilliams, line 19: How is Corretto being used?

core/src/androidTest/java/com/amplifyframework/devmenu/DevMenuMainViewInstrumentationTest.java
```diff
@@ -60,7 +59,7 @@ public void resetView() {
      * Tests that the app is navigated to the environment information
      * screen when the view environment information button is pressed.
      */
-    @Test
+    @Ignore("DeviceFarm does not appear to recognize ignore at the class level.")
```
jamesonwilliams, line 62: Do we know how/why these tests fail? Can we add that description here, instead of this note?

scripts/pull_backend_config.sh
```diff
@@ -0,0 +1,30 @@
+base_dir=`pwd`
+project_name=$1
+module_name=$2
+echo "base_dir=$base_dir"
```
jamesonwilliams, line 4: You can kill these lines, use `set -x` while debugging your script.

scripts/pull_backend_config.sh
```diff
@@ -0,0 +1,30 @@
...
+cd $tmp_amplify_dir
+res_dir="$base_dir/$module_name/src/androidTest/res/raw"
+echo "Resource directory for $module_name: $res_dir"
+
+app_id=`aws amplify list-apps --query "apps[?name=='$project_name'].appId" --output text`
+echo "Amplify app_id=$app_id"
+amplify pull \
+  --amplify "{\"projectName\":\"$project_name\",\"appId\":\"$app_id\",\"envName\":\"main\",\"defaultEditor\":\"code\"}" \
+  --frontend "{\"frontend\":\"android\",\"config\":{\"ResDir\":\"config_files\"}}" \
+  --providers "{\"awscloudformation\":{\"configLevel\":\"general\",\"useProfile\":false,\"region\":\"us-east-1\"}}" \
```
jamesonwilliams, line 23: I wonder if you can write these JSON blurbs in more more a structured format, somehow?
```bash
read -d '' amplify_arg << EOF
{
    "key": "value"
}
EOF
```

scripts/codebuild_build.sh
```diff
@@ -0,0 +1,193 @@
...
+    if [ $OSTYPE == "msys" ]
+    then
+        return 0
+    else
+        return 1
+    fi
+}
+
+function usage {
+    echo "usage: codebuild_build.sh [-i image_name] [-a artifact_output_directory] [options]"
```
jamesonwilliams, line 32: You can use a here-doc:
```bash
cat <<- EOF
    stuff without echo
    just regular text
EOF
```

rjuliano:
Bump

richardmcclellan (commented):
Have you thought about adding logic to retry up to 3 times if there are failures (mostly for some of our transient tests)?  This helped a lot on CircleCI.    Maybe we can monitor this for a week and see if it's needed?

aws-datastore/src/test/java/com/amplifyframework/datastore/AWSDataStorePluginTest.java
```diff
@@ -106,7 +106,7 @@ public void setup() {
      * crashing the calling thread. Basic. 🙄
      * @throws AmplifyException Not expected; on failure to configure of initialize plugin.
      */
-    @Test
+    @Ignore("Needs investigation as it fails sporadically.")
```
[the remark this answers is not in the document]
rjuliano: This was supposed to be temporary. I'm going to remove the ignore. I was having issues with this one test failing at times both locally and inside the container. Need to do some investigating

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageDownloadTest.java
```diff
@@ -40,7 +40,6 @@
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
-import org.junit.Test;
```
jamesonwilliams, line 43: You can leave the `@Test`s in place, and just add the `@Ignore`, right?

scripts/pull_backend_config_from_s3
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
jamesonwilliams, line 1: This file looks mostly like the one in `.circleci`, right? Can we remove the copy in `.circleci`, and just use this one?

jamesonwilliams (approved):

PR #920, opened 2020-10-21
chore(aws-api): move owner-based auth logic out of plugin

raphkim:
*Description of changes:*
Simple refactor to move the owner-based auth logic out of API plugin.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #930, opened 2020-10-26
feature(aws-api): support custom group claim

raphkim:
*Description of changes:*
Support custom group claim for OIDC `@auth` use-case.

*No tests are included in this PR, but the changes were tested in a sample app. Explicit unit tests will follow in a separate PR unless requested otherwise.*

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -292,7 +292,7 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
             try {
                 AppSyncGraphQLRequest<R> appSyncRequest = (AppSyncGraphQLRequest<R>) request;
                 AuthRule ownerRuleWithReadRestriction = null;
-                Map<String, List<String>> readAuthorizedGroupsMap = new HashMap<>();
+                Map<String, Set<String>> readAuthorizedGroupsMap = new HashMap<>();
 
                 // Note that we are intentionally supporting only one owner rule with a READ operation at this time.
                 // If there is more than one, the operation will fail because AppSync generates a parameter for each
@@ -309,13 +309,14 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
                             return null;
                         }
                     } else if (isReadRestrictingStaticGroup(authRule)) {
+                        // Group read-restricting groups by the claim name
                         String groupClaim = authRule.getGroupClaimOrDefault();
                         List<String> groups = authRule.getGroups();
-                        List<String> readAuthorizedGroups = readAuthorizedGroupsMap.get(groupClaim);
+                        Set<String> readAuthorizedGroups = readAuthorizedGroupsMap.get(groupClaim);
                         if (readAuthorizedGroups != null) {
                             readAuthorizedGroups.addAll(groups);
                         } else {
-                            readAuthorizedGroupsMap.put(groupClaim, new ArrayList<>(groups));
+                            readAuthorizedGroupsMap.put(groupClaim, new HashSet<>(groups));
                         }
                     }
                 }
@@ -404,14 +405,22 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
     }
 
     private boolean userNotInReadRestrictingGroups(
-            Map<String, List<String>> readAuthorizedGroupsMap,
+            Map<String, Set<String>> readAuthorizedGroupsMap,
             AuthorizationType authType
     ) throws ApiException {
-        for (Map.Entry<String, List<String>> entry : readAuthorizedGroupsMap.entrySet()) {
+        // Iterate through map of "group claim" -> "read-restricting groups from that claim". e.g.
+        // {
+        //   "https://myapp.com/claims/groups" -> {Admins}
+        //   "https://differentapp.com/claims/groups" -> {Moderators, Editors}
+        // }
+        for (Map.Entry<String, Set<String>> entry : readAuthorizedGroupsMap.entrySet()) {
             String groupClaim = entry.getKey();
-            List<String> readAuthorizedGroups = entry.getValue();
+            // Get a list of groups that user belongs in for a given claim.
+            // e.g. [Admins, User]
             List<String> userGroups = getUserGroups(groupClaim, authType);
-            if (!Collections.disjoint(readAuthorizedGroups, userGroups)) {
+            Set<String> readAuthorizedGroups = entry.getValue();
+            // If user belongs in any group for a given group claim, return false.
+            if (!Collections.disjoint(userGroups, readAuthorizedGroups)) {
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -291,7 +292,7 @@ public String getVersion() {
             try {
                 AppSyncGraphQLRequest<R> appSyncRequest = (AppSyncGraphQLRequest<R>) request;
                 AuthRule ownerRuleWithReadRestriction = null;
-                ArrayList<String> readAuthorizedGroups = new ArrayList<>();
+                Map<String, List<String>> readAuthorizedGroupsMap = new HashMap<>();
```
jamesonwilliams, line 295: The data type here is starting to look like a poor-man's data structure. Does it make sense to create a `class ReadAuthorizedGroups`, to encapsulate the data?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -361,65 +363,61 @@ private boolean isReadRestrictingOwner(AuthRule authRule) {
...
-                    "This should never happen - see attached exception for more details and report to us on GitHub."
+                    "Failed to parse group claim from the token.",
+                    AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION
             );
         }
 
         return groups;
     }
 
+    private boolean userNotInReadRestrictingGroups(
```
jamesonwilliams, line 406: How can we carve more of this logic out of the `AWSApiPlugin`?

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
```diff
@@ -68,6 +68,9 @@
     private static final String GRAPHQL_API_WITH_COGNITO = "graphQlApi_cognito";
     private static final String GRAPHQL_API_WITH_OIDC = "graphQlApi_oidc";
 
+    private static final String GROUP_A = "http://myapp1.com/claims/groups";
+    private static final String GROUP_B = "http://myapp2.com/claims/groups";
```
jamesonwilliams, line 72: How are group A and B different? Can  you give semantic names, `PRIVILEGED_GROUP`, `NO_ACCESS_GROUP` Or `ALICE_GROUP` and `BOB_GROUP`, with a little comment about the security relationship between Alice and Bob?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: > poor-man's data structure.
🤣  yep that's exactly right. I can do that, but I think that "map of lists" is a common enough data structure that it is passable as is. I think the gains from creating an entirely new class for this will be very little in terms of readability and intuitiveness. what do you think?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: this will have to be merge-conflicted in my [other PR](#920), ugh. This helper method will definitely not live in the plugin code for long.

aws-api/src/test/java/com/amplifyframework/api/aws/auth/OwnerBasedAuthTest.java
raphkim: This actually doesn't get used at the moment, but snuck its way in the PR... I'll take out the unused items from the unit test.

richardmcclellan:
It would help to have some comments / documentation that describe what use case this PR addresses, and how, for the benefit of future developers.  Perhaps you could provide a sample schema, and describe how it will be processed?  It's not really apparent from just looking at the code right now.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
jamesonwilliams: Ah right, I forgot that wasn't in, yet. I was like "damn, I thought we moved some of this out already..."

Sounds good. I remember you saying you'd rebase that one onto this. That plan works.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
jamesonwilliams: Yea, it's not a big deal either way. Up to you. The benefit of doing the class is that you'll get to give a semantic label to the key and to the values. Looking at this `readAuthorizedGroupsMap` right now, I'm not certain what the key is, until I scroll down and look at the impl.

TrekSoft: Yeah I feel the benefit to seeing right away the structure of the data from the map outweighs the benefit of creating a custom class for naming purposes.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -361,65 +363,61 @@ private boolean isReadRestrictingOwner(AuthRule authRule) {
 
     private boolean isReadRestrictingStaticGroup(AuthRule authRule) {
         return AuthStrategy.GROUPS.equals(authRule.getAuthStrategy())
-            && authRule.getGroups() != null && !authRule.getGroups().isEmpty()
+            && !authRule.getGroups().isEmpty()
```
TrekSoft, line 366: Why remove the null check?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -361,65 +363,61 @@ private boolean isReadRestrictingOwner(AuthRule authRule) {
...
-            identityValue = CognitoJWTParser
+            return CognitoJWTParser
                     .getPayload(getAuthToken(authType))
                     .getString(identityClaim);
         } catch (JSONException | CognitoParameterInvalidException error) {
-            // Could not read identity value from the token...
-            // Exception will be thrown so do nothing for now
-        }
-
-        if (identityValue == null || identityValue.isEmpty()) {
```
TrekSoft, line 380: The reason this was done the way it was is because an error won't be thrown if the identityClaim key is present but empty. And it seemed to make sense to throw the error in that case.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: It's redundant because `AuthRule#getGroups()` explicitly has `@NonNull` return value :)

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
raphkim: i don't think that's something that should be caught in the client side. It should be treated the same as having an incorrect/non-existent username, which should be caught by the server.

TrekSoft: Fair enough

PR #931, opened 2020-10-27
chore(aws-datastore): clarify hub event names

jamesonwilliams:
The `DataStoreChannelEventName` enumeration contained two generations of
event names:

 1. An initial set, which included `RECEIVED_FROM_CLOUD`, and
    `PUBLISHED_TO_CLOUD`;
 2. A larger set, containing many more, exposing various internal
    details of the synchronization engine.

The later work partly duplicated the original work.

This commit *removes* the `PUBLISHED_TO_CLOUD`, which was duplicated by
the `OUTBOX_MUTATION_PROCESSED`,  and *renames* the
`RECEIVED_FROM_CLOUD` event to `SUBSCRIPTION_DATA_PROCESSED`.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
@@ -30,7 +30,7 @@ import com.amplifyframework.datastore.appsync.AppSyncClient;
 import com.amplifyframework.datastore.appsync.ModelMetadata;
 import com.amplifyframework.datastore.appsync.ModelWithMetadata;
 import com.amplifyframework.datastore.appsync.SynchronousAppSync;
-import com.amplifyframework.datastore.syncengine.PendingMutation;
+import com.amplifyframework.datastore.syncengine.OutboxMutationEvent;
 import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.hub.HubEvent;
 import com.amplifyframework.hub.HubEventFilter;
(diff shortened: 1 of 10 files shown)


rjuliano (approved):
Looks good!!! 🚢

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
```diff
@@ -154,11 +153,7 @@ public void canDrainMutationOutbox() throws DataStoreException {
         mutationProcessor.startDrainingMutationOutbox();
 
         // Assert: the event was published
-        List<HubEvent<?>> events = accumulator.await();
-        assertEquals(1, events.size());
-        @SuppressWarnings("unchecked")
-        PendingMutation<BlogOwner> mutation = (PendingMutation<BlogOwner>) events.get(0).getData();
-        assertEquals(createTony, mutation);
```
raphkim, line 161: wouldn't this have been a stronger assertion?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -180,22 +180,4 @@
...
-         */
-        IGNORE_PENDING_MUTATIONS,
-
-        /**
-         * When merging, the contents of the mutation outbox will be considered.
-         * If there is already a pending mutation in the mutation outbox, for a model of the
-         * same ID as the model being merged -- then the merge will *not* modify the existing model.
-         */
-        CONSIDER_PENDING_MUTATIONS
-    }
```
raphkim, line 200: where was this being used before?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
```diff
@@ -195,34 +195,42 @@ public void syncDownFromCloudIsWorking() throws AmplifyException {
...
-            if (modelWithMetadata == null) {
-                return false;
-            } else if (!model.getClass().isAssignableFrom(modelWithMetadata.getModel().getClass())) {
+            if (!DataStoreChannelEventName.SUBSCRIPTION_DATA_PROCESSED.equals(actualEventName)) {
                 return false;
             }
-            return model.getId().equals(modelWithMetadata.getModel().getId());
+            return hasModelData(model, (ModelWithMetadata<? extends Model>) event.getData());
         };
     }
```
raphkim, line 213: should we put all these inside `TestHubEventFilters` as `isProcessed()` and `isReceived()` for consistency?

raphkim (approved):
Nice PR!

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
jamesonwilliams: I wish. The problem is that `TestHubEventFilters` lives in `aws-datastore/src/test`, and this file is in `aws-datastore/src/androidTest`. I evaluated a few workaround to that hurdle, but none were very elegant. If I moved it to `testutils`, then `testutils` needs to depend on the DataStore plugin module, and that doesn't seem right.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
jamesonwilliams: There use to be some logic to change the behavior depending on whether the change was outbound, or inbound.

If outbound, the mutation outbox would have a pending mutation, while being merged. So, we'd have to not consider it.

The issue was later remedied by changing the sequence of events, so that the event gets removed from the outbox before being merged.

But, this enum apparently didn't removed in that previous change.

core/src/main/java/com/amplifyframework/datastore/DataStoreChannelEventName.java
```diff
@@ -51,14 +42,16 @@
...
      */
     SUBSCRIPTIONS_ESTABLISHED("subscriptionsEstablished"),
 
     /**
-     * The DataStore as a whole (not just the sync piece) is ready. At this point all data is available.
+     * The server sent the client data over the WebSocket subscription. The data was
+     * successfully melded back into the local store.
      */
-    READY("ready"),
+    SUBSCRIPTION_DATA_PROCESSED("subscriptionDataProcessed"),
```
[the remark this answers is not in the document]
jamesonwilliams: I think we can achieve that by omitting it from the public docs, for now. It might make sense for the other platforms to add this, later?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
jamesonwilliams: It was duplicative logic. The `isProcessed` filter already checks for the model type and ID. So if the accumulator even gets an event, it means we did find this match, already.

In other words, this assertion is sort of just testing the `isProcessed` filter again.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AWSDataStorePluginInstrumentedTest.java
raphkim: Ah shoot, I think I vaguely remember this being an issue when I was writing the tests :/

PR #937, opened 2020-10-27
Add access to orchestrator targetMode by DataStorePlugin

teoremasail:
Start orchestrator in LOCAL_ONLY mode and allow the datastore user to set the orchestrator in online mode SYNC_VIA_API once he has successfully signed in by AWSCognitoAuthPlugin.

I might miss alot of things with this pull request or even break expected behaviour but it helps me to get the apiSync running when the user is not logged in before I start the activity.


rjuliano (commented):
Thanks for the contribution. I think think you will be able to accomplish the same thing you're trying to accomplish once #909 gets published. That PR adds a start/stop method to address the authentication use case you mentioned.

It also modifies the plugin such that it does not start the orchestrator from the configure method. Instead, the orchestrator is started either by an explicit call to start() or an interaction with the DataStore.

Let me follow up on that PR and see how close we are to pushing it.

teoremasail:
Hi @rjuliano thanks you for the feedback. I will give the #909 a try as soon as its released.
Thanks aswell for the great support on Discord, especially to @jamesonwilliams

PR #938, opened 2020-10-28
Add Builder for AuthRule

fjnoyp:
PR for Datastore module of Amplify Flutter 

Provides a public constructor for AuthRule. 

When transmitting ModelSchema from Dart to Android, we need to recreate the AuthRule object.  However the existing API only allows AuthRule objects to be created from the annotation object..

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
+++ b/core/src/main/java/com/amplifyframework/core/model/AuthRule.java
@@ -54,6 +54,28 @@ public final class AuthRule {
     }
 
     /**
+     * Construct the AuthRule object from the builder.
+     * For internal use only.
+     */
+    private AuthRule(@NonNull AuthRule.Builder builder) {
+        this.authStrategy = builder.authStrategy;
+        this.ownerField = builder.ownerField;
+        this.identityClaim = builder.identityClaim;
+        this.groupClaim = builder.groupClaim;
+        this.groups = builder.groups;
+        this.groupsField = builder.groupsField;
+        this.operations = builder.operations;
+    }
+
+    /**
+     * Return the builder object.
+     * @return the builder object.
+     */
+    public static AuthRule.Builder builder() {
+        return new AuthRule.Builder();
+    }
+
+    /**
      * Returns the type of strategy for this {@link AuthRule}.
      * @return the type of strategy for this {@link AuthRule}
      */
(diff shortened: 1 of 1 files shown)


jamesonwilliams (commented):
Looks pretty good to me. A few improvements to this builder:

**Collections**. Regarding the `groups` and `operations`: you can mark these as `final`. Add a `private Builder() { ... }` in which you assign these values as `new ArrayList<>();`. In the builder configurators, do `if (provided != null) { groups.clear(); groups.addAll(provided); }`. This way, the `groups` and `operations` will never be null, only possibly empty.

**Validation.** Do you need to validate that the values being passed to the builder are not-null? You can make every assignment as `this.ownerField = Objects.requireNonNull(this.ownerField);`. That will prevent the caller from supplying a null on accident and setting up a bogus rule. Lastly, in the `build()` method, you should validate the provided values, to ensure that they make sense. If all of the options are null, the `AuthRule` as a whole probably shouldn't be built, since it would be useless.

**Annotations.** Please add `@NonNull` or `@Nullable` to each and every return type / input parameter added here. These are crucial to get proper nullability cues in Kotlin (e.g., to have a `?` or a `!!`, or ideally, neither of those.)

richardmcclellan:
The only property that must be specified is `allow`, so I would suggest putting `Objects.requireNonNull(allow);` at the top of the build() method.   

Tagging onto @jamesonwilliams's feedback, I would suggest that all of the properties should be `@NonNull`, since they are all non-null when created with an `@AuthRule` annotation.  Unset Strings should default to empty `String` and unset Lists should default to empty Lists, if they aren't provided to the builder.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -161,4 +186,104 @@ public String toString() {
...
+            this.operations = Objects.requireNonNull(operations);
+            return this;
+        }
+
+        /**
+         * Builds an immutable AuthRule instance using
+         * builder object.
+         * @return AuthRule instance
+         */
+        public AuthRule build() {
```
jamesonwilliams, line 284:
```suggestion
        @NonNull
        public AuthRule build() {
```

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -53,6 +56,28 @@ public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
...
+        this.groups = builder.groups;
+        this.groupsField = builder.groupsField;
+        this.operations = builder.operations;
+    }
+
+    /**
+     * Return the builder object.
+     * @return the builder object.
+     */
+    public static AuthRule.Builder builder() {
```
jamesonwilliams, line 77:
```suggestion
    @NonNull
    public static AuthRule.Builder builder() {
```

jamesonwilliams (approved):

PR #949, opened 2020-11-02
fix: consider deletion predicate when requested

jamesonwilliams:
The `DataStoreCategoryBehavior` provides a `delete(...)` method which
evaluates a predicate condition. This condition was never being passed
to the `LocalStorageAdapter`, though. It was just ignored.

This fix wires the predicate into the `LocalStorageAdapter` so that it
will actually be considered.


richardmcclellan (approved):
Should we remove the `delete` method of the `SQLiteStorageAdapter` that takes no predicate, since `AWSDataStorePlugin` is not calling it any more?  Or is it still being used by some other code?

Amplifiyer (approved):

jamesonwilliams:
> Should we remove the `delete` method of the `SQLiteStorageAdapter` that takes no predicate, since `AWSDataStorePlugin` is not calling it any more? Or is it still being used by some other code?

Sounds like a fair idea. Will update!

richardmcclellan (approved):

PR #952, opened 2020-11-03
fix(datastore) only fire DataStore Hub READY event if not already started

richardmcclellan:
A common `DataStore` use case is to call `DataStore.query` after the `READY` event has fired, like this:

```
Amplify.Hub.subscribe(HubChannel.DATASTORE,
    { DataStoreChannelEventName.READY.toString().equals(it.name) },
    { query() }
)

fun query() {
    Amplify.DataStore.query(Blog::class.java,
        { LOG.debug("query succeeded") },
        { LOG.error("Error querying: " + it.message, it) }
    )
}
```

This results in an infinite loop though, because `query` calls `start`, which publishes another `READY` event, etc.

This PR modifies the `start` method to not fire the `READY` event if DataStore has already been previously started, so the READY event will only be fired once.


rjuliano (approved):
Ship it!!! 🚢

PR #953, opened 2020-11-05
User defined error handling

raphkim:
*Description of changes:*
Implements custom error handling to be invoked upon failure to publish a mutation.


aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -508,6 +509,7 @@ public void deltaSyncRequestedIfLastSyncIsRecent() throws DataStoreException {
      * @throws DataStoreException On failure to build GraphQLRequest for sync query.
      */
     @Test
+    @Ignore("Custom error handler no longer handles initial cloud sync failures.")
```
jamesonwilliams, line 512: You can delete the test! :0D

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
```diff
@@ -221,4 +223,46 @@ public void conflictHandlerInvokedForUnhandledConflictError() throws DataStoreEx
...
     }
+
+    /**
+     * If the AppSync response to the mutation contains not-empty GraphQLResponse error
+     * list without any ConflictUnhandled error, then the user-provided error handler
+     * should be invoked.
+     * @throws DataStoreException On failure to obtain configuration from the provider
+     */
+    @Test
+    public void errorHandlerInvokedForError() throws DataStoreException {
```
jamesonwilliams, line 234:
```suggestion
    public void errorHandlerInvokedForPublicationError() throws DataStoreException {
```

aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
```diff
@@ -137,7 +139,7 @@ public void testInvalidKeyThrowsException() throws DataStoreException, JSONExcep
         public void onConflictDetected(
                 @NonNull ConflictData<? extends Model> conflictData,
                 @NonNull Consumer<ConflictResolutionDecision<? extends Model>> onDecision) {
-            onDecision.accept(ConflictResolutionDecision.retry(null));
+            onDecision.accept(ConflictResolutionDecision.retry(RandomString::string));
```
jamesonwilliams, line 142: Hm. Isn't this field supposed to store a model instance?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
```diff
@@ -221,4 +223,46 @@ public void conflictHandlerInvokedForUnhandledConflictError() throws DataStoreEx
...
+        // Arrange a user-provided error handler.
+        CountDownLatch handlerInvocationsRemainingCount = new CountDownLatch(1);
+        when(configurationProvider.getConfiguration())
+            .thenReturn(DataStoreConfiguration.builder()
+                .errorHandler(error -> handlerInvocationsRemainingCount.countDown())
```
jamesonwilliams, line 239: Right now, you test that the config provided was asked to supply an error handler. That's related to what you want to test, but not all the way there. _Ideally_, you would latch on the interior of an error handler closure, itself.

```console
errorToBeHandled -> {
    // Aha! The handler was actually invoked
}
```

Otherwise, this test seems pretty good!

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreErrorHandler.java
```diff
@@ -15,10 +15,40 @@
...
+    @NonNull
+    static DataStoreErrorHandler defaultHandler() {
+        return new DefaultDataStoreErrorHandler();
+    }
+
+    /**
+     * Default instance for handling error response from the GraphQL server. Error message
+     * is logged, but no further action is taken.
+     */
+    final class DefaultDataStoreErrorHandler implements DataStoreErrorHandler {
```
jamesonwilliams, line 44: I think I had named these originally. And I apologize. I now have come up with a better idea. 😆 

Maybe we should more clearly name these:
```suggestion
    @NonNull
    static LoggingErrorHandler loggingHandler() {
        return new LoggingErrorHandler();
    }

    /**
     * Default instance for handling error response from the GraphQL server. Error message
     * is logged, but no further action is taken.
     */
    final class LoggingErrorHandler implements DataStoreErrorHandler {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
```diff
@@ -0,0 +1,82 @@
...
+    private final GraphQLResponse.Error error;
+    private final M local;
+    private final M remote;
+
+    /**
+     * Constructs a new error instance containing details regarding
+     * a failed publication to the cloud and involved models.
+     * @param error the error response from the server
+     * @param local the local model pending mutation
+     * @param remote the remote model if applicable
```
jamesonwilliams, line 43: We should bottom out on this. Right now, the `DataStoreError` looks _alot_ like the `ConflictData` bundle used by the conflict handler. I think the two concepts are getting increasingly close to eachother.

jamesonwilliams (commented):
Let's try to capture some `GraphQLResponse.Error` JSONs for the various error scenarios we're trying to intercept. Do they actually include a server model, for the `DataStoreError` shape?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
```diff
@@ -145,7 +145,40 @@ public String toString() {
...
+         * conflict handler because the supplied resource LambdaConflictHandlerArn is not found.
+         */
+        CONFLICT_ERROR("ConflictError"),
+
+        /**
+         * Max retry attempts were reached for conflict resolution.
+         * Example: Too many concurrent requests on the same object. Before the conflict is resolved,
+         * the object is updated to a new version by another client.
+         */
+        MAX_CONFLICTS("MaxConflicts"),
```
rjuliano, line 159: Are you tracking retry attempts or number of conflicts? If it's retry attempts, maybe naming it `MaxRetries` would make more sense.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -154,15 +153,9 @@ private Completable createHydrationTask(Class<? extends Model> modelClass) {
...
-                LOG.warn("Initial cloud sync failed.", failureToSync);
-                DataStoreErrorHandler dataStoreErrorHandler =
-                    dataStoreConfigurationProvider.getConfiguration().getErrorHandler();
-                dataStoreErrorHandler.accept(new DataStoreException(
-                    "Initial cloud sync failed.", failureToSync,
-                    "Check your internet connection."
-                ));
-            })
+            .doOnError(failureToSync ->
+                LOG.warn("Initial cloud sync failed.", failureToSync)
```
rjuliano, line 157: Why would we not want to call the error handler if something happens during the full/delta sync?

rjuliano (commented):
Looks good. Left one question and one minor comment.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
raphkim: I simply listed all of the possible server [errors from AppSync](https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html#errors) in case that the customer wants to know :)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
raphkim: For now, Amplify only handles errors from mutation, and anything during query/subscription is purposefully ignored (across all platforms)

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
raphkim: Luckily, the error handler is obtained from the configuration _ONLY_ when the handler is being invoked during runtime, so this is ultimately testing both haha. I can make it more explicit as you say though.

rjuliano (approved):
All my questions were addressed, so I'm giving it a 👍  🚢

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
jamesonwilliams: Yea! The "explicit" part, that's all. Thanks :-D

aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
raphkim: Not really sure what it was, but my Android Studio was complaining that a `null` value is being supplied into `@NonNull` parameter, so I just put in a fake Model haha

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MutationProcessorTest.java
raphkim: On second thought, this test is actually doing exactly that! The given lambda _is_ the implementation of error handler, so the fact that latch was counted down means the handler itself was invoked.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreErrorHandler.java
raphkim: I will rename `DefaultDataStoreErrorHandler` to `LoggingErrorHandler` as you suggest to make the class name make more sense, but I will keep the static method `defaultHandler()` name because we may decide to change the default behavior in the future.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
raphkim: I think it's _fine_.

As I mentioned before, I tried consolidating all of these into generic AppSync related objects, and frankly, it wasn't super difficult since they contain similar data. However, I'm not sure if introducing breaking changes at this point is a good idea or even worth the effort. We already make it clear that `ConflictUnhandled` error is handled specifically by the conflict handler, so a little bit of separation might make it more clear what we are doing internally.

aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
jamesonwilliams: Well but it says `RandomString::string` -- not `SomeModel.instance()`, that was my point.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
jamesonwilliams: We should aim to have our error handler look like the JS and iOS ones. If we're going to do something different, let's loop them in.

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
raphkim: Exposing `remote` model was my attempt to mirror the JS error handler input. The rest of the parameters are exposed in the form of `GraphQLResponse.Error`, because the user should be able to obtain the errorType and such from it. I thought about exposing it directly, but I didn't do it in case that the GraphQL error was not a direct AppSync error. 

Speaking of which, my current implementation doesn't pass back the operation type! I will add that in before I merge the changes.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Gets the remote item in the server if provided by the
+     * error payload.
+     * @return the remote model.
+     */
+    @Nullable
+    public M getRemote() {
+        return remote;
+    }
```
jamesonwilliams, line 81: Oh, `equals`, `hashCode`, `toString`!

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
```diff
@@ -32,37 +33,59 @@
...
-     * @param remote the remote model if applicable
+     *
+     * @param error     the error response from the server
+     * @param operation the type of operation
+     * @param local     the local model pending mutation
+     * @param remote    the remote model if applicable
      */
     public DataStoreError(
             @NonNull GraphQLResponse.Error error,
+            @NonNull Operation operation,
```
jamesonwilliams: If the `DataStoreError` is packaged under `com.amplifyframework.datastore`, then we should try to treat these two fields as implementation concerns (or, in other words, do not provide getters for them.)

Right now, the `DataStoreError` sort of looks like an API thing by a different name. You could also call this an `AppSyncError`, or a `MutationError`, or something like that?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: My understanding is that this is the long term plan as well.  The general philosophy is that we should not make the client deal with errors if at all possible.  The only thing the client likely cares about is whether datastore is in sync, or not, which they can get from the [Hub's networkStatus event](https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/android).   So, for sync and subscription failures, we should just handle these internally, by retrying with exponential backoff (already exists for iOS/JS, and something we should try to add in our next sprint)

Mutations are a special case though.  If a mutation fails for an unrecoverable reason (e.g. user is `Unauthorized`), we don't want it keep it in the persistent mutation queue, since it will just continually fail over and over again and block the queue.  We also can't remove it though, because we don't want to delete customer data.  We can, however, delete it after giving it back to the customer via the error handler.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreError.java
raphkim: Considering how similar the error handler is to the conflict handler in the same package, I figured they should be grouped together. I considered making this error class an internal class inside `DataStoreErrorHandler` like `ConflictData` is inside `DataStoreConflictHandler`, but I left it as a standalone class for arbitrary reasons.

As for the not exposing `GraphQLResponse.Error` and `Operation`, what is the purpose? Wouldn't they be the most important bits of information that the customer might want to know when handling the error? Besides, I am trying to reach functional parity with JS here by exposing all of these properties.

Lastly, I decided to leave the name more generic (I did initially name it `MutationError` like you suggested) in case that we might want to support generic error handling for any GraphQL error in the future. Also, I invoke the error handler for any GraphQL error from mutation publication at the moment (not sure if JS does for only AppSync-related error or not), so `AppSyncError` would be less appropriate.

raphkim:
As per last discussion, error handling will be done via Hub. Closing the PR.

PR #959, opened 2020-11-13
feat(datastore) selective sync

richardmcclellan:
This PR adds the selective sync feature to DataStore, which consists of the following parts:
 - **Configuration**: Clients configure a `DataStoreSyncExpression`, which can resolve a `QueryPredicate`. 
 - **Sync**: Sync requests sent to AppSync include this `QueryPredicate`, so that data is filtered out server side.
 - **Subscriptions**:  AppSync's subscription operation does not currently support a `filter` argument, so this PR filters out incoming subscriptions that don't match the predicate on the client side.  In the future, if AppSync adds this, we can switch to server-side filtering, which will help reduce network traffic on the subscription socket.
 - **Runtime Modification**: Clients can force a re-evaluation of their `DataStoreSyncExpression` at runtime by calling `DataStore.stop()`, `DataStore.start()`, if they wish to modify the filter to obtain a different set of data.

DX:
```
Amplify.addPlugin(new AWSDataStorePlugin(DataStoreConfiguration.builder()
    .syncExpression(Blog.class, () -> Blog.NAME.equals("foo"))
    .syncExpression(Post.class, () -> Post.TITLE.beginsWith("bar"))
    .syncExpression(Comment.class, () -> Comment.CONTENT.contains("baz"))
    .build()));
Amplify.configure(context);
```

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
@@ -177,9 +177,10 @@ public final class DataStoreConfiguration {
     }
 
     /**
-     * Returns a QueryPredicate that should be used to filter the data that is synced.
+     * Returns a QueryPredicate that should be used to filter data received from AppSync,
+     * either during a base sync or over the real-time subscription.
      * @param modelClass Class of the {@link Model}.
-     * @return a QueryPredicate that should be used to filter the data that is synced.
+     * @return a QueryPredicate that should be used to filter the data received from AppSync
      */
     @NonNull
     public QueryPredicate getSyncQueryPredicate(Class<? extends Model> modelClass) {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
@@ -22,8 +22,6 @@ import androidx.annotation.VisibleForTesting;
 import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.core.model.Model;
-import com.amplifyframework.core.model.query.predicate.QueryPredicate;
-import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -177,23 +175,9 @@ public final class DataStoreConfiguration {
     }
 
     /**
-     * Returns a QueryPredicate that should be used to filter data received from AppSync,
-     * either during a base sync or over the real-time subscription.
-     * @param modelName class name of the {@link Model}.
-     * @return a QueryPredicate that should be used to filter the data received from AppSync
-     */
-    @NonNull
-    public QueryPredicate getSyncQueryPredicate(String modelName) {
-        DataStoreSyncExpression expression = syncExpressions.get(modelName);
-        if (expression != null) {
-            return expression.resolvePredicate();
-        }
-        return QueryPredicates.all();
-    }
-
-    /**
-     * Returns the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
-     * @return the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
+     * Returns the Map of all {@link DataStoreSyncExpression}s used to filter data received from AppSync, either during
+     * a sync or over the real-time subscription.
+     * @return the Map of all {@link DataStoreSyncExpression}s.
      */
     @NonNull
     public Map<String, DataStoreSyncExpression> getSyncExpressions() {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
@@ -21,10 +21,16 @@ import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import androidx.core.util.ObjectsCompat;
 
+import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.query.predicate.QueryPredicate;
+import com.amplifyframework.core.model.query.predicate.QueryPredicates;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
@@ -44,21 +50,16 @@ public final class DataStoreConfiguration {
     private final DataStoreConflictHandler conflictHandler;
     private final Integer syncMaxRecords;
     private final Integer syncPageSize;
-    private Long syncIntervalInMinutes;
+    private final Map<Class<? extends Model>, DataStoreSyncExpression> syncExpressions;
+    private final Long syncIntervalInMinutes;
 
-    private DataStoreConfiguration(
-            DataStoreErrorHandler errorHandler,
-            DataStoreConflictHandler conflictHandler,
-            Long syncIntervalInMinutes,
-            Integer syncMaxRecords,
-            Integer syncPageSize) {
-        this.errorHandler = errorHandler;
-        this.conflictHandler = conflictHandler;
-        this.syncMaxRecords = syncMaxRecords;
-        this.syncPageSize = syncPageSize;
-        if (syncIntervalInMinutes != null) {
-            this.syncIntervalInMinutes = syncIntervalInMinutes;
-        }
+    private DataStoreConfiguration(Builder builder) {
+        this.errorHandler = builder.errorHandler;
+        this.conflictHandler = builder.conflictHandler;
+        this.syncMaxRecords = builder.syncMaxRecords;
+        this.syncPageSize = builder.syncPageSize;
+        this.syncIntervalInMinutes = builder.syncIntervalInMinutes;
+        this.syncExpressions = builder.syncExpressions;
     }
 
     /**
@@ -175,6 +176,29 @@ public final class DataStoreConfiguration {
         return this.syncPageSize;
     }
 
+    /**
+     * Returns a QueryPredicate that should be used to filter the data that is synced.
+     * @param modelClass Class of the {@link Model}.
+     * @return a QueryPredicate that should be used to filter the data that is synced.
+     */
+    @NonNull
+    public QueryPredicate getSyncQueryPredicate(Class<? extends Model> modelClass) {
+        DataStoreSyncExpression expression = syncExpressions.get(modelClass);
+        if (expression != null) {
+            return expression.resolvePredicate();
+        }
+        return QueryPredicates.all();
+    }
+
+    /**
+     * Returns the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
+     * @return the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
+     */
+    @NonNull
+    public Map<Class<? extends Model>, DataStoreSyncExpression> getSyncExpressions() {
+        return this.syncExpressions;
+    }
+
     @Override
     public boolean equals(@Nullable Object thatObject) {
         if (this == thatObject) {
@@ -196,7 +220,13 @@ public final class DataStoreConfiguration {
         if (!ObjectsCompat.equals(getSyncPageSize(), that.getSyncPageSize())) {
             return false;
         }
-        return ObjectsCompat.equals(getSyncIntervalInMinutes(), that.getSyncIntervalInMinutes());
+        if (!ObjectsCompat.equals(getSyncIntervalInMinutes(), that.getSyncIntervalInMinutes())) {
+            return false;
+        }
+        if (!ObjectsCompat.equals(getSyncExpressions(), that.getSyncExpressions())) {
+            return false;
+        }
+        return true;
     }
 
     @Override
@@ -206,6 +236,7 @@ public final class DataStoreConfiguration {
         result = 31 * result + (getSyncMaxRecords() != null ? getSyncMaxRecords().hashCode() : 0);
         result = 31 * result + (getSyncPageSize() != null ? getSyncPageSize().hashCode() : 0);
         result = 31 * result + (getSyncIntervalInMinutes() != null ? getSyncIntervalInMinutes().hashCode() : 0);
+        result = 31 * result + (getSyncExpressions() != null ? getSyncExpressions().hashCode() : 0);
         return result;
     }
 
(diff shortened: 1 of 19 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -175,6 +176,29 @@ public Integer getSyncPageSize() {
         return this.syncPageSize;
     }
 
+    /**
+     * Returns a QueryPredicate that should be used to filter the data that is synced.
+     * @param modelClass Class of the {@link Model}.
+     * @return a QueryPredicate that should be used to filter the data that is synced.
```
jamesonwilliams, line 182: Perhaps:

```suggestion
     * Returns a QueryPredicate that should be used to filter data received from AppSync,
     * either during a base sync or over the real-time subscription.
     * @param modelClass Class of the {@link Model}.
     * @return a QueryPredicate that should be used to filter the data received from AppSync
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -175,6 +176,29 @@ public Integer getSyncPageSize() {
...
+        }
+        return QueryPredicates.all();
+    }
+
+    /**
+     * Returns the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
+     * @return the Map of all {@link DataStoreSyncExpression}s used for filtering the data that is synced.
+     */
+    @NonNull
+    public Map<Class<? extends Model>, DataStoreSyncExpression> getSyncExpressions() {
```
jamesonwilliams, line 198: If you have the one above, do you need this `Map`-returning one, as well?

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -305,6 +339,23 @@ public Builder syncPageSize(@IntRange(from = 0) Integer syncPageSize) {
             return Builder.this;
         }
 
+        /**
+         * Sets a sync expression for a particular model to filter which data is synced locally.  The expression
+         * is evaluated each time DataStore is started.  The QueryPredicate is applied on both sync and subscriptions.
```
jamesonwilliams, line 344: From the PR description:
> Runtime Modification: Clients can force a re-evaluation of their DataStoreSyncExpression at runtime by calling DataStore.stop(), DataStore.start(), if they wish to modify the filter to obtain a different set of data.

However, if I need to provide the sync expressions before calling `Amplify.configure(...)`, how could I change the expressions later? (Perhaps the sync expression itself has to be dynamic?)

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreSyncExpression.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore;
```
jamesonwilliams, line 16: Arguably, it could live under `.syncengine`? It's sort of both a sync-thing, and a top-level-user-provided-thing.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -97,7 +98,11 @@ private AppSyncRequestFactory() {}
             if (limit != null) {
                 builder.variable("limit", "Int", limit);
             }
-
+            if (!QueryPredicates.all().equals(predicate)) {
+                String modelName = ModelSchema.fromModelClass(modelClass).getName();
```
jamesonwilliams, line 102: If you're able to get it from the `ModelSchemaRegistry`, instead, that would prevent another reflection over the model class. Alternately, `modelClass.getSimpleName()` probably does the trick, too.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -66,23 +68,20 @@
...
      */
-    SubscriptionProcessor(
-            @NonNull AppSync appSync,
-            @NonNull ModelProvider modelProvider,
-            @NonNull Merger merger) {
-        this.appSync = Objects.requireNonNull(appSync);
-        this.modelProvider = Objects.requireNonNull(modelProvider);
-        this.merger = Objects.requireNonNull(merger);
+    private SubscriptionProcessor(Builder builder) {
+        this.appSync = Objects.requireNonNull(builder.appSync);
```
jamesonwilliams, line 81: As sort of a philosophical point, I'd argue the non-null checks should be in the `build()`, not in the `private` constructor. It is to say "We won't even attempt construction of the object, if user-provided options `foo`, `bar`, are null."

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -299,4 +310,75 @@ static SubscriptionMethod subscriptionMethodFor(
...
+    interface DataStoreConfigurationProviderStep {
+        @NonNull
+        BuildStep dataStoreConfigurationProvider(DataStoreConfigurationProvider dataStoreConfiguration);
+    }
+
+    interface BuildStep {
+        @NonNull
+        SubscriptionProcessor build();
+    }
+
```
jamesonwilliams, line 383:
```suggestion
```

aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
```diff
@@ -96,11 +100,13 @@ public void testDefaultOverriddenFromConfigurationAndObject() throws JSONExcepti
         DummyConflictHandler dummyConflictHandler = new DummyConflictHandler();
         DataStoreErrorHandler errorHandler = DefaultDataStoreErrorHandler.instance();
 
+        DataStoreSyncExpression syncExpression = () -> BlogOwner.ID.beginsWith("a");
```
jamesonwilliams, line 103: The `RandomString.string()` would have the same behavior here, but it reads like "oh, the value of the ID isn't important for the test, so long as it's there."

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessorTest.java
```diff
@@ -84,7 +86,13 @@ public void setup() throws AmplifyException {
         this.models = sortedModels(modelProvider);
         this.appSync = mock(AppSync.class);
         this.merger = mock(Merger.class);
-        this.subscriptionProcessor = new SubscriptionProcessor(appSync, modelProvider, merger);
+        this.subscriptionProcessor = SubscriptionProcessor.builder().appSync(appSync)
+                .modelProvider(modelProvider)
```
jamesonwilliams, line 90:
```suggestion
        this.subscriptionProcessor = SubscriptionProcessor.builder()
                .appSync(appSync)
                .modelProvider(modelProvider)
```

jamesonwilliams (approved):
Looks good! Left some comments, of no particular importance.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
richardmcclellan: I considered the same question.  The main reason I need this though is for obtaining the `Map` from a user-provided `DataStoreConfiguration`, while merging with the default configuration.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
richardmcclellan: Here's how I'm doing it in my sample app:

```
AmplifyFactory.setBlogQueryPredicate(Blog.NAME.contains("a"));
Amplify.DataStore.stop(...);
Amplify.DataStore.start(...);

public class AmplifyFactory {
    private static QueryPredicate blogQueryPredicate = QueryPredicates.all();

    public static void setBlogQueryPredicate(QueryPredicate queryPredicate) {
        blogQueryPredicate = queryPredicate;
    }

    public static void configure(Context context) {
        try {
            Amplify.addPlugin(new AWSDataStorePlugin(DataStoreConfiguration.builder()
                    .syncExpression(Blog.class, () -> blogQueryPredicate)
                    .build()));
            Amplify.configure(context);
        } catch (AmplifyException exception){
            LOG.error(exception.getMessage(), exception);
        }
    }
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
jamesonwilliams: Ah yes, that makes sense. 👍

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
jamesonwilliams: Seems ~reasonable. Or someone really wanted, they could make like a:

```java
final class MutableSyncExpression implements DataStoreSyncExpression {
    private QueryPredicate queryPredicate;

    MutableSyncExpression() {
        this.queryPredicate = QueryPredicates.all();
    }

    @Override public QueryPredicate resolvePredicate() {
        return queryPredicate;
    }

    void updatePredicate(QueryPredicate queryPredicate) {
        this.queryPredicate = queryPredicate;
    }
    // equals/hashCode/toString ...
}
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
richardmcclellan: That philosophy makes sense!    I'll remove these here in the constructor.  I actually don't think I need to add them to `build()`, because the step methods call `Objects.requireNonNull(...)` already, and the step builder makes all arguments required.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreSyncExpression.java
richardmcclellan: Hmm, I think you could make an argument for either one (`.datastore`, or `.datastore.syncengine`).  I think it makes a little more sense in `.datastore` though, so that it's in the same place as `DatastoreConflictHandler`, `DataStoreConfiguration`, etc.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
richardmcclellan: :+1: I'll go with `modelClass.getSimpleName()`

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -44,21 +50,16 @@
...
-            Integer syncMaxRecords,
-            Integer syncPageSize) {
-        this.errorHandler = errorHandler;
-        this.conflictHandler = conflictHandler;
-        this.syncMaxRecords = syncMaxRecords;
-        this.syncPageSize = syncPageSize;
-        if (syncIntervalInMinutes != null) {
-            this.syncIntervalInMinutes = syncIntervalInMinutes;
-        }
+    private final Map<Class<? extends Model>, DataStoreSyncExpression> syncExpressions;
```
jamesonwilliams, line 53: Can you make this a map from model name to sync expression, `Map<String, DataStoreSyncExpression>`? This would be more likely to work for Flutter, if/when that gets implemented in `amplify-flutter`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -206,6 +213,11 @@ private boolean isUnauthorizedException(DataStoreException exception) {
         .subscribeOn(Schedulers.io())
         .observeOn(Schedulers.io())
         .map(SubscriptionProcessor::unwrapResponse)
+        .filter(modelWithMetadata -> {
+            Class<? extends Model> clazz = modelSchema.getModelClass();
```
jamesonwilliams, line 217: If you do the `String modelName` instead of the `Class<? extends Model> modelClass`, could be just `modelSchema.getName()`

jamesonwilliams (approved):
Clean stuff. Nice.

jamesonwilliams (approved):

richardmcclellan:
I just updated this PR to ensure the `DataStoreSyncExpression` objects are only evaluated once, on `DataStore.start()`.    This required creating a `QueryPredicateProvider`, which essentially caches the `QueryPredicate` objects, and then is used by the `SyncProcessor` and `SubscriptionProcessor` when needed.   @jamesonwilliams or anyone else, please review this one last change!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -69,20 +69,16 @@
...
-            DataStoreConfigurationProvider dataStoreConfigurationProvider) {
-        this.modelProvider = Objects.requireNonNull(modelProvider);
-        this.modelSchemaRegistry = Objects.requireNonNull(modelSchemaRegistry);
-        this.syncTimeRegistry = Objects.requireNonNull(syncTimeRegistry);
-        this.appSync = Objects.requireNonNull(appSync);
-        this.merger = Objects.requireNonNull(merger);
-        this.dataStoreConfigurationProvider = dataStoreConfigurationProvider;
+    private final QueryPredicateProvider queryPredicateProvider;
+
+    private SyncProcessor(Builder builder) {
```
TrekSoft, line 74: Not that this matters or is worth changing but is there a reason you switched from passing in params to passing in the Builder instance? We do the opposite elsewhere in the code.

TrekSoft (approved):
Cat26u3

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: I've seen both patterns in our code base.  I like the single `Builder` parameter approach a little better, because then we don't end up with a constructor that takes a ton of parameters, and one of the main reasons for using the Builder pattern is to avoid constructors with lots of parameters :).   I'm open to other thoughts on why the other pattern might be better though.

jamesonwilliams: My argument for all-args constructor is that the constructor should not know any detail about the builder, or how it is being constructed. However, if the constructor is going to be `private` anyway, then it really doesn't matter. @richardmcclellan's point about not having to maintain the constructor argument list makes sense to me.

PR #961, opened 2020-11-14
fix: properly support flutter online scenarios

jamesonwilliams:
Previous work to support online-sync mode from Flutter had been vetted
with the `HybridCloudSyncInstrumentationTest`. This test was using
`ModelSchema.fromClass()`, which would ___still___ store a `Class<T>` in the
`modelClass`. As a result, the test was giving false optimism as to the
functionality in the code base.

A change is made to pull `ModelSchema` from JSON files, in the `./assets`
directory, instead. Driven by this updated tests, a number of additional
classes are updated to become aware of `ModelSchema` and `SerializedModel`.


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -224,7 +224,19 @@ void startDrainingMutationBuffer(Action onPipelineBroken) {
...
                     LOG.info("Starting processing subscription data buffer.")
                 )
-                .flatMapCompletable(mutation -> merger.merge(mutation.modelWithMetadata()))
+                .flatMapCompletable(mutation -> {
+                    ModelWithMetadata<? extends Model> original = mutation.modelWithMetadata();
+                    if (original.getModel() instanceof SerializedModel) {
+                        SerializedModel originalModel = (SerializedModel) original.getModel();
+                        SerializedModel newModel = SerializedModel.builder()
+                            .serializedData(originalModel.getSerializedData())
+                            .modelSchema(mutation.modelSchema())
```
jamesonwilliams, line 233: The response from AppSync doesn't contain any information about what _type_ it is. Previously, we used to know this because of the `<T>`. Now, I have to hack the `modelSchema` back in at this higher level. The `SubscriptionProcessor` knows which subscription the data came back on, so it can append the schema, here.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -140,9 +141,20 @@ private Completable createHydrationTask(ModelSchema schema) {
...
                 return syncModel(schema, lastSyncTime)
                     // For each ModelWithMetadata, merge it into the local store.
-                    .flatMapCompletable(modelWithMetadata ->
-                        merger.merge(modelWithMetadata, metricsAccumulator::increment)
-                    )
+                    .flatMapCompletable(original -> {
+                        ModelWithMetadata<? extends Model> updated;
+                        if (original.getModel() instanceof SerializedModel) {
+                            SerializedModel originalModel = (SerializedModel) original.getModel();
+                            SerializedModel newModel = SerializedModel.builder()
```
jamesonwilliams, line 148: Same basic situation as in the `SubscriptionProcessor`.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/HybridCloudSyncInstrumentationTest.java
```diff
@@ -116,11 +116,12 @@ public void setup() throws AmplifyException {
         api = SynchronousApi.delegatingTo(apiCategory);
         appSync = SynchronousAppSync.using(AppSyncClient.via(apiCategory));
 
+        schemaProvider = SchemaLoader.loadFromAssetsDirectory("schemas/commentsblog");
```
jamesonwilliams, line 119: The key change in this PR is that the schema are being read from JSON. The JSON files no longer say `com.path.to.some.generated.java.model.JavaPojo`. (Previously, the test had passed, with those queues.)

Now, the stated type for a model is a serialized model.

testutils/src/main/java/com/amplifyframework/testutils/Assets.java
```diff
@@ -79,5 +81,30 @@ public static Bitmap readAsBitmap(final String name) throws RuntimeException {
...
     }
-}
 
+    /**
+     * Lists the files in the assets directory.
+     * @param path Directory path under assets/.
+     * @return A list of files found
+     */
+    @NonNull
+    public static List<String> list(String path) {
```
jamesonwilliams, line 91: This utility is used by the `SchemaLoader`. The schema loader reads the schema JSON from disk with this `list(...)` utility. It will then build `ModelSchema` instances, from each.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
rjuliano: Just a clarification...are the terms new and original referring to the type of model you're working with (i.e. SerializedModel vs POJO) or is it referring to the data (i.e. original = existing data and new = updated data)?

jamesonwilliams: `originalModel` means "before I patch this `SerializedModel` instance's `modelSchema` field," and `newModel` means "the `SerializedModel` we got back from the sync command, but with a patched-up `modelSchema` field."

rjuliano (approved):
Looks good 👍  Both test suites are passing so I'd say it's a 🚢 IT from me.

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelWithMetadataAdapter.java
```diff
@@ -60,7 +61,15 @@ public static void register(@NonNull GsonBuilder builder) {
             throw new JsonParseException("Expected a parameterized type during ModelWithMetadata deserialization.");
         }
 
-        Model model = context.deserialize(json, modelClassType);
+        final Model model;
```
richardmcclellan, line 64: Hmm, why is this change necessary?  Shouldn't the `SerializedModelAdapter` already handle this?

If this is needed here, then I think we would probably also need a similar change in any other adapter that deserializes a Class that could contain a Model (e.g. `PaginatedResult`, `Iterable`, `GraphQLResponse`)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
richardmcclellan: Not a blocker for this PR, but it would be nice to see a `newBuilder()` method on the `SerializedModel` class, so you could simplify this to: 

```
SerializedModel newModel = originalModel.newBuilder()
    .modelSchema(mutation.modelSchema())
    .build();
```

richardmcclellan (commented):
I don't really like the approach of patching up the `ModelSchema` in the `SyncProcessor`, `SubscriptionProcessor`.   Just seems like we are scattering similar logic around.  Wouldn't we also need to do something similar in the `MutationProcessor`?  Also, when we launch the Flutter API category, we may need some other solution to populate `ModelSchema` on deserialization there.

Could we figure out a way to do this in the `SerializedModelAdapter` instead?   We could add a constructor to `SerializedModelAdapter`, which takes in the `AppSyncGraphQLRequest` used to make the request.   Then, register the adapter in `GsonGraphQLResponseFactory` (where we have access to the request).  Then, just call `getSchema()` to get the schema, and populate it in the `SerializedModel`.  

Making it work for requests with nested selection sets would require some additional work, but we probably don't need to support that just yet.

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelWithMetadataAdapter.java
jamesonwilliams: Alright, I studied this change a bit more. I think, for now at least, it probably makes sense to leave it as is.

The `SerializedModelAdapter` currently stores `modelSchema` and `serializedData`,into a JSON structure like this:
```json
{
    "modelSchema": {
        ...
    },
    "serializedData": {
    }
}
```

However, the format of the `ModelWithMetadata` is like:
```json
{
    "flatDataKey1": "dataVal",
    "flatDataKey2": "dataVal",
    "_metadataStuff1": "metadata",
    "_metadataStuff2": "metadata"
}
```

The GSON deserialization will cascade _through_ the adapters you mention, and will try to deserialize a `GraphQLResponse<Iterable<ModelWithMetadata<SerializedModel>>>`. It knows how to do all of that just fine, except when it gets to the inner-most adapter. At that point, it looks for a field called `serializedData` in that second JSON structure (which doesn't exist.)

So I think TLDR: I _am_ actually doing some custom deserialization only for `ModelWithMetadata`.

And, it isn't needed for other adapters, since DataStore will _always_ envelope in the `ModelWithMetadata`.

--------

I can think of a few others ways we could do this.

1. At a higher level, pass the `T` as `Map<String, Object>`, s.t. we try to deserialize `GraphQLResponse<Iterable<ModelWithMetadata<Map<String, Object>>>>`. Then, we could rectify the map data back into a `SerializedModel.builder().serializedData(theResponseContent).build()` at a higher level in the stack.

2. Rip `<T>` out of almost all of the system, in future work. This is what I _actually_ want to do. With this, the SQL layer and the AppSync sync layer will just know about `SerializedModel` only, but not any model types. (I also propose that we rename `SerializedModel` to maybe just `Blob`, but that's a separate matter.) Only the highest layer of the system (just under the DataStore API) would know about Java-language data-modeling. At that high-level, it would quickly convert to schema/blob object pairs. Sync engine & storage adapter would operate wholly on those.

richardmcclellan: Got it, thanks for investigating it :). I'm good to leave it as is for now.

> Rip <T> out of almost all of the system, in future work.

I agree this is the right direction to move towards.

> I also propose that we rename SerializedModel to maybe just Blob, but that's a separate matter.

:+1:

I wonder if we could also eliminate `ModelWithMetadata` entirely, and just store the metadata as a part of the `Blob`.  It would be closer to how the data is modeled on the AppSync side anyway.   Seems like this could simplify a lot too.

richardmcclellan (approved):
As discussed offline, I think we can merge this now, and come back later and try to update the `SerializedModelAdapter` with my suggestion later.

PR #968, opened 2020-11-17
 fix(aws-predictions): expose label instances' bounding boxes

raphkim:
*Issue #, if available:*

*Description of changes:*
Exposed bounding boxes for objects of identified label. This property is already available in [iOS](https://github.com/aws-amplify/amplify-ios/blob/eb5b7ff98ae9ac090dc8c71e374b8f3bcb0ee472/Amplify/Categories/Predictions/Result/IdentifyLabelsResult.swift#L23) and [JS](https://github.com/aws-amplify/amplify-js/blob/e85640a4e18fa1d55411038fee58919ad73121fa/packages/predictions/src/types/Predictions.ts#L236) libraries, but was not included in Android library as an undersight.

### DIFF
--- a/core/src/main/java/com/amplifyframework/predictions/models/Label.java
+++ b/core/src/main/java/com/amplifyframework/predictions/models/Label.java
@@ -15,6 +15,7 @@
 
 package com.amplifyframework.predictions.models;
 
+import android.graphics.RectF;
 import androidx.annotation.NonNull;
 
 import java.util.ArrayList;
@@ -34,10 +35,12 @@ public final class Label extends ImageFeature<String> {
     public static final String FEATURE_TYPE = Label.class.getSimpleName();
 
     private final List<String> parentLabels;
+    private final List<RectF> boxes;
 
     private Label(final Builder builder) {
         super(builder);
         this.parentLabels = builder.getParentLabels();
+        this.boxes = builder.getBoxes();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/amplifyframework/predictions/models/Label.java
```diff
@@ -34,10 +35,12 @@
     public static final String FEATURE_TYPE = Label.class.getSimpleName();
 
     private final List<String> parentLabels;
+    private final List<RectF> boxes;
```
jamesonwilliams, line 38: Should probably add this new field into the `hashCode()`, `toString()` and `equals()`, as well.

core/src/main/java/com/amplifyframework/predictions/models/Label.java
```diff
@@ -120,5 +146,10 @@ public Label build() {
         List<String> getParentLabels() {
             return Objects.requireNonNull(parentLabels);
         }
+
+        @NonNull
+        List<RectF> getBoxes() {
+            return Objects.requireNonNull(boxes);
```
jamesonwilliams, line 152:
```suggestion
            return Objects.requireNonNull(Immutable.of(boxes));
```

jamesonwilliams (commented):
Looks good, just a few suggestions.

rjuliano (approved):
I think after addressing the couple of comments Jameson left, this should be good to go. I'm approving it since I'll be out the rest of the week. Once you fix that stuff, I'd say you can 🚢  it!!!

core/src/main/java/com/amplifyframework/predictions/models/Label.java
raphkim: this is something that needs to be applied to every result model in predictions category... will revisit this in a separate PR.

core/src/main/java/com/amplifyframework/predictions/models/Label.java
raphkim: good call! will apply this to the other variable in this class too.

Issue #823, opened 2020-09-11
Item entered offline refuses to sync to backend


msiewert:
**Development Environment**

Android Studio 4.0.1
`minSdkVersion` 16
`targetSdkVersion` 30
`sourceCompatibility` `JavaVersion.VERSION_1_8`
`targetCompatibility` `JavaVersion.VERSION_1_8`

```gradle
implementation 'com.amplifyframework:core:1.3.1'
implementation 'com.amplifyframework:aws-datastore:1.3.1'
implementation 'com.amplifyframework:aws-api:1.3.1'
```

Amplify CLI 4.29.1

**Execution Environment**

Famoco FX100 - https://help.famoco.com/developers/devices/FX100/
Android 24

**Description**

I have a simple todo application based on the tutorials.

I can demonstrate a case where I save an item in the datastore while offline (WebSocket connection errors occur) and it refuses to sync back to the backend even after connectivity is re-established. 

The behavior is intermittent and does not always manifest itself, but basic steps are:

- put device in airplane mode
- add multiple todos
- reboot device
- add more todos
- turn off airplane mode

msiewert:
@jamesonwilliams

msiewert:
2020-09-14 08:03:34.224 5054-5054/? I/amplify_androi: Not late-enabling -Xcheck:jni (already on)
2020-09-14 08:03:34.267 5054-5054/? I/amplify_androi: Unquickening 12 vdex files!
2020-09-14 08:03:34.268 5054-5054/? W/amplify_androi: Unexpected CPU variant for X86 using defaults: x86
2020-09-14 08:03:34.314 5054-5054/com.example.amplify_android I/amplify_androi: The ClassLoaderContext is a special shared library.
2020-09-14 08:03:34.331 5054-5054/com.example.amplify_android D/NetworkSecurityConfig: No Network Security Config specified, using platform default
2020-09-14 08:03:34.332 5054-5054/com.example.amplify_android D/NetworkSecurityConfig: No Network Security Config specified, using platform default
2020-09-14 08:03:34.402 5054-5054/com.example.amplify_android I/amplify: Initialized Amplify
2020-09-14 08:03:34.464 5054-5086/com.example.amplify_android D/libEGL: loaded /vendor/lib/egl/libEGL_emulation.so
2020-09-14 08:03:34.465 5054-5086/com.example.amplify_android D/libEGL: loaded /vendor/lib/egl/libGLESv1_CM_emulation.so
[...]

PR #976, opened 2020-11-21
feat(aws-datastore): add mutation failed hub event

raphkim:
*Description of changes:*
Currently, there is no logic in place to handle a failed attempt at publishing a mutation. If the failed mutation was due to an unrecoverable cause such as `@auth` violation, then it will stay in the mutation outbox and block the queue. 

To unblock the outbox, the failing mutation must be removed from the queue, but there needs to be a mechanism to alert this action to the customer since customer data is being modified as a result. This PR adds:
1. logic to remove failing mutation
2. a hub event containing details of failed mutation and underlying error response

~Note: the hub event payload has not yet been finalized across other platforms. The PR will stay as a draft until it is.~

Final format: 
```
OutboxMutationFailedEvent:
{
    errorType: Enum // error type from the server 
    operation: Enum // type of mutation that failed (Create vs Update vs Delete)
    model: Model // current local model
    modelName: String // model name
}
```


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -287,15 +296,24 @@ void stopDrainingMutationOutbox() {
...
+        // mutation from the queue.
+        return Completable.fromAction(() -> announceMutationFailed(pendingMutation, errors))
+            .doOnError(error -> LOG.warn("Failed to create an outboxMutationFailedEvent. ", error))
+            .andThen(mutationOutbox.remove(pendingMutation.getMutationId()))
+            .andThen(Single.error(new DataStoreException(
+                    "Mutation failed. Failed mutation = " + pendingMutation + ". " +
+                            "AppSync response contained errors = " + errors,
+                    "Verify that your AppSync endpoint is able to store " +
+                            pendingMutation.getMutatedItem() + "."
+            )));
```
richardmcclellan, line 315: Hmm.  I think this error is going to bubble up all the way to `MutationProcessor::startDrainingMutationOutbox`, and stop all subsequent mutations from being processed.  This is an existing problem, but it would be ideal if we can address it in this PR.

I would suggest these changes:
1. Instead of emitting a `DataStoreException`, emit a `DataStoreException.GraphQLResponseException`.  That will allow us to identify it as an error that should be swallowed higher up in the stack.
2. In `MutationProcessor::processOutboxItem`, if `publishToNetwork` gets an error, swallow it, and complete.  You can do that with something like this: 

```
.onErrorComplete(error -> error instanceof DataStoreException.GraphQLResponseException)
```

3. After making the first two changes, you could move the logic for the removing the item from the queue and calling `announceMutationFailed` into `processOutboxItem`.  Since `DataStoreException.GraphQLResponseException` is able to hold a list of `GraphQLResponse.Error` objects, they will be available to build a `OutboxMutationFailedEvent`.   It would be a little cleaner than doing it down in `handleResponseErrors`, since it would be right next to the call to `announceMutationProcessed`.  You'd add something like this:

```
.doOnError(error -> {
        .doOnComplete(() -> announceMutationFailed(mutationOutboxItem, error))
}
```

Note that even with all these suggested changes, I think there is still a problem with the way non-AppSync errors are handled (e.g. network errors) (https://github.com/aws-amplify/amplify-android/issues/823).  We can address that separately though.

raphkim: Thanks for the detailed feedback! 
I will address the network errors in a separate PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationEvent.java
```diff
@@ -31,7 +32,7 @@
  * and {@link DataStoreChannelEventName#OUTBOX_MUTATION_PROCESSED} event.
  * @param <M> The class type of the model in the mutation outbox.
  */
-public final class OutboxMutationEvent<M extends Model> {
+public final class OutboxMutationEvent<M extends Model> implements HubEvent.Data<OutboxMutationEvent<M>> {
```
jamesonwilliams, line 35: Let's either make _all_ of the payloads do this, or none. Can we save this for a follow up PR where we switch over all payload structures at once?

jamesonwilliams (commented):
Looks good overall. How can we get some coverage over the new source?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
jamesonwilliams: While doing some unrelated testing, I noticed that we have this existing problem around [line 120](https://github.com/aws-amplify/amplify-android/blob/8ccb280c0418853731f46fe709bda857d7b7765e/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L120), too. If we fail to publish an item from the outbox (for example, if the app is offline), the exception tears down the observer, _and_ crashes the app. 😢

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationEvent.java
raphkim: this event is from an older PR and I probably shouldn't have made this change in this PR since it is unrelated. But yea, that was what I was going for (to make every payload implement `HubEvent.Data`).

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -111,8 +96,83 @@ public M getModel() {
...
+                return UNKNOWN;
+            }
+        }
+
+        // Look at the first error to extract error type.
+        private static MutationErrorType fromGraphQLErrors(List<GraphQLResponse.Error> errors) {
+            Iterator<GraphQLResponse.Error> iterator = errors.iterator();
+            if (iterator.hasNext()) {
+                GraphQLResponse.Error firstError = iterator.next();
+                AppSyncExtensions extensions = new AppSyncExtensions(firstError.getExtensions());
```
jamesonwilliams, line 172: We recently got bit by this. `getExtension()` is nullable, but `AppSyncExtensions` requires non null. See https://github.com/aws-amplify/amplify-android/pull/1000.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -111,8 +96,83 @@ public M getModel() {
...
+        /**
+         * Returns an enum value of matching error type value.
+         * @param value the type of error that caused mutation publication failure
+         * @return An enum value of matching error type. If there is no match,
+         *          then return {@link MutationErrorType#UNKNOWN}.
+         */
+        @NonNull
+        public static MutationErrorType fromErrorType(@Nullable String value) {
+            try {
+                return MutationErrorType.valueOf(value);
```
jamesonwilliams, line 161: Hm, does `valueOf` check the `name()` (an uppercase String)? How would it know to inspect the `value` property, which is Pascal-cased? 

[One reference](https://www.tutorialspoint.com/java/lang/enum_valueof.htm) states:

> The java.lang.Enum.valueOf() method returns the enum constant of the specified enum type with the specified name. The name must match exactly an identifier used to declare an enum constant in this type.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -111,8 +96,83 @@ public M getModel() {
         return model;
     }
 
+    /**
+     * Gets the name of the model.
+     * @return the model name
+     */
+    @NonNull
+    public String getModelName() {
+        return model.getClass().getSimpleName();
```
jamesonwilliams, line 105: This won't work when we are using `ModelSchema` on flutter. The new approach is to always ask `modelSchema.getName()`. I had to update this logic in https://github.com/aws-amplify/amplify-android/issues/1012. You can most likely use a similar technique to access the model name externally, and then pass it in when you are constructing the `OutboxMutationFailedEvent`.

TLDR, `getClass().getSimpleName()` isn't a thing for us anymore, we have to make them all say `modelSchema.getName()`.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -55,46 +55,31 @@ private OutboxMutationFailedEvent(
...
-            default:
-                throw new DataStoreException(
-                    "Invalid operation type.",
-                    AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION
-                );
-        }
-        return new OutboxMutationFailedEvent<>(errors, opType, pendingMutation.getMutatedItem());
+        MutationErrorType errorType = MutationErrorType.fromGraphQLErrors(errors);
+        String operation = pendingMutation.getMutationType().name();
+        MutationType opType = MutationType.valueOf(operation);
```
jamesonwilliams, line 68: Hm, I'm confused by this logic. You're converting an _error type_ (`UNAUTHORIZED`) to a _`MutationType`_ (`UPDATE`, etc.)?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -35,16 +35,16 @@
  */
 public final class OutboxMutationFailedEvent<M extends Model>
         implements HubEvent.Data<OutboxMutationFailedEvent<M>> {
-    private final List<GraphQLResponse.Error> errors;
+    private final MutationErrorType errorType;
     private final MutationType operation;
     private final M model;
```
jamesonwilliams, line 41: As below, you're probably going to want to make `modelName` an explicit property of the shape, as opposed to trying to pull it from the `model` in the accessor.

jamesonwilliams (commented):
Let's show the final shape to @iartemiev & @royjit as a last follow up from our design meeting.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
raphkim: nice catch! will fix.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
raphkim: It is merely converting it from `PendingMutation.Type` internal enum to a customer-exposed `MutationType` enum. However, I realized that this enum comes from `aws-api-appsync` module, which is a problem on its own since we want to abstract away GraphQL from here. I will create another enum type within datastore module.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
raphkim: caught the whole thing in `try-catch` for convenience.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationEvent.java
raphkim: Addressed in [this commit](https://github.com/aws-amplify/amplify-android/pull/976/commits/01ee3ee914060394f95ac602fa3b84a3c4284fc3) (Needs review).

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
raphkim: Addressed in [this commit](https://github.com/aws-amplify/amplify-android/pull/976/commits/01ee3ee914060394f95ac602fa3b84a3c4284fc3) (Needs review).

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -0,0 +1,183 @@
...
+            return UNKNOWN;
+        }
+
+        // Look at the first error to extract error type.
+        private static MutationErrorType fromGraphQlErrors(List<GraphQLResponse.Error> errors) {
+            try {
+                GraphQLResponse.Error firstError = errors.get(0);
+                AppSyncExtensions extensions = new AppSyncExtensions(firstError.getExtensions());
+                return enumerate(extensions.getErrorType().getValue());
+            } catch (RuntimeException parsingError) {
```
richardmcclellan, line 178: It's generally best practice to catch the lowest level of exception that you need, not all `RuntimeException`.   Otherwise, it's unclear to future developers as to what _could_ go wrong.    Can you just do a null check (`Empty.check(firstError.getExtensions())`), and return `UNKNOWN` for that case?

richardmcclellan (commented):
Left one minor comment, but otherwise looks good!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
raphkim: So this code snippet can throw error in following scenarios:
* `errors` is `null`
* `errors` is `empty`
* first item in `errors` is `null`
* error type inside first error object is `null`

Rather than excessively performing a null check, I decided to do a catch-all. However, I did fix the `RuntimeException` to something more specific (`NullPointerException | IndexOutOfBoundsException`) for clarification.

richardmcclellan (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -0,0 +1,183 @@
...
+        return modelName;
+    }
+
+    @Override
+    public HubEvent<OutboxMutationFailedEvent<M>> toHubEvent() {
+        return HubEvent.create(DataStoreChannelEventName.OUTBOX_MUTATION_FAILED, this);
+    }
+
+    @Override
+    public String toString() {
```
jamesonwilliams, line 118: `equals` & `hashCode` please!

raphkim: oopsie. added!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationFailedEvent.java
```diff
@@ -51,6 +51,31 @@ private OutboxMutationFailedEvent(
...
+        OutboxMutationFailedEvent<?> that = (OutboxMutationFailedEvent<?>) obj;
+        return getErrorType() == that.getErrorType() &&
+            getOperation() == that.getOperation() &&
+            getModelName().equals(that.getModelName()) &&
+            getModel().equals(that.getModel());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(
```
jamesonwilliams, line 71: [`Objects.hash(..)`](https://developer.android.com/reference/java/util/Objects#hash(java.lang.Object...)) is only supported from API level 19. This should say [`ObjectsCompat.hash(...)`](https://developer.android.com/reference/androidx/core/util/ObjectsCompat#hash(java.lang.Object...)).

I thought maybe it could be desugared in the Android 4.x tools, but that doesn't look to be the case, either.

Issue #896, opened 2020-10-12
Support for Scoped Storage


jamesonwilliams:
When upgrading the Android Gradle Plugin to 4.1.0, the Android Lint now complains:
> /Users/jhwill/amplify-android/aws-storage-s3/src/main/AndroidManifest.xml:19: Error: WRITE_EXTERNAL_STORAGE no longer provides write access when targeting Android 11+, even when using requestLegacyExternalStorage [ScopedStorage]

To continue upgrading Amplify (and to properly support the latest versions of Android), we will need to add support for [Scoped Storage](https://developer.android.com/training/data-storage#scoped-storage):
> To give users more control over their files and to limit file clutter, apps that target Android 10 (API level 29) and higher are given scoped access into external storage, or scoped storage, by default. Such apps have access only to the app-specific directory on external storage, as well as specific types of media that the app has created.

sdhuka:
@richardmcclellan Could we close this issue?

richardmcclellan:
@sdhuka I've opened https://github.com/aws-amplify/amplify-android/pull/1377 which should resolve this issue.

PR #981, opened 2020-11-22
chore: update to gradle plugin 4.1.1

richardmcclellan:
In 4.1.0, the [version properties are removed from the BuildConfig class in library projects](https://developer.android.com/studio/releases/gradle-plugin#version_properties_removed_from_buildconfig_class_in_library_projects
).   Our library currently uses `BuildConfig.VERSION_NAME` for building user agents, and for displaying the current library version in the developer menu.   In order to continue using this, I created custom `buildConfig` properties of the same name.

This also removes the `WRITE_EXTERNAL_STORAGE` permission from the S3 plugin's `AndroidManifest.xml`, since [it was deprecated in Android 11](https://developer.android.com/about/versions/11/privacy/storage#permissions-target-11).  If TransferUtility does rely on this permission, it's already broken for apps built with API 30 and above, irrespective of this PR.    From what I can tell though, TransferUtility should still work.  It creates a SQLite DB using [SQLiteOpenHelper](https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper), which creates the DB in the app sandbox, so permissions should not be required.

This PR resolves https://github.com/aws-amplify/amplify-android/issues/896

### DIFF
--- a/aws-storage-s3/src/main/AndroidManifest.xml
+++ b/aws-storage-s3/src/main/AndroidManifest.xml
@@ -16,8 +16,6 @@
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.amplifyframework.storage.s3" >
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-
     <application>
         <service android:name="com.amazonaws.mobileconnectors.s3.transferutility.TransferService" android:enabled="true" />
     </application>
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.1'
+        classpath 'com.android.tools.build:gradle:4.1.1'
     }
 }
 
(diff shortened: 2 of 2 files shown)


jamesonwilliams (approved):

jamesonwilliams (commented):
Is everything cool with the integ tests?

richardmcclellan:
> Is everything cool with the integ tests?

No, there is an issue to be resolved :/.  I wasn't able to figure it out yesterday, but will take another look.

rjuliano:
> > Is everything cool with the integ tests?
> 
> No, there is an issue to be resolved:
> 
> ```
> Could not determine the dependencies of task ':core:l8DexDesugarLibDebugAndroidTest'.
> Cannot query the value of this property because it has no value available.
> ```
> 
> I wasn't able to figure it out yesterday, but will take another look.



> > Is everything cool with the integ tests?
> 
> No, there is an issue to be resolved:
> 
> ```
> Could not determine the dependencies of task ':core:l8DexDesugarLibDebugAndroidTest'.
> Cannot query the value of this property because it has no value available.
> ```
> 
> I wasn't able to figure it out yesterday, but will take another look.

I'm stumped on this one. Found [this issue](https://issuetracker.google.com/issues/171364505) that could possibly related. However, the workaround suggested is to remove multidex, but unfortunately we can't do that. We may have to wait to update to 4.2 since [it doesn't sound like they're cherry picking the fix to 4.1](https://issuetracker.google.com/issues/171364505#comment8)

richardmcclellan:
Skipping 4.1 and waiting for 4.2 seems like a reasonable solution to me!

richardmcclellan:
Closing this since this not a critical issue, and we haven't been able to get the integration tests passing with it.    Once the gradle plugin 4.2 is available, we can update to that.

PR #988, opened 2020-11-24
fix: associated Flutter models can sync to cloud

jamesonwilliams:
Version 1.6.3 of Amplify fixed an issue where associated/connected
models failed to sync down from AppSync, when being used with
Flutter-style `ModelSchema`.

Unfortunately, 1.6.3 stopped short of supporting associated models when
syncing *up* to AppSync, from the client.

This current change expands the `HybridCloudSyncInstrumentationTest`, to
demonstrate successful sync up and down from AppSync.

To support this, request-forming logic is extracted from the
`ModelSchema`, and split across the `AppSyncRequestFactory` (for DataStore)
and the `AppSyncGraphQLRequestFactory` (for API category). The DataStore
flavor of the factory is expanded to properly build requests when
dealing with associations in Flutter-style `SerializedModel`s.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
@@ -24,7 +24,11 @@ import com.amplifyframework.api.graphql.MutationType;
 import com.amplifyframework.api.graphql.PaginatedResult;
 import com.amplifyframework.api.graphql.QueryType;
 import com.amplifyframework.api.graphql.SubscriptionType;
+import com.amplifyframework.core.model.AuthRule;
+import com.amplifyframework.core.model.AuthStrategy;
 import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.ModelAssociation;
+import com.amplifyframework.core.model.ModelField;
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.predicate.BeginsWithQueryOperator;
 import com.amplifyframework.core.model.query.predicate.BetweenQueryOperator;
@@ -44,6 +48,7 @@ import com.amplifyframework.datastore.DataStoreException;
 import com.amplifyframework.util.Casing;
 import com.amplifyframework.util.TypeMaker;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -315,4 +302,81 @@ private static Object appSyncOpValue(QueryOperator<?> qOp) throws DataStoreExcep
                     amplifyException, "Validate your model file.");
         }
     }
+
+    private static Map<String, Object> getMapOfFieldNameAndValues(
```
richardmcclellan, line 306: What's the reasoning for moving this out of `ModelSchema`?    And why is the implementation of this method not identical to the similar method created in `AppSyncGraphQLRequestFactory`?

Ultimately, I think `AppSyncRequestFactory` (used for DataStore) and `AppSyncGraphQLRequestFactory` (used for API) should be combined, or at least have a common superclass maybe.

richardmcclellan:
Looks like integration tests are failing:
```
Execution failed for task ':aws-datastore:runTestsInDeviceFarm'.
Execution failed for task ':aws-predictions:runTestsInDeviceFarm'.
```

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
rjuliano: Maybe have some of the common functionality could go in the `aws-api-appsync` package since it's used by both api and datastore?

jamesonwilliams: These are good points, thanks guys. The original point of the `aws-api-appync` package was to contain AppSync-specific code. I agree that we should probably have just one of these, and it should live in that module. Will make this change.

richardmcclellan: That doesn't _have_ to be in this PR if you wanted to split it out.  But, go for it if you want!

I was more interested in the reasoning for moving this method out of `ModelSchema`?

jamesonwilliams: The two request factories have little differences laced throughout - the response type, whether to include _version, the request options object, etc. … :cry:

Long term, I would like to see all of the AppSync-specific stuff go into `aws-api-appsync`; I still think that's a good direction.

I had moved the request formation logic out of the `ModelSchema` since `ModelSchema` lives in core. I didn’t want core doesn’t know about `SerializedModel`s. This `getMapOfFieldNameAndValues(...)` method really _is_ about forming an outbound request, I think, not something having to do with the schema in-and-of-itself.

rjuliano (approved):
Looks good. The one failure on the integration tests is related to predictions, so not related to this PR. Ship it 🚢

richardmcclellan (approved):

Issue #994, opened 2020-11-25
[Flutter support] Remote delete mutations are not getting saved locally


Amplifiyer:
In a flutter app using `SerializedModel`. when a delete mutation is received by the app running in android, it is not currently saved/merged in the local store. No error/exceptions are observed when this happens. 

To reproduce this:

* Run a flutter app, or an android app with serialized models.
* Create a delete mutation outside of the android app.
* Observe that the delete mutation is not persisted in local storage and consequently no event is received through `observe()`

It looks like the issue could be with how library queries (See line 178 below) for an incoming delete mutation with the class SerializedModel rather with the actual String modelName. this results in onNotPresent to be called and so the library [doesn’t delete ](https://github.com/aws-amplify/amplify-android/blob/9db52862c3c7174877ce7247624999e67cb32a47/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java#L135-L146) anything from the localStorage

https://github.com/aws-amplify/amplify-android/blob/9db52862c3c7174877ce7247624999e67cb32a47/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java#L176-L185

jamesonwilliams:
Thanks for doing the research, here, @Amplifiyer. It made it fast for me to resolve the issue.

https://github.com/aws-amplify/amplify-android/pull/1004

PR #1004, opened 2020-11-27
fix: honor deletions from AppSync when used by Flutter

jamesonwilliams:
When the DataStore is used from Flutter, Java-language model classes
don't exist.

The Merger component is responsible for applying remote mutations onto
the local store. It contained an ifPresent() check, which checked for
the presence of a model by its Java class. This method would fail to
work as expected when the SerializedModel envelope type was used.

ifPresent() has been updated to query the local store by model name,
instead of using the Java class.

Resolves: https://github.com/aws-amplify/amplify-android/issues/994


rjuliano (approved):
Looks good. 🚢

PR #1006, opened 2020-11-27
fix: fix field label in toString() of ModelField

jamesonwilliams:
When the formerly-existing `type` member was renamed to
`javaClassForValue`, the label in `toString()` was not updated.

Here's a fix for that.


rjuliano (approved):
Let it sail!!! ⛵

jamesonwilliams:
> Let it sail!!! ⛵

(Hoists the canvas, hull begins to move through water ....)

Issue #1011, opened 2020-11-30
[Flutter Support] Failing to sync down a delete mutation for connected model


Amplifiyer:
In a typical `Blog` & `Post` schema, when I delete a post instance (belonging to a Blog) from let's say a iOS device, I get an error while (android DataStore is) querying for that item in the local datastore before deleting from it.

```
android.database.sqlite.SQLiteException: ambiguous column name: id (code 1 SQLITE_ERROR): , while compiling: SELECT `Post`.`id` AS `Post_id`, `Post`.`created` AS `Post_created`, `Post`.`rating` AS `Post_rating`, `Post`.`title` AS `Post_title`, `Post`.`blogID` AS `Post_blogID`, `Blog`.`id` AS `Blog_id`, `Blog`.`name` AS `Blog_name` FROM `Post` LEFT JOIN `Blog` ON `Post`.`blogID`=`Blog`.`id` WHERE id = ? LIMIT ? OFFSET ?;
at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
        at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:986)
at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:593)
        at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:590)
at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:61)
        at android.database.sqlite.SQLiteQuery.<init>(SQLiteQuery.java:37)
at android.database.sqlite.SQLiteDirectCursorDriver.query(SQLiteDirectCursorDriver.java:46)
        at android.database.sqlite.SQLiteDatabase.rawQueryWithFactory(SQLiteDatabase.java:1443)
at android.database.sqlite.SQLiteDatabase.rawQuery(SQLiteDatabase.java:1382)
        at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.getQueryAllCursor(SQLiteStorageAdapter.java:825)
at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.lambda$query$5$SQLiteStorageAdapter(SQLiteStorageAdapter.java:406)
at com.amplifyframework.datastore.storage.sqlite.-$$Lambda$SQLiteStorageAdapter$YrR7B1Rj3JmC_omOKlzMTn0U5hM.run(Unknown Source:10)
[...]

PR #1014, opened 2020-12-01
fix(aws-datastore): idempotent deletions from merger

jamesonwilliams:
When deletions are merged to the local store, the storage adapter might
throw an error, if there's nothing left to delete.

The existing code tries to work around this by first checking if there
is a matching instance before attempting to delete it.

Instead, we will swallow deletion failures and log a verbose message.

Resolves: https://github.com/aws-amplify/amplify-android/issues/1011

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
@@ -17,17 +17,14 @@ package com.amplifyframework.datastore.syncengine;
 
 import androidx.annotation.NonNull;
 
-import com.amplifyframework.core.Action;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.NoOpConsumer;
 import com.amplifyframework.core.model.Model;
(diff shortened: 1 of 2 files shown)


Amplifiyer (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -128,63 +126,35 @@
...
-        }));
+    private <T extends Model> Completable delete(T model, Consumer<StorageItemChange.Type> changeTypeConsumer) {
+        return Completable.create(emitter ->
+            localStorageAdapter.delete(model, StorageItemChange.Initiator.SYNC_ENGINE, QueryPredicates.all(),
+                storageItemChange -> {
+                    changeTypeConsumer.accept(storageItemChange.type());
+                    emitter.onComplete();
+                },
+                failure -> {
+                    LOG.verbose(
```
jamesonwilliams, line 137: This PR has two main changes:
1. Remove the `ifPresent(...)` check;
2. Only log this `failure`, instead of calling `emitter.onError(...)`.

richardmcclellan (approved):

Issue #1012, opened 2020-12-01
DataStore Event "outboxMutationProcessed" is malformed


haverchuck:
The[ documentation states](https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/android#outboxmutationprocessed) that the outboxMutationProcessed DataStore Hub event payload contains the modelName and an element, which in turn contains the model and syncMetadata.

However, the actual payload contains a key for 'model' and a key for 'element'.  The element value contains some model data, but not the model schema - therefore it is not possible to infer the true model name when it is an instance of 'SerializedModel'.

Expected behavior:  as defined [in documentation](https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/android#outboxmutationprocessed).

Actual behavior:
<img width="732" alt="Screen Shot 2020-11-30 at 4 04 35 PM" src="https://user-images.githubusercontent.com/10621276/100680129-c7926a00-3325-11eb-97f1-0682d3828bf6.png">

Bug summary: modelName is not present.

jamesonwilliams:
> 7:47 PM
Dustin Noyes hmm - i see the modelName field in the hub payload now... but it is null
> 7:53 PM
Jameson Williams Ugh 🤦‍♂️ damn

PR #1021, opened 2020-12-02
fix(aws-datastore): null modelName in OutboxMutationProcessed

jamesonwilliams:
When the `MutationProcessor` successfully processes an item from the
Mutation Outbox, it will publish an `OutboxMutationProcessed` event to the
Hub.

The AppSync response itself contains no information about *what type* of
model has just been published. That information is only available before
making the request.

As a consequence, we cannot inspect the model name from the AppSync
response, and instead, we should pull it out of the `MutationProcessor`'s
request context.

To validate this fix, the various instrumentation tests are updated to
use the `getModelName()` code path on the `OutboxMutationProcessed` event.

Resolves: https://github.com/aws-amplify/amplify-android/issues/1012

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
@@ -141,7 +141,10 @@ final class MutationProcessor {
                     // if there are outstanding mutations in the outbox.
                     mutationOutbox.remove(mutationOutboxItem.getMutationId())
                         .andThen(merger.merge(modelWithMetadata))
-                        .doOnComplete(() -> announceMutationProcessed(modelWithMetadata))
+                        .doOnComplete(() -> {
+                            String modelName = mutationOutboxItem.getModelSchema().getName();
+                            announceMutationProcessed(modelName, modelWithMetadata);
+                        })
                 )
             )
             .doOnComplete(() -> {
@@ -159,9 +162,9 @@ final class MutationProcessor {
      * @param modelWithMetadata A model that was successfully mutated and its sync metadata
      * @param <T> Type of model
      */
-    private <T extends Model> void announceMutationProcessed(ModelWithMetadata<T> modelWithMetadata) {
-        OutboxMutationEvent<T> mutationEvent =
-            OutboxMutationEvent.fromModelWithMetadata(modelWithMetadata);
+    private <T extends Model> void announceMutationProcessed(
+            String modelName, ModelWithMetadata<T> modelWithMetadata) {
+        OutboxMutationEvent<T> mutationEvent = OutboxMutationEvent.create(modelName, modelWithMetadata);
         HubEvent<OutboxMutationEvent<T>> hubEvent =
             HubEvent.create(DataStoreChannelEventName.OUTBOX_MUTATION_PROCESSED, mutationEvent);
         Amplify.Hub.publish(HubChannel.DATASTORE, hubEvent);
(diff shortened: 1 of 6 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -141,7 +141,10 @@ private Completable drainMutationOutbox() {
                     // if there are outstanding mutations in the outbox.
                     mutationOutbox.remove(mutationOutboxItem.getMutationId())
                         .andThen(merger.merge(modelWithMetadata))
-                        .doOnComplete(() -> announceMutationProcessed(modelWithMetadata))
+                        .doOnComplete(() -> {
+                            String modelName = mutationOutboxItem.getModelSchema().getName();
```
jamesonwilliams, line 145: This line is the actual fix. Everything else is just wiring or test changes.

richardmcclellan (commented):
I think this approach works, though instead of passing in a new `modelName` parameter to `OutboxMutationEvent.create`, why can't you get it from the `ModelSchema` on the `SerializedModel` like this?

```
SerializedModel serializedModel =  modelWithMetadata.getModel();
String modelName = serializedModel.getModelSchema().getName();
```

jamesonwilliams:
> I think this approach works, though instead of passing in a new `modelName` parameter to `OutboxMutationEvent.create`, why can't you get it from the `ModelSchema` on the `SerializedModel` like this?
> 
> ```
> SerializedModel serializedModel =  modelWithMetadata.getModel();
> String modelName = serializedModel.getModelSchema().getName();
> ```

I believe this is the same basic issue we had in the `SubscriptionProcessor` and `SyncProcessor`. The `ModelSchema` isn't known when we get a response back from AppSync. So even though `T` is `SerializedModel`, it's just a wrapper around some data, without an associated `ModelSchema`.

In the case of `SyncProcessor` and SubscripitonProcessor` my workaround was to ["patch" the `ModelSchema`](https://github.com/aws-amplify/amplify-android/blob/8ccb280c0418853731f46fe709bda857d7b7765e/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java#L145-L155) back into the `SerializedModel`. But I don't think either of us really liked that solution, :-(.

richardmcclellan (approved):

richardmcclellan:
> I believe this is the same basic issue we had in the `SubscriptionProcessor` and `SyncProcessor`. The `ModelSchema` isn't known when we get a response back from AppSync. So even though `T` is `SerializedModel`, the instance is just a wrapper around some data, without an associated `ModelSchema`.
> 
> In the case of `SyncProcessor` and `SubscriptionProcessor` my workaround was to ["patch" the `ModelSchema`](https://github.com/aws-amplify/amplify-android/blob/8ccb280c0418853731f46fe709bda857d7b7765e/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java#L145-L155) back into the `SerializedModel`. But I don't think either of us really liked that solution, :-(.


Got it, that makes sense.  Something to improve in the future!

jamesonwilliams:
> > I believe this is the same basic issue we had in the `SubscriptionProcessor` and `SyncProcessor`. The `ModelSchema` isn't known when we get a response back from AppSync. So even though `T` is `SerializedModel`, the instance is just a wrapper around some data, without an associated `ModelSchema`.
> > In the case of `SyncProcessor` and `SubscriptionProcessor` my workaround was to ["patch" the `ModelSchema`](https://github.com/aws-amplify/amplify-android/blob/8ccb280c0418853731f46fe709bda857d7b7765e/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java#L145-L155) back into the `SerializedModel`. But I don't think either of us really liked that solution, :-(.
> 
> Got it, that makes sense. Something to improve in the future!

Certainly, there is no shortage of such opportunities, haha. :-D

Issue #995, opened 2020-11-25
Android's mutation selection set is incompatible with iOS' subscription handling


jamesonwilliams:
From @Amplifiyer: 

While testing the sync between local and remote, @jamesonwilliams and I found some weird behavior between how it works in iOS and Android. The selection set of the create mutations in iOS seem to include the full nested model fields, while in android it’s only the ID. So for instance if I create two mutations from AppSync as follows:

```graphql
mutation OnlyIdMutation {
  createPost(input: {blogID: "6a83a3b4-2b82-4dfa-8b58-5e32326865a2", rating: 10, title: "Remote blog 6"}) {
    _deleted
    _lastChangedAt
    _version
    blogID
    created
    createdAt
    id
    rating
    title
    updatedAt
    blog {
      id
    }
  }
}
```

and

```graphql
mutation FullNestedModelMutation {
  createPost(input: {blogID: "6a83a3b4-2b82-4dfa-8b58-5e32326865a2", rating: 10, title: "Remote blog 6"}) {
    _deleted
    _lastChangedAt
    _version
    blogID
    created
    createdAt
    id
    rating
    title
    updatedAt
    blog {
      id
      updatedAt
      name
      createdAt
      _version
      _lastChangedAt
      _deleted
    }
  }
}
```

Both are successful on Android, but only the later works on iOS. The first one will result in some GraphQL errors on the iOS client:

```json
{
  "id": "2ACF986E-7607-4CD7-AEE3-148F34B01183",
  "type": "data",
  "payload": {
    "data": {
      "onCreatePost": {
        "id": "909fccc6-7ad3-4408-8b49-23616123413e",
        "created": null,
        "rating": 3,
        "title": "Post 2 android\t",
        "blog": null,
        "__typename": "Post",
        "_version": 1,
        "_deleted": null,
        "_lastChangedAt": 1606251462643
      }
    },
    "errors": [
      {
        "message": "Cannot return null for non-nullable type: 'String' within parent 'Blog' (/onCreatePost/blog/name)",
        "path": [
          "onCreatePost",
          "blog",
          "name"
        ]
      },
      {
        "message": "Cannot return null for non-nullable type: 'Int' within parent 'Blog' (/onCreatePost/blog/_version)",
        "path": [
          "onCreatePost",
          "blog",
          "_version"
        ]
      },
      {
[...]

PR #1023, opened 2020-12-03
fix(aws-datastore): include deep relationships in selection set

jamesonwilliams:
When the Android DataStore performs a mutation against AppSync, it will
build a selection set defining how AppSync responds.

When the mutation involves a model that has a relationship to another
model, the creation of the selection set differed between Android and
iOS.

iOS would request _all fields_ of the related model, as well as _all
fields_ from the parent. Android, on the other hand, would request _all
fields_ from the parent, but only the ID of the child.

When AppSync receives the Android mutation, it will create subscription
event. The payload of the subscription event is influenced by the
selection set specified in the mutation. iOS is only able to process
subcription payloads where all fields are specified. Due to Android's
decision to only include the related model's ID, iOS was not able to
process the (incomplete) child data.

The simplest way to resolve this is to include *all fields* in the
mutation selection set, so that iOS will get a "complete" child, and
process it successfully.

The code base contained an existing concept for this, called "leaf
serialization behavior." With this change, there is no longer any use
case for the "just id" behavior. So it is removed.

*All* selection sets in the code base will now include all fields from
in the related models.

Resolves: https://github.com/aws-amplify/amplify-android/issues/995


aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -355,18 +355,14 @@ private static boolean isCustomType(@NonNull Field field) {
...
             for (Map.Entry<String, ModelField> entry : modelSchema.getFields().entrySet()) {
                 String fieldName = entry.getKey();
                 ModelAssociation association = modelSchema.getAssociations().get(fieldName);
                 if (association != null) {
                     if (depth >= 1) {
                         String associatedModelName = association.getAssociatedType();
-                        ModelSchema associateModelSchema = ModelSchemaRegistry.instance()
-                                .getModelSchemaForModelClass(associatedModelName);
+                        ModelSchema associateModelSchema =
+                            modelSchemaRegistry.getModelSchemaForModelClass(associatedModelName);
```
jamesonwilliams, line 365: This change necessitated all of the rewiring of the `ModelSchemaRegistry`. `ModelSchemaRegistry.instance()` creates a new empty registry, but we need the instance that got setup by the storage adapter during initialization.

core/src/main/java/com/amplifyframework/core/model/ModelSchemaRegistry.java
```diff
@@ -119,4 +121,29 @@ public static synchronized ModelSchemaRegistry instance() {
...
+        if (thatObject == null || getClass() != thatObject.getClass()) {
+            return false;
+        }
+        ModelSchemaRegistry registry = (ModelSchemaRegistry) thatObject;
+        return getModelSchemaMap().equals(registry.getModelSchemaMap());
+    }
+
+    @Override
+    public int hashCode() {
+        return ObjectsCompat.hash(getModelSchemaMap());
```
rjuliano, line 139: Only edge case I can see here is if you happen to have two objects with identical schemas:

```
Thing1(id, name)
Thing2(id, name)
```
Would these two evaluate to the same hashCode?

rjuliano: nvm...This would be comparing two registries. 🤦

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
```diff
@@ -296,6 +330,7 @@ public Builder variable(@NonNull String key, String type, Object value) {
         public <R> AppSyncGraphQLRequest<R> build() throws AmplifyException {
             Objects.requireNonNull(this.operation);
             Objects.requireNonNull(this.responseType);
+            Objects.requireNonNull(this.modelSchemaRegistry);
```
richardmcclellan, line 333: I see that `AppSyncRequestFactory` (used by `DataStore`) has been updated to pass this new required parameter.  However, it looks like `AppSyncGraphQLRequestFactory` (used by `API`) is not updated yet.  I think that needs to be updated as well, otherwise, this is going to throw an NPE for any usage of the API category.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -67,7 +69,22 @@
...
  */
 final class AppSyncRequestFactory {
-    private AppSyncRequestFactory() {}
+    private final ModelSchemaRegistry modelSchemaRegistry;
+
+    /**
+     * Constructs an AppSyncRequestFactory using an empty schema registry.
+     */
+    AppSyncRequestFactory() {
+        this.modelSchemaRegistry = ModelSchemaRegistry.instance();
```
richardmcclellan, line 78: What's the purpose of this constructor?  Shouldn't a `ModelSchemaRegistry` always be passed in, instead of just creating an empty one?

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequest.java
richardmcclellan: Hmm, actually I guess it won't throw an NPE since I see you are initing it to an empty registry in the `Builder` constructor.   So I guess that's fine, but we may need to update `AppSyncGraphQLRequestFactory` in the future, if Flutter is going to depend on it for their API category?

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
```diff
@@ -186,6 +187,11 @@ public Builder modelSchema(@NonNull ModelSchema modelSchema) {
             return Builder.this;
         }
 
+        public Builder modelSchemaRegistry(@NonNull ModelSchemaRegistry modelSchemaRegistry) {
+            this.modelSchemaRegistry = modelSchemaRegistry;
```
richardmcclellan, line 191: Should we add `Objects.requireNotNull(modelSchemaRegistry)` in the `build()` method below?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
jamesonwilliams: I started out requiring it to be passed. I added this default constructor only to reduce the amount of changes needed to snake the repository around. But, I agree with your concern - this would be hard to track down later, and could cause sneaky bugs. I'll remove this version of the constructor.

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/SelectionSet.java
jamesonwilliams: Sounds good!

richardmcclellan:
@jamesonwilliams what's your thinking on this change?  I think we should still ship this so that we can fix the issue where iOS doesn't receive subscription events for mutations generated on Android.

jamesonwilliams:
> @jamesonwilliams what's your thinking on this change? I think we should still ship this so that we can fix the issue where iOS doesn't receive subscription events for mutations generated on Android.

Agreed. I had a bunch of unfinished work hanging around, so just closed some of it out. I'll re-open this when I'm able to apply needed polish. Also needs a thorough round of testing.

PR #1024, opened 2020-12-03
fix: adaptive timeouts should apply for Flutter too

jamesonwilliams:
The sync engine uses "adaptive" timeouts, that scale up or down
depending on the number of models that need to be sync'd.

To determine the number of models, `modelProvider.models().size()` was
being used. This will return 0 when being used from Flutter. Instead, we
can consider `modelProvider.modelNames().size()`, which will work for
both ModelSchema and Model-based usages of the DataStore.

Refer: https://github.com/aws-amplify/amplify-android/issues/854


jamesonwilliams:
cc: @fjnoyp

jamesonwilliams:
Upon looking at the existing logic, I'm not at all convinced this actually helps handle large data sets.

The existing logic says "If you have up to 6 model types, use a timeout of 10 seconds. If you have six or more model types, add two seconds for each (6 get 12s, 7 get 14s, 8 get 16s.)"

But this logic does nothing for a more common scenario, when there are thousands of instances of one or two model types.

PR #1025, opened 2020-12-04
fix(amplify-tools): Add --forcePush to init

rjuliano:
*Issue #, if available:*

*Description of changes:*
CLI requires that we use `forcePush` if we want to deploy changes during the init command. This PR incorporates this change into the build tools to address issues users found while walking through the getting started guide.


richardmcclellan (approved):
Looks good, just two quick questions:
 - Are we still planning to get rid of this tool entirely?
 - Is the integration test failure related to this change?

rjuliano:
> Looks good, just two quick questions:
> 
> * Are we still planning to get rid of this tool entirely?
> * Is the integration test failure related to this change?

Not entirely sure on the first. The idea has been floated, but no decision on it thus far. I'm looking into the integration test failures. Doesn't appear related to this. It started failing consistently yesterday.

jpignata:
> * Are we still planning to get rid of this tool entirely?

I think we should mirror what iOS has done and walk customers through using the CLI and amplify-app. This requires a bit more alignment with them before we rejigger the tutorial as they are reworking theirs.

jamesonwilliams (approved):

PR #1033, opened 2020-12-09
fix(aws-datastore): deleting nonexistent model instance no longer fails

raphkim:
*Description of changes:*
SQLiteStorageAdapter now checks if item exists before attempting delete. If not, then operation "succeeds" and returns a dummy item change instance, but does not publish any change.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
@@ -481,6 +481,20 @@ public final class SQLiteStorageAdapter implements LocalStorageAdapter {
                 final SQLiteTable sqliteTable = SQLiteTable.fromSchema(modelSchema);
                 final String primaryKeyName = sqliteTable.getPrimaryKeyColumnName();
 
+                if (!dataExistsInSQLiteTable(sqliteTable.getName(), primaryKeyName, item.getId())) {
+                    LOG.warn(modelName + " model with id = " + item.getId() + " does not exist.");
+                    // Pass back item change instance without publishing it.
+                    onSuccess.accept(StorageItemChange.<T>builder()
+                        .changeId(item.getId())
+                        .item(item)
+                        .modelSchema(modelSchema)
+                        .type(StorageItemChange.Type.DELETE)
+                        .predicate(predicate)
+                        .initiator(initiator)
+                        .build());
+                    return;
+                }
+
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -499,22 +513,18 @@ public void query(
...
-                            "Wanted to delete one row, but deleted " + rowsDeleted + " rows.",
-                            "This is likely a bug. Please report to AWS."
-                        );
-                    }
                     compiledSqlStatement.clearBindings();
-                    if (problem != null) {
-                        throw problem;
+                    if (rowsDeleted == 0) {
+                        throw new DataStoreException(
+                            "Failed to meet condition. Model was not deleted.",
```
richardmcclellan, line 525: Much better error message here!

richardmcclellan (approved):
🚢

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -481,6 +481,20 @@ public void query(
                 final SQLiteTable sqliteTable = SQLiteTable.fromSchema(modelSchema);
                 final String primaryKeyName = sqliteTable.getPrimaryKeyColumnName();
 
+                if (!dataExistsInSQLiteTable(sqliteTable.getName(), primaryKeyName, item.getId())) {
```
jamesonwilliams, line 484: Two questions here:

Does this mean we perform two operations on SQL? A lookup and the a delete? Performance wise that’s probably okay for our delete API. Might have implications for atomicity of the operation.

Second more important - should we fire an event on the observe method here? Would the sync engine care about observing this case?

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: > Does this mean we perform two operations on SQL? A lookup and the a delete?

This lookup is already performed in `save` to determine whether item should be created or updated. There hasn't been an issue there so far, so it will be fine.

> should we fire an event on the observe method here? 

No change took place in the database, so I don't see the need for the sync engine to worry about it.

Issue #1028, opened 2020-12-07
DELETE rest request fails with and without body (when using v4 signer)


patrickhoette:
When making a DELETE REST request using the Amplify SDK if you include a body with it the SDK throws an exception. When you create a DELETE request without a body Okhttp3 adds a body of `byte[0]`. Later down the chain the `AppSyncSigV4SignerInterceptor` sets the content type to JSON as it only checks if the body is `null`.

I'm using version 1.6.6. of Amplify Android.

```kotlin
override fun makeDeleteCall(path: String): Completable {
    return Completable.create {
        Amplify.API.delete(
            createCallOptions(path),
            it.consumeSuccess(),
            it.consumeError()
        )
    }
}

private fun createCallOptions(path: String): RestOptions {
    return RestOptions.builder().run {
        addPath(path)
        build()
    }
}
```

With body:
```
ApiException{message=HTTP method does not support data object! DELETE, cause=null, recoverySuggestion=Try sending the request without any data in the options.}
```

Without body:
```
java.io.IOExceptionL: Network request not successful. Code{statusCode=400}: {"statusCode":400,"code":"FST_ERR_CTP_EMPTY_JSON_BODY","error":"Bad Request","message":"Body cannot be empty when content-type is set to 'application/json'"}
```

### Longer explanation

The bug is that creating a `DELETE` request without a body creates an invalid request. If you do that you pass the check the Amplify SDK does in `com.amplifyframework.api.aws.AWSApiPlugin` line 605-617. Then it goes on to OkHttp which in `okhttp3.Request` on line 239-241 (method is `delete()`) does:
```java
    public Builder delete(@Nullable RequestBody body) {
      return method("DELETE", body);
    }

    public Builder delete() {
      return delete(Util.EMPTY_REQUEST);
    }
```
and in the `okhttp3.internal.Util` class on lines 59 and 63:
```java
  public static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
  public static final RequestBody EMPTY_REQUEST = RequestBody.create(null, EMPTY_BYTE_ARRAY);
```

So later on in `com.amplifyframework.api.aws.sigv4.AppSyncSigV4SignerInterceptor` the following happens on line 229-231:
```java
//Set the URL and Method
final RequestBody requestBody = req.body() != null ?
                RequestBody.create(bodyBytes, JSON_MEDIA_TYPE) : null;
```
This adds the header `content-type: application/json` to the request. Which then leads to  request like the following:
[...]

PR #1037, opened 2020-12-09
[aws-api] Fix DELETE calls not working with v4 signer

patrickhoette:
`OkHttp3` sets the body of a `DELETE` call to a byte array of length 0. The
v4 signer interceptor only checks if the body is null and if its not it sets the
content type to `application/json`. This causes an invalid REST request to
be created. Fixed by expanding the check to also make sure the body has a
content length greater than 0.

Github Issue: https://github.com/aws-amplify/amplify-android/issues/1028

Tested in the app that had the original problem and verified to fix the issue.


jamesonwilliams (approved):
Thanks!

patrickhoette:
No worries!

richardmcclellan (approved):

PR #1039, opened 2020-12-09
fix(aws-datastore): Resolves crash when going offline

richardmcclellan:
Fixes https://github.com/aws-amplify/amplify-android/issues/541 (DataStore crashes when device goes offline)

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -241,7 +241,8 @@ final class SubscriptionEndpoint {
 
         // Only do this if the subscription was NOT pending.
         // Otherwise it would probably fail since it was never established in the first place.
-        if (!wasSubscriptionPending) {
+
+        if (!wasSubscriptionPending && !webSocketListener.isDisconnectedState()) {
             try {
                 webSocket.send(new JSONObject()
                     .put("type", "stop")
@@ -257,8 +258,11 @@ final class SubscriptionEndpoint {
             subscription.awaitSubscriptionCompleted();
         }
 
+        subscriptions.remove(subscriptionId);
+
         // If we have zero subscriptions, close the WebSocket
         if (subscriptions.size() == 0) {
+            LOG.info("No more active subscriptions. Closing web socket.");
             timeoutWatchdog.stop();
             webSocket.close(NORMAL_CLOSURE_STATUS, "No active subscriptions");
         }
@@ -512,6 +516,9 @@ final class SubscriptionEndpoint {
                 return new Connection("Thread interrupted waiting for connection acknowledgement");
             }
             LOG.debug("Current endpoint status: " + endpointStatus.get());
+            if (EndpointStatus.CONNECTION_FAILED.equals(endpointStatus.get())) {
+                return new Connection("Connection failed.");
+            }
             return new Connection();
         }
 
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -376,7 +376,7 @@ final class SubscriptionEndpoint {
                 }
             } catch (InterruptedException interruptedException) {
                 dispatchError(new ApiException(
-                    "Failure awaiting acknowledgement of subscription completion.",
+                    "Thread interrupted awaiting subscription completion.",
                     interruptedException,
                     AmplifyException.TODO_RECOVERY_SUGGESTION
                 ));
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
@@ -111,7 +111,7 @@ final class SubscriptionOperation<T> extends GraphQLOperation<T> {
         } else if (subscriptionFuture != null && subscriptionFuture.cancel(true)) {
             LOG.debug("Subscription attempt was canceled.");
         } else {
-            LOG.debug("Nothing to cancel. Subscription not yet created.");
+            LOG.debug("Nothing to cancel. Subscription not yet created, or already cancelled.");
         }
     }
 
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
@@ -88,11 +88,8 @@ final class SubscriptionOperation<T> extends GraphQLOperation<T> {
                 },
                 onNextItem,
                 apiException -> {
-                    // Guard against calling something that's been cancelled already.
-                    if (!canceled.get()) {
-                        canceled.set(true);
-                        onSubscriptionError.accept(apiException);
-                    }
+                    cancel();
+                    onSubscriptionError.accept(apiException);
                 },
                 onSubscriptionComplete
             );
@@ -104,6 +101,7 @@ final class SubscriptionOperation<T> extends GraphQLOperation<T> {
         if (subscriptionId != null && !canceled.get()) {
             canceled.set(true);
             try {
+                LOG.debug("Cancelling subscription: " + subscriptionId);
                 subscriptionEndpoint.releaseSubscription(subscriptionId);
             } catch (ApiException exception) {
                 onSubscriptionError.accept(exception);
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -42,8 +42,8 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.rxjava3.core.Completable;
-import io.reactivex.rxjava3.core.Scheduler;
 import io.reactivex.rxjava3.disposables.CompositeDisposable;
+import io.reactivex.rxjava3.functions.Action;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 
 /**
@@ -64,7 +64,6 @@ public final class Orchestrator {
     private final AtomicReference<State> currentState;
     private final MutationOutbox mutationOutbox;
     private final CompositeDisposable disposables;
-    private final Scheduler startStopScheduler;
     private final long adjustedTimeoutSeconds;
     private final Semaphore startStopSemaphore;
 
@@ -131,7 +130,6 @@ public final class Orchestrator {
         this.currentState = new AtomicReference<>(State.STOPPED);
         this.targetState = targetState;
         this.disposables = new CompositeDisposable();
-        this.startStopScheduler = Schedulers.single();
 
         // Operation times out after 10 seconds. If there are more than 5 models,
         // then 2 seconds are added to the timer per additional model count.
@@ -149,7 +147,7 @@ public final class Orchestrator {
      *      and started (asynchronously) the transition to SYNC_VIA_API, if an API is available.
      */
     public synchronized Completable start() {
-        return performSynchronized(Completable.fromAction(() -> {
+        return performSynchronized(() -> {
             switch (targetState.get()) {
                 case LOCAL_ONLY:
                     transitionToLocalOnly();
@@ -161,7 +159,7 @@ public final class Orchestrator {
                 default:
                     break;
             }
-        }));
+        });
     }
 
     /**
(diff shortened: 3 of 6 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -288,91 +294,77 @@ private void stopObservingStorageChanges() {
...
+
+                LOG.debug("About to hydrate...");
+                try {
+                    boolean subscribed = syncProcessor.hydrate()
+                            .blockingAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+                    if (!subscribed) {
+                        throw new TimeoutException("Timed out while performing initial model sync.");
+                    }
+                } catch (Throwable failure) {
+                    if (!emitter.isDisposed()) {
```
jamesonwilliams, line 316: 😺

jamesonwilliams (approved):
Code changes look promising. Do we have sufficient automated coverage of this stuff? Which scenarios were you able to test manually through the [sample app](https://github.com/richardmcclellan/app-datastore)?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -105,14 +104,11 @@ public static AppSyncStep builder() {
...
      */
-    synchronized void startSubscriptions() {
+    synchronized void startSubscriptions(Action onPipelineBroken) {
         int subscriptionCount = modelProvider.modelNames().size() * SubscriptionType.values().length;
         // Create a latch with the number of subscriptions are requesting. Each of these will be
         // counted down when each subscription's onStarted event is called.
         CountDownLatch latch = new CountDownLatch(subscriptionCount);
-        // Need to create a new buffer so we can properly handle retries and stop/start scenarios.
-        // Re-using the same buffer has some unexpected results due to the replay aspect of the subject.
-        buffer = ReplaySubject.create();
```
richardmcclellan, line 115: Before making any of the changes in this PR, I was _not_ able to reproduce this crash if only one subscription is active (just ON_CREATE, for 1 model).  The basic issue causing the crash is due to the following sequence of events:

 1. Going offline closes the socket, which calls `AmplifyWebSocketListener::onFailure`
 2. `onSubscriptionFailure` is called for the first subscription
 3. The error is passed to the buffer (`buffer.onError()`)
 4. `onPipelineBroken::call` (the Action that the Orchestrator calls on error)
 5. `Orchestrator::stopApiBlocking` is called
 6. `SubscriptionProcessor::stopAllSubscriptionActivity` is called, which disposes of the TWO disposables in SubscriptionProcessor: the `buffer`, and the `subscriptionObservables`.
 7.  `onSubscriptionFailure` is called for the second subscription
 8. At this point, the `buffer` is already disposed, and the app crashes with: `io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with`

This PR solves the problem by removing the buffer entirely, and just streaming `SubscriptionEvent` objects from the subscriptionObservables directly to the `Merger`.  The `SubscriptionProcessor` is now only managing  one `Disposable` (the subscriptionObservable stream) instead of two.   The same crash doesn't occur with the `subscriptionObservable` stream, because there is an explicit check to see if the stream is disposed before emitting errors on it.  I don't think a similar check was possible with the buffer though, since it was a `ReplaySubject`, which I don't believe has a way to check if it has been disposed or not.  

The one side effect of this change is that `SubscriptionEvent` are sent to the Merger as they arrive, instead of waiting for all subscriptions to be started.  I don't believe there is any problem with this, but just calling it out.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -185,50 +183,56 @@ private Completable performSynchronized(Completable completable) {
...
     }
 
-    private Completable transitionToStopped() {
-        LOG.info("DataStore orchestrator stopping. Current mode = " + currentState.get().name());
-        disposables.clear();
+    private void transitionToStopped() throws DataStoreException {
         switch (currentState.get()) {
             case SYNC_VIA_API:
-                return stopApiSync().doFinally(this::stopObservingStorageChanges);
+                LOG.info("Orchestrator transitioning from SYNC_VIA_API to STOPPED");
```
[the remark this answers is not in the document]
richardmcclellan: I had the logs outside the switch initially, but was seeing a ton of "Orchestrator transitioning from SYNC_VIA_API to SYNC_VIA_API" logs.  I moved them into the switch, so it would only log the 6 transitions that actually matter :)

I've been thinking about log levels and feel like the following makes sense:
 - `Verbose` should be used when logging something for every instance of a model
 - `Debug` should be used when logging something for every model type
 - `Info` should be used for general state changes in the orchestrator or any of the processors, that only result in ~1 instance of the log message

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
richardmcclellan: This was already there, I just moved it a bit :). I believe the root cause though is that we didn't have a similar check for the `ReplaySubject buffer` before.  We did check `hasThrowable`, but I don't think that's quite the same.  So, the solution I've implemented is to just simplify and get rid of the buffer entirely.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -375,6 +375,7 @@ void awaitSubscriptionCompleted() {
                     ));
                 }
             } catch (InterruptedException interruptedException) {
+                LOG.warn("Thread interrupted awaiting subscription completion.", interruptedException);
                 dispatchError(new ApiException(
                     "Failure awaiting acknowledgement of subscription completion.",
```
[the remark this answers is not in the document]
richardmcclellan: Updated!

richardmcclellan:
> Code changes look promising. Do we have sufficient automated coverage of this stuff? Which scenarios were you able to test manually through the [sample app](https://github.com/richardmcclellan/app-datastore)?

Here's what I verified manually:
1. Start app while online
2. Turn off WiFi
3. Assert app doesn't crash, and log messages indicate a successful transition to LOCAL_ONLY
4. Create a Todo.  Verify via the logs that the app attempts to start API sync, but fails, and transitions back to LOCAL_ONLY.
5. Turn on WiFi
6. Create another Todo (needed to trigger DataStore.start()).
7. Verify via the logs that API sync is started succesfully
 and two `outboxMutationProcessed` events are observed, confirming both the Todo created offline, and the Todo created online are synced.

The integration tests aren't working for me at all right now.  Let me see if I can get those working, as it would be good to do that before this gets merged in.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -241,7 +241,8 @@ synchronized void releaseSubscription(String subscriptionId) throws ApiException
 
         // Only do this if the subscription was NOT pending.
         // Otherwise it would probably fail since it was never established in the first place.
-        if (!wasSubscriptionPending) {
+
+        if (!wasSubscriptionPending && !webSocketListener.isDisconnectedState()) {
```
richardmcclellan, line 245: Don't try to send a "stop" message if the web socket is disconnected.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -257,8 +258,11 @@ synchronized void releaseSubscription(String subscriptionId) throws ApiException
             subscription.awaitSubscriptionCompleted();
         }
 
+        subscriptions.remove(subscriptionId);
```
richardmcclellan, line 261: Previously there was no logic for cleaning up the subscriptions array.  Adding this here ensures that:
 - the subscription objects are properly removed from `subscriptions`
 - the websocket is closed when `subscriptions.size() == 0`.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -512,6 +516,9 @@ public Connection waitForConnectionReady() {
                 return new Connection("Thread interrupted waiting for connection acknowledgement");
             }
             LOG.debug("Current endpoint status: " + endpointStatus.get());
+            if(EndpointStatus.CONNECTION_FAILED.equals(endpointStatus.get())) {
+                return new Connection("Connection failed.");
+            }
```
richardmcclellan, line 521: For the following use case:
 - WiFi is off.
 - Client attempts to save a mutation.

This change ensures that we fail immediately, instead of waiting 10 seconds for the start_ack.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionOperation.java
```diff
@@ -88,11 +88,8 @@ public synchronized void start() {
                 },
                 onNextItem,
                 apiException -> {
-                    // Guard against calling something that's been cancelled already.
-                    if (!canceled.get()) {
-                        canceled.set(true);
-                        onSubscriptionError.accept(apiException);
-                    }
+                    cancel();
```
richardmcclellan, line 91: This was a bug.  We can't just set `canceled` to true.  We have to actually do the cancellation too.

I've also removed the `!canceled.get()` guard, because the `cancel` method itself is already `idempotent`, and once a subscription is cancelled, it won't throw any additional errors.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -185,15 +186,12 @@ private boolean isUnauthorizedException(DataStoreException exception) {
...
-                    if (!emitter.isDisposed()) {
-                        if (isUnauthorizedException(dataStoreException)) {
-                            LOG.warn("Unauthorized failure for " + subscriptionType + " " + modelSchema.getName());
-                        } else {
-                            emitter.onError(dataStoreException);
-                        }
+                    if (isUnauthorizedException(dataStoreException)) {
+                        LOG.warn("Unauthorized failure for " + subscriptionType.name() + " " + modelSchema.getName());
+                    } else {
+                        onFailure.accept(dataStoreException);
```
richardmcclellan, line 194: This is the most important change in this whole PR.  Instead of passing errors to the `emitter`, and then to the `buffer`, all errors are sent directly to an `onFailure` callback which the Orchestrator passes in to the `SubscriptionProcessor` on initialization.  

Basically, we are no longer sending errors to any stream, since those streams end up getting disposed after receiving an error.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -512,6 +516,9 @@ public Connection waitForConnectionReady() {
                 return new Connection("Thread interrupted waiting for connection acknowledgement");
             }
             LOG.debug("Current endpoint status: " + endpointStatus.get());
+            if (EndpointStatus.CONNECTION_FAILED.equals(endpointStatus.get())) {
```
jamesonwilliams, line 519:
```suggestion
            if (EndpointStatus.CONNECTION_FAILED.equals(endpointStatus.get())) {
```

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -512,6 +516,9 @@ public Connection waitForConnectionReady() {
...
+                return new Connection("Connection failed.");
+            }
```
richardmcclellan, line 521: This change ensures that if there is the device has no network connection, the subscription will fail immediately, instead of failing due to timeout awaiting the `start_ack` after 10 seconds.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
jamesonwilliams: This comment is out of date, right?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -388,7 +393,12 @@ public SubscriptionProcessor build() {
...
     interface QueryPredicateProviderStep {
         @NonNull
-        BuildStep queryPredicateProvider(QueryPredicateProvider queryPredicateProvider);
+        OnFailureStep queryPredicateProvider(QueryPredicateProvider queryPredicateProvider);
+    }
+
+    interface OnFailureStep {
+        @NonNull
+        BuildStep onFailure(Consumer<Throwable> onFailure);
     }
```
jamesonwilliams, line 402:
```suggestion
    }

```

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
richardmcclellan: Yes, it is out of date.  I put the buffer back, because I believe it is needed to prevent a data inconsistency scenario.      Instead, this PR solves the crash by only using the buffer for SubscriptionEvents, not errors.  Errors are kept out of any RX stream since it may be disposed, and instead sent to a error callback provided by the Orchestrator.

Issue #1031, opened 2020-12-08
[Flutter Support] DataStore should support a double instead of a float


Amplifiyer:
As per this [code](https://github.com/aws-amplify/amplify-android/blob/c425a4f1073b6245cf8fef18096c553e8fec99f2/aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java#L28) and [CLI model generation](https://github.com/aws-amplify/amplify-cli/blob/b26c40f433d7d703878b6df25dd312f5de34afdb/packages/amplify-codegen-appsync-model-plugin/src/scalars/index.ts#L7), amplify-android requires GraphQL `float` type to be `Java`'s `Float` type. Based on [appsync guide ](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) a float is

> The Float scalar type represents signed double-precision fractional values as specified by IEEE 754. 

This means that we lose precision when converting a dart `double` to a java `Float`.

Expectation: amplify-android to support `double` type to properly accommodate the appsync `Float`

Originated from: https://github.com/aws-amplify/amplify-flutter/issues/242

jamesonwilliams:
While this fix seems trivial at first glance, it is more challenging than initially meets the eye. In order to support models that currently use `Float`, we have to do this in a few phases:
1. Conditionally support both `java.lang.Float` and `java.lang.Double` in the library. Choose the type based on the type of the passed data value.
2. Update the CLI to use `java.lang.Double` in the Java language models.
3. Later, remote support for the `java.lang.Float` type.

Note for Flutter integrators: after this change, you should pass `javaClassForValue` as `java.lang.Double`.  The type of the value in the map data should also be `java.lang.Double`.

PR #1040, opened 2020-12-09
fix(aws-datastore): store schema float as a java double

jamesonwilliams:
The Amplify DataStore leverages GraphQL to define its data models.
Therefor, our intention is to support a strict superset of the primitive
types described by the GraphQL specification.

While implementing the DataStore, we overlooked that the GraphQL "Float"
type is a double-precision value. The SQL adapter is able to accomodate
this, currently. However, our Java codegen stores a float value, and we
look for values of Float type while serializing and deserializing object
data.

Unfortunately, we need to continue to support that behavior, since
customers will have Float values in their models. None-the-less, the
Flutter consumer of this library rightly wants to use the actual GraphQL
scalar types, and see the same behavior accross iOS and Android. As a
result of these constraints, we must support both types in the library,
for now.

The decision to express a floating point value as a Java Float or Double
will be taken from the type of the value in the model object. For our
Java models, it will be a Float. When Map data is passed from Flutter,
it will be a Double. Both SQL and AppSync will treat both as a double,
which is large enough to accomodate both types. Or in other words, both
types are supported, now.

The next step will be to update the Java codegen, so that it represents
GraphQL Float types as Java Doubles. This change will allow new
customers to use the fix, while we continue to support existing
customers' models.

Resolves: https://github.com/aws-amplify/amplify-android/issues/1031

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
@@ -47,6 +47,11 @@ public enum JavaFieldType {
     FLOAT(Float.class),
 
     /**
+     * Represents the double data type.
+     */
+    DOUBLE(Double.class),
+
+    /**
      * Represents the String data type.
      */
     STRING(String.class),
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncConflictUnhandledErrorTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncConflictUnhandledErrorTest.java
@@ -119,9 +119,8 @@ public final class AppSyncConflictUnhandledErrorTest {
             AppSyncConflictUnhandledError.findFirst(Note.class, response.getErrors());
         assertNotNull(conflictUnhandledError);
 
-        // TODO: The JSON document has '1601499066604' as the time. These differ by 26604,
-        // also the TimeUnit appears to be wrong. Should be MILLISECONDS.
-        Temporal.Timestamp lastChangedAt = new Temporal.Timestamp(1601499040000L, TimeUnit.SECONDS);
+        // TODO: the TimeUnit appears to be wrong. Should be MILLISECONDS?
+        Temporal.Timestamp lastChangedAt = new Temporal.Timestamp(1601499066604L, TimeUnit.SECONDS);
         assertEquals(
             new ModelWithMetadata<>(
                 new Note("KoolId22", "Resurecting the dataz"),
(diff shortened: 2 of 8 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/appsync/AppSyncConflictUnhandledErrorTest.java
```diff
@@ -119,9 +119,8 @@ public void conflictErrorExtractedIfPresent() {
             AppSyncConflictUnhandledError.findFirst(Note.class, response.getErrors());
         assertNotNull(conflictUnhandledError);
 
-        // TODO: The JSON document has '1601499066604' as the time. These differ by 26604,
-        // also the TimeUnit appears to be wrong. Should be MILLISECONDS.
-        Temporal.Timestamp lastChangedAt = new Temporal.Timestamp(1601499040000L, TimeUnit.SECONDS);
+        // TODO: the TimeUnit appears to be wrong. Should be MILLISECONDS?
```
jamesonwilliams, line 122: Accidentally fixed one of the bugs here. Appears to have been caused by an overflow.

The milliseconds issue is unrelated. That's about AppSync using `AWSTimestamp` as:
> the number of seconds that have elapsed since 1970-01-01T00:00Z

But then it returns a value in milliseconds in the conflict handler data, in production 🤷 .

richardmcclellan:
Should we also update codegen to generate `Double` instead of `Float` now?   The mapping is right here: https://github.com/aws-amplify/amplify-cli/blob/b624e0fff58659d0aeb13bc3e79b437071295aa3/packages/amplify-codegen-appsync-model-plugin/src/scalars/index.ts#L7

aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
```diff
@@ -46,6 +46,11 @@
      */
     FLOAT(Float.class),
```
richardmcclellan, line 47: Could we get rid of the `FLOAT` type at some point in the future?  We can't now, since the codegen'd models all have `Float` objects, but if we change codegen to generate Double objects, then we could, right?

richardmcclellan (approved):

aws-api-appsync/src/main/java/com/amplifyframework/core/model/types/JavaFieldType.java
jamesonwilliams: Yes! Down the road, when the majority of folks are using the new modelgen, we could drop the Floats. Truthfully I suspect we could do it now, if we showed people how to update their models. But, it would technically be a breaking change.

jamesonwilliams:
> Should we also update codegen to generate `Double` instead of `Float` now?   The mapping is right here: https://github.com/aws-amplify/amplify-cli/blob/b624e0fff58659d0aeb13bc3e79b437071295aa3/packages/amplify-codegen-appsync-model-plugin/src/scalars/index.ts#L7
> 
Yes! I think I'll queue that change up immediately after we release this dual-support in the library. Technically speaking, the CLI would become only compatible with that version and forward.

PR #1041, opened 2020-12-10
chore: remove circleci configs

jamesonwilliams:
We don't use CircleCI anymore, so remove the configuration files.
They'll be in the git log, if we need them ;-).


richardmcclellan (commented):
I think we need to keep `.circleci/copy-configs`, right?

jamesonwilliams:
> I think we need to keep `.circleci/copy-configs`, right?

It has been forked into `scripts/pull_backend_config_from_s3`. The new script can get resources from any bucket, but they have to be in a specific layout (that mirrors the `src/androidTest/res/raw` pathing.)

There is a __new__ S3 bucket, that is _not_ the one in the `copy-configs`, that has that layout.

We can download resources from it using that new script:
```console
the_shell --command --we_use --to_get --internal_credentials
./scripts/pull_backend_config_from_s3 a_bucket_ill_tell_you_about_offline
```

richardmcclellan:
Oh!  Huh.  So, is that why some of my integration tests are failing?  Maybe because I'm pulling configs from the wrong bucket?

jamesonwilliams:
> Oh! Huh. So, is that why some of my integration tests are failing? Maybe because I'm pulling configs from the wrong bucket?

Yes! I was surprised to learn about the new bucket, too. Definitely, I think we only need one bucket.

richardmcclellan (approved):

rjuliano:
> > Oh! Huh. So, is that why some of my integration tests are failing? Maybe because I'm pulling configs from the wrong bucket?
> 
> Yes! I was surprised to learn about the new bucket, too. Definitely, I think we only need one bucket.

The content of the config files didn't change. I just copied them over to the ops account where the builds are running now, so we wouldn't have to deal with cross-account permissions. 

That other script (`pull_backend_from_amplify`) is ideally where we end up. With the amplify projects deployed somewhere, that script will just do an `amplify pull` as part of the build. I paused on that because the backend projects were in a couple different accounts.

PR #1043, opened 2020-12-10
chore: simplify implementation of config pull script

jamesonwilliams:
The `pull_backend_config_from_s3` was forked from an old script,
`./scripts/copy-config`. However, it was updated to not rely on the
`sources` array. Instead, S3 keys must have now have the same folder
hierarchy as local file paths. As a consequence of that change, we no
longer need to maintain a sources array, and can further simplify the
iteration and download process.


rjuliano (approved):
Ship it

PR #1051, opened 2020-12-11
fix(aws-datastore): MutationProcessor - Resolve out-of-sync model  version

### LINKED ISSUE #1050: DataStore model _version gets out of sync with backend
DataStore models versions will occasionally get out of sync with the backend.
amplify-android: 1.6.7

This sync problem occurs only on the device that made the mutation. Most of the time it will go unnoticed since that device also has the updated model stored locally. However, the next model mutation sent to the backend will have the wrong _version number in the request. The side effects are an occasional discard of an update by the backend or merging of old/new items in a list. 

Turning on full AppSync logging is the easiest way to see this. The following shows a partial log of the input/output. All updates were made from the same client so _version should always be in sync.
```json
{
  "arguments": {
    "input": {
      "id": "fb544757-3cbc-49e0-a396-2f209d8abdec",
      "testInt": 14,
      "testString": "string-14",
      "_version": 12,
      "updatedAt": "2020-12-11T18:48:02.958Z",
      "__typename": "TestModel"
    }
  },
  "result": {
    "createdAt": "2020-12-11T18:17:41.318Z",
    "_lastChangedAt": 1607712483041,
    "__typename": "TestModel",
    "testString": "string-11",
    "testInt": 11,
    "id": "fb544757-3cbc-49e0-a396-2f209d8abdec",
    "_version": 16,
    "updatedAt": "2020-12-11T18:46:56.791Z"
  }
}
```
A simple Flutter test app can be found here: https://github.com/kjones/datastore_test

kjones:
Fixes #1050

DataStore model mutation responses from AppSync are being discarded and the model _version is not being updated properly in the local store. Once that is fixed, outbox mutations for SerializedModels will crash during save because they are missing the modelSchema property.

The discard is happening due to code that removes outstanding mutations and code that checks for outstanding mutations before proceeding are not being called in the opposite order as desired. 

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
@@ -134,12 +134,13 @@ final class MutationProcessor {
         return mutationOutbox.markInFlight(mutationOutboxItem.getMutationId())
             // Then, put it "into flight"
             .andThen(publishToNetwork(mutationOutboxItem)
+                .map(modelWithMetadata -> ensureModelHasSchema(mutationOutboxItem, modelWithMetadata))
                 .flatMapCompletable(modelWithMetadata ->
                     // Once the server knows about it, it's safe to remove from the outbox.
                     // This is done before merging, because the merger will refuse to merge
                     // if there are outstanding mutations in the outbox.
                     mutationOutbox.remove(mutationOutboxItem.getMutationId())
-                        .andThen(merger.merge(modelWithMetadata))
+                        .andThen(Completable.defer(() -> merger.merge(modelWithMetadata)))
                         .doOnComplete(() -> {
                             String modelName = mutationOutboxItem.getModelSchema().getName();
                             announceMutationProcessed(modelName, modelWithMetadata);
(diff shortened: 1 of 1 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -134,12 +134,13 @@ private Completable drainMutationOutbox() {
...
             // Then, put it "into flight"
             .andThen(publishToNetwork(mutationOutboxItem)
+                .map(modelWithMetadata -> ensureModelHasSchema(mutationOutboxItem, modelWithMetadata))
                 .flatMapCompletable(modelWithMetadata ->
                     // Once the server knows about it, it's safe to remove from the outbox.
                     // This is done before merging, because the merger will refuse to merge
                     // if there are outstanding mutations in the outbox.
                     mutationOutbox.remove(mutationOutboxItem.getMutationId())
-                        .andThen(merger.merge(modelWithMetadata))
+                        .andThen(Completable.defer(() -> merger.merge(modelWithMetadata)))
```
kjones, line 143: You could also fix `merger.merge()` to not do anything until subscribed to.

jamesonwilliams (approved):

PR #1052, opened 2020-12-12
fix(aws-datastore): Send Unix epoch in OutboxMutationEvent instead of Temporal.Timestamp

kjones:
*Description of changes:*
DataStore Hub [event documentation](https://docs.amplify.aws/lib/datastore/datastore-events/q/platform/android#outboxmutationprocessed) states that the OutboxMutationProcessed event will contain a `_lastChangedAt` field of type `Int`. Currently the event is being sent with `_lastChangedAt` as type `Temporal.Timestamp`. This PR switches that to sending a Unix epoch (type `java.lang.Long`) instead.

*Notes:*
- This is a breaking change to existing Android apps. The odds do seem extremely low that any Android app has been digging deep into the internals of the OutboxMutationProcessed DataStore hub event.
- This now matches the iOS format.
- Fixes a crash in the Amplify Flutter SDK that is assuming the value passed matches the documentation. I'm not sure this event has ever been seen by a Flutter app since other bugs, that will be fixed in #1051, prevented it from being published.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationEvent.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OutboxMutationEvent.java
@@ -20,7 +20,6 @@ import androidx.annotation.Nullable;
 import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.core.model.Model;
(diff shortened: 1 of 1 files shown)


jamesonwilliams (approved):
Thanks, Ken, looks good!

Issue #605, opened 2020-06-26
RFC: Kotlin Extensions for Amplify Libraries


jpignata:
The Amplify Libraries for Android are written in Java. Increasingly, we're hearing from developers that you are using Kotlin. In recognition of this, we would like to provide a more idiomatic interface for using Amplify from Kotlin. We're considering building a Kotlin Extensions library. The extensions would provide a Kotlin-first experience to developers integrating the Amplify Libraries in their Kotlin-based Android applications.

We're requesting your feedback on this general product direction, and have some specific questions:

- Are you building applications in Java or Kotlin?
- If you're using the Amplify Libraries in a Kotlin application, what are the points of friction you've encountered?
- How would you expect to interact with a Kotlin facade for the Amplify Libraries? Are there any patterns you'd prefer to see implemented?

Please comment on this thread with your thoughts, suggestions, nits, complaints, or anything else on your mind. 

Reactions (👍 , 👎 , etc.) are also welcomed, as they help us weigh the relative importance of this work.

Thank you!

shichonghuotian:
I hope amplify can support the coroutine

Blu3Parr0t:
I am building applications in Kotlin. I'm also an Android Contractor. A lot of jobs I get are requiring Kotlin experience due to most companies wanting to migrate to Kotlin. There is still Java needed for maintaining old frameworks that aren't being migrated over, but most new features are being created with Kotlin in mind.

The main friction I'm having right now is making a wrapper utility function that would be able to handle all of Amplify calls from sign up, login, queries, and mutations in a more generic way. Since I'm used to using Retrofit I've been able to pass in suspend functions and wrap them with the ServiceResult sealed class (https://developer.android.com/jetpack/guide). This has helped me to be able to handle all types of calls in a more generic manner in order to catch errors.

Facade could work. As mentioned above the way I'd develop to handle the Amplify calls would be to implement a singleton utility function that would expect a suspend fun and return a Response<T> where that may be for sing up, log in, mutation, queries. There might be more to unpack on how to effectively deliver the payload to a client, but one of the biggest issues we have at least in Android is being able to cover error cases. Most companies want us to implement some sort of mechanism that will catch Errors. That requires us to keep in mind not only when the API is down, but if the API returns null, user is has wireless off, etc. Most companies I've been to have had devs create base model classes that have the error fields and are extended to their unique model of expected data in order to handle false positives from the back end. Outside of that they implement something like this: 

```
object GraphQLHandler {

[...]

PR #1057, opened 2020-12-16
A Kotlin-centric Facade for Amplify Android

jamesonwilliams:
This commit introduces a new, optional `core-kotlin` module.
The module improves the experience of using Amplify from Kotlin.

### Usage
To use the Kotlin facade, include this dependency:
```gradle
dependencies {
    implementation "com.amplifyframework:core-kotlin:$version"
}
```
And import the Kotlin facade *instead of* the similar one at
`com.amplifyframework.core.Amplify`:

```kotlin
import com.amplifyframework.kotlin.core.Amplify
```

### Overview
core-kotlin introduces an alternate `Amplify` facade, which provides new
means of interacting with the various Amplify categories (API, Auth,
DataStore, Predictions, Storage.) The new facade include several
improvements to the Kotlin developer experience, mainly by adding
support for Coroutines.

Amplify has several types of APIs. Some are synchronous calls, which
immediately return a value. However, most Amplify behaviors are
asynchronous calls. Among the async calls, there are a few broad
categories:

 1. Functions that return a single value, and cannot be canceled;
 2. Functions that return a single value, and can be canceled;
 3. Functions that emit a stream of values
 4. Functions that emit multiple types of values

### Single-valued functions

Most Amplify behaviors are of type (1). Auth and Predictions are
entirely comprised of type (1). The Kotlin flavors of these behaviors
are expressed as suspending functions, e.g.:

```kotlin
suspend fun signOut() {
   ...
}
```

The simplest developer experience to invoke this method will be:
```kotlin
activityScope.launch {
    Amplify.Auth.signOut()
}
```

### Cancelable, single-valued functions
Some single-valued functions, such as the `mutate(...)` behavior in the
API category, may be canceled before rendering a result. The Kotlin
version of `mutate(...)` is expressed this way:

```kotlin
suspend fun <R> ApiCategory.mutate(
        request: GraphQLRequest<R>, apiName: String? = null)
        : GraphQLResponse<R> {
    ...
}
```

The user may cancel the behavior via Kotlin's `Job` construct:
```kotlin
val job = activityScope.launch {
    Amplify.API.mutate(request)
}
...
job.cancel()
```

### Multi-valued functions

There are also some Amplify behaviors which emit a stream of values.
DataStore's `observe()` is a canonical example. The Kotlin facade
function is:

```kotlin
suspend fun observe(): Flow<DataStoreItemChange<out Model>> {
    ...
}
```

A developer may interact with the flow in this way:
```kotlin
activityScope.launch {
    Amplify.DataStore.observe()
        .collect { print(it) }
}
```

Note that `observe()` itself is a suspending function. This function
suspends until the observation is ready to be used, e.g.:
```kotlin
val changes: Flow<DataStoreItemChange<out Model>> =
    runBlocking { Amplify.DataStore.observe() }
print("Ready to see DataStore changes!")
activityScope.launch {
    dataStoreChanges.collect { print("Found one: $it") }
}
```

### Mixed-valued functions
API's `subscribe()` also emits a stream of values. Like DataStore, there is
some non-zero cost associated with establishing a subscription. So
`subscribe()` also suspends until it establishes a connection.

```kotlin
suspend fun <T> ApiCategory.subscribe(
        request: GraphQLRequest<T>, apiName: String? = null)
        : Flow<GraphQLResponse<T>> {
    ...
}
```

A developer can access the flow:
```kotlin
val events = runBlocking { Amplify.API.subscribe(request) }
print("Ready to see subscription events!")
activityScope.launch {
    events.collect { print("Found one: $it") }
}
```

The Storage category's various upload and download functions provide
two interesting pieces of information. We want obtain the result of the
transfer, but we may also like to observe a stream of progress updates.
The signature looks like:
```kotlin
fun StorageCategory.downloadFile(
        key: String,
        local: File,
): InProgressStorageOperation<StorageDownloadFileResult> {
```

A developer can access the result of the download via a suspend
function exposed on the download operation:
```kotlin
val result = runBlocking { download.result() }
```
Or get periodic progress updates by doing:
```kotlin
activityScope.async {
    download.progress().collect {
        print("Download made some progress! $it")
    }
}
```

Refer: https://github.com/aws-amplify/amplify-android/issues/605


jamesonwilliams:
cc @brady-aiello

kotlin-extensions/src/test/java/com/amplifyframework/ktx/DataStoreCategoryExtTest.kt
```diff
@@ -0,0 +1,99 @@
...
+        Amplify.addPlugin(mockPlugin)
+        val config = AmplifyConfiguration.fromJson(JSONObject())
+        Amplify.configure(config, getApplicationContext())
+    }
+
+    @Test
+    fun highlightDeveloperExperience() {
+        runBlocking {
+            Amplify.DataStore.save(blogOwner)
+            val results = Amplify.DataStore.query(BlogOwner::class.java)
```
jamesonwilliams, line 95: We're going to have to start generating Kotlin models, too. We'll need to add a "which language?" prompt to the CLI.

aajtodd: What makes you say we will "have" to? Not that I necessarily disagree, just curious

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+@file:Suppress("unused")
+
+package com.amplifyframework.ktx // TODO: or .datastore?
```
aajtodd, line 18: Yeah I don't love `ktx` for some reason.

You could just use the original package name since these are extensions to those models or perhaps `coroutines`. 

All just bike shedding though.

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
```diff
@@ -0,0 +1,157 @@
...
+    return suspendCoroutine { continuation ->
+        delete(item, predicate,
+            { continuation.resume(Unit) },
+            { continuation.resumeWithException(it) }
+        )
+    }
+}
+
+@Throws(DataStoreException::class)
+suspend fun <T: Model> DataStoreCategory.query(clazz: Class<T>): Iterator<T> {
```
aajtodd, line 78: You can also do something like:

```kt
suspend inline fun <reified T: Model> DataStoreCategory.query(): Iterator<T> {
    return suspendCoroutine { continuation ->
        query(
            T::class.java,
            { continuation.resume(it) },
            { continuation.resumeWithException(it) }
        )
    }
}
```

Which allows callers to do:
```
val results: Iterator<MyItem> = store.query()
// OR
val results = store.query<MyItem>()
```


See: https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
```diff
@@ -0,0 +1,157 @@
...
+        query(
+            clazz,
+            options,
+            { continuation.resume(it) },
+            { continuation.resumeWithException(it) }
+        )
+    }
+}
+
+// TOOD: Flow is from kotlinx, which requires another dependency,
```
aajtodd, line 112: This is fine. If you're using coroutines you're already depending on the `kotlinx-coroutines-core`

aajtodd (commented):
Overall looks good. Seems like it's reasonably easy to adapt the API to coroutines.

kotlin-extensions/src/test/java/com/amplifyframework/ktx/DataStoreCategoryExtTest.kt
jamesonwilliams: "Have" was the wrong word. "Should," I mean. I think they'd make more sense as data classes.

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
jamesonwilliams: ^ As I've thought about this more, I don't think it makes sense to make a separate library for this work. I can add these extensions directly to our `core` Java library. Our existing APIs will continue to work, but people who are using Kotlin will automatically get the goodness.

aajtodd: The only downside to that is the additional dependency/dependencies introduced (I'm guessing that's `coroutines` and `kotlin-stdlib`?).

jamesonwilliams: ^ I keep going back and forth on this. The current version of the PR keeps the extensions as a separate module, mostly because of the new dependencies. The Android platform also tends to segregate out its Kotlin extension libraries. So, that's probably a good way to go for cultural consistency.

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
jamesonwilliams: In the current version of the PR, I have updated `query()` to return a `Flow<T>`.

I do still have the `Class<T>` as an argument. My concern is that `storage.query()` isn't explicit enough about the user's intention, since the behavior can change depending on the _implicit_ type.

This would be a good topic to discuss in design review.

kotlin-extensions/src/main/java/com/amplifyframework/ktx/DataStoreCategoryExt.kt
jamesonwilliams: I ended up changing this to `.kotlin`.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        override fun cancel() {
+            delegate?.cancel()
+        }
+
+        @Suppress("UNCHECKED_CAST")
+        suspend fun result(): T {
+            // We want to take the first item from either one,
+            // without waiting for the other.
+            // Maybe there's a cleaner way to achieve this.
```
richardmcclellan, line 107: Instead of having two `Flow`'s, (`results` and `errors`), could you just have `results`, and in the facade implementation emit success, as an item, or failure, as an error, directly to `results`?

Then, make `result` public, and you don't need this method at all any more.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
```diff
@@ -0,0 +1,118 @@
...
+    @Throws(StorageException::class)
+    suspend fun list(
+        path: String,
+        options: StorageListOptions = StorageListOptions.defaultInstance()
+    ): StorageListResult
+
+    @FlowPreview
+    data class InProgressStorageOperation<T>(
+        private val results: Flow<T>,
+        private val progress: Flow<StorageTransferProgress>,
```
richardmcclellan, line 94: Make this public, so customers can access it?  I don't see a way for customers to access progress, currently.
```suggestion
       val progress: Flow<StorageTransferProgress>,
```

richardmcclellan (commented):
Great work @jamesonwilliams, this is really cool!   I'm excited to start using this in my own sample apps!  

### Threading Question

One issue that I've noticed with Amplify in general is that callbacks are not run on the same thread that initiated them.  For example, the following code crashes because the callback is not run on the main thread:
```
// from main thread
Amplify.API.query(request, { results ->
    this.items = results.iterator().asSequence().toList()
    listAdapter.notifyDataSetChanged() // crash because can't update UI from bg thread
},  { })
```

To solve this, customers must use `runOnUiThread` before updating their UI with the results. 
 In my opinion, we should change this behavior to have all callbacks run on the same thread that initiated the call, to make it easier for customers.

The reason I'm bring it up on this PR though is - how does this affect coroutines?  If the callbacks are run on a different thread, will the suspend functions automatically figure it out and return the user to the calling thread?  Or will they be on a different thread and need to use `runOnUiThread` before?

### Conventional Commit
Can you make sure the commit starts with `feat:` before doing squash and merge?   That way, our release script will pick it up and know we should do a minor version bump.

aajtodd:
> ### Threading Question
> 
> One issue that I've noticed with Amplify in general is that callbacks are not run on the same thread that initiated them. For example, the following code crashes because the callback is not run on the main thread:
> 
> ```
> // from main thread
> Amplify.API.query(request, { results ->
>     this.items = results.iterator().asSequence().toList()
>     listAdapter.notifyDataSetChanged() // crash because can't update UI from bg thread
> },  { })
> ```
> 
> To solve this, customers must use `runOnUiThread` before updating their UI with the results.
> In my opinion, we should change this behavior to have all callbacks run on the same thread that initiated the call, to make it easier for customers.
> 
> The reason I'm bring it up on this PR though is - how does this affect coroutines? If the callbacks are run on a different thread, will the suspend functions automatically figure it out and return the user to the calling thread? Or will they be on a different thread and need to use `runOnUiThread` before?

Coroutines (without additional input from the user) can pretty much run on whatever thread they want. In fact it may switch threads. Of course this depends on the context they are running in (and the dispatcher used in particular). 

In Kotlin you would do:
```
with(Dispatchers.Main) {
    // update ui
}
```
See [Dispatchers.Main](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html)

There is also some documentation on this [here](https://developer.android.com/kotlin/coroutines#use-coroutines-for-main-safety) as well (apparently they also define a `viewModelScope`)

core-ktx/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+dependencies {
+    implementation dependency.kotlin.stdlib
+    implementation dependency.kotlin.coroutines
+    implementation dependency.androidx.core_ktx
+    implementation project(':core')
+
+    testImplementation dependency.androidx.test.core
+    testImplementation dependency.junit
+    testImplementation dependency.mockk
```
rjuliano, line 29: very klever 😆

rjuliano (commented):
Overall I like the direction. I think it's something we should consider getting it out in front of devs (external or ourselves building apps) as an alpha release component.

One comment I have is that it would be interesting to think about whether the generation of some of this facade code can be auto-generated somehow. For example, if an API is added to a category, the corresponding Kotlin interface would be updated. That would at least force us to keep the facades in sync.

Another topic for future consideration is whether it makes sense to create full blown plugin implementations in Kotlin. It may be a good hackathon exercise.

build.gradle
```diff
@@ -63,6 +71,7 @@ ext {
             annotation: 'androidx.annotation:annotation:1.1.0',
             appcompat: 'androidx.appcompat:appcompat:1.2.0',
             core: 'androidx.core:core:1.3.2',
+            core_ktx: 'androidx.core:core-ktx:1.3.2',
```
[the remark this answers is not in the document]
jamesonwilliams: @aajtodd Also had this same feedback. And particularly now that this PR _doesn't even use_ **K**o**T**lin E**x**tensions, ktx probably doesn't make sense. Resolution: I'll update this.

core-ktx/src/main/java/com/amplifyframework/kotlin/Hub.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.amplifyframework.hub.HubEventFilters
+import kotlinx.coroutines.flow.Flow
+
+interface Hub {
+    /**
+     * Publish an event to a particular channel of the Hub.
+     * @param channel Hub channel onto which to publish
+     * @param event Event to publish
+     */
+    suspend fun publish(channel: HubChannel, event: HubEvent<*>)
```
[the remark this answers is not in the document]
jamesonwilliams: Yes! Good catch, thanks.

core-ktx/src/main/java/com/amplifyframework/kotlin/Amplify.kt
```diff
@@ -0,0 +1,106 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.kotlin
```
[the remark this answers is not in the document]
jamesonwilliams: `.api`, `.datastore`, etc. _are_ in the core module, in the existing codebase.

For this PR, I've arranged all of the new API specifications into the same package, com.amplifyframework.kotlin`. So, for example, core uses `com.amplifyframework.api` for API contracts in core, but this package uses `com.amplifyframework.kotlin` instead.

If I do update the package names, I think I'd want to do like:
1. com.amplifyframework.core.kotlin.Amplify
2. com.amplifyframework.auth.kotlin.Auth
3. com.amplifyframework.datastore.kotlin.DataStore
4. etc.

Personally I'm not sure if its an improvement. What do you think?

core-ktx/src/main/java/com/amplifyframework/kotlin/Auth.kt
```diff
@@ -0,0 +1,269 @@
...
+     *                If not provided, default options will be used
+     * @return A sign-in result. The nextStep field may indicate additional actions to be taken
+     *         to confirm the sign-in, or it may show isSignInComplete as true, in which case
+     *         an authenticated session is available
+     */
+    @Throws(AuthException::class)
+    suspend fun signIn(
+        username: String? = null,
+        password: String? = null,
+        options: AuthSignInOptions? = null
```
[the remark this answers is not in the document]
jamesonwilliams: Yes! Good eye. Most of the other APIs in this PR have some kind of default object that I can use as a default parameter.

[`AuthSignInOptions` is an abstract class](https://github.com/aws-amplify/amplify-android/blob/72b5820ff217f5ca6cde715aeb2ece523f79b274/core/src/main/java/com/amplifyframework/auth/options/AuthSignInOptions.java#L24). As is right now, I can't instantiate it, so I sued `null` instead.

One alternative would be to add a factory method to that abstract class, like:
```kotlin
AuthSignInOptions.defaults()
```
It would create a new sentinel object, `DefaultAuthSignInOptions`. We could look for that type inside our implementation and treat it specially.

On second thought, this does seem like the better approach. I can update the code base to add this mechanism, if it sounds good to you.

core-ktx/src/main/java/com/amplifyframework/kotlin/Auth.kt
```diff
@@ -0,0 +1,269 @@
...
+        callingActivity: Activity,
+        options: AuthWebUISignInOptions = AuthWebUISignInOptions.builder().build()
+    ):
+        AuthSignInResult
+
+    /**
+     * Handles the response which comes back from {@link #signInWithWebUI(Activity, Consumer, Consumer)}.
+     * @param intent The app activity's intent
+     */
+    @Throws(AuthException::class)
```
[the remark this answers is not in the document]
jamesonwilliams: It doesn't, you're right -- good catch. Will remove.

core-ktx/src/main/java/com/amplifyframework/kotlin/DataStore.kt
```diff
@@ -0,0 +1,112 @@
...
+ */
+interface DataStore {
+    /**
+     * Save an item into the DataStore.
+     * @param item Item to save
+     * @param predicate Conditions that must be true before save can succeed.
+     *                  If not specified, an "allow all" predicate is used.
+     */
+    @Throws(DataStoreException::class)
+    suspend fun <T : Model> save(item: T, predicate: QueryPredicate = QueryPredicates.all())
```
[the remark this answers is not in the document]
jamesonwilliams: In more recent versions of the Amplify API, we have decided _not_ to include this change param on `save` or `delete`. iOS does not include anything like this. Rx bindings returns a value-less `Completable`, here. Flutter returns a `Future<void>`. Even though this is technically _more than just wrapping the low-level API_, I'm inclined to drop the `DataStoreItemChange` to be more like iOS & Flutter.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
jamesonwilliams: Whoops! I actually have another commit, where I go back and update this mechanism. Yes, you're absolutely right. This current PR doesn't expose any ability for the customer to view progress. Will update.

core-ktx/src/main/java/com/amplifyframework/kotlin/facades/KotlinStorageFacade.kt
```diff
@@ -0,0 +1,157 @@
...
+                operation as Cancelable
+            )
+        }
+
+    @ExperimentalCoroutinesApi
+    @FlowPreview
+    override fun uploadFile(key: String, local: File, options: StorageUploadFileOptions):
+        InProgressStorageOperation<StorageUploadFileResult> {
+            val progress = MutableSharedFlow<StorageTransferProgress>(replay = 0)
+            val results = MutableSharedFlow<StorageUploadFileResult>(replay = 1)
```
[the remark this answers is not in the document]
jamesonwilliams: A "cold" flow is initiated by the collector, whereas a "hot" flow has its own lifecycle that the collector can't change.

In our code base, these are (intuitively) cold:
1. DataStore query

And there are hot:
1. DataStore observe (lifecycle calls on someone else calling save/delete -- or subscription event)
2. Hub subscribe (lifecycle depends on someone else calling publish)
3. Download progress events (lifecycle depends on the network)
4. API subscribe (lifecycle depends on external forces.)

`MutableSharedFlow` and `MutableSharedState` are the current generation of utilities for hot flows in Kotlin. Even the `ConflatedBroadcastChannel` is marked obsolete and shouldn't be used.

I'm publish an update to this PR which makes everything hot.

-----------

Second matter: this is something Tim and I deliberated while working on the Rx Bindings & Combine support. Both of those API surfaces do work this way. 

One reasonable alternative is to use a single flow which emits different types of events on the same flow. Then, the use could initiate the thing at `collect` time. We could use a Kotlin `sealed class` to build out the various events on the single flow. I believe this gets to your other point about the `result()` in `Storage.kt`.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
jamesonwilliams: The Rx Bindings and Combine APIs work this way, today. The way you're suggesting is also a very reasonable mechanism. We can go the route you're suggesting, I'm open to it! It will mean changing all of our `Flow` APIs in this PR to use a single `Flow`, emitting `sealed class`es. This will ultimately work differently from any other API in iOS, Flutter, or Android -- but it might be the most ergonomic for the user?

If I understand you correctly, you're suggesting we do something like this?
```kotlin
Amplify.Storage.download(remoteKey, localFile)
    .collect {
        when (it) {
            is ProgressUpdate -> {
            },
            is TransferComplete -> {
            },
            is Error -> {
            }
    }

sealed class StorageEvent {
    data class ProgressUpdate(val increment, val total): StorageEvent
    data class TransferComplete<T>(result: T): StorageEvent
    data class Error(cause: StorageException): StorageEvent
}
```

jamesonwilliams:
> getting it out in front of devs (external or ourselves building apps) as an alpha release component.

Agreed! I think we can give it a little bake time before putting a 1.x on it.

> the generation of some of this facade code can be auto-generated somehow. [...] That would at least force us to keep the facades in sync.

Absolutely. There's a ton of duplication here. We could look at future work to cut out all of the `suspendCoroutine`-based functions by using a [Kotlin Poet preprocessor](https://square.github.io/kotlinpoet/), perhaps.

> Another topic for future consideration is whether it makes sense to create full blown plugin implementations in Kotlin. It may be a good hackathon exercise.

100%, in my view. Any new plugin we do should be pure Kotlin.

jamesonwilliams:
> To solve this, customers must use `runOnUiThread` before updating their UI with the results.

As @aajtodd, I believe the coroutine-y way of dealing with this would be to write:
```kotlin
activityScope.launch(Dispatchers.IO) {
    Amplify.DataStore.query(Post::class.java)
        .collect { aPost ->
            withContext(Dispatchers.Main) {
                dataSet.add(aPost)
                listAdapter.notifyDataSetChanged()
            }
        }
}
```
> In my opinion, we should change this behavior to have all callbacks run on the same thread that initiated the call, to make it easier for customers.

It could be a good change to explore. We'd of course have to make a larger update to our current set of APIs, too.

> Can you make sure the commit starts with `feat:` before doing squash and merge? That way, our release script will pick it up and know we should do a minor version bump.

Yessir! Will add this before the final merge.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
rjuliano: I haven't done much with `Flow`s so take this with a grain of salt. It's mostly academic.

With regards to errors, one approach could be to allow customers to use the [catch](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/catch.html) operator of the `Flow`. Unless we're doing any other processing of that error internally, I don't know that it makes sense to emit a special value for errors. So I think you would end up with something like this:

```kotlin
Amplify.Storage.download(remoteKey, localFile)
    .collect {
        when (it) {
            is ProgressUpdate -> {
            },
            is TransferComplete -> {
            }
    }
    .catch { storageExceptioon -> /* do whatever */ }
```

For progress updates, the sealed class approach seems solid. Only other approach I could think of, would be to make `InProgressStorageOperation<T>` a child class `MutableSharedFlow` and internally have the separate flow for progress updates. Invocations might look something like this:

```kotlin
Amplify.Storage.download(remoteKey, localFile)
    .collect { }
    .onProgress { }
    .catch { storageExceptioon -> /* do whatever */ }
```

I do understand the concern with keeping it consistent with the other platforms, but we should also keep in mind what approach feels more idiomatic for each of the dev stacks.

jamesonwilliams: FWIW, the current code should facilitate use of `.catch { }` as you mention. DataStore query & observe and API subscribe will currently emit errors like this. Storage upload/download will throw the error from the `result()` suspending function (it's like a single valued `Flow`, I suppose), but the `progress()` Flow never throws.

rjuliano: Ah...got itt. So they would do something like this, correct?

```
Amplify.Storage.download(remoteKey, localFile)
    .results
    .collect {
        when (it) {
            is ProgressUpdate -> {
            },
            is TransferComplete -> {
            }
    }
    .catch { storageExceptioon -> /* do whatever */ }
```

core-ktx/src/main/java/com/amplifyframework/kotlin/Amplify.kt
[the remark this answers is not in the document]
richardmcclellan: Thanks for your detailed explanation!  

This PR adds a `core-ktx` module, which provides Kotlin extensions for the `core` module.  In the future, we might want to add a `aws-datastore-ktx` module, which provides Kotlin extensions for the `aws-datastore` module.  For example, we might want let customers provide a coroutine that returns a `QueryPredicate`, when configuring sync expressions for selective sync in their DataStoreConfiguration.  Basically I just want to make sure we have a logical path to support more Kotlin extensions in the future.

Putting all Kotlin extension interfaces in the same package (`com.amplifyframework.kotlin`) seems like an anti-pattern though, for all the same reasons that putting your entire project in a single package would be a bad idea (name collisions, limiting scope, etc).

After you laid it out above though, I don't like my original suggestion to insert `.kotlin` as the 4th component of the package.  What if, instead, we insert it as the 3rd component, like this?

| `core` interface | `core-ktx` interface |
|----|-----|
| `com.amplifyframework.core.Amplify` | `com.amplifyframework.kotlin.core.Amplify` |
| `com.amplifyframework.auth.AuthCategoryBehavior` | `com.amplifyframework.kotlin.auth.Auth` |
| `com.amplifyframework.datastore.DataStoreCategoryBehavior` | `com.amplifyframework.kotlin.datastore.DataStore` |
| etc... | etc... |

Then, there is no ambiguity around what the package name for a Kotlin extension is, given the package of the corresponding Java interface.

> If a future all-Kotlin plugin is created, I doubt it would use `kotlin` in its namespace, either.

I agree.   My suggestion above only applies to these Kotlin extensions where we are providing a layer on top of the Java interfaces.

build.gradle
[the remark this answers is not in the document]
richardmcclellan: Ahh, right!  This doesn't use [Kotlin extensions](https://kotlinlang.org/docs/extensions.html), so yeah, we probably shouldn't use `ktx`.

core-ktx/src/main/java/com/amplifyframework/kotlin/Auth.kt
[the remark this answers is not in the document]
richardmcclellan: > I can update the code base to add this mechanism, if it sounds good to you.

Sounds reasonable.  I think the current approach is fine too.  Either way seems fine to me!

core-ktx/src/main/java/com/amplifyframework/kotlin/DataStore.kt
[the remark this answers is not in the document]
richardmcclellan: Got it!  Given that iOS/Flutter and our Rx bindings don't return a `DataStoreItemChange`, I agree we don't need to return it here.

core-ktx/src/main/java/com/amplifyframework/kotlin/Storage.kt
jamesonwilliams: The download result is the odd-one-out. It works like a regular suspending function, e.g.:
```kotlin
val download = Amplify.Storage.downloadFile(remoteKey, localFile)
try {
    val result = download.result()
} catch (error: StorageException) {
    Log.w("DemoApp", "Failed to complete download", error)
}
```

The API subscribe has two `Flow`s in the operation. It *does* work the way you suggested:
```kotlin
Amplify.API.subscribe(request)
    .subscriptionData
    .catch { Log.w("DemoApp", "Error ended subscription", it) }
    .collect { Log.i("DemoApp", "New subscription event! $it") }
```

core-ktx/src/main/java/com/amplifyframework/kotlin/Amplify.kt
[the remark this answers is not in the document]
jamesonwilliams: We chatted about this offline and agree to go with the `com.amplifyframework.kotlin.<whatever_was_in_core>` approach.

fatinWasta:
Hello, I am trying to consume this inside my Kotlin Multiplatform Mobile project so that I can create kinda my own aplify layer which will talk to android and iOS. But when I make project it's throwing error as `Could not find com.amplifyframework:core-kotlin:1.16.13.`
Have attached screenshot for more details. Can anyone please help on this?
PS:I am new at Kotlin.
Screenshot 2021-02-18 at 5 54 20 PM

jamesonwilliams:
> Hello, I am trying to consume this inside my Kotlin Multiplatform Mobile project so that I can create kinda my own[...]

Hi! You must be [Fatin from StackOverflow](https://stackoverflow.com/q/66218993/695787). Come [join us on the Amplify Discord chat server](https://discord.gg/amplify) and we can support your Kotlin Multiproject Mobile (KMM) investigation. Let's avoid a discussion on this GitHub Pull Request.

core-ktx/src/main/java/com/amplifyframework/kotlin/DataStore.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Delete an item from the DataStore.
+     * @param item Item to delete
+     * @param predicate Conditions that must be true before delete can succeed.
+     *                  If not specified, an "allow all" predicate is used.
+     */
+    @Throws(DataStoreException::class)
+    suspend fun <T : Model> delete(item: T, predicate: QueryPredicate = QueryPredicates.all())
```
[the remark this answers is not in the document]
jamesonwilliams: (Same here as in `save(...)`: we decided to be more like iOS/Flutter and to drop the result value.)

core-ktx/src/main/java/com/amplifyframework/kotlin/DataStore.kt
```diff
@@ -0,0 +1,112 @@
...
+
```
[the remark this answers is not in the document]
jamesonwilliams: Yes! Good catch. Will add now.

core-ktx/src/main/java/com/amplifyframework/kotlin/facades/KotlinStorageFacade.kt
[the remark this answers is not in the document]
richardmcclellan: In this PR's current state, these are cold:
 - DataStore.query
 - DataStore.observe
 - Hub.subscribe

and these are hot:
 - API.subscribe
 - Storage progress

Can you clarify what your plan is on this?  I think you wanted to leave this PR as is, but maybe open another PR that makes DataStore.observe and Hub.subscribe hot to A/B against?

core-ktx/src/main/java/com/amplifyframework/kotlin/DataStore.kt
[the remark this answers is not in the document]
jamesonwilliams: On closer inspection, that new `delete` API in the Java API surface isn't ready for prime time. It can delete many things, but it has a return type for a single deletion. I'm inclined to not include it in the Kotlin API surface right now, since it will need some rework. https://github.com/aws-amplify/amplify-android/pull/1106

core-kotlin/src/test/java/com/amplifyframework/kotlin/datastore/KotlinDataStoreFacadeTest.kt
```diff
@@ -100,11 +100,11 @@ class KotlinDataStoreFacadeTest {
     }
 
     /**
-     * When delete() coroutine is called, it passes through to the delegate.
-     * When delegate succeeds, so too does the coroutine API.
+     * When item-based delete() coroutine is called, it passes through to
+     * the delegate When delegate succeeds, so too does the coroutine API.
```
richardmcclellan, line 104:
```suggestion
     * the delegate.  When delegate succeeds, so too does the coroutine API.
```

core-ktx/src/main/java/com/amplifyframework/kotlin/facades/KotlinDataStoreFacade.kt
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+
+    @ExperimentalCoroutinesApi
+    @Throws(DataStoreException::class)
+    override fun observe(): Flow<DataStoreItemChange<out Model>> {
+        return callbackFlow {
+            val cancelable = AtomicReference<Cancelable?>()
+            delegate.observe(
+                { cancelable.set(it) },
```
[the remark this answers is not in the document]
jamesonwilliams: Ah, yea. Actually I just prototyped this, and __it is awesome__.

With Rx, It would have been like a `Single<Observable<DataStoreItemChange<T>>`, and it would be kind of ugly to deal with.

But since coroutines get rid of the await-style boiler plate, I implemented your solution and literally didn't have to change my unit tests. Everything was already getting launched in a coroutine scope and it just worked.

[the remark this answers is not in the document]
richardmcclellan: Awesome!  Would a similar approach work for `API.subscribe` too?

core-ktx/src/main/java/com/amplifyframework/kotlin/GraphQL.kt
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Subscribe to realtime events observed on a GraphQL API.
+     * @param request Subscription request
+     * @param apiName The name of an API as configured in your configuration file;
+     *                if not provided, the first GraphQL API in your config is used
+     * @return A subscription object. Inspect its connectionState and subscriptionData.
+     */
+    @ExperimentalCoroutinesApi
+    @FlowPreview
+    fun <T> subscribe(request: GraphQLRequest<T>, apiName: String? = null): Subscription<T>
```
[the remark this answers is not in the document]
jamesonwilliams: This one worked out, too. Done.

core-ktx/src/main/java/com/amplifyframework/kotlin/facades/KotlinStorageFacade.kt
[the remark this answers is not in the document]
jamesonwilliams: I had to read and think more about this, I don't think I got this completely right.

It's true that we don't control the timing of those external forces (network events, Hub publications, DataStore changes).

However, here's how the `callbackFlow` defines "cold":
> The resulting flow is cold, which means that block is called every time a terminal operator is applied to the resulting flow.

All this means is that we start watching what's going on at the time the user calls `collect`. That does seem like a desirable behavior. What's more, the `SharedFlow`/`StateFlow` have some annoying properties, like not being able to explicitly close them.

I think the lifecycle we care about is the setup/teardown lifecycle -- not the timing of the emissions.

core-kotlin/src/main/java/com/amplifyframework/kotlin/datastore/DataStore.kt
```diff
@@ -47,6 +48,18 @@ interface DataStore {
...
 
+    /**
+     * Delete item(s) of a given class from the DataStore.
+     * @param byClass The class of item(s) being deleted
+     * @param filter Items must additionally match this filter, to be targeted for deletion.
+     *               If no filter is specified, an "allow all" predicate is used.
+     */
+    @Throws(DataStoreException::class)
+    suspend fun <T : Model> delete(
+        byClass: KClass<T>,
```
richardmcclellan, line 59: Maybe name this variable `itemClass`, for consistency with the `query` method below?

core-kotlin/src/main/java/com/amplifyframework/kotlin/datastore/DataStore.kt
```diff
@@ -73,33 +73,45 @@ interface DataStore {
...
      * @param itemClass The class of the model being observed
      * @param uniqueId The ID of the item being observed
      * @return A flow of changes to the requested model
      */
     @ExperimentalCoroutinesApi
     @Throws(DataStoreException::class)
-    fun <T : Model> observe(itemClass: Class<T>, uniqueId: String): Flow<DataStoreItemChange<T>>
+    suspend fun <T : Model> observe(
+        itemClass: Class<T>,
+        uniqueId: String
```
richardmcclellan, line 98: Maybe call this `itemId`?  Just a thought, though I don't feel strongly one way or the other.

```suggestion
        itemId: String
```

core-kotlin/src/main/java/com/amplifyframework/kotlin/datastore/KotlinDataStoreFacade.kt
```diff
@@ -170,4 +176,38 @@ class KotlinDataStoreFacade(private val delegate: Delegate = Amplify.DataStore)
...
+
+    internal class Observation<T>(
+        internal val starts: MutableSharedFlow<Cancelable> = MutableSharedFlow(1),
+        internal val changes: MutableSharedFlow<T> = MutableSharedFlow(1),
+        internal val failures: MutableSharedFlow<DataStoreException> = MutableSharedFlow(1),
+        internal val completions: MutableSharedFlow<Unit> = MutableSharedFlow(1)
+    ) {
+        @Suppress("UNCHECKED_CAST")
+        @FlowPreview
+        internal suspend fun waitForStart(): Flow<T> {
```
richardmcclellan, line 188: This is super awesome! 😎

jamesonwilliams: Thanks! Appreciate you Insisting on High Standards here. With our back and forth on these `Flow` APIs, I do feel like we ultimately ended up with a better result for the customer.

core-kotlin/src/main/java/com/amplifyframework/kotlin/datastore/DataStore.kt
jamesonwilliams: Will do! Update coming ...

core-kotlin/src/main/java/com/amplifyframework/kotlin/datastore/DataStore.kt
jamesonwilliams: Done!

richardmcclellan (approved):
Great work!  I'm excited to see this shipping!

jamesonwilliams:
> Great work! I'm excited to see this shipping!

Thanks @richardmcclellan! Really appreciate all the time you put into your thoughtful review of this work.

rjuliano (approved):
This is awesome. Let's get this merged! 🚢

Issue #1013, opened 2020-12-01
Deleting a parent in datastore doesn't delete the children from remote, only locally


Amplifiyer:
In DataStore category, if you have a connected model such as `Blog` and `Post` where a `Post` belongs to a `Blog` and a `Blog` has several `Post`s if a `Blog` instance is deleted, all the children `Post` instances are deleted in the local datastore, however these children remove Mutations are not sent to appsync. 

Since appsync itself doesn’t reinforce foreign key constraints, the orphaned items (Posts) will stay in the server and when a new device tries to sync it will result in a Foreign Key Constraint error while saving these orphaned `Post`s since their `BlogId` doesn't exist in the `Blog` table. See the exception below

```
[...]

PR #1058, opened 2020-12-17
fix(aws-datastore): ignore foreign key error during sync

raphkim:
*Issue #, if available:*
Partially addresses #1013 

By ignoring foreign key constraint violation during sync, #1013 issue no longer crashes DataStore.

Even with the best effort to ensure that the foreign key constraints are met on the client side, it is possible that the remote storage simply does not enforce foreign key violation. This guarantees the possibility of encountering a foreign key constraint violation during sync to be non-zero (e.g. cascading deletion failed halfway while publishing; a race condition arose from multi-device operation; or an entry was inserted directly to remote store). There are two options to deal with a failure during initial sync:

1. Issue a DELETE mutation on the orphaned remote model that caused the crash.
2. Ignore the error.

The first option is not ideal because this directly assumes that the orphaned entry in remote store is invalid, which may not always be true. DataStore client should _never_ delete customer data unless the DELETE operation was directly issued.

The second option is more feasible because it leaves the data safely in cloud while avoiding a fatal crash. Since it does not directly affect customer data, we do not need to provide an explicit error handler to the customer.

*Description of changes:*
Checks if sync failure was caused by SQLite constraint violation, and if so log the details and continue without crashing.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
@@ -28,10 +28,10 @@ import com.amplifyframework.datastore.appsync.ModelMetadata;
 import com.amplifyframework.datastore.appsync.ModelWithMetadata;
 import com.amplifyframework.datastore.storage.LocalStorageAdapter;
 import com.amplifyframework.datastore.storage.StorageItemChange;
+import com.amplifyframework.datastore.utils.ErrorInspector;
 import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.hub.HubEvent;
 import com.amplifyframework.logging.Logger;
-import com.amplifyframework.util.Error;
 
 import java.util.Objects;
 
@@ -114,7 +114,7 @@ final class Merger {
             // Remote store may not always respect the foreign key constraint, so
             // swallow any error caused by foreign key constraint violation.
             .onErrorComplete(failure -> {
-                if (!Error.containsCause(failure, SQLiteConstraintException.class)) {
+                if (!ErrorInspector.contains(failure, SQLiteConstraintException.class)) {
                     return false;
                 }
                 LOG.warn("Failed to sync due to foreign key constraint violation: " + modelWithMetadata, failure);
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
@@ -15,6 +15,7 @@
 
 package com.amplifyframework.datastore.syncengine;
 
+import android.database.sqlite.SQLiteConstraintException;
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.core.Amplify;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/amplifyframework/util/Error.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.util;
```
jamesonwilliams, line 16: I think we should avoid putting stuff in the core module unless it'l be used widely. Even though this is a generic utility, it might be better suited to the DataStore module, where it won't add size to the core artifact (if, for example, someone uses Auth, without DataStore.)

core/src/main/java/com/amplifyframework/util/Error.java
```diff
@@ -0,0 +1,51 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.Objects;
+
+/**
+ * Utility class to provide helpful functions to use on throwable instances.
+ */
+public final class Error {
```
richardmcclellan, line 26: I would expect a class named `Error` to be instantiable, and to represent a particular error.  What about calling this something like `ErrorInspector`, or `ThrowableInspector`, or `ErrorCause` (and rename `containsCause` to just `contains`)?   Just a thought.  I think I like `ErrorInspector` the best, though I don't feel too strongly.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -109,6 +111,15 @@
                 announceSuccessfulMerge(modelWithMetadata);
```
richardmcclellan, line 111: Swallowing `SQLiteConstraintException` errors makes sense, but I think we also need to make sure that we do NOT announce a successful merge.

I think you could do this by doing the `announceSuccessfulMerge` with `andThen`, instead of `doOnComplete`.

It could even make sense to broadcast a new `SUBSCRIPTION_DATA_FAILED` event if a `SQLiteConstraintException` is thrown, but I don't think we need that just yet.

core/src/main/java/com/amplifyframework/util/Error.java
raphkim: I too was uncomfortable with the name `Error` but I didn't want to call it `ErrorUtils` because that's just lame haha. I'll go with `ErrorInspector` as you suggested.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
raphkim: order matters! since `Completable` has already encountered error, it has moved onto the `Completable` that was _returned_ by `doOnComplete()`. At this stage, the `SQLiteConstraintException` is caught by `.onErrorComplete()`, which means that `announceSuccessfulMerge()` will not be invoked.

richardmcclellan: Thanks for double checking! Rx is hard :)

richardmcclellan (approved):

PR #1059, opened 2020-12-17
fix(aws-datastore): publish each cascading delete

raphkim:
*Issue #, if available:*
#1013

*Description of changes:*
* Identify every descendant of a deleted item to publish its deletion to remote store.
* Add `MatchNoneQueryPredicate` to act as `not(MatchAllQueryPredicate)`


jamesonwilliams (commented):
`SQLiteStorageAdapter` is already quite long. We should be sensitive to this and aim to _shorten_ it. In this PR, there's a bunch of new logic coming into it. I suspect, though, that you can factor this logic into some additional classes to formulate the family tree, iterator over it, etc. Perhaps there is a data structure here, `FamilyTree`, in which you can encapsulate some of the new operations you've added, which are prerequisite to the eventual deletions?

raphkim:
i agree! `SQLiteStorageAdapter` is quite a mess at the moment and I certainly want to refactor it just the way you describe it. How about I get the current PRs (#1058 + #1059) + implementation of 1..n deletion in (WIP) merged first and _then_ work on the refactor? I would prefer to have all the code available before i start refactoring, because otherwise things can get even more disorganized.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -941,7 +876,6 @@ private String getModelName(@NonNull Model model) {
     }
 
     @SuppressWarnings("WeakerAccess")
-    @VisibleForTesting
     Cursor getQueryAllCursor(@NonNull String tableName,
```
jamesonwilliams, line 879: Can this be private?
```suggestion
    private Cursor getQueryAllCursor(@NonNull String tableName,
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
```diff
@@ -0,0 +1,140 @@
...
+    private void recurseTree(
+            Map<ModelSchema, Set<String>> tree,
+            ModelSchema modelSchema,
+            Collection<String> parentIds
+    ) {
+        SQLiteTable parentTable = SQLiteTable.fromSchema(modelSchema);
+        for (ModelAssociation association : modelSchema.getAssociations().values()) {
+            switch (association.getName()) {
+                case "HasOne":
+                case "HasMany":
```
jamesonwilliams, line 88: Can we get these relationship strings into an enum somewhere?

Also, the contents of the case statements looks like a good candidate for some smaller helper method(s). That will also help to reduce the nesting.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
```diff
@@ -0,0 +1,140 @@
...
+    private final ModelSchemaRegistry registry;
+    private final SQLiteStorageAdapter storage;
+
+    /**
+     * Constructs a model family tree traversing utility.
+     * @param registry model registry to search schema from
+     * @param storage SQLite storage engine
+     */
+    ModelTreeHelper(ModelSchemaRegistry registry,
+                    SQLiteStorageAdapter storage) {
```
jamesonwilliams, line 53: This `SQLiteStorageAdapter` looks like a bad coupling. The adapter owns a model tree helper, and the model tree helper owns a storage. But in reality, you're only ever using it to access a SQL cursor. So, a step better would be to pass just that, directly.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Utility class to help traverse a tree of models by relationship.
+ */
+final class ModelTreeHelper {
```
jamesonwilliams, line 43: Now that you have this in its own unit, it will be a lot easier to write a unit test!

----------------

`SQLiteModelTree`?
```suggestion
final class SQLiteModelTree {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Returns a map of descendants of a set of models (of same type).
+     * A model is a child of its parent if it uses its parent's ID as foreign key.
+     * @param root Collection of models to query its descendants of.
+     * @return Map of descendants keyed by model schema. The value contains a set of
+     *          descendants' IDs for that model type.
+     */
+    <T extends Model> Map<ModelSchema, Set<String>> descendantsOf(Collection<T> root) {
```
jamesonwilliams, line 65: This type still kind of looks like a data structure in-and-of itself. You could make a `class Descendent`, and then return a `List<Descendent>`, here.
```
    Map<ModelSchema, Set<String>>
```

Any Map other than `Map<PrimitiveA, PrimitiveB>` is usually waiting to be made into a complex type of its own.

raphkim: I still do need to be able to search for the set of IDs by model, which warrants the use of a map. Only way to simplify this further would be to use `Map<PrimitiveA, NewType>` where `NewType` wraps a set, but that would be an overkill. I think a "Map of Collections" is an intuitive enough data structure that it won't need another level of simplification.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
raphkim: Replaced storage adapter with database handle.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ModelTreeHelper.java
raphkim: The content of the case statement _does_ feel long, but I believe it may be even more difficult to follow if they are comparmentalized into helper methods. With how many local variables there are being used in each step, it will become messier to try to move them away from the main body.

raphkim: With regards to turning `ModelAssociation#name` variable to enum, I think it's a good idea, but I'm afraid that it may potentially break some customers that are not relying on codegened models (custom models) in their project.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -503,47 +515,28 @@ public void query(
...
-                            "Please verify the current state of saved item."
-                        );
+                // Use ModelTreeHelper to identify the models affected by cascading delete.
+                Map<ModelSchema, Set<String>> modelFamilyTree =
+                        sqLiteModelTree.descendantsOf(Collections.singleton(item));
+
+                for (ModelSchema schema : modelFamilyTree.keySet()) {
+                    for (String id : modelFamilyTree.get(schema)) {
+                        // Publish DELETE mutation for each affected item.
+                        String dummyJson = String.format("{\"id\":%s}", id);
```
richardmcclellan, line 525: Quotes around %s?

richardmcclellan: Actually, instead of manually writing JSON, create a `Map`, and then use gson to convert it to a `String`, like this:
```suggestion
                        String dummyJson = gson.toJson(Collections.singletonMap("id", id));
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Returns a map of descendants of a set of models (of same type).
+     * A model is a child of its parent if it uses its parent's ID as foreign key.
+     * @param root Collection of models to query its descendants of.
+     * @return Map of descendants keyed by model schema. The value contains a set of
+     *          descendants' IDs for that model type.
+     */
+    <T extends Model> Map<ModelSchema, Set<String>> descendantsOf(Collection<T> root) {
```
richardmcclellan, line 71: Does this method need to take a `Collection<T>`?  It looks like it is only ever taking a Collection of 1 item, so maybe you could just pass a `T` here?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private void recurseTree(
+            Map<ModelSchema, Set<String>> tree,
+            ModelSchema modelSchema,
+            Collection<String> parentIds
+    ) {
+        SQLiteTable parentTable = SQLiteTable.fromSchema(modelSchema);
+        for (ModelAssociation association : modelSchema.getAssociations().values()) {
+            switch (association.getName()) {
+                case "HasOne":
```
richardmcclellan, line 93: Instead of a raw `String` here, could you do `HasOne.class.getSimpleName()`, and same for the cases?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+                case "HasMany":
+                    String childModel = association.getAssociatedType(); // model name
+                    ModelSchema childSchema = registry.getModelSchemaForModelClass(childModel);
+                    SQLiteTable childTable = SQLiteTable.fromSchema(childSchema);
+                    String childPrimaryKey = childTable.getPrimaryKey().getAliasedName();
+                    QueryField queryField = QueryField.field(parentTable.getPrimaryKeyColumnName());
+
+                    // Chain predicates with OR operator.
```
richardmcclellan, line 101: Can you add a test that exercises this predicate building logic?  I think your current test is too simple to verify it.   Maybe add another test that deletes a `Blog`, which has 3 `Post`s, each of which has 3 `Comment`s, and then verify that a total of 13 `StorageItemChange` items are observed?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+     * A model is a child of its parent if it uses its parent's ID as foreign key.
+     * @param root Collection of models to query its descendants of.
+     * @return Map of descendants keyed by model schema. The value contains a set of
+     *          descendants' IDs for that model type.
+     */
+    <T extends Model> Map<ModelSchema, Set<String>> descendantsOf(Collection<T> root) {
+        if (Empty.check(root)) {
+            throw new IllegalArgumentException("Cannot traverse tree from an empty root.");
+        }
+        Map<ModelSchema, Set<String>> descendants = new LinkedHashMap<>();
```
richardmcclellan, line 75: Does this need to be a `LinkedHashMap`, or would a regular `HashMap` work the same way?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+                    Set<String> childrenIds = new HashSet<>();
+                    try (Cursor cursor = queryAll(childModel, options)) {
+                        if (cursor != null && cursor.moveToFirst()) {
+                            int index = cursor.getColumnIndexOrThrow(childPrimaryKey);
+                            do {
+                                childrenIds.add(cursor.getString(index));
+                            } while (cursor.moveToNext());
+                        }
+                    } catch (DataStoreException exception) {
+                        // Don't cut the search short. Populate rest of the tree.
```
richardmcclellan, line 127: Looks like this will only be thrown if `SQLPredicate::parsePredicate` encounters a `QueryPredicate` that it can't parse, which should only happen if there is a bug in the library.  Wouldn't it be better to throw an exception back to the user, so they can report the bug to us, instead of failing silently?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+                    String childPrimaryKey = childTable.getPrimaryKey().getAliasedName();
+                    QueryField queryField = QueryField.field(parentTable.getPrimaryKeyColumnName());
+
+                    // Chain predicates with OR operator.
+                    // No predicate = Match NONE.
+                    // 1 predicate = Match SELF.
+                    // 2 or more predicates = Match ANY.
+                    QueryPredicate predicate = QueryPredicates.none();
+                    for (String parentId : parentIds) {
+                        QueryPredicateOperation<Object> operation = queryField.eq(parentId);
```
richardmcclellan, line 107:
```suggestion
                        QueryPredicateOperation<String> operation = queryField.eq(parentId);
```

This should be the data type of the field being evaluates, which is `String` for the model id.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+        Set<String> rootIds = new HashSet<>();
+        for (T model : root) {
+            rootIds.add(model.getId());
+        }
+        recurseTree(descendants, rootSchema, rootIds);
+        return descendants;
+    }
+
+    private void recurseTree(
+            Map<ModelSchema, Set<String>> tree,
```
richardmcclellan, line 86: This is just a `Map`, not a tree, so maybe call it `map`?.

```suggestion
            Map<ModelSchema, Set<String>> map,
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -0,0 +1,157 @@
...
+                            int index = cursor.getColumnIndexOrThrow(childPrimaryKey);
+                            do {
+                                childrenIds.add(cursor.getString(index));
+                            } while (cursor.moveToNext());
+                        }
+                    } catch (DataStoreException exception) {
+                        // Don't cut the search short. Populate rest of the tree.
+                    }
+
+                    // Add queried result to the tree
```
richardmcclellan, line 130:
```suggestion
                    // Add queried result to the map
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -503,47 +515,28 @@ public void query(
...
+                        Model dummyItem = gson.fromJson(dummyJson, schema.getModelClass());
+                        itemChangeSubject.onNext(StorageItemChange.builder()
+                                .changeId(id)
```
richardmcclellan, line 528: Unrelated to this PR, but this seems wrong.  Shouldn't we not set `changeId` to anything here, so that `StorageItemChange::build` just randomly generates one?  From the `changeId` Javadoc, this is supposed to be a UUID representing this change, not the id of the model that it relates to.

Maybe this is something that we could follow up on in a separate PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
```diff
@@ -106,19 +108,25 @@ private void addBinding(Object value) {
...
-            throw new DataStoreException(
-                    "Tried to parse an unsupported QueryPredicate",
-                    "Try changing to one of the supported values: " +
-                            "QueryPredicateOperation, QueryPredicateGroup."
-            );
         }
+        throw new DataStoreException(
+                "Tried to parse an unsupported QueryPredicate",
+                "Try changing to one of the supported values: " +
+                        "QueryPredicateOperation, QueryPredicateGroup."
```
richardmcclellan, line 128:
```suggestion
                        "QueryPredicateOperation, QueryPredicateGroup, MatchAllQueryPredicate, MatchNoneQueryPredicate"
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -503,47 +515,28 @@ public void query(
...
-                    // executeUpdateDelete returns the number of rows affected.
-                    final int rowsDeleted = compiledSqlStatement.executeUpdateDelete();
-                    compiledSqlStatement.clearBindings();
-                    if (rowsDeleted == 0) {
-                        throw new DataStoreException(
-                            "Failed to meet condition. Model was not deleted.",
-                            "Please verify the current state of saved item."
-                        );
+                // Use ModelTreeHelper to identify the models affected by cascading delete.
+                Map<ModelSchema, Set<String>> modelFamilyTree =
```
richardmcclellan, line 519: This is a `Map` not a tree, so maybe just call it `descendants` or `modelFamilyMap` or something?

```suggestion
                Map<ModelSchema, Set<String>> descendants =
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: Java compiler doesn't allow that because it needs to be a simple constant value

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: it can also happen from failure to query the SQLite database, which can result from various situations (cursor size overflow, too many cursors, corrupt row in cursor, etc.).

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: I want to put `String` there as well, but because `QueryField#eq(Object)` was originally declared that way, we can't do that. I considered refactoring `QueryField` class to use generics instead of `Object`, but that is potentially a breaking change, and also outside of this PR's scope.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: hmm, i also noticed that but it seems that other operations seem to treat change ID as model ID as well. I'll leave it as is for now and we can revisit it later as you said.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: I want to reuse this algorithm when implementing 1..n deletion, which also needs to account for publishing cascaded deletes.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: Ah, I used to think the order matters, but later realized that the order does not matter. I don't think it's a big deal anyways, and it _is_ nice that the map happens to present the entries in the order of generation.

richardmcclellan: Fair point.  The fact that it is ordered makes it deterministic, which might make unit tests easier to write.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
richardmcclellan: Hmm.  I still think a customer would probably want to know about all of those failures, so they can report them to us, so we can fix those issues.  How about adding a `LOG.error` here?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
richardmcclellan: Got it, makes sense!  I like the idea of refactoring `QueryField` to use generics as well.  There was a customer recently who tried to pass a `String` where they should have passed a `Temporal.DateTime` object.  If `QueryField` was generic, it would help customers avoid errors like that.  Like you said, it would be a breaking change though.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: clever! I prefer this approach more.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: Sure thing. It's never bad to provide some more insight i guess

richardmcclellan (approved):
:shipit:

PR #1062, opened 2020-12-18
feat(api) response deserialization should only skip top level for specific response types

richardmcclellan:
### Current Android Behavior 
A customer can build their own `GraphQLRequest` like this:
```
val document = ("query getTodo(\$id: ID!) { "
    + "getTodo(id: \$id) { "
        + "id "
        + "name "
      + "}"
    + "}")
return SimpleGraphQLRequest(
      document,
      mapOf("id" to id),
      String::class.java,
      GsonVariablesSerializer())
}
```
In the onResponse callback, response.getData() will return a String like:
```
{
    "id": "068ac891-77a3-4653-89de-b05d46bcc60a",
    "name": "A Todo"
}
```

### Current iOS Behavior
A customer can build their own `GraphQLRequest` like this:
```
let operationName = "getTodo"
let document = """
query getTodo($id: ID!) {
  \(operationName)(id: $id) {
    id
    name
  }
}
"""
return GraphQLRequest<Todo>(document: document,
                            variables: ["id": id],
                            responseType: String.self)
```

response.getData() will be the following `String`:
```
{
    "getTodo": {
        "id": "068ac891-77a3-4653-89de-b05d46bcc60a",
        "name": "A Todo"
    }
}
```
### What is this PR about?

As illustrated above, iOS and Android have different response deserialization behavior when requesting a `String` response type (rather than a `Model` or `PaginatedResult<Model>` type).  This PR aims to align Android to the iOS behavior.

Android strips out the top level JSON element, "getTodo", and returns the first child.  If there are 0 or >1 children, response deserialization currently [throws an exception](https://github.com/aws-amplify/amplify-android/blob/fef0cc31b50c427c60e144346a896c20a2f440e4/aws-api-appsync/src/main/java/com/amplifyframework/api/graphql/GsonResponseAdapters.java#L110).  This prevents us from supporting multiple operations in the same request, because the response would contain multiple top level operations.  iOS does not strip out this top level JSON element.

This PR modifies the Android behavior to only strip out the top level JSON element if the provided responseType is either `Iterable<?>`, `ModelWithMetadata<?>`, or `? extends Model`.  

Note: This is a breaking change for customers currently using the low level `SimpleGraphQLRequest`, and using a `responseType` other than the types mentioned above, such as `String.class`.


jamesonwilliams (approved):
Wow this was fast, nice @richardmcclellan!

PR #1063, opened 2020-12-18
chore(api) remove responseType from GraphQLOperation, since we already know it from the GraphQLRequest

richardmcclellan:
No functional changes, just a small refactor to simplify the `GraphQLOperation`.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java
@@ -55,9 +55,9 @@ final class GsonGraphQLResponseFactory implements GraphQLResponse.Factory {
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/GsonGraphQLResponseFactory.java
```diff
@@ -55,9 +55,9 @@
     }
 
     @Override
-    public <T> GraphQLResponse<T> buildResponse(GraphQLRequest<T> request, String responseJson, Type typeOfT)
+    public <T> GraphQLResponse<T> buildResponse(GraphQLRequest<T> request, String responseJson)
```
richardmcclellan, line 58: This is the main change.  We don't need to pass `request` _and_ `typeOfT` because we can get `typeOfT` via `request.getResponseType()`.

jamesonwilliams (approved):

PR #1068, opened 2020-12-22
chore(datastore) missing column name should be a verbose log since it is expected for relationship fields

richardmcclellan:
Several customers (and internal team members) have been concerned about log messages like:
```
Column with name posts does not exist
```
which are logged when querying for a Post object.   

For a typical Blog / Post / Comment schema, the Blog table does not contain any references to its `Post`'s because each `Post` contains a reference to its `blogID`.

This log could be useful if there is in fact a missing field that should be there.  For this reason, I'm going to leave it, but change the log level to verbose.


jamesonwilliams (approved):
🥳

PR #1072, opened 2020-12-23
fix(aws-api): dont fire error if rest op canceled

jamesonwilliams:
iOS doesn't fire any terminal event when a REST operation is canceled.

When an Rx subscription is disposed, neither onError nor onComplete is
called.

In those cases, there is no notification of cancelation/disposal.

This commit aligns Android's behavior, so that an operation will
terminate silently, when canceled.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
@@ -60,6 +60,10 @@ public final class RestRequestFactory {
             .host(url.getHost())
             .addPathSegment(stripLeadingSlashes(url.getPath()));
 
+        if (url.getPort() != -1) {
+            builder.port(url.getPort());
+        }
+
         if (path != null) {
             builder.addPathSegments(stripLeadingSlashes(path));
         }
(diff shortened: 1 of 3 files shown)


jamesonwilliams:
cc: @fjnoyp

aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
```diff
@@ -58,6 +58,7 @@ public static URL createURL(
         HttpUrl.Builder builder = new HttpUrl.Builder()
             .scheme(url.getProtocol())
             .host(url.getHost())
+            .port(url.getPort())
```
jamesonwilliams, line 61: This is an unrelated bugfix. Since AWS always operates on HTTP(S) ports, we never exploited this. Still, I'm unsure how our existing `MockWebServer` tests were ever to function. But the new `AWSApiOperationTest` s certainly did _not_ work without this fix.

rjuliano (approved):
Ship it!!!

raphkim (approved):

aws-api/src/test/java/com/amplifyframework/api/aws/operation/AWSRestOperationTest.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Tests the happy path, wherein the server returns a response, the
+     * operation hasn't been canceled, and we get a {@link RestResponse}
+     * at the end of it.
+     * @throws ApiException
+     *         A possible outcome of the operation. This is not
+     *         expected, and would constitute a test failure.
+     */
+    @Test
+    public void responseEmittedIfOperationWhenOperationSucceeds() throws ApiException {
```
jamesonwilliams, line 90:
```suggestion
    public void responseEmittedWhenOperationSucceeds() throws ApiException {
```

PR #1073, opened 2020-12-25
chore: add explicit error for signin cancelation

jamesonwilliams:
During sign-in with the Cognito hosted web UI, a Chrome Custom tab will
pop up. The user may fill in their sign-in information. However, they
may also exit out of the tab, in which case the sign-in will be
cancelled.

The current behavior is to fire an exception on the onError callback
when this occurs. That exception looks like this:
```
AuthException{
  message=Sign in with web UI failed
  cause=[...mobile client...] AuthNavigationException{
    message=user cancelled
  }
  recoverySuggestion=See attached exception for more details
}
```

This commit updates the exception that gets returned, to something more
specific, that a user could catch explicitly:
```
ActionCancelledException{
  message=The user cancelled the sign-in attempt, so it did not complete.
  cause=[...mobile client...] AuthNavigationException {
    message=user cancelled
  }
  recoverySuggestion=To recover: catch this error, and show the sign-in screen again.
}
```

With this change, the caller may catch the cancellation exception in a
`when` block:
```kotlin
Amplify.Auth.signInWithWebUI(this, { /* result .. */ }, { error ->
    when (error) {
        is ActionCancelledException -> Log.w("TAG", "sign-in cancelled")
        else ->  Log.e("xx", "Login failed.", error)
    }
}
```

This commit also applies similar updates to the sign-out method.

Related: https://github.com/aws-amplify/amplify-ios/pull/982

### DIFF
--- a/core/src/main/java/com/amplifyframework/auth/AuthException.java
+++ b/core/src/main/java/com/amplifyframework/auth/AuthException.java
@@ -452,16 +452,16 @@ public class AuthException extends AmplifyException {
     /**
      * Could not complete an action because it was cancelled by the user.
      */
-    public static class ActionCancelledException extends AuthException {
+    public static class UserCancelledException extends AuthException {
         private static final long serialVersionUID = 1L;
 
         /**
-         * Constructs an {@link ActionCancelledException}.
+         * Constructs an {@link UserCancelledException}.
          * @param message Describes why the error has occurred
          * @param cause An underlying cause of the error
          * @param recoverySuggestion How to remedy the error, if possible
          */
-        public ActionCancelledException(String message, Throwable cause, String recoverySuggestion) {
+        public UserCancelledException(String message, Throwable cause, String recoverySuggestion) {
             super(message, cause, recoverySuggestion);
         }
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -449,6 +449,23 @@ public FailedAttemptsLimitExceededException(Throwable cause) {
         }
     }
 
+    /**
+     * Could not complete an action because it was cancelled by the user.
+     */
+    public static class ActionCancelledException extends AuthException {
```
richardmcclellan, line 455: Maybe call this `UserCancelledException` instead?  That seems a little more clear to me, and also [matches iOS](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Error/AWSCognitoAuthError.swift#L69).

richardmcclellan (approved):

core/src/main/java/com/amplifyframework/auth/AuthException.java
jamesonwilliams: Done!

richardmcclellan (approved):

PR #1076, opened 2020-12-30
[DO NOT MERGE] Bare Auth Plugin PoC

jamesonwilliams:
## What is this?

This is a proof-of-concept of a new auth plugin that is **not** built on
the AWS Mobile SDK. Instead, this plugin calls Cognito Identity Provider directly through a
low-level code-generated client. Specifically, this PoC uses the [V2 Java SDK](https://github.com/aws/aws-sdk-java-v2#aws-sdk-for-java-20), which _[kind of works](https://gist.github.com/jamesonwilliams/e5a78e09cbe6e81a5c0b1ccc7f1ff599#using-the-v2-aws-java-sdk-from-android)_ on Android.

## What's been implemented?
The PR includes **two new modules**:

1. `aws-auth`, a library module that implements a new subclass of `AuthPlugin` in Kotlin
2. `auth-sample-app`, a sample app that navigates a UI from sign-up to confirm-code to sign-in to done, calling the plugin functionalities as it goes along.

The PoC implements sign-up, confirmation with email code, and sign-in with
SRP_A. Aside from the Cognito interactions needed to fulfill SRP_A, this
PoC **does not include support for MFA** sign-in.

## Key Findings

A few findings from this exercise:

1. It's not that bad, really! >50% of our current implementation is truly just layers of meaningless garbage.
2. Frankly, _we don't even really need the Cognito IDP SDK_, and all of its cascading dependencies. We could just call the ~handful of Cognito APIs directly, like @borisirota does in his [JavaScript SRP client](https://github.com/borisirota/amazon-user-pool-srp-client).
3. SRP and hashing are hairy implementation details.
4. The system works like a state machine.

## Realizations for our Auth Design

The realization of point (4) is the biggest and supports @palpatim's focus on unidirectional flows.

Either:

1. We need an architecture to accommodate a variety of interrelated
flows, each of which can recycle implementation details, or

2. We should just make a lot of different little plugins. It might be easier
for the customer if they can just install a different plugin for
different auth flow, instead of trying to sell some grand, all-encompassing behemoth.

## Next Steps & Follow-up Experiments
0. Needs a little more thinking around how one state flows into the next. The challenge responses can go on for an arbitrary numbers of steps, and some of them may be customer-provided.
1. A lot of missing functionality. Resend/update passwords, device tracking, MFA, token refreshes, yikes.
2. There's a file called `AuthenticationHelper` that needs to be well-factored into other pieces.
3. I really think we can implement a slimmed down interface/impl pair that makes straight HTTP calls to Cognito. This PoC only uses four APIs: [`SignUp`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SignUp.html), [`ConfirmSignUp`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ConfirmSignUp.html), [`InitiateAuth`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html), [`RespondToAuthChallenge`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RespondToAuthChallenge.html). Why do we need the entire SDK dependency tree just for that? It could be done in 100-200 lines of `UrlHttpConnection` code, at a fraction of the code size.
4. Well, tests, duh.
5. Identify a contact in the Cognito who can support this project (we'll have questions.)
6. Find some SDEs in AWS Mobile who want to take this PoC and run with it.


palpatim:
Thanks for putting this together @jamesonwilliams! Responses below

---

> 1. It's not that bad, really! >50% of our current implementation is truly just layers of meaningless garbage.

I disagree with the characterization of the implementation as "meaningless garbage", simply because I've written my fair share of stuff that we needed at the time but that turned out to be the wrong thing in the long term. I wholeheartedly agree that there's an awful lot of this that's not useful to us now. :)

> 2. Frankly, _we don't even really need the Cognito IDP SDK_, and all of its cascading dependencies. We could just call the ~handful of Cognito APIs directly, like @borisirota does in his [JavaScript SRP client](https://github.com/borisirota/amazon-user-pool-srp-client).

Comments on this below.

> 3. SRP and hashing are hairy implementation details.

+1. I'll be interested to chat with you about some strategies we can use to get this right.

> 4. The system works like a state machine.
> 
> ## Realizations for our Auth Design
> The realization of point (4) is the biggest and supports @palpatim's focus on unidirectional flows.
> 
> Either:
> 
> 1. We need an architecture to accommodate a variety of interrelated
>    flows, each of which can recycle implementation details, or
> 2. We should just make a lot of different little plugins. It might be easier
>    for the customer if they can just install a different plugin for
>    different auth flow, instead of trying to sell some grand, all-encompassing behemoth.

Your second approach is interesting, and I hadn't considered it before--auth has been a monolith for so long that it hadn't occurred to me to split it apart. I'd want to gather some data on how often customer apps support different authentication types. The other concern is simplicity--making customers understand which plugin matches to which authentication scheme could be challenging. If we adopted this approach, I'd want to make sure we support client-side dependency configuration from the CLI so that we can automatically inject the appropriate plugin dependency declarations after the customer configures their Auth category during `amplify add auth`. Otherwise, we'd have to make customers go through a relatively unfamiliar process of mapping the choices they make during CLI config to our plugin names.


> ## Next Steps & Follow-up Experiments
> 1. Needs a little more thinking around how one state flows into the next. The challenge responses can go on for an arbitrary numbers of steps, and some of them may be customer-provided.

The state machine approach plays nicely with this. I've done some architecture prototyping to prove out the idea, and I'm optimistic about it.

> 2. A lot of missing functionality. Resend/update passwords, device tracking, MFA, token refreshes, yikes.

I'm always tempted to say "But we can add these in later," but in our hard-won experience, it's these edge cases and little bits of functionality that really throw us for a loop. I think you picked the right flow to prove the concept--we just need to prove it out by adding a meaningful representation of the little edge cases and unhappy paths.

> 3. There's a file called `AuthenticationHelper` that needs to be well-factored into other pieces.
> 4. I really think we can implement a slimmed down interface/impl pair that makes straight HTTP calls to Cognito. This PoC only uses four APIs: [`SignUp`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SignUp.html), [`ConfirmSignUp`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ConfirmSignUp.html), [`InitiateAuth`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html), [`RespondToAuthChallenge`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RespondToAuthChallenge.html). Why do we need the entire SDK dependency tree just for that? It could be done in 100-200 lines of `UrlHttpConnection` code, at a fraction of the code size.

This is going to need some thought. There's a pretty compelling argument that we should use common infrastructure to make calls to the code-generated SDK--common invocation patterns, automatically getting updated features & functionality upon new service releases, extended API surface to support additional functionality beyond what's exposed in the REST API. That said, Auth is a very special case, and it's reasonable to suggest that we want to get this as tuned as possible without bringing in those extra dependencies and call patterns.

---

Thanks again for putting this together! Looking forward to more in-depth discussions.

PR #1080, opened 2020-12-31
chore(api): Fix integration tests

rjuliano:
*Issue #, if available:*

*Description of changes:*
A few tweaks related to get the `subscriptionReceivesMutationOverApiKey` test passing again. The previous backend being used was created using the console and relied on some custom resolvers (Events app sample from the AppSync console). These changes allow the test to work with the resolvers generated by the CLI. 

Updated the name of the Amplify app used by the DataStore tests.

### DIFF
--- a/aws-api/src/androidTest/assets/create-comment.graphql
+++ b/aws-api/src/androidTest/assets/create-comment.graphql
@@ -1,5 +1,5 @@
-mutation CommentOnEvent($eventId: ID!, $content: String!, $createdAt: String!) {
-  commentOnEvent(eventId: $eventId, content: $content, createdAt: $createdAt) {
+mutation CommentOnEvent($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
+   createComment(input: {commentId: $commentId, eventId: $eventId, content: $content, createdAt: $createdAt}) {
       eventId
       content
       createdAt
--- a/aws-api/src/androidTest/assets/create-comment.graphql
+++ b/aws-api/src/androidTest/assets/create-comment.graphql
@@ -1,8 +1,7 @@
-mutation CommentOnEvent($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
+mutation CreateComment($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
    createComment(input: {commentId: $commentId, eventId: $eventId, content: $content, createdAt: $createdAt}) {
       eventId
       content
       createdAt
   }
 }
-
--- a/aws-api/src/androidTest/assets/create-event.graphql
+++ b/aws-api/src/androidTest/assets/create-event.graphql
@@ -1,5 +1,5 @@
 mutation CreateEvent($name: String!, $when: String!, $where: String!, $description: String!) {
-  createEvent(name: $name, when: $when, where: $where, description: $description) {
+  createEvent(input: {name: $name, when: $when, where: $where, description: $description}) {
     id
     name
     where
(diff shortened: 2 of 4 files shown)


aws-api/src/androidTest/assets/create-comment.graphql
```diff
@@ -1,5 +1,5 @@
-mutation CommentOnEvent($eventId: ID!, $content: String!, $createdAt: String!) {
-  commentOnEvent(eventId: $eventId, content: $content, createdAt: $createdAt) {
+mutation CommentOnEvent($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
```
richardmcclellan, line 1: I think the request will work either way, but the name of the mutation should probably be changed to `CreateEvent` as well.

```suggestion
mutation CreateEvent($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
```

aws-api/src/androidTest/assets/create-event.graphql
```diff
@@ -1,5 +1,5 @@
 mutation CreateEvent($name: String!, $when: String!, $where: String!, $description: String!) {
-  createEvent(name: $name, when: $when, where: $where, description: $description) {
+  createEvent(input: {name: $name, when: $when, where: $where, description: $description}) {
```
richardmcclellan, line 2: I think this is usually supposed to be a `CreateEventInput!` object, but if it's working as is, maybe I'm wrong?

```suggestion
  createEvent(input: CreateEventInput!) {
```

Also, if you make this change, don't you also need to update the code in `GraphQLInstrumentationTest::createEvent` to add an "input" variable, instead of the other 4 variables (name, when, where, description)?

aws-api/src/androidTest/assets/create-comment.graphql
```diff
@@ -1,8 +1,7 @@
-mutation CommentOnEvent($eventId: ID!, $content: String!, $createdAt: String!) {
-  commentOnEvent(eventId: $eventId, content: $content, createdAt: $createdAt) {
+mutation CreateComment($eventId: ID!, $commentId:String!, $content: String!, $createdAt: String!) {
+   createComment(input: {commentId: $commentId, eventId: $eventId, content: $content, createdAt: $createdAt}) {
```
richardmcclellan, line 2: This is usually a `CreateCommentInput!` object.   But if it's working as is, maybe I'm wrong?

```suggestion
   createComment(input: CreateCommentInput!) {
```

jamesonwilliams (approved):
The tests are green again! 🙏

aws-api/src/androidTest/assets/create-event.graphql
rjuliano: I was confused by this as well. I think the API that used to be hit by this test was created via the AppSync console using one of the sample apps they provide, so it differs from the schema that gets generated by the Amplify CLI. Since we're now generating via the CLI, some of the calls had to be tweaked a bit.

At first, I tried adding "input" to the call from `GraphQLInstrumentationTest::createEvent`, but turns out that those parameter names have to match what's specified in the mutation declaration `mutation CreateEvent($name: String!, $when: String!, $where: String!, $description: String!)`

aws-api/src/androidTest/assets/create-comment.graphql
rjuliano: Same deal as the one above.

aws-api/src/androidTest/assets/create-event.graphql
rjuliano: Also, `CreateEventInput` doesn't exist in the context of this query. I think it would have to be declared as part of the `create-event.graphql` file.

richardmcclellan: Got it, thanks for the explanation!

PR #1082, opened 2020-12-31
fix(datastore) Defer merger.merge to avoid failure if outbox has mutation

richardmcclellan:
Previously, there was an issue where mutation responses were not merged because the merge was attempted before the item was removed from the queue.  This was fixed in #1051 by wrapping the merge in `Completable.defer()`.   Thanks @kjones!

This PR moves the `Completable.defer()` into the merge method itself.  This ensures that the fix will be effective for any calls to the `merge` method, such as calls from the `SyncProcessor`.  This PR also adds a test that fails without this fix, or the fix from #1051.

### DIFF
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MergerTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MergerTest.java
@@ -242,6 +242,51 @@ public final class MergerTest {
     }
 
     /**
+     * When processing a mutation response, the pending mutation should be removed from the outbox, and the mutation
+     * should be merged to local storage.
+     * @throws InterruptedException If interrupted while awaiting terminal result in test observer
+     * @throws AmplifyException On failure to arrange model schema
+     */
+    @Test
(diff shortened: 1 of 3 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/MergerTest.java
```diff
@@ -241,6 +241,45 @@ public void itemIsNotMergedWhenOutboxHasPendingMutation() throws AmplifyExceptio
         assertEquals(blogOwner, blogOwnersInStorage.get(0));
     }
 
+    @Test
+    public void itemIsMergedAfterPendingMutationRemovedFromOutbox() throws AmplifyException, InterruptedException {
+        // Arrange: some model with a well known ID exists on the system.
+        // We pretend that the user has recently updated it via the DataStore update() API.
+        String knownId = RandomString.string();
+        BlogOwner blogOwner = BlogOwner.builder()
+                .name("Jameson")
```
jamesonwilliams, line 250: 👋🏻

jamesonwilliams (approved):

Issue #812, opened 2020-09-09
Wanted 1 metadata for item with id MODEL_ID, but had 0.


azisoli:
```console
W/amplify:aws-datastore: Failed to publish a local change = PendingMutation{mutatedItem=Profile {id=94adedda-c8ee-4817-bcbf-13f6a0a8c0f8, username=94adedda-c8ee-4817-bcbf-13f6a0a8c0f8, state=CHANGE, spendleeId=azadehsoli+990h, avatarImage=null, scansPerDay=null}, classOfMutatedItem=class com.amplifyframework.datastore.generated.model.Profile, mutationType=UPDATE, mutationId=4948cfce-f2a7-11ea-a24f-b9ff350d8ffc, predicate=MatchAllQueryPredicate}
    AmplifyException {message=Wanted 1 metadata for item with id = 94adedda-c8ee-4817-bcbf-13f6a0a8c0f8, but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}
        at com.amplifyframework.datastore.syncengine.VersionRepository.extractVersion(VersionRepository.java:89)
```

richardmcclellan:
I thought this issue had been resolved, but looks like we need to take another look.   Here are the previous reports:

https://github.com/aws-amplify/amplify-android/issues/603
https://github.com/aws-amplify/amplify-android/issues/588
https://github.com/aws-amplify/amplify-android/issues/602

azisoli:
@richardmcclellan  do you have any suggestion that i can do ?

msiewert:
2020-09-15 11:40:31.259 6937-7088/com.example.amplify_android I/amplify:aws-datastore: Successfully enqueued PendingMutation{mutatedItem=Todo {id=2c3029e2-9b15-46da-8510-340863d32588, name=1, description=1}, classOfMutatedItem=class com.amplifyframework.datastore.generated.model.Todo, mutationType=DELETE, mutationId=8d066a4e-f77a-11ea-89dc-07779c1e8fe5, predicate=MatchAllQueryPredicate}
[...]

PR #1085, opened 2021-01-04
fix(datastore) Make PersistentMutationOutbox operations synchronized

richardmcclellan:
Previously, if a customer were to try to create the same object twice, like this:

```
val blog = Blog.builder()                    
    .name("My Blog")
    .build()
Amplify.DataStore.save(blog, {}, {})
Amplify.DataStore.save(blog, {}, {})
```

both would successfully be added to the mutation outbox.  This was because the second mutation was checking for existence of the item in the queue before the first mutation was complete.  This PR moves the logic in `enqueue` which checks existence of the item into the `Completable`, so that it happens _after_ the first mutation completes. 

It also makes a similar change to the `remove` method, so that it is synchronized as well.

Lastly, I removed any `Completeable.defer` that just wrapped a `Completable.create`, because `Completable.create` already defers the operation until it is subscribed to, so the `defer` is redundant.   This was removed from `save` and `load`.

I believe this should fix https://github.com/aws-amplify/amplify-android/issues/812 as well.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
@@ -87,17 +87,21 @@ final class PersistentMutationOutbox implements MutationOutbox {
     @Override
     public <T extends Model> Completable enqueue(@NonNull PendingMutation<T> incomingMutation) {
         Objects.requireNonNull(incomingMutation);
-        // If there is no existing mutation for the model, then just apply the incoming
-        // mutation, and be done with this.
-        String modelId = incomingMutation.getMutatedItem().getId();
-        @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


raphkim (approved):

suryaishnavi:
The issue still not fixed. Getting error DataStoreException{message=Wanted 1 metadata for item with id = "e4ba9d5f-7fa9-4320-ac10-3807664bd3e0"#"d1634dea-d001-70c6-9418-775db1032248", but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}

ridakaddir:
> The issue still not fixed. Getting error DataStoreException{message=Wanted 1 metadata for item with id = "e4ba9d5f-7fa9-4320-ac10-3807664bd3e0"#"d1634dea-d001-70c6-9418-775db1032248", but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}

I am having the issue

PR #1106, opened 2021-01-14
datastore(feat): support delete by model type with predicate

raphkim:
*Issue #, if available:*

*Description of changes:*
* Added method to `DataStoreCategoryBehavior` interface for deleting models by type & predicate
* Implemented it on DataStore plugin and local storage engine
* Added tests

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
@@ -18,6 +18,7 @@ package com.amplifyframework.datastore.storage;
 import android.content.Context;
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.core.Action;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelSchema;
@@ -27,6 +28,7 @@ import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.DataStoreException;
 import com.amplifyframework.testutils.Await;
+import com.amplifyframework.testutils.VoidResult;
 
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -234,6 +236,28 @@ public final class SynchronousStorageAdapter {
                     onResult,
                     onError
                 )
+        );
+    }
+
+    /**
+     * Delete every model of given type that matches predicate.
+     * @param modelType Model type to delete
+     * @param predicate Filter to apply for deletion
+     * @param <T> Type of model being deleted
+     * @throws DataStoreException On any failure to delete the models
+     */
+    public <T extends Model> void delete(@NonNull Class<T> modelType, @NonNull QueryPredicate predicate)
+            throws DataStoreException {
+        Await.result(
+                operationTimeoutMs,
+                (Consumer<Object> onResult, Consumer<DataStoreException> onError) ->
+                    asyncDelegate.delete(
+                            modelType,
+                            StorageItemChange.Initiator.DATA_STORE_API,
+                            predicate,
+                            () -> onResult.accept(VoidResult.instance()),
+                            onError
+                    )
         );
     }
 
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
@@ -18,7 +18,6 @@ package com.amplifyframework.datastore.storage;
 import android.content.Context;
 import androidx.annotation.NonNull;
 
-import com.amplifyframework.core.Action;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelSchema;
@@ -249,15 +248,15 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> void delete(@NonNull Class<T> modelType, @NonNull QueryPredicate predicate)
             throws DataStoreException {
         Await.result(
-                operationTimeoutMs,
-                (Consumer<Object> onResult, Consumer<DataStoreException> onError) ->
-                    asyncDelegate.delete(
-                            modelType,
-                            StorageItemChange.Initiator.DATA_STORE_API,
-                            predicate,
-                            () -> onResult.accept(VoidResult.instance()),
-                            onError
-                    )
+            operationTimeoutMs,
+            (Consumer<Object> onResult, Consumer<DataStoreException> onError) ->
+                asyncDelegate.delete(
+                    modelType,
+                    StorageItemChange.Initiator.DATA_STORE_API,
+                    predicate,
+                    () -> onResult.accept(VoidResult.instance()),
+                    onError
+                )
         );
     }
 
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
@@ -31,6 +31,7 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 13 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
```diff
@@ -237,6 +239,28 @@ public void terminate() throws DataStoreException {
...
+     * @param modelType Model type to delete
+     * @param predicate Filter to apply for deletion
+     * @param <T> Type of model being deleted
+     * @throws DataStoreException On any failure to delete the models
+     */
+    public <T extends Model> void delete(@NonNull Class<T> modelType, @NonNull QueryPredicate predicate)
+            throws DataStoreException {
+        Await.result(
+                operationTimeoutMs,
+                (Consumer<Object> onResult, Consumer<DataStoreException> onError) ->
```
jamesonwilliams, line 253: Oh, interesting. I usually provide the type bounds in `Await.<Here, here>`, but it looks like they are inferred if you add them to the lambda, too, eh.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -203,7 +207,95 @@ public void deleteModelWithPredicateDeletesConditionally() throws DataStoreExcep
...
+        final BlogOwner mark = BlogOwner.builder()
+                .name("Mark")
+                .build();
+        adapter.save(john);
+        adapter.save(jane);
+        adapter.save(mark);
+
+        // Delete everybody whose names start with "J" (i.e. John & Jane)
+        final QueryPredicate predicate = BlogOwner.NAME.beginsWith("J");
+        adapter.delete(BlogOwner.class, predicate);
```
jamesonwilliams, line 235:
```suggestion
        adapter.delete(BlogOwner.class, BlogOwner.NAME.beginsWith("J"));
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -203,7 +207,95 @@ public void deleteModelWithPredicateDeletesConditionally() throws DataStoreExcep
...
+        assertEquals(12, deleted.size());
+        assertEquals(expected, deleted);
+
+        // Get the BlogOwner from the database. Should not have been deleted.
+        final List<BlogOwner> blogOwners = adapter.query(BlogOwner.class);
+        assertEquals(Collections.singletonList(ownerModel), blogOwners);
+
+        // Get the Blog from the database. Should be deleted.
+        final List<Blog> blogs = adapter.query(Blog.class);
+        assertTrue(blogs.isEmpty());
```
jamesonwilliams, line 295:
```suggestion
        assertTrue(adapter.query(Blog.class).isEmpty());
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -203,7 +207,95 @@ public void deleteModelWithPredicateDeletesConditionally() throws DataStoreExcep
...
+
+        // Get the Post from the database. Should be deleted
+        final List<Post> posts = adapter.query(Post.class);
+        assertTrue(posts.isEmpty());
```
jamesonwilliams, line 299:
```suggestion
        assertTrue(adapter.query(Post.class).isEmpty());
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/LocalStorageAdapter.java
```diff
@@ -132,6 +132,25 @@ void query(
...
+     * specific conditions. A {@link Consumer} will be invoked when the
+     * save operation is completed, to notify the caller of success or failure.
+     * @param <T> The type of item being deleted
+     * @param itemClass Item to delete
+     * @param initiator An identification of the actor who initiated this deletion
+     * @param predicate Predicate condition for conditional delete
+     * @param onSuccess A callback that will be invoked when deletion succeeds
+     * @param onError A callback that will be invoked when deletion fails with an error
+     */
+    <T extends Model> void delete(
```
jamesonwilliams, line 146: I think we should try to _reduce_ the number of APIs we have on the local storage adapter.

Would it be possible to consolidate our various public `delete(...)` behaviors, so that they call one single primitive method on the storage adapter?

We want to provide various APIs on the public DataStore contract, to make things easier for customers. But to help keep our storage layer lean and mean, it would be better to centralize its entry points.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -521,6 +521,70 @@ public void query(
         });
     }
 
+    public <T extends Model> void delete(
```
jamesonwilliams, line 524:
```suggestion
    @Override
    public <T extends Model> void delete(
```

Any way we can achieve this functionality on the storage adapter without forking/cloning a new method instance?

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -26,10 +26,14 @@
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.QueryOptions;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
+import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.DataStoreException;
 
+import com.google.common.base.Predicates;
```
jamesonwilliams, line 32: No `com.google.common`, please. We can use Gson, itself, since we have a dependency on it. But we ought not to use its transitive dependencies. This one might be from Guava, I guess?

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
```diff
@@ -187,6 +191,43 @@ public void query(
         onSuccess.accept(deletion);
     }
 
+    @SuppressWarnings("unchecked") // item.getClass() -> Class<?>, but type is T. So cast as Class<T> is OK.
+    @Override
+    public <T extends Model> void delete(
```
jamesonwilliams, line 196: Any opportunity to consolidate the implementation with the storage adapter's other delete method(s)?

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
```diff
@@ -97,6 +97,22 @@
             @NonNull Consumer<DataStoreException> onFailureToDelete
     );
 
+    /**
+     * Deletes every item of given type from the DataStore that meets the provided
+     * conditions.
```
jamesonwilliams, line 102: What's the behavior if there are no matches? Nothing gets deleted? Can you document?

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStoreBinding.java
```diff
@@ -72,6 +72,13 @@
             dataStore.delete(item, predicate, onResult, onError));
     }
 
+    @NonNull
+    @Override
+    public <T extends Model> Completable delete(@NonNull Class<T> itemClass, @NonNull QueryPredicate predicate) {
+        return toCompletable((onResult, onError) ->
+                dataStore.delete(itemClass, predicate, () -> {}, onError));
```
jamesonwilliams, line 79: For the `() -> {}`, I think we have a `NoOpConsumer.create()` or something of that sort, which is a minor readability enhancement. Makes it look more deliberate.

jamesonwilliams (commented):
Look great, Raph! Love the coverage added here, as well. Only thought I had was if we can reduce the footprint being added to the LocalStorageAdapter. It would be nice to have only one entry point for deletions at the storage layer, even while we keep various convenience APIs on the public interface, for customers.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
raphkim: that shouldn't have made it here... must have been auto imported from when i accidentally auto-completed. I removed them though.

core/src/main/java/com/amplifyframework/datastore/DataStoreCategoryBehavior.java
raphkim: Nothing is deleted and operation succeeds. I will add it to the doc.

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStoreBinding.java
raphkim: this overload of `delete` takes in `Action` rather than a `Consumer`. There doesn't appear to be a no-op Action available outside of `EmptyAction` which resides in `testutils` package.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/InMemoryStorageAdapter.java
raphkim: this one behaves completely differently than the other delete method unfortuantely. If anything, the all the other methods where a single data is being written/removed (i.e. `save(Model)` and `delete(Model)`)  are similar enough to each other that it's plausible to consolidate those. I made an effort to do something along that line actually in my last refactoring PR, but didn't remove any methods.

changxu0306 (approved):
Overall it looks good!

rxbindings/src/main/java/com/amplifyframework/rx/RxDataStoreBinding.java
raphkim: added `NoOpAction` to address this.

PR #1107, opened 2021-01-14
feat(core): add an api to check if amplify is configured

jamesonwilliams:
Resolves: https://github.com/aws-amplify/amplify-flutter/issues/124

### DIFF
--- a/core-kotlin/src/main/java/com/amplifyframework/kotlin/core/Amplify.kt
+++ b/core-kotlin/src/main/java/com/amplifyframework/kotlin/core/Amplify.kt
@@ -82,6 +82,14 @@ class Amplify {
         }
 
         /**
+         * Checks if Amplify has been configured, yet.
+         * @return true if so, false, otherwise.
+         */
+        fun isConfigured(): Boolean {
+            return delegate.isConfigured()
+        }
+
+        /**
          * Add a plugin to the Amplify framework.
          * You must add or remove plugins before calling configure();
          * it is an error to try to add a plugin after calling configure().
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -116,6 +116,18 @@ public final class Amplify {
     }
 
     /**
+     * Check is Amplify has been configured. Amplify cannot be used until it has been configured.
+     * {@link #configure(Context)} will throw an {@link AmplifyException} if you try to re-configure
+     * Amplify. To guard against this, you can wrap the call inside an isConfigured() check.
+     * @return True if Amplify has been configured, false otherwise
+     */
+    public static boolean isConfigured() {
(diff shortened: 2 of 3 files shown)


jamesonwilliams:
We met and decided *not* to add this API. Instead, Android will created a purpose-tailored exception for this use case, as discussed in #1108.

jamesonwilliams:
> We met and decided _not_ to add this API. Instead, Android will created a purpose-tailored exception for this use case, as discussed in #1108.

We did add the purpose-built exception type, but customers were left unsatisfied. New resolution: we'll make both available.

richardmcclellan:
> > We met and decided _not_ to add this API. Instead, Android will created a purpose-tailored exception for this use case, as discussed in #1108.
> 
> We did add the purpose-built exception type, but customers were left unsatisfied. New resolution: we'll also make this `isConfigured()` API available.

Do you have any more background you could share?   Why were customers left unsatisfied?

richardmcclellan:
> > > We met and decided _not_ to add this API. Instead, Android will created a purpose-tailored exception for this use case, as discussed in #1108.
> > 
> > 
> > We did add the purpose-built exception type, but customers were left unsatisfied. New resolution: we'll also make this `isConfigured()` API available.
> 
> Do you have any more background you could share? Why were customers left unsatisfied?

Context provided offline:  For Flutter customers (using Dart), catching this purpose built exception goes against the fundamentals of the language because checked exceptions are only supposed to be thrown if there is a bug, or a truly unexpected state.  It also causes major issues for Flutter developers because it interrupts the debugger during development.

richardmcclellan (approved):

richardmcclellan:
@jamesonwilliams I'm trying to decide if we should merge this change, or not.   Let me know your thoughts!

### Reasons not to merge:
 - While Flutter is definitely adding `isConfigured`, they are not dependent on this change to implement it, because they are tracking the `isConfigured` track themselves.
 - I'm not aware of a reason an Android customer would need this.  I think that catching the `AlreadyConfiguredException` is a better approach, and matches Android/Java paradigms better.

### Reasons to merge:
 - Since Flutter is adding this, perhaps we should just add it for platform parity.

jamesonwilliams:
> @jamesonwilliams I'm trying to decide if we should merge this change, or not. Let me know your thoughts!
> 
> ### Reasons not to merge:
> * While Flutter is definitely adding `isConfigured`, they are not dependent on this change to implement it, because they are tracking the `isConfigured` track themselves.
> * I'm not aware of a reason an Android customer would need this.  I think that catching the `AlreadyConfiguredException` is a better approach, and matches Android/Java paradigms better.  Also, this introduces two ways of doing the same thing, which goes against our design tenets.
> 
> ### Reasons to merge:
> * Since Flutter is adding this, perhaps we should just add it for platform parity.

I think we can close it for now and await further customer input. This API seems to have a platform specific need on Flutter, but not on Android.

PR #1110, opened 2021-01-15
feat(datastore) only include changed fields in update mutations

richardmcclellan:
### Problem
Currently when performing an update mutation, all fields are sent in the request to AppSync.  This prevents customers from being able to benefit from the AutoMerge conflict resolution strategy.  If AppSync receives a mutation with an outdated version, it attempt to intelligently merge the update, if the fields that have changed do not overlap with the fields that changed in the conflicting mutation.  However, since the client sends all fields, the benefits of AutoMerge will never occur.   Instead, all mutations with outdated versions will silently fail.  The update will not occur, and the original model is returned back to the client.

### Solution
This PR updates DataStore so that only the fields that have _changed_ will be sent to the server.   A comparison is done with the current version of the model in SQLite to determine which fields are different.   
 
### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
@@ -53,7 +53,7 @@ public final class ModelConverter {
 
     /**
      * Convert a Map&lt;String, Object&gt; to a Model.
-     * @param map a a Map&lt;String, Object&gt;
+     * @param map a Map&lt;String, Object&gt;.
      * @param modelClass Class of Model to convert the Map to.
      * @param <T> type of the Model instance to convert to.
      * @return a Model instance created from the provided Map.
@@ -64,4 +64,3 @@ public final class ModelConverter {
         return gson.fromJson(jsonString, modelClass);
     }
 }
-
--- a/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.datastore.appsync;
+
+import com.amplifyframework.core.model.Model;
+import com.amplifyframework.util.GsonFactory;
+import com.amplifyframework.util.GsonObjectConverter;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utility for converting a Model to/from a Map&lt;String, Object&gt;.
+ */
+public final class ModelConverter {
+
+    private ModelConverter() { }
+
+    /**
+     * Convert a Model to a Map&lt;String, Object&gt;.
+     * @param model a Model instance.
+     * @param <T> type of the Model instance.
+     * @return a Map&lt;String, Object&gt; representation of the provided Model instance.
+     */
+    public static <T extends Model> Map<String, Object> toMap(T model) {
+        if (model == null) {
+            return new HashMap<>();
+        }
+        if (model instanceof SerializedModel) {
+            return ((SerializedModel) model).getSerializedData();
+        } else {
+            Gson gson = GsonFactory.instance();
+            JsonElement jsonElement = gson.toJsonTree(model);
+            return GsonObjectConverter.toMap(jsonElement.getAsJsonObject());
+        }
+    }
+
+    /**
+     * Convert a Map&lt;String, Object&gt; to a Model.
+     * @param map a a Map&lt;String, Object&gt;
+     * @param modelClass Class of Model to convert the Map to.
+     * @param <T> type of the Model instance to convert to.
+     * @return a Model instance created from the provided Map.
+     */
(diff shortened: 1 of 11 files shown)


aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+            Gson gson = GsonFactory.instance();
+            JsonElement jsonElement = gson.toJsonTree(model);
+            return GsonObjectConverter.toMap(jsonElement.getAsJsonObject());
+        }
+    }
+
+    /**
+     * Convert a Map&lt;String, Object&gt; to a Model.
+     * @param map a a Map&lt;String, Object&gt;
```
raphkim, line 56:
```suggestion
     * @param map a Map&lt;String, Object&gt;.
```

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utility for converting a Model to/from a Map&lt;String, Object&gt;.
+ */
+public final class ModelConverter {
+
+    private ModelConverter() { }
```
raphkim, line 33: nit: we don't enforce "whitespace around" checkstyle rules for empty constructors so we can just do `{}` instead of `{ }`

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -754,18 +756,30 @@ private synchronized void bindValueToStatement(
...
+
+        Iterator<? extends Model> result = Single.<Iterator<? extends Model>>create(emitter -> {
+            if (model instanceof SerializedModel) {
+                query(modelName, Where.matches(condition), emitter::onSuccess, emitter::onError);
+            } else {
+                query(model.getClass(), Where.matches(condition), emitter::onSuccess, emitter::onError);
+            }
+        }).blockingGet();
+        return result.hasNext() ? result.next() : null;
     }
```
raphkim, line 783: the reason for implementing `modelExist()` method explicitly rather than using `query()` was to remove the workload required to serialize queried models when quickly checking if an item exists in the database. Not sure how much of a boost you get out of it though.

raphkim (approved):

changxu0306 (approved):
Ship it!🚢

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
```diff
@@ -0,0 +1,66 @@
...
```
richardmcclellan, line 33:
```suggestion
    private ModelConverter() {}
```

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/SerializedModel.java
```diff
@@ -44,6 +44,45 @@ private SerializedModel(
...
+     * Computes the difference between two Models, comparing equality of each field value for each Model, and returns
+     * the difference as a SerializedModel.
+     * @param updated the updated Model, whose values will be used to build the resulting SerializedModel.
+     * @param original the original Model to compare against.
+     * @param modelSchema ModelSchema for the Models between compared.
+     * @param <T> type of the Models being compared.
+     * @return a SerializedModel, containing only the values from the updated Model that are different from the
+     *         corresponding values in original.
+     */
+    public static <T extends Model> SerializedModel difference(T updated, T original, ModelSchema modelSchema) {
```
rjuliano, line 71: Should modelSchema be a parameter or could we just infer it from the models passed in?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
richardmcclellan: That is a valid performance concern!  For the use case addressed by this PR, I need to serialize the queried models, in order to compute which fields have changed, for any save initiated by the client.  That said, this isn't necessary when saving models that come from the server, via the `SyncProcessor`, or `SubscriptionProcessor`.   We actually know this based on the `initiator` parameter of the `save` method.

Let me see what I can do to eliminate any unnecessary work based on that parameter.

rjuliano (commented):
Overall I think it looks good. I do have one concern (which may or may not be relevant).

Say, for example, that the `wea` field of `BlogOwner` is a required field. If you create a mutation that changes the `name`, you would have only `id` and `name` in the graphql statement (which is correct). My concern has to do with the response: does the resolver return all the fields as a result of the update so it can serialize back into a `BlogOwner` object.

richardmcclellan:
> Say, for example, that the `wea` field of `BlogOwner` is a required field. If you create a mutation that changes the `name`, you would have only `id` and `name` in the graphql statement (which is correct). My concern has to do with the response: does the resolver return all the fields as a result of the update so it can serialize back into a `BlogOwner` object.

Yes!   This PR only changes what fields are included in the "input" variable.  It does not make any changes to the selection set, which is what determines what is in the response.

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/SerializedModel.java
richardmcclellan: This would be pretty easy if T was always a generated Java model, but it's more complex since it could be a `SerializedModel` now too.  If we don't pass in a `ModelSchema`, we'd probably need to pass in a `ModelSchemaRegistry`, which doesn't seem much better, so I think it's best to leave as is.

rjuliano (approved):
Ship it 🚢

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
richardmcclellan: I just pushed a commit which reverts the changes I made to `modelExists()`, and adds a new `query` method, which is only called when we know this is an update mutation, _and_ it was initiated by `DATASTORE_API`.

This should ensure there won't be any negative performance impact on saves from the `SyncProcessor` or `SubscriptionProcessor`, or when creating an object via the DataStore API.

raphkim (approved):

PR #1117, opened 2021-01-20
chore(devmenu) use java.util.Date instead of java.time for dev menu logging

richardmcclellan:
The developer menu logging functionality used `java.time.LocalDateTime` to store time of log messages.  The `java.time` library either requires API 26 or above, or requires customers to be on Android Studio 4.0 and above, and make changes in their `build.gradle` to enable desugaring.  There no particular reason to use `java.time.*` over `java.util.Date` in this case, so this PR changes the dev menu to use `java.util.Date`.

### DIFF
--- a/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
+++ b/core/src/main/java/com/amplifyframework/devmenu/LogEntry.java
@@ -22,8 +22,8 @@ import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.logging.LogLevel;
 
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
+import java.text.SimpleDateFormat;
+import java.util.Date;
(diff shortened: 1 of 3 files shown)


jamesonwilliams:
If the customer will need support for Studio 4+ for other things, shouldn't we just drop support entirely for lower versions?

jamesonwilliams (approved):

richardmcclellan:
> If the customer will need support for Studio 4+ for other things, shouldn't we just drop support entirely for lower versions?

Any customer using any of the `Temporal` types is still required to use Studio 4+.  But, for those customers who can't upgrade to Studio 4+ yet, as a workaround, they can still use `Amplify` as long as they aren't using any of the AWS Temporal types in their schema.

Essentially, this PR prevents the app from crashing when the dev menu launches, for customers using less than Studio 4.+

PR #1121, opened 2021-01-21
fix(datastore): fix crash when deleting model with many children

raphkim:
*Issue #, if available:*

*Description of changes:*
SQLite complains when query contains complex predicate (e.g. `WHERE id = 1 OR id = 2 OR ... OR id = 1000`).
Normally this won't be an issue, but cascading delete uses above logic to identify every children of a set of models. When deleting 1000+ items from a single model type (which is possible), this error will trigger.

This PR fixes this error during cascading delete by chaining maximum of 900 predicates together at max per query and then merging multiple query results together.


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -160,11 +137,65 @@ private void recurseTree(
...
+        String childPrimaryKey = childTable.getPrimaryKey().getAliasedName();
+
+        // Chain predicates with OR operator.
+        // Beware the SQLite expression tree size limit of 1000
+        List<QueryOptions> options = new ArrayList<>();
+        QueryPredicate predicate = QueryPredicates.none();
+        int currentSize = 0;
+        for (String parentId : parentIds) {
+            // Stop appending if size exceeds 900
+            if (currentSize > 900) {
```
richardmcclellan, line 155: So....why is 900 the magic number? 🤣

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -160,11 +137,65 @@ private void recurseTree(
...
+                currentSize = 0;
+            }
+            QueryPredicate operation = parentIdField.eq(parentId);
+            predicate = predicate.or(operation);
+            currentSize++;
+        }
+        options.add(Where.matches(predicate));
+
+        // Collect every children one level deeper than current level
+        // SELECT * FROM <CHILD_TABLE> WHERE <PARENT> = <ID_1> OR <PARENT> = <ID_2> OR ...
```
richardmcclellan, line 167: Instead of chaining with SQL `OR` keyword, could you build the query like this?
```suggestion
// SELECT * FROM <CHILD_TABLE> WHERE <PARENT> IN (<ID_1>, <ID_2>, ...)
```
I think there are two benefits to this approach:
1. It would prevent the SQLite expression tree size limit of 1000, without having to break things up into multiple queries
2. It would probably improve performance.  I don't know for sure, but I would guess that SQLite represents the list as a hash table internally.  If that is true, then this approach would improve the overall efficiency of this query from `O(n^2)` to `O(n)`.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: 999 didn't work :(

jamesonwilliams: Haha, ah.... Hm.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: Turns out [SQLite optimizer](https://www.sqlite.org/optoverview.html) converts `column = a OR column = b OR column = c` automatically into `column IN (a, b, c)`. Cool!

This approach solved the tree depth limit problem since it is a single expression rather than a combination of 1000+. However, it turns out that Android SQLit _also_ has a hard limit on the number of `?` that you can use per query expression (same as tree depth limit of 1000). 

I resolved it by simply using ID strings directly without binding them, which could potentially be a problem.
As of right now, ID cannot be anything than UUID format, so there is no concern for SQL injection. However, there is a conversation floating around in favor of allowing _any_ string to be assignable to ID field (like iOS does), so I'm curious if this change would potentially introduce a vulnerability.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -275,22 +274,18 @@ public SqlCommand insertFor(@NonNull ModelSchema modelSchema) {
...
-                .append("(");
-        for (int i = 0; i < columns.size(); i++) {
-            if (i == columns.size() - 1) {
-                stringBuilder.append("?");
-            } else {
-                stringBuilder.append("?, ");
-            }
-        }
-        stringBuilder.append(")");
+                .append("(")
```
richardmcclellan, line 284: Potential for `Wrap.inParentheses(...)` here?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -275,22 +274,18 @@ public SqlCommand insertFor(@NonNull ModelSchema modelSchema) {
...
+                .append(Collections.nCopies(columns.size(), SqlKeyword.VARIABLE.toString())
```
richardmcclellan, line 285: > I resolved it by simply using ID strings directly without binding them, which could potentially be a problem.
As of right now, ID cannot be anything than UUID format, so there is no concern for SQL injection. However, there is a conversation floating around in favor of allowing any string to be assignable to ID field (like iOS does), so I'm curious if this change would potentially introduce a vulnerability.

If you are no longer binding ID strings, why are you still creating this prepared statement with variables?  Quite likely, I just don't fully understand something here.

richardmcclellan (commented):
> Turns out SQLite optimizer converts column = a OR column = b OR column = c automatically into column IN (a, b, c). Cool!

Very cool!

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -301,4 +304,43 @@ public void deleteNoneDoesNotFail() throws DataStoreException {
         final List<BlogOwner> blogOwners = adapter.query(BlogOwner.class);
         assertEquals(Collections.singletonList(john), blogOwners);
     }
+
+    /**
+     * Test deleting item with many children does not trigger SQLite exception.
+     * SQLite places a hard limit of 1000 on expression tree depth, so chaining too many
```
jamesonwilliams, line 310: Is this limit exposed with a constant by any of the SQL APIs? `andorid.sql.SqlGizmo.SOME_LIMIT`, that kind of thing?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -47,6 +47,7 @@
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
```
jamesonwilliams, line 50: We have talked about using Java 8 Stream APIs before, and decided against it. 

It looks like you're using it to join items by a comma `,`. For that, please use Android's [`TextUtils.join(CharSequence, Iterable)`](https://developer.android.com/reference/android/text/TextUtils#join(java.lang.CharSequence,%20java.lang.Iterable)):

```java
String fooBar = TextUtils.join(", ", Arrays.asList("foo", "bar"))
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -160,14 +146,34 @@ private void recurseTree(
...
+        // SELECT <child_id>, <parent_id> FROM <child_table> WHERE <parent_id> IN (<id_1>, <id_2>, ...)
+        String queryString = String.valueOf(SqlKeyword.SELECT) +
+                SqlKeyword.DELIMITER +
+                Wrap.inBackticks(childIdField) +
+                SqlKeyword.SEPARATOR +
+                Wrap.inBackticks(parentIdField) +
+                SqlKeyword.DELIMITER +
+                SqlKeyword.FROM +
+                SqlKeyword.DELIMITER +
+                Wrap.inBackticks(childTable) +
```
jamesonwilliams, line 164: Maybe we can create a little help method for these repeated groupings of tokens, like
```java
from(childTable)
```

Sort of like a domain specific language for SQL, but using static factory methods?

Overall I feel like our use of these consts is making it harder to see what the actual SQL commands will look like, when done.

We can make more, like:

```java
where(parentIdField).in(parentIds)
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -160,14 +146,34 @@ private void recurseTree(
...
+                SqlKeyword.DELIMITER +
+                SqlKeyword.WHERE +
+                SqlKeyword.DELIMITER +
+                Wrap.inBackticks(parentIdField) +
+                SqlKeyword.DELIMITER +
+                SqlKeyword.IN +
+                SqlKeyword.DELIMITER +
+                Wrap.inParentheses(parentIds
+                    .stream()
```
jamesonwilliams, line 173: Let's avoid `stream()` for now. Technically as of Studio 4+, I think this gets desugared at build time. So there isn't actually any hard technical reason not to use it, anymore.

However, we don't want to mix too many different functional paradigms in the code base. Rx has a way to express most everything that the Collections APIs do -- and usually more.

FWIW, I do sort of prefer using the Stream APIs natively on the backend that support Oracle/Corretto JVM, as opposed to Rx. 🤷

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
```diff
@@ -259,19 +262,19 @@ public void deleteModelTypeWithPredicateCascades() throws DataStoreException {
...
         // Triggers a delete of all blogs.
         // All posts will be deleted by cascade.
         adapter.delete(Blog.class, QueryPredicates.all());
 
-        // Assert that cascaded deletions are observed.
-        assertEquals(12, deleted.size());
-        assertEquals(expected, deleted);
+        // Assert 3 blogs and 9 posts are deleted.
+        deleteObserver.assertValueCount(12);
+        assertEquals(expected, new HashSet<>(deleteObserver.values()));
```
jamesonwilliams, line 277: There is actually an `assertValueSetOnly` which may be appropriate here.

It does **not** work like a `HashSet`, de-duplicating values. It **does** ignore the order in which events arrived on the stream, which I believe is the real reason that a Set was being used in the test to begin with.

Might be a little cleaner.

Actually come to think of it, this would remove the need for the count check, as well.
```suggestion
        deleteObserver.assertValueSetOnly(expected);
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -45,6 +41,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Utility class to help traverse a tree of models by relationship.
```
jamesonwilliams, line 47: We missed getting a `SQLiteModelTreeTest.java` when this class was first introduced, but let's please get one started.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
raphkim: `assertValueSetOnly` was deprecated and is no longer supported by rxjava3 :(

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: this is a fantastic idea! I think it deserves its own PR because there is a bunch of similar code in `SQLiteCommandFactory` that will be made much cleaner with the addition of this utility.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
raphkim: Haha that's the utility I was looking for. Didn't know that Android provides it.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
raphkim: oh oops. none of the changes in `SQLiteCommandFactory` is relevant for this PR. I will remove them for clarity.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterDeleteTest.java
raphkim: i believe this variable lives internally in sqlite3, so it isn't directly accessible from android code.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: added a test!

richardmcclellan (approved):

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTreeTest.java
```diff
@@ -0,0 +1,161 @@
...
+     * up to two levels deep.
+     */
+    @Test
+    public void testDescendantsOfTwoLevelsDeepRelationship() {
+        // Insert items into database
+        //      A1
+        //     /  \
+        //    B2  B3
+        //   /  \   \
+        //  C4  C5  C6
```
jamesonwilliams, line 89: Woo-hoo!

jamesonwilliams (approved):

Issue #987, opened 2020-11-24
Disable synchronization to specific models with selective sync


upachecog:
Hi @richardmcclellan as we discussed in #944 in this [comment](https://github.com/aws-amplify/amplify-android/issues/944#issuecomment-731855877) 

It would be useful for me to disable synchronization to specific models in my GraphQL schema because there are some use cases in which the user is just allowed to create data, so queries and subscriptions are not required.

Your suggestion was to do something like this:

```
.syncExpression(Location::class.java) { QueryPredicates.NONE }
```

PS-MS:
Is there any update on this issue? I have just tried with the latest (amplify-android 1.26.0 and amplify-kotlin 0.10.0) but run into the logs mentioned in PR #1126

`Caused by: DataStoreException{message=Tried to parse an unsupported QueryPredicate, cause=null, recoverySuggestion=Try changing to one of the supported values: QueryPredicateOperation, QueryPredicateGroup.}
        at com.amplifyframework.datastore.appsync.AppSyncRequestFactory.parsePredicate(AppSyncRequestFactory.java:199)
        at com.amplifyframework.datastore.appsync.AppSyncRequestFactory.buildSyncRequest(AppSyncRequestFactory.java:110)
        	... 16 more`

raphkim:
Hey @PS-MS @upachecog , 

This use-case is already supported via following sync expression:

```
.syncExpression(Location::class.java) { QueryPredicates.none() }
```

Please let me know if this is not working for your use-case.

PS-MS:
Hi @raphkim ,

I no longer need this but tried just in case and still have the same DataStoreException as above
```
W/amplify:aws-datastore: Initial cloud sync failed.
    DataStoreException{message=Failed to get fields for model., cause=DataStoreException{message=Tried to parse an unsupported QueryPredicate, cause=null, recoverySuggestion=Try changing to one of the supported values: QueryPredicateOperation, QueryPredicateGroup.}, recoverySuggestion=Validate your model file.}
```
        
This is when using
`.syncExpression(Answer::class.java) { QueryPredicates.none() }`

eeatonaws:
Hi @PS-MS, thank you for trying that sync expression and reporting the error you observed. We are looking into the cause of the error.

raphkim:
Thank you for pointing that out! I have opened a PR to address this issue.

eeatonaws:
[...]

PR #1126, opened 2021-01-22
[POC] feat(datastore) Allow QueryPredicates.none() in syncExpressions to enable local only models

richardmcclellan:
Resolves this feature request: https://github.com/aws-amplify/amplify-android/issues/987

Similar feature request for JS: https://github.com/aws-amplify/amplify-js/issues/6593

Today, customers can use [selective sync](https://docs.amplify.aws/lib/datastore/sync/q/platform/android#selectively-syncing-a-subset-of-your-data) to only sync a subset of data for a given model, based on a `QueryPredicate`, like this:  
```
Amplify.addPlugin(
    AWSDataStorePlugin(DataStoreConfiguration.builder()
        .syncExpression(Blog::class.java) { Blog.NAME.contains("Amplify") }
        .build()
    )
)
```

However, there is not a way to disable syncing entirely for a given type, even though a `QueryPredicates.none()` does exist.  This PR adds support for the following approach:
```
Amplify.addPlugin(
    AWSDataStorePlugin(DataStoreConfiguration.builder()
        .syncExpression(Blog::class.java) { QueryPredicates.none() }
        .build()
    )
)
```

If you try the above today _without_ this PR, DataStore fails to start with this error:
```
Caused by: DataStoreException{message=Tried to parse an unsupported QueryPredicate, cause=null, recoverySuggestion=Try changing to one of the supported values: QueryPredicateOperation, QueryPredicateGroup.}
        at com.amplifyframework.datastore.appsync.AppSyncRequestFactory.parsePredicate(AppSyncRequestFactory.java:199)
        at com.amplifyframework.datastore.appsync.AppSyncRequestFactory.buildSyncRequest(AppSyncRequestFactory.java:110)
        	... 16 more
```

When DataStore is started with the configuration above, the Blog type will be excluded from the `SyncProcessor` (no sync query will be made, for `Blog`) and the `SubscriptionProcessor` (no subscriptions will be started, for `Blog`).

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
@@ -29,6 +29,7 @@ import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelProvider;
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
+import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.AmplifyDisposables;
 import com.amplifyframework.datastore.DataStoreChannelEventName;
 import com.amplifyframework.datastore.DataStoreException;
@@ -42,6 +43,7 @@ import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.hub.HubEvent;
 import com.amplifyframework.logging.Logger;
 import com.amplifyframework.util.Empty;
+import com.amplifyframework.util.ForEach;
 
 import java.util.Arrays;
 import java.util.HashSet;
@@ -111,7 +113,16 @@ final class SubscriptionProcessor {
      * Start subscribing to model mutations.
      */
     synchronized void startSubscriptions() throws DataStoreException {
-        int subscriptionCount = modelProvider.modelNames().size() * SubscriptionType.values().length;
+        List<ModelSchema> modelSchemas = Observable.fromIterable(modelProvider.modelSchemas().values())
+            // Don't subscribe to models that have an associated syncExpression of QueryPredicates.none().
+            .filter(schema -> {
+                QueryPredicate predicate = queryPredicateProvider.getPredicate(schema.getName());
+                return !QueryPredicates.none().equals(predicate);
+            })
+            .toList()
+            .blockingGet();
+
+        int subscriptionCount = modelSchemas.size() * SubscriptionType.values().length;
         // Create a latch with the number of subscriptions are requesting. Each of these will be
         // counted down when each subscription's onStarted event is called.
         AbortableCountDownLatch<DataStoreException> latch = new AbortableCountDownLatch<>(subscriptionCount);
@@ -121,7 +132,7 @@ final class SubscriptionProcessor {
         buffer = ReplaySubject.create();
 
         Set<Observable<SubscriptionEvent<? extends Model>>> subscriptions = new HashSet<>();
-        for (ModelSchema modelSchema : modelProvider.modelSchemas().values()) {
+        for (ModelSchema modelSchema : modelSchemas) {
             for (SubscriptionType subscriptionType : SubscriptionType.values()) {
                 subscriptions.add(subscriptionObservable(appSync, subscriptionType, latch, modelSchema));
             }
@@ -150,7 +161,8 @@ final class SubscriptionProcessor {
                                 HubEvent.create(DataStoreChannelEventName.SUBSCRIPTIONS_ESTABLISHED));
             LOG.info(String.format(Locale.US,
                 "Started subscription processor for models: %s of types %s.",
-                modelProvider.modelNames(), Arrays.toString(SubscriptionType.values())
+                ForEach.inCollection(modelSchemas, ModelSchema::getName),
+                Arrays.toString(SubscriptionType.values())
             ));
         } else {
             throw new DataStoreException("Timed out waiting for subscription processor to start.", "Retry");
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
@@ -27,6 +27,7 @@ import com.amplifyframework.core.model.ModelProvider;
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.ModelSchemaRegistry;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
+import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.AmplifyDisposables;
 import com.amplifyframework.datastore.DataStoreChannelEventName;
 import com.amplifyframework.datastore.DataStoreConfigurationProvider;
@@ -50,6 +51,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
 import io.reactivex.rxjava3.processors.BehaviorProcessor;
 import io.reactivex.rxjava3.schedulers.Schedulers;
@@ -71,7 +73,6 @@ final class SyncProcessor {
     private final AppSync appSync;
     private final Merger merger;
     private final DataStoreConfigurationProvider dataStoreConfigurationProvider;
-    private final String[] modelNames;
     private final QueryPredicateProvider queryPredicateProvider;
 
     private SyncProcessor(Builder builder) {
@@ -82,9 +83,6 @@ final class SyncProcessor {
         this.merger = builder.merger;
         this.dataStoreConfigurationProvider = builder.dataStoreConfigurationProvider;
         this.queryPredicateProvider = builder.queryPredicateProvider;
-        this.modelNames =
-            ForEach.inCollection(modelProvider.modelSchemas().values(), ModelSchema::getName)
-                .toArray(new String[0]);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -100,22 +98,31 @@ public static ModelProviderStep builder() {
...
         // And sort them all, according to their model's topological order,
         // So that when we save them, the references will exist.
         TopologicalOrdering ordering =
             TopologicalOrdering.forRegisteredModels(modelSchemaRegistry, modelProvider);
         Collections.sort(modelSchemas, ordering::compare);
+
         for (ModelSchema schema : modelSchemas) {
-            hydrationTasks.add(createHydrationTask(schema));
+            QueryPredicate predicate = queryPredicateProvider.getPredicate(schema.getName());
+            if (!QueryPredicates.none().equals(predicate)) {
```
raphkim, line 115: Hmm, why does `MatchNoneQueryPredicate` get special treatment here? Should other things that don't meet the predicate logic filter still pass through? (i.e. `predicate.evaluate()`)

raphkim: (Same applies to `filter()` above)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -100,22 +98,31 @@ public static ModelProviderStep builder() {
      */
     Completable hydrate() {
         final List<Completable> hydrationTasks = new ArrayList<>();
-        List<ModelSchema> modelSchemas = new ArrayList<>(modelProvider.modelSchemas().values());
+        List<ModelSchema> modelSchemas = Observable.fromIterable(modelProvider.modelSchemas().values())
+            // Don't sync models that have an associated syncExpression of QueryPredicates.none().
+            .filter(schema -> !QueryPredicates.none().equals(queryPredicateProvider.getPredicate(schema.getName())))
```
raphkim, line 103: instead of doing a direct comparison on `QueryPredicates.none()`, maybe it's better to use `x instanceof MatchNoneQueryPredicate` instead? 

There is no guarantee that `QueryPredicates.none()` is a constant (even though it is) since it's a static method result right? Whereas `MatchNoneQueryPredicate.class` is a constant.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: > why does MatchNoneQueryPredicate get special treatment here?

There is no way to serialize a `MatchNoneQueryPredicate`, and send it to AppSync.  This makes sense, because there's no point in making an API call if you know the the response will just be zero results.    Basically, we don't want to perform the sync query at all, to eliminate an unnecessary API call.

> (Same applies to filter() above)

Woops, I actually don't need to filter twice!   I had debated doing the filter here and above, but accidentally left both of them in the code.  I'm going to do the filter at the top, and remove this one.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: Even if it's not a constant, it should always return the correct result, because of the [implementation of `equals`](https://github.com/aws-amplify/amplify-android/blob/b29c39ee9b73f359297f93567541e11a81812fc1/core/src/main/java/com/amplifyframework/core/model/query/predicate/MatchNoneQueryPredicate.java#L48), which returns true as long as `obj instanceof MatchNoneQueryPredicate`.

I did a quick search, and it looks like we currently use the `QueryPredicates.all().equals(queryPredicate)` approach in 8 places, and the `queryPredicate instanceof MatchAllQueryPredicate` approach in 3 places.  I actually think we should make the `MatchAllQueryPredicate` and `MatchNoneQueryPredicate` classes package private.  I think they are implementation details, and we should only access them via the `QueryPredicates` factory.

richardmcclellan: Here's a separate PR that makes those classes package-private: https://github.com/aws-amplify/amplify-android/pull/1127

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/StorageObserver.java
```diff
@@ -41,13 +42,16 @@
 
     private final LocalStorageAdapter localStorageAdapter;
     private final MutationOutbox mutationOutbox;
+    private final QueryPredicateProvider queryPredicateProvider;
     private final CompositeDisposable ongoingOperationsDisposable;
 
     StorageObserver(
```
aimiebell, line 48: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

richardmcclellan: I agree that method names should have the first letter lowercase, but this is not a method, this is a constructor, so I don't believe this is problematic.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -128,7 +128,7 @@ public Orchestrator(
                 .queryPredicateProvider(queryPredicateProvider)
                 .onFailure(this::onApiSyncFailure)
                 .build();
-        this.storageObserver = new StorageObserver(localStorageAdapter, mutationOutbox);
+        this.storageObserver = new StorageObserver(localStorageAdapter, mutationOutbox, queryPredicateProvider);
```
jamesonwilliams, line 131: IMO, the `StorageObserver` is just a dumb little utility to watch the DataStore and funnel stuff into the `MutationOutbox`.

Instead of giving the StorageObserver awareness of the filtering rules, how about giving this responsibility to the  **`MutationProcessor`**?

When the MutationProcessor is evaluating a mutation, it can say "whoa! Hold the phone. I'm not going to send _this_, are you kidding me? It's sync predicate is all wrong."

----

Related question: Are sync predicates for outgoing mutations, today? Or just inbound?

Seems actually like this is not a unique problem to the `.none()` filter?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -109,7 +111,13 @@ public static AppSyncStep builder() {
      * Start subscribing to model mutations.
      */
     synchronized void startSubscriptions() throws DataStoreException {
-        int subscriptionCount = modelProvider.modelNames().size() * SubscriptionType.values().length;
+        List<ModelSchema> modelSchemas = Observable.fromIterable(modelProvider.modelSchemas().values())
+            // Don't subscribe to models that have an associated syncExpression of QueryPredicates.none().
+            .filter(schema -> !QueryPredicates.none().equals(queryPredicateProvider.getPredicate(schema.getName())))
```
jamesonwilliams, line 116: It's getting to be a long line -- 

```suggestion
            .filter(schema -> {
                String name = schema.getName();
                QueryPredicate predicate = queryPredicateProvider.getPredicate(name);
                return !QueryPredicates.none().equals(predicate);
            })
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
richardmcclellan: > Related question: Are sync predicates for outgoing mutations, today? Or just inbound?

No, they are not applied at all for mutations today.   Today, sync predicates are applied to the sync query (on the server side), and subscriptions (today, on the client side, but in the future, server side, when AppSync support it)

> Seems actually like this is not a unique problem to the .none() filter?

This is a really good point that I've been thinking about.  This PR only filters mutations if the syncExpression is set to `.none()`.  We really shouldn't treat `.none()` as a special case though.  We have a design tenet that going something like, "Before adding support for a special case, think twice, because you probably don't _really_ need it."

Then the question becomes - should we apply all sync expressions to mutations, or not?     I actually think we should _not_.  The primary use case for sync expressions is to speed up the sync process and reduce data usage by reducing the total data that is actually synced.   Customers should not expect all of the data in DataStore to evaluate to true against the sync expressions.  For example, a customer starts DataStore with syncExpression A, then stops and starts again with syncExpression B, the local database will contain the union of A & B.  It is not intended to be used as an implicit filter on `DataStore.query`, even though it may sometimes work that way.  `DataStore.query` takes a `QueryPredicate` specifically for that purpose.

The one reason _to_ apply mutations to sync expressions is to support the use case where a customer wants some of their models to sync with their API, and others to just use to save data locally.  I don't think we should invest in supporting this use case though, because it feels outside the scope of Amplify. Customers wanting to save data locally should just use their own SQLite DB, or any other Android library for local persistence.

TL;DR: I'm going to remove the mutation filtering from this PR.

jamesonwilliams: Sounds good! Another (rather hairy) aspect to this:

If we don't do a server-to-device sync for some models, we won't know their correct version. Now, let's say we *do* allow device-to-server sync ("mutations") for those same models. The mutations will go out with incorrect versions. In the best case (optimistic concurrency) the mutations will fail with an error and the conflict handler will be invoked. But in the 80% case (Auto Merge), the mutation will fail silently and we will lose customer data without any error being raised.

richardmcclellan: Hmm, that is a good point.  I believe that scenario is a problem today, irrespective of this PR. 

1. Client 1 starts DataStore with sync predicate X
2. Client 1 creates item A, version 1, which does **not** match sync predicate X.
3. Client 2 receives item A, version 1, via a subscription.
4. Client 2 updates item A, incrementing `_version` to 2.
5. Client 1 now has stale version of item A,  _version 1
5. Client 1 updates item A, version 1.  For Auto Merge, the mutation will fail silently, and we will lose customer data.

To solve this, maybe we _should_ actually have a filter in the `MutationProcessor`?  And if the mutation doesn't match the sync predicate, then, we could fire an `outboxMutationFailed` event, containing the item, to prevent customer data loss.  Currently, `outboxMutationFailed` is only fired for API errors, but perhaps we could also use it for this use case.

jamesonwilliams: That sounds potentially like a mitigation, yes. Maybe we should also talk about the silent Auto Merge failure w/ the DataStore Working Group / Jonny P? I'm worried that it may be causing errors that we're not even aware of -- although, we've identified a few failure modes, to date. Perhaps we can optionally turn on "strict version matching" via a request header or something?

PR #1129, opened 2021-01-25
fix(datastore) Remove timeouts in SyncProcessor and SubscriptionProcessor

richardmcclellan:
This PR removes the timeouts in the `SyncProcessor` and `SubscriptionProcessor` when DataStore is started.  I don't think these timeouts are really necessary because for the sync queries, `OkHttp` already has a default timeout of 10 seconds per network request, which I believe is a reasonable default.  Additionally, `SubscriptionEndpoint` times out automatically if it doesn't receive a `start_ack` within 10 seconds.  I don't think we should be adding timeout on top of these.

In a sample app, I created ~600 model objects, and was unable to use DataStore, because the sync process timed out every time (after a 10 second timeout). I had the syncPageSize set to 100 items, so 6 network requests were required to fetch all pages.  Each request completed in ~1 second.  Given this, you would think that the sync would complete in around 6 seconds, or certainly within 10 seconds, but it actually takes ~20 seconds.  I believe the majority of this time is due to inefficiencies in saving the data to disc locally.  For each item, a query is done to read the current `ModelMetadata` table, and determine if the new version is newer, before saving that item.  This is quite expensive.

We should definitely spend some time improving performance, but in the meantime, this PR removes the timeouts so that DataStore is at least useable with larger (500+ model instances) datasets.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/AbortableCountDownLatch.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/AbortableCountDownLatch.java
@@ -41,6 +41,13 @@ final class AbortableCountDownLatch<E extends Throwable> extends CountDownLatch 
         }
     }
 
+    public void abortableAwait() throws InterruptedException, E {
+        super.await();
+        if (error != null) {
+            throw error;
+        }
+    }
+
     public boolean abortableAwait(long timeout, TimeUnit unit) throws InterruptedException, E {
         final boolean success = super.await(timeout, unit);
         if (error != null) {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -52,8 +52,6 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
  */
 public final class Orchestrator {
     private static final Logger LOG = Amplify.Logging.forNamespace("amplify:aws-datastore");
-    private static final long TIMEOUT_SECONDS_PER_MODEL = 2;
-    private static final long NETWORK_OP_TIMEOUT_SECONDS = 10;
     private static final long LOCAL_OP_TIMEOUT_SECONDS = 2;
 
     private final SubscriptionProcessor subscriptionProcessor;
@@ -65,8 +63,8 @@ public final class Orchestrator {
     private final AtomicReference<State> currentState;
     private final MutationOutbox mutationOutbox;
     private final CompositeDisposable disposables;
-    private final long adjustedTimeoutSeconds;
     private final Semaphore startStopSemaphore;
+    private Long syncStartTime;
 
     /**
      * Constructs a new Orchestrator.
@@ -133,12 +131,6 @@ public final class Orchestrator {
         this.targetState = targetState;
         this.disposables = new CompositeDisposable();
 
-        // Operation times out after 10 seconds. If there are more than 5 models,
-        // then 2 seconds are added to the timer per additional model count.
-        this.adjustedTimeoutSeconds = Math.max(
-            NETWORK_OP_TIMEOUT_SECONDS,
-            TIMEOUT_SECONDS_PER_MODEL * modelProvider.models().size()
-        );
         this.startStopSemaphore = new Semaphore(1);
 
     }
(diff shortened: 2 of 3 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/AbortableCountDownLatch.java
```diff
@@ -41,6 +41,13 @@ public void abort(E error) {
         }
     }
 
+    public void abortableAwait() throws InterruptedException, E {
+        super.await();
```
jamesonwilliams, line 45: ⚠️

I think we should use a default timeout. If we wait indefinitely, we may hang the test harness with no clear rationale provided.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -65,8 +63,8 @@
     private final AtomicReference<State> currentState;
     private final MutationOutbox mutationOutbox;
     private final CompositeDisposable disposables;
-    private final long adjustedTimeoutSeconds;
     private final Semaphore startStopSemaphore;
+    private Long syncStartTime;
```
jamesonwilliams, line 67: Little `l`?
```suggestion
    private long syncStartTime;
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -335,8 +323,10 @@ private void startApiSync() {
 
                 emitter.onComplete();
             })
+            .doOnSubscribe(disposable -> syncStartTime = System.currentTimeMillis())
             .doOnError(error -> LOG.error("Failure encountered while attempting to start API sync.", error))
-            .doOnComplete(() -> LOG.info("Started the orchestrator in API sync mode."))
+            .doOnComplete(() -> LOG.info("Started the orchestrator in API sync mode in "
+                + (System.currentTimeMillis() - syncStartTime) + "ms"))
```
jamesonwilliams, line 329:
```suggestion
            .doOnComplete(() -> {
                private long elapsedTimeMs = System.currentTimeMillis() - syncStartTime
                LOG.info("Started the orchestrator in API sync mode in " + elapsedTimeMs + "ms")
            })
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -134,25 +123,20 @@ synchronized void startSubscriptions() throws DataStoreException {
             .subscribe(buffer::onNext, buffer::onError, buffer::onComplete)
         );
 
-        boolean subscriptionsStarted;
         try {
             LOG.debug("Waiting for subscriptions to start.");
-            subscriptionsStarted = latch.abortableAwait(adjustedTimeoutSeconds, TimeUnit.SECONDS);
+            latch.abortableAwait();
```
jamesonwilliams, line 128: Same concern here, as stated above. This has the potential to hang the sync engine, and without any error rendered.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/AbortableCountDownLatch.java
richardmcclellan: That's a fair point.  I'm good with leaving the timeout here, for now, but would propose we raise it from 10 seconds to 60 seconds or so.  

This timeout is what limits the number of items that can be synced with `DataStore`, which is bad, so I do think we need a better solution.  Another approach would be to split the timeout into each of the Processors, rather than the `Orchestrator`.  Or perhaps, we could use a watchdog approach, where the timeout is reset whenever a new page of models has been returned from the server.

jamesonwilliams: I think the timeouts for the subscription and sync processors might warrant different values. The sync can be very long, but setting up the subscriptions ought to be shorter. I'd always envisioned that we'd make the sync timeout user-configurable. I agree that 10 seconds is too short of a default. 60 does sound like a better starting place. I think our hearts were in the right place with the `(number of model types) * value` approach, but in practice I'm not sure it has helped much.

richardmcclellan:
Closing in favor of https://github.com/aws-amplify/amplify-android/pull/1161

Issue #908, opened 2020-10-16
message=Error in querying the model., cause=android.database.sqlite.SQLiteException: ambiguous column name: id (code 1 SQLITE_ERROR)


shirui0413:
This is my schema  ↓↓↓

```graphql
type User @model {
  id: ID!
  password: String!
  card: [Card] @connection(name: "createdUser")
}

type Card @model {
  id: ID!
  user: User @connection(name: "createdUser")
  date: AWSDateTime!
  category: Int!
  goalItem: Int
  goalValue: Float
  achieve: Float
  reaction: [Reaction] @connection(name: "getReaction")
}

type Reaction @model {
  id: ID!
  card: Card @connection(name: "getReaction")
  category: Int!
  item: Int
}
```

When I used the ID of card as a predicate, there is an error.   ↓↓
![image](https://user-images.githubusercontent.com/27470444/96240619-15ece300-0fdc-11eb-9c2c-a9637d05f125.png)

When I use another field for the predicate, eg. `User` of `Card` as a keyword, there is no problem:
![image](https://user-images.githubusercontent.com/27470444/96241359-015d1a80-0fdd-11eb-9bb8-379603c7e9e9.png)

I don't know why. Who can help me?

raphkim:
Hi @shirui0413 ,

There is currently an issue with the way query works with relational models. Please see the [`**NOTE**` in the following doc](https://docs.amplify.aws/lib/datastore/data-access/q/platform/android#predicates) to work around this issue.

Rather than using `Card.ID` and `Reaction.ID`, please try using `QueryField.field("card.id")` and `QueryField.field("reaction.id")`

jamesonwilliams:
Could we resolve this by changing [this line](https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java#L418):
```java
.append(ownedTable.getPrimaryKeyColumnName())
```
to:
```java
.append(ownedTable.getName() + "_" + ownedTable.getPrimaryKeyColumnName())
```
?

richardmcclellan:
The workaround posted above by @raphkim should work, though I am re-opening this issue to track fixing the actual issue, which requires updating the code that generates our Java models, as @jamesonwilliams said.

upachecog:
@richardmcclellan I found an issue in **version 1.6.10.**

When I upgraded the version my queries stopped working:

[...]

PR #1130, opened 2021-01-25
fix(datastore): dont use where.id() function

raphkim:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/908#issuecomment-766284487

*Description of changes:*
Using `Where.id()` internally to match model ID broke query function for models that had `@BelongsTo` relationship with another model. This PR reverts that change and adds a test case.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
@@ -852,11 +852,17 @@ public final class SQLiteStorageAdapter implements LocalStorageAdapter {
      * @return the Model instance from SQLite, if it exists, otherwise null.
      */
     private Model query(Model model) {
+        final String modelName = getModelName(model);
+        final ModelSchema schema = modelSchemaRegistry.getModelSchemaForModelClass(modelName);
+        final SQLiteTable table = SQLiteTable.fromSchema(schema);
+        final String primaryKeyName = table.getPrimaryKeyColumnName();
+        final QueryPredicate matchId = QueryField.field(primaryKeyName).eq(model.getId());
+
         Iterator<? extends Model> result = Single.<Iterator<? extends Model>>create(emitter -> {
             if (model instanceof SerializedModel) {
-                query(getModelName(model), Where.id(model.getId()), emitter::onSuccess, emitter::onError);
+                query(modelName, Where.matches(matchId), emitter::onSuccess, emitter::onError);
             } else {
-                query(model.getClass(), Where.id(model.getId()), emitter::onSuccess, emitter::onError);
+                query(model.getClass(), Where.matches(matchId), emitter::onSuccess, emitter::onError);
             }
         }).blockingGet();
         return result.hasNext() ? result.next() : null;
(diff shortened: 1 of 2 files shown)


richardmcclellan (approved):
🚢

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -852,11 +852,17 @@ private boolean modelExists(Model model, QueryPredicate predicate) throws DataSt
      * @return the Model instance from SQLite, if it exists, otherwise null.
      */
     private Model query(Model model) {
+        final String modelName = getModelName(model);
+        final ModelSchema schema = modelSchemaRegistry.getModelSchemaForModelClass(modelName);
+        final SQLiteTable table = SQLiteTable.fromSchema(schema);
+        final String primaryKeyName = table.getPrimaryKeyColumnName();
+        final QueryPredicate matchId = QueryField.field(primaryKeyName).eq(model.getId());
```
jamesonwilliams, line 859: Seems like we always have to build out some scaffolding to parse the model. Maybe we need a helper data structure. Could be like `SqlStorageData data = SqlStorageData.from(model)`? Not for this PR, of course.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -852,11 +852,17 @@ private boolean modelExists(Model model, QueryPredicate predicate) throws DataSt
...
+
         Iterator<? extends Model> result = Single.<Iterator<? extends Model>>create(emitter -> {
             if (model instanceof SerializedModel) {
-                query(getModelName(model), Where.id(model.getId()), emitter::onSuccess, emitter::onError);
+                query(modelName, Where.matches(matchId), emitter::onSuccess, emitter::onError);
             } else {
-                query(model.getClass(), Where.id(model.getId()), emitter::onSuccess, emitter::onError);
+                query(model.getClass(), Where.matches(matchId), emitter::onSuccess, emitter::onError);
```
jamesonwilliams, line 865: How are they different, internally?

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: `Where.id` uses hardcoded `"id"` as column name when building query statement, which causes ambiguous column name SQLiteException. `SQLiteColumn#getPrimaryKeyColumnName()` appends column name to the table name like following: `<model>.<column>`.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
raphkim: I agree! Building a whole `SQLiteTable` instance from model schema every time a model is being queried or saved is such an unnecessary amount of work. It was meant to be a temporary solution while in a rush to release it a year ago, but it definitely doesn't cut it now.

PR #1133, opened 2021-01-26
fix(core): query field now stores model name

raphkim:
*Issue #, if available:*

*Description of changes:*
Add an additional parameter to `QueryField` instance called `modelName` so that SQL queries can be constructed without ambiguous column names.

This PR aims to address #908 without introducing breaking changes for existing customers. While this PR itself does not yet fix the issue, it allows for model codegen to be enhanced to support better usage of built-in query fields from code-generated models. This way, the enhanced codegen will fix the issue and allow all customers to rely directly on the built-in query fields instead of needing to construct their own.

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
+++ b/core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
@@ -34,7 +34,7 @@ import java.util.Objects;
  * {@link QueryField#descending()} helper methods (e.g. Todo.DESCRIPTION.ascending() or Todo.DESCRIPTION.descending())
  */
 public final class QuerySortBy {
-    private final String model;
+    private final String modelName;
     private final String field;
     private final QuerySortOrder sortOrder;
 
@@ -51,12 +51,12 @@ public final class QuerySortBy {
     /**
      * Constructor for {@code QuerySortBy}.
      *
-     * @param model name of the model being sorted.
+     * @param modelName name of the model being sorted.
      * @param field name of field to sort by.
      * @param sortOrder order to sort by, either ASCENDING or DESCENDING.
      */
-    public QuerySortBy(@Nullable String model, @NonNull String field, @NonNull QuerySortOrder sortOrder) {
-        this.model = model;
+    public QuerySortBy(@Nullable String modelName, @NonNull String field, @NonNull QuerySortOrder sortOrder) {
+        this.modelName = modelName;
         this.field = Objects.requireNonNull(field);
         this.sortOrder = Objects.requireNonNull(sortOrder);
     }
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -32,6 +34,7 @@
  * {@link QueryField#descending()} helper methods (e.g. Todo.DESCRIPTION.ascending() or Todo.DESCRIPTION.descending())
  */
 public final class QuerySortBy {
+    private final String model;
```
jamesonwilliams, line 37:
```suggestion
    private final String modelName;
```
This is the identifier usually used for this concept elsewhere in the code base 

Same update suggested for all uses

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -73,19 +99,21 @@ public boolean equals(Object object) {
...
     public int hashCode() {
-        return ObjectsCompat.hash(field, sortOrder);
+        return ObjectsCompat.hash(model, field, sortOrder);
     }
 
     @Override
     public String toString() {
         return "QuerySortBy{" +
-                "field='" + field + '\'' +
+                "model=" + (model == null ? null : Wrap.inSingleQuotes(model)) +
```
jamesonwilliams, line 115: This is a clever and welcome use of the Wrap utility.

However it looks weird to only use it on one line and not the others.

I would update all lines to use this style, or stick with the existing approach for this new line.

jamesonwilliams (approved):
Nice @raphk! This is a tasteful solution. Appreciate you thinking through this and coming up with an incremental approach that won't break customers.

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
raphkim: the fields on the other lines are `@NonNull`, so I figured there isn't a need to do that.

core/src/main/java/com/amplifyframework/core/model/query/QuerySortBy.java
```diff
@@ -32,6 +34,7 @@
  * {@link QueryField#descending()} helper methods (e.g. Todo.DESCRIPTION.ascending() or Todo.DESCRIPTION.descending())
  */
 public final class QuerySortBy {
+    private final String modelName;
```
richardmcclellan, line 37: Rather than add a `String modelName` property, what do you think about replacing the `String field` property with a `QueryField queryField` property (which contains the `modelName`, and `fieldName`)?   

Maybe slightly cleaner, though not a huge deal either way as this is just an internal implementation detail.

PR #1141, opened 2021-01-29
feat(datastore): support sorting by related models

raphkim:
*Issue #, if available:*

*Description of changes:*
DataStore's `query` supports filtering by field of a related model (e.g. "Filter `Blog` by `BlogOwner`'s name"). However, we don't currently support sorting queried result by related model (e.g. "Sort `Blog` by `Blog`'s name" is *supported* but "Sort `Blog` by `BlogOwner`'s name" is *not*). This PR uses new `QueryField` property to allow this.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
@@ -17,6 +17,7 @@ package com.amplifyframework.datastore.storage.sqlite;
 
 import com.amplifyframework.core.model.query.Page;
 import com.amplifyframework.core.model.query.Where;
+import com.amplifyframework.core.model.query.predicate.QueryField;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.DataStoreException;
 import com.amplifyframework.datastore.StrictMode;
@@ -39,6 +40,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import io.reactivex.rxjava3.core.Observable;
 
@@ -414,6 +416,38 @@ public final class SQLiteStorageAdapterQueryTest {
                 .reversed()
                 .thenComparing(BlogOwner::getWea)
         );
+        assertEquals(sorted, result);
+    }
+
+    /**
+     * Test query with order by.  Validate that a list of Blog can be sorted by the names of BlogOwners.
+     * @throws DataStoreException On failure to arrange items into store, or from the query action itself
+     */
+    @Test
+    public void queryWithOrderByRelatedModel() throws DataStoreException {
+        // Expect: Create BlogOwners and their respective blogs
+        List<String> names = Arrays.asList("Joe", "Bob", "Dan", "Jane");
+        List<Blog> blogs = new ArrayList<>();
+
+        for (String name : names) {
+            BlogOwner owner = BlogOwner.builder()
+                    .name(name)
+                    .build();
+            adapter.save(owner);
+            Blog blog = Blog.builder()
+                    .name("")
+                    .owner(owner)
+                    .build();
+            adapter.save(blog);
+            blogs.add(blog);
+        }
+
+        // Act: Query Blogs sorted by owner's name
+        List<Blog> result = adapter.query(Blog.class, Where.sorted(BlogOwner.NAME.ascending()));
+
+        // Verify: Query result is sorted by owner's name
+        List<Blog> sorted = new ArrayList<>(blogs);
+        Collections.sort(sorted, Comparator.comparing(blog -> blog.getOwner().getName()));
         assertEquals(sorted, result);
     }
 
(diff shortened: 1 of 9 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -220,8 +220,13 @@ public SqlCommand queryFor(@NonNull ModelSchema modelSchema,
             Iterator<QuerySortBy> sortByIterator = sortByList.iterator();
             while (sortByIterator.hasNext()) {
                 final QuerySortBy sortBy = sortByIterator.next();
-                SQLiteColumn column = table.getColumn(sortBy.getField());
-                rawQuery.append(Wrap.inBackticks(column.getAliasedName()))
+                String modelName = Wrap.inBackticks(sortBy.getModelName());
```
richardmcclellan, line 223: Will this work if a custom is using models generated the old way?   Don't we need to default this to the current modelName being queried, if `sortBy.getModelName()` is null?

testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/Comment.java
```diff
@@ -20,9 +20,9 @@
 @SuppressWarnings("all")
 @ModelConfig(pluralName = "Comments")
 public final class Comment implements Model {
-  public static final QueryField ID = field("id");
-  public static final QueryField CONTENT = field("content");
-  public static final QueryField POST = field("commentPostId");
+  public static final QueryField ID = field("Comment", "id");
```
jamesonwilliams, line 23: Do you still have some test coverage older the old, `null` modelName code path, now?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
raphkim: Yep, there is a null-check on `modelName` right after this logic to default to the current model right after this :)

testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/Comment.java
raphkim: I only updated this particular package, but I agree that it's ideal to test that path as well. Will fix.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
richardmcclellan: Oh!  Not sure why I missed that the first time!  Looks good.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -449,6 +449,39 @@ public void queryWithOrderByRelatedModel() throws DataStoreException {
...
 
+    /**
+     * Test that new QueryField with explicit model name produces the same result as old QueryField.
+     * @throws DataStoreException On failure to arrange items into store, or from the query action itself
+     */
+    @Test
+    public void queryFieldsAreBackwardsCompatible() throws DataStoreException {
+        BlogOwner blogOwner = BlogOwner.builder().name("Test Dummy").build();
+        adapter.save(blogOwner);
+        Blog blog = Blog.builder().name("Blogging for Dummies").owner(blogOwner).build();
```
jamesonwilliams, line 460: haha

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -449,6 +449,39 @@ public void queryWithOrderByRelatedModel() throws DataStoreException {
...
+                    .blog(blog)
+                    .build();
+            adapter.save(post);
+        }
+
+        // Assert that using QueryField without model name yields same results if there is no column ambiguity
+        assertEquals(
+            adapter.query(Post.class, Where.matches(field("Post", "title").contains("4"))),
+            adapter.query(Post.class, Where.matches(field("title").contains("4")))
+        );
```
jamesonwilliams, line 478: There's probably some cool way to start from just the query fields, and then build towards an `allSame()` predicate. I've got a start here but this still isn't exactly it....

```java
Observable.fromArray(field("Post", 'title), field("title))
    .map(queryField -> Where.matches(queryField).contains("4"))
    .toList()
    .map(HashSet::new)
    .awaitCount(1)
    .assertValue(set -> set.size() == 0);
```

jamesonwilliams (approved):

richardmcclellan (approved):

testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/Comment.java
jamesonwilliams: Thanks!

jamesonwilliams (approved):

Issue #1142, opened 2021-01-30
notContains not available in QueryPredicate


anandanwar4:
The documentation says that there exists a `QueryPredicate` called `notContains`. But `notContains` cannot be resolved as one of the functions of `QueryField`.

**Version Info**:
```gradle
implementation 'com.amplifyframework:core:1.16.11'
implementation 'com.amplifyframework:aws-datastore:1.16.11'
implementation 'androidx.multidex:multidex:2.0.0'
```
**Code** :
```java
Where.matches(SampleTable.SAMPLE_LIST.notContains("sampleId"))
```

**Error** :
> Error: Cannot resolve method 'notContains' in 'QueryField'

**Reference Documentation**:
https://docs.amplify.aws/start/getting-started/integrate/q/integration/android#query-todos

**schema.graphql**:
```graphql
type SampleTable @model @key(fields: ["id"]) {
  id: ID!
  name: String!
  sampleList: [ID!]
}
```

raphkim:
Hi @anandanwar4 ,

Will using `QueryPredicateOperation#not()` method to negate `contains()` operator work for your use-case?
```
Where.matches(not(SampleTable.SAMPLE_LIST.contains("sampleId")))
```

anandanwar4:
Hi @raphkim ,

Yes that should do it. Will use this then. Thanks.

Regards,
Ankit

PR #1145, opened 2021-02-01
feat(datastore) add support for notContains query operator

richardmcclellan:
Adds support for `notContains` query operator.  [Our documentation](https://docs.amplify.aws/start/getting-started/integrate/q/integration/android#query-todos) indicates that we do support this, but it looks like it was missed during the initial implementation.   This PR adds support for it.

On amplify-ios, it is also [not supported](https://github.com/aws-amplify/amplify-ios/blob/e2f7f64e2a501f3b2807481ca921e71710c82b22/Amplify/Categories/DataStore/Query/QueryField.swift#L34-L44), so I opened https://github.com/aws-amplify/amplify-ios/issues/1049.
On amplify-js, it [is supported](https://github.com/aws-amplify/amplify-js/blob/6e28eaf37525ce231d7793bf82a960046fc7f8f4/packages/datastore/src/types.ts#L295-L299).  

Note: I also discovered there are 3 other AppSync operators not supported by Amplify: `attributeExists`, `attributeType`, and `size`, though I don't think we need to add support for them unless there is a specific customer use case or request.

Resolves https://github.com/aws-amplify/amplify-android/issues/1142

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/core/model/query/predicate/GsonPredicateAdapters.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/core/model/query/predicate/GsonPredicateAdapters.java
@@ -64,6 +64,8 @@ public final class GsonPredicateAdapters {
             switch (QueryOperator.Type.valueOf(operatorType)) {
                 case CONTAINS:
                     return context.deserialize(json, ContainsQueryOperator.class);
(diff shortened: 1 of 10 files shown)


jamesonwilliams (approved):
Beautiful stuff!

richardmcclellan:
Integration tests passed locally.

Issue #1146, opened 2021-02-02
Amplify.DataStore.delete not working


anandanwar4:
An exception is thrown when trying to delete a record from database.

**Version Info**:
```
implementation 'com.amplifyframework:core:1.16.11'
implementation 'com.amplifyframework:aws-datastore:1.16.11'
implementation 'androidx.multidex:multidex:2.0.1'
```

**Code**:
```
    @Override
    public void deleteItem(Item itemToBeDeleted, Consumer<Item> deletionSuccessAction,
                           Consumer<Throwable> deletionFailureAction) {
        Amplify.DataStore.delete(itemToBeDeleted,
                deletedItem -> deletionSuccessAction.accept(deletedItem.item()),
                deletionFailureAction::accept
        );
    }
```

**Error Log**:
```
E/SQLiteLog: (1) no such column: Item.id
    DataStoreException{message=Error in deleting the model., cause=android.database.sqlite.SQLiteException: no such column: Item.id (code 1): , while compiling: SELECT `Category`.`id` AS `Category_id`, `Category`.`isCustom` AS `Category_isCustom`, `Category`.`name` AS `Category_name` FROM `Category` WHERE `Item`.`id` = ?;, recoverySuggestion=See attached exception for details.}
        at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.lambda$delete$6$SQLiteStorageAdapter(SQLiteStorageAdapter.java:523)
        at com.amplifyframework.datastore.storage.sqlite.-$$Lambda$SQLiteStorageAdapter$XngSjLlcHwo0KebM60Fw5B2MrsM.run(Unknown Source:12)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:457)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: android.database.sqlite.SQLiteException: no such column: Item.id (code 1): , while compiling: SELECT `Category`.`id` AS `Category_id`, `Category`.`isCustom` AS `Category_isCustom`, `Category`.`name` AS `Category_name` FROM `Category` WHERE `Item`.`id` = ?;
[...]

PR #1149, opened 2021-02-03
fix(datastore): ignore sqlite exceptions during cascading delete

raphkim:
*Issue #, if available:*

*Description of changes:*
Ignore `SQLiteException` during cascading delete for general stability. (local deletions will still be carried out, but remote deletions are cut short)


richardmcclellan:
This change looks good!  Could you also add a unit test for the scenario that this fixes?

jamesonwilliams (commented):
> Ignore SQLiteException during cascading delete for general stability. (local deletions will still be carried out, but remote deletions are cut short)

Can you expand upon this a bit more in the PR description? Is it not possible to try enqueuing the remote deletion, even if the local one fails?

Also, wouldn't it be better to fire an error on the `onError` callback, if there is a failure to delete? Otherwise, the customer could be given the impression that the content was deleted locally and remotely, when perhaps neither is true.

raphkim:
> Also, wouldn't it be better to fire an error on the onError callback, if there is a failure to delete? Otherwise, the customer could be given the impression that the content was deleted locally and remotely, when perhaps neither is true.

The items that customer directly ordered the deletion of _will_ be guaranteed to be deleted both locally and remotely if they receive the onResult callback. However, this implementation does not guarantee that all of the cascaded deletes were remotely sent. 
The philosophy of giving "best effort" to make sure that there is no orphaned items in cloud, but being "okay" if there are some is described [here](https://github.com/aws-amplify/amplify-android/pull/1058).

> Could you also add a unit test for the scenario that this fixes?

There is currently no specific issue that this PR addresses directly. Because we are "okay" with leaving orphaned items in remote storage, we simply decide to ignore any unexpected error that results while computing model relationship.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -140,9 +141,9 @@ private void recurseTree(
                                 childrenIds.add(cursor.getString(index));
                             } while (cursor.moveToNext());
                         }
-                    } catch (DataStoreException exception) {
+                    } catch (DataStoreException | SQLiteException exception) {
```
richardmcclellan, line 144: Should this be a more specific SQLite foreign key constraint exception?

richardmcclellan:
> There is currently no specific issue that this PR addresses directly. Because we are "okay" with leaving orphaned items in remote storage, we simply decide to ignore any unexpected error that results while computing model relationship.

Now I'm a little confused.  If SQL's `ON DELETE CASCADE` is smart enough to identify the items that need to be deleted locally, why would we not be able to identify them in a similar way for the remote deletion?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
raphkim: This aims to ignore _any_ unexpected exception that can be thrown here, so if anything I would rather catch generic `Exception` here, but that would be code smell...
For example:
* Any SQLite specific behavior that was unexpected
  * 1000 variable/expression limit that we didn't know about would be an example of this (#1121 aims to fix this explicitly though)
* Incorrectly formatted graphQL schema that generated malformed models
  * #1146 would be an example of this. If customers started off with malformed models that happened to work out, we don't want to suddenly break their code by catching them here

raphkim:
> Now I'm a little confused. If SQL's `ON DELETE CASCADE` is smart enough to identify the items that need to be deleted locally, why would we not be able to identify them in a similar way for the remote deletion?

SQLite does seem to be pretty efficient at discovering all related models... but I wasn't able to find a way to extract this information

richardmcclellan:
> > Could you also add a unit test for the scenario that this fixes?
> 
> There is currently no specific issue that this PR addresses directly. Because we are "okay" with leaving orphaned items in remote storage, we simply decide to ignore any unexpected error that results while computing model relationship.


Doesn't this PR specifically address https://github.com/aws-amplify/amplify-android/issues/1146?  Couldn't we add a test case for it?

raphkim:
> > > Could you also add a unit test for the scenario that this fixes?
> > 
> > 
> > There is currently no specific issue that this PR addresses directly. Because we are "okay" with leaving orphaned items in remote storage, we simply decide to ignore any unexpected error that results while computing model relationship.
> 
> Doesn't this PR specifically address #1146? Couldn't we add a test case for it?

This PR is not a direct solution to #1146. I do not think that a test case to see if deliberately breaking the schema will not crash cascading delete does not provide much value.

richardmcclellan (approved):
🚢  Approved, under the condition that we follow up by file a codegen bug to automatically generate bi-directional models, even if the customer provided a uni-directional schema.

raphkim:
https://github.com/aws-amplify/amplify-codegen/issues/102
Addressing above codegen error will eliminate a major error case that this PR "ignores"

PR #1153, opened 2021-02-05
Fastlane deployment scripts for amplify-android

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR creates two Fastlane deployment lanes for android. One for unstable builds and one for release builds. 

Unstable builds can be executed by running:

## Unstable lane
```
bundle exec fastlane android unstable
```

The `android unstable` lane automatically determines a new version number by looking at the latest git tag. If the latest tag has a format of `release_v<major>.<minor>.<patch>`, it increments patch and creates a version called `release_v<major>.<minor>.<patch+1>-unstable.0`. If the format is `release_v<major>.<minor>.<patch>-unstable.X` it creates a new version by incrementing X. 

## Release lane
```
bundle exec fastlane android release
```
The `android release` lane automatically determines the version by pulling the latest release tag (excluding unstable tags). It looks at the commit history and applies the following logic:
- If there are breaking changes (denoted by ! in the slub or BREAKING CHANGE in the commit body), it increments the major version.
- If there are `feat` commits it increments the minor version
- Otherwise, it increments patch.

## Push to maven
The last step is to push the release to the maven staging repo for manual approval.


richardmcclellan:
> If there are breaking changes (denoted by ! in the slub or BREAKING CHANGE in the commit body), it increments the major version.
If there are feat commits it increments the minor version
Otherwise, it increments patch.

Is this what we want?  I believe what we had discussed is this:
> Never increment major.  If there are breaking changes, increment minor, otherwise increment patch.

rjuliano:
> > If there are breaking changes (denoted by ! in the slub or BREAKING CHANGE in the commit body), it increments the major version.
> > If there are feat commits it increments the minor version
> > Otherwise, it increments patch.
> 
> Is this what we want? I believe what we had discussed is: Never increment major. If there are breaking changes, increment minor, otherwise increment patch.

I'm working on moving that logic some place where we can override it. It's currently shared by iOS. Right now, I don't think we mark anything as a breaking change, but I agree that we should be explicit about it.

richardmcclellan:
> > > If there are breaking changes (denoted by ! in the slub or BREAKING CHANGE in the commit body), it increments the major version.
> > > If there are feat commits it increments the minor version
> > > Otherwise, it increments patch.
> > 
> > 
> > Is this what we want? I believe what we had discussed is: Never increment major. If there are breaking changes, increment minor, otherwise increment patch.
> 
> I'm working on moving that logic some place where we can override it. It's currently shared by iOS. Right now, I don't think we mark anything as a breaking change, but I agree that we should be explicit about it.

Oh, I see, I didn't realize it was shared with iOS.  Ideally, we should just use the same versioning strategy as iOS.  Thinking about it now, I think that strategy might actually work.   Our mandate is _not_ to make breaking changes without incrementing the major version anyway.

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,165 @@
...
+                                                release_tag_prefix: RELEASE_TAG_PREFIX,
+                                                from_tag: latest_tag )
+    UI.message("Next version: #{next_version}")
+
+    # Update version information in gradle.properties
+    increment_versions(version: next_version)
+
+    gradle(
+      project_dir: gradle_project_root,
+      task: "assemble",
```
richardmcclellan, line 25: Should this be `build`, or `clean build`?  `assemble` doesn't run unit tests, which `build` does.
```suggestion
      task: "build",
```

richardmcclellan:
This looks like a great approach overall!  One thing I don't understand is - what actually triggers a release?

rjuliano:
> This looks like a great approach overall! One thing I don't understand is - what actually triggers a release?

Release triggers are not setup yet, so in the short term, the goal would be to replace the manual steps we perform today, with the execution of one of the Fastlane lanes.

I was thinking we could trigger unstable releases every time something gets merged into main, but wanted to discuss with the team to see what everybody thought.

richardmcclellan:
> Release triggers are not setup yet, so in the short term, the goal would be to replace the manual steps we perform today, with the execution of one of the Fastlane lanes.

:+1:

> I was thinking we could trigger unstable releases every time something gets merged into main, but wanted to discuss with the team to see what everybody thought.

Would unstable releases still require someone to manually login to Sonatype to hit the release button?  If so, that seems a bit cumbersome.

rjuliano:
> > Release triggers are not setup yet, so in the short term, the goal would be to replace the manual steps we perform today, with the execution of one of the Fastlane lanes.
> 
> 👍
> 
> > I was thinking we could trigger unstable releases every time something gets merged into main, but wanted to discuss with the team to see what everybody thought.
> 
> Would unstable releases still require someone to manually login to Sonatype to hit the release button? If so, that seems a bit cumbersome.

Correct...that's the part the gives me a little pause. I believe we can change so it publishes directly, but I'm not sure we want to do that right off the bat.

rjuliano:
Closing this one in favor of an updated version

PR #1154, opened 2021-02-08
fix(core): Improve error messaging when category configuration or initialization fails

richardmcclellan:
Currently, when a customer tries to use a plugin that is not configured, they get this error:

`Tried to get a plugin before it was configured.  Make sure you call Amplify.configure() first.`

This error makes sense if the current state is `NOT_CONFIGURED` or `CONFIGURING`.  However, if the state is `FAILED`, it could be because either configuration or initialization failed, so we should throw a more useful error.

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/category/Category.java
+++ b/core/src/main/java/com/amplifyframework/core/category/Category.java
@@ -33,6 +33,7 @@ import org.json.JSONObject;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -45,7 +46,6 @@ import java.util.concurrent.atomic.AtomicReference;
  *            support, e.g. StoragePlugin, AnalyticsPlugin, etc.
  */
 public abstract class Category<P extends Plugin<?>> implements CategoryTypeable {
-
     /**
      * Map of the { pluginKey => plugin } object.
      */
@@ -56,12 +56,15 @@ public abstract class Category<P extends Plugin<?>> implements CategoryTypeable 
      */
     private final AtomicReference<State> state;
 
+    private final AtomicReference<CategoryInitializationResult> categoryInitializationResult;
+
     /**
      * Constructs a new, not-yet-configured, Category.
      */
     public Category() {
         this.plugins = new ConcurrentHashMap<>();
         this.state = new AtomicReference<>(State.NOT_CONFIGURED);
+        this.categoryInitializationResult = new AtomicReference<>(null);
     }
 
     /**
@@ -89,7 +92,7 @@ public abstract class Category<P extends Plugin<?>> implements CategoryTypeable 
                 }
                 state.set(State.CONFIGURED);
             } catch (Throwable anyError) {
-                state.set(State.FAILED);
+                state.set(State.CONFIGURATION_FAILED);
                 throw anyError;
             }
         }
@@ -130,8 +133,9 @@ public abstract class Category<P extends Plugin<?>> implements CategoryTypeable 
 
         final CategoryInitializationResult result =
             CategoryInitializationResult.with(pluginInitializationResults);
+        categoryInitializationResult.set(result);
         if (result.isFailure()) {
-            state.set(State.FAILED);
+            state.set(State.INITIALIZATION_FAILED);
         } else {
             state.set(State.INITIALIZED);
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -221,6 +227,13 @@ protected P getSelectedPlugin() throws IllegalStateException {
                 "Tried to get a default plugin but there are more than one to choose from in this category. " +
                 "Call getPlugin(pluginKey) to choose the specific plugin you want to use in this case."
             );
+        } else if (State.FAILED.equals(state.get())) {
+            throw new IllegalStateException(
+                "Failed to get plugin because either configuration or initialization failed.  Check for " +
+                "configuration failures by logging any exceptions thrown by Amplify.configure().  Check for " +
+                "initialization failures by subscribing to the InitializationStatus.FAILED event on the Hub channel " +
+                "for the respective category with Amplify.Hub.subscribe()"
```
jamesonwilliams, line 235: I wonder if we should add a troubleshooting page in our docs, and link to it here?

core/src/main/java/com/amplifyframework/core/category/Category.java
```diff
@@ -184,6 +183,13 @@ public final P getPlugin(@NonNull final String pluginKey) throws IllegalStateExc
                 "Tried to get a plugin but that plugin was not present. " +
                 "Check if the plugin was added originally or perhaps was already removed."
             );
+        } else if (State.FAILED.equals(state.get())) {
+            throw new IllegalStateException(
+                "Failed to get plugin because either configuration or initialization failed.  Check for " +
+                "configuration failures by logging any exceptions thrown by Amplify.configure().  Check for " +
+                "initialization failures by subscribing to the InitializationStatus.FAILED event on the Hub channel " +
+                "for the respective category with Amplify.Hub.subscribe()"
+            );
```
jamesonwilliams, line 192:
```suggestion
            throw new IllegalStateException(
                "Failed to get plugin because either configuration or initialization had previously failed. " +
                "Check for configuration failures by logging any exceptions thrown by Amplify.configure(). " +
                "Check for initialization failures by subscribing to the InitializationStatus.FAILED event on "
                " the Hub channel for the respective category with Amplify.Hub.subscribe()."
            );
```

jamesonwilliams (approved):

core/src/main/java/com/amplifyframework/core/category/Category.java
richardmcclellan: I kind of see this PR as a short term fix to be honest.  I think ideally, it would be better to just log the errors, inline, either when they occur, or here, when trying to get a plugin.

It is difficult to log the errors at the time that they occur, because the `Logging` plugin may not have been configured yet.   Maybe we could fix that by configuring it first though.

There is a `CategoryInitializationResult` object that contains the `Throwable` that caused an initialization failure, but we don't store this object currently.  Perhaps we could store it on `Category`, and then pass it back to the customer as the cause of the `IllegalStateException`?

@jamesonwilliams do you have any historical context on this on this `CategoryInitializationResult`?  Would it be reasonable to store it?

jamesonwilliams: Sure, here's a bit of background on that `CategoryInitializationResult`. The original assumption was that a category may have k-many plugins associated to it. (In practice: no, it's only ever one.)

Multiple plugins initialize in parallel, and the result of their initializations are captured into this `CategoryInitializationResult` container object.

The category as a whole is said to fail if any of the plugins failed to initialize. The container also provides some utility methods to inspect failures from the different plugins and see which plugins succeeded and failed.

In retrospect, it has turned out that 1 plugin implements 1 category spec, and we haven't ever needed to support multiple plugins per category. So this container object is overengineered, and we could simplify this part of the code considerably if we enforce a plugin count of 0 or 1 at a framework level.

richardmcclellan: I just pushed a couple updates: 
 - Replaced `FAILED` with `CONFIGURATION_FAILED` and `INITIALIZATION_FAILED`, so that we can keep track of what failed, and provide a more specific error to the customer.
 - If the state is `INITIALIZATION_FAILED`, and the customer tries to get a plugin,  an `IllegalStateException` is thrown.  Now, that exception will also include the `Throwable` with the specific failure that occurred during initialization.  This is made possible by saving the `CategoryInitializationResult` on the `Category` object. 
 - Consolidated logic from the two very similar `getPlugin` methods into one helper method.

rjuliano (approved):
🚢  it!!! 

The integration tests are failing, but it should just be a matter to merging from main. The buildspec file in your branch is missing the part that creates the build report so it's failing in the post build stage.

PR #1159, opened 2021-02-10
chore: add logging and analytics to rx bindings

jamesonwilliams:
Even though there are no actual Rx-enabled methods in the Logging or
Analytics categories, it is convenient to have all of the various
Amplify categories available in the single RxAmplify facade. Doing so
means that a user can use RxAmplify at all times, without having to mix
in miscellaneous calls to the original Amplify facade as well.

### DIFF
--- a/rxbindings/src/main/java/com/amplifyframework/rx/RxAmplify.java
+++ b/rxbindings/src/main/java/com/amplifyframework/rx/RxAmplify.java
@@ -19,26 +19,28 @@ import android.content.Context;
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.AmplifyException;
+import com.amplifyframework.analytics.AnalyticsCategory;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.AmplifyConfiguration;
 import com.amplifyframework.core.plugin.Plugin;
+import com.amplifyframework.logging.LoggingCategory;
 
 /**
  * A collection of bindings to facilitate use of the Amplify framework with Rx-centered APIs.
  */
-@SuppressWarnings({"WeakerAccess", "unused"})
+@SuppressWarnings({"unused", "RedundantSuppression"})
 public final class RxAmplify {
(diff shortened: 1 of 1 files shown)


rjuliano (approved):
Ship it!!! 🚢

rxbindings/src/main/java/com/amplifyframework/rx/RxAmplify.java
```diff
@@ -19,26 +19,28 @@
...
     // Breaking the rules, here. Don't look.
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxAuthCategoryBehavior Auth = new RxAuthBinding();
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxApiCategoryBehavior API = new RxApiBinding();
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxDataStoreCategoryBehavior DataStore = new RxDataStoreBinding();
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxHubCategoryBehavior Hub = new RxHubBinding();
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxStorageCategoryBehavior Storage = new RxStorageBinding();
-    @SuppressWarnings({"checkstyle:all", "unused"}) public static final RxPredictionsCategoryBehavior Predictions = new RxPredictionsBinding();
-    // Logging currently has no callback/async behaviors
-    // Analytics currently has no callback/async behaviors
+    @SuppressWarnings({"checkstyle:all"}) public static final AnalyticsCategory Analytics = new AnalyticsCategory();
```
richardmcclellan, line 36: Should this just be the interface, rather than the class, like the others?   Like this:

```suggestion
    @SuppressWarnings({"checkstyle:all"}) public static final AnalyticsCategoryBehavior Analytics = new AnalyticsCategory();
```

Same suggestion for `LoggingCategory` - replace with `LoggingCategoryBehavior`

Issue #944, opened 2020-10-31
[DataStore] CursorWindowAllocationException crash for large datasets


upachecog:
Hey guys, I have some issues and usage questions:

What I am doing is just calling **the first query method on the DataStore of a single model**, and at that moment Amplify starts to **sync/download all the data from all the models**. 

**Issue description:**

The problem I noticed is that when the amount of data in the DynamoDB started to increase, the first synchronization started to fail. 

We have two main models, Location and Activities. Each location has _N_ activities and they will be assigned to each user on a daily basis. 

We made some calculations and found that each user will have approximately 1350 activities assigned per month within 150 different locations. We want to previously populate the database with locations and activities for a full month for each user. So each month will have 1500 rows in total. With this amount of data, I found the following issues just in the first sync:

_First sync: When the user logs-in for the first time and I perform the first query, after that everything seems to work fine._

_First query:_

```
fun getTaskList(
        onResponse: (value: List<Ubicacion>) -> Unit,
        onFailure: (value: DataStoreException) -> Unit
    ) {
        Amplify.DataStore.query(Ubicacion::class.java,
            Where.matches(Ubicacion.FECHA.eq(Utils.getCurrentDateString())),
            { matches ->
                Log.i(
                    "TaskDataStore",
                    "Obtained today's ${Utils.getCurrentDateString()} task list: $matches"
                )
                onResponse(matches.asSequence().toList())
            },
            { onFailure(it) }
        )
 }
```
**Issues:**

1.  In low-range smartphones (Samsung Galaxy A8, A10 & LGE LM-K410), I found this issue:

>     CursorWindowAllocationException
>     Cursor window allocation of 2097152 bytes failed. # Open Cursors=442 (# cursors opened by this proc=442)

   -  More information in: https://sentry.io/share/issue/78f9af09c61b4a6e925035f6819ae6cd/

2.  In mid-range smartphones, sometimes I have a _TimeOutException_ like in this issue: #563

**Expected result:**

Amplify only syncs/download the locations and activities of this single day, for this user. So the app doesn't have memory issues. OR Amplify syncs/downloads by small chunks of data.

**Current result**

Amplify syncs/downloads all the information previously registered information and in low-mid-range smartphones the application crashes, even though I only queried today's locations.

**Usage question:**

Is there a way to implement this with the current version of the API or this would be something like a new feature?

jamesonwilliams:
Hi @upachecog!

[...]

PR #1161, opened 2021-02-10
fix(datastore) improve storage adapter performance

richardmcclellan:
Performance Improvements:
 - Querying pages of results from the API now happens in parallel with saving items to local storage.   This improves performance by about 0.5 seconds.  The benefit is small right now, only because SQL write performance is still pretty bad. 
 - For `save` and `delete` without a predicate (e.g. `QueryPredicates.all()`), I eliminated a redundant query to see if the model exists for a given predicate, by short circuiting if the provided predicate is `QueryPredicates.all()`
 - Use `SELECT EXISTS(SELECT 1 FROM tableName...` to check if a modelExists instead of doing a full query and checking if the Cursor has any results.   Improves performance from ~1.5ms to ~0.5ms for checking if a model exists.  

Other Improvements:
 - Reduced `SQLiteStorageAdapter` thread pool size to equal number of available processors.   Previously, syncing 1000 items resulted in creation of ~5000 threads.   This change doesn't have a measurable performance improvement, but should help prevent out of memory errors or errors due to too many SQLite cursors being open (https://github.com/aws-amplify/amplify-android/issues/944)
 - Increased orchestrator timeout from 10 seconds to 60 seconds.  We still have work to do to improve SQL transaction performance.  In the meantime, we should raise the timeout so that customers can at least use larger datasets.

Minor refactorings:
 - Added `default getModelName()` to `Model.java`, as a convenience so that `getModelName()` works not just on a `SerializedModel`, but also on any `Model`
 - Separated code for executing `SQLiteStatement` objects into a new `SQLiteCommandProcessor` class, since `SQLiteStorageAdapter` is getting quite large.
 - Moved responsibility for compiling `SQLiteStatement`s from `SQLCommandFactory` to `SQLiteCommandProcessor`.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
@@ -17,6 +17,7 @@ package com.amplifyframework.datastore.storage.sqlite;
 
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.QueryOptions;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
@@ -59,6 +60,17 @@ interface SQLCommandFactory {
                         @NonNull QueryOptions options) throws DataStoreException;
 
     /**
+     * Generates the SELECT EXISTS command in a raw string representation from
+     * the {@link ModelSchema}.
+     *
+     * @param modelSchema schema of the model
+     * @return the QUERY SQL command
+     */
+    @NonNull
+    SqlCommand existsFor(@NonNull ModelSchema modelSchema,
+                         @NonNull QueryPredicate predicate) throws DataStoreException;
+
+    /**
      * Generates the INSERT INTO command in a raw string representation and a compiled
      * prepared statement that can be bound later with inputs.
      *
@@ -66,7 +78,7 @@ interface SQLCommandFactory {
      * @return the SQL command that encapsulates the INSERT INTO command
      */
     @NonNull
-    SqlCommand insertFor(@NonNull ModelSchema modelSchema);
+    <T extends Model> SqlCommand insertFor(@NonNull ModelSchema modelSchema, @NonNull T item) throws DataStoreException;
 
     /**
      * Generates the UPDATE command in a raw string representation and a compiled
@@ -76,8 +88,7 @@ interface SQLCommandFactory {
      * @return the SQL command that encapsulates the UPDATE command
      */
     @NonNull
-    SqlCommand updateFor(@NonNull ModelSchema modelSchema,
-                         @NonNull QueryPredicate predicate) throws DataStoreException;
+    <T extends Model> SqlCommand updateFor(@NonNull ModelSchema modelSchema, @NonNull T item) throws DataStoreException;
 
     /**
      * Generates the DELETE command in a raw string representation.
@@ -86,6 +97,5 @@ interface SQLCommandFactory {
      * @return the SQL command that encapsulates the DELETE command
      */
     @NonNull
-    SqlCommand deleteFor(@NonNull ModelSchema modelSchema,
-                         @NonNull QueryPredicate predicate) throws DataStoreException;
+    SqlCommand deleteFor(@NonNull ModelSchema modelSchema, @NonNull QueryPredicate predicate) throws DataStoreException;
 }
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
@@ -53,6 +53,7 @@ interface SQLCommandFactory {
      * the {@link ModelSchema}.
      *
      * @param modelSchema schema of the model
+     * @param options options to apply to the query
      * @return the QUERY SQL command
      */
     @NonNull
@@ -64,6 +65,7 @@ interface SQLCommandFactory {
      * the {@link ModelSchema}.
      *
      * @param modelSchema schema of the model
+     * @param predicate predicate to filter by
      * @return the QUERY SQL command
      */
     @NonNull
@@ -75,6 +77,7 @@ interface SQLCommandFactory {
      * prepared statement that can be bound later with inputs.
      *
      * @param modelSchema schema of the model
+     * @param item the Model to insert
      * @return the SQL command that encapsulates the INSERT INTO command
      */
     @NonNull
@@ -85,6 +88,7 @@ interface SQLCommandFactory {
      * prepared statement that can be bound later with inputs.
      *
      * @param modelSchema schema of the model
+     * @param item the Model to update
      * @return the SQL command that encapsulates the UPDATE command
      */
     @NonNull
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessor.java
@@ -44,7 +44,6 @@ final class SQLCommandProcessor {
     }
 
     Cursor rawQuery(SqlCommand command) throws DataStoreException {
-        SQLiteStatement sqliteStatement = sqliteDatabase.compileStatement(command.sqlStatement());
         try {
             long startTime = System.currentTimeMillis();
             Cursor result = sqliteDatabase.rawQuery(command.sqlStatement(), command.getBindingsAsArray());
@@ -52,7 +51,7 @@ final class SQLCommandProcessor {
                     + " ms inTransaction: " + sqliteDatabase.inTransaction() + " SQL: " + command.sqlStatement());
             return result;
         } catch (SQLException sqlException) {
-            throw dataStoreException(sqlException, sqliteStatement);
+            throw dataStoreException(sqlException, command.sqlStatement());
         }
     }
 
@@ -66,7 +65,7 @@ final class SQLCommandProcessor {
                     + " ms inTransaction: " + sqliteDatabase.inTransaction() + " SQL: " + command.sqlStatement());
             return result;
         } catch (SQLException sqlException) {
-            throw dataStoreException(sqlException, sqliteStatement);
+            throw dataStoreException(sqlException, command.sqlStatement());
         }
     }
 
@@ -79,13 +78,13 @@ final class SQLCommandProcessor {
             LOG.verbose("SQLCommandProcessor execute in " + (System.currentTimeMillis() - startTime)
                     + " ms inTransaction: " + sqliteDatabase.inTransaction() + " SQL: " + command.sqlStatement());
         } catch (SQLException sqlException) {
-            throw dataStoreException(sqlException, sqliteStatement);
+            throw dataStoreException(sqlException, command.sqlStatement());
         }
     }
 
-    private DataStoreException dataStoreException(SQLException sqlException, SQLiteStatement sqliteStatement) {
+    private DataStoreException dataStoreException(SQLException sqlException, String sqlStatement) {
         return new DataStoreException(
-                "Invalid SQL statement: " + sqliteStatement,
+                "Invalid SQL statement: " + sqlStatement,
                 sqlException,
                 AmplifyException.REPORT_BUG_TO_AWS_SUGGESTION
         );
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessor.java
@@ -29,6 +29,11 @@ import com.amplifyframework.logging.Logger;
 import java.util.List;
 import java.util.Objects;
 
+/**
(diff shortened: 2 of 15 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
```diff
@@ -76,8 +88,7 @@ SqlCommand queryFor(@NonNull ModelSchema modelSchema,
      * @return the SQL command that encapsulates the UPDATE command
      */
     @NonNull
-    SqlCommand updateFor(@NonNull ModelSchema modelSchema,
-                         @NonNull QueryPredicate predicate) throws DataStoreException;
+    <T extends Model> SqlCommand updateFor(@NonNull ModelSchema modelSchema, @NonNull T item) throws DataStoreException;
```
jamesonwilliams, line 91: Looks like the predicate was not previously being unused? It wasn't in the JavaDoc either. Can you update the JavaDoc with the new param (and if any others were missed in this file)?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessor.java
```diff
@@ -0,0 +1,130 @@
...
+
+import com.amplifyframework.AmplifyException;
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.datastore.DataStoreException;
+import com.amplifyframework.logging.Logger;
+
+import java.util.List;
+import java.util.Objects;
+
+final class SQLCommandProcessor {
```
jamesonwilliams, line 32: Can you a class doco describing the single responsibility for the component? We have a number of various SQL gizmos now & it would help us to keep track in the future.

core/src/main/java/com/amplifyframework/core/model/Model.java
```diff
@@ -30,4 +30,12 @@
      */
     @NonNull
     String getId();
+
+    /**
+     * Returns the name of this model as a String.
+     * @return the name of this model as a String.
+     */
+    default String getModelName() {
```
jamesonwilliams, line 38: Can we put `@NonNull` on this?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -142,9 +137,6 @@ public SqlCommand createTableFor(@NonNull ModelSchema modelSchema) {
 
     /**
      * {@inheritDoc}
```
jamesonwilliams, line 139: You could probably just kill the entire comment and let the interface doco come through.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -254,19 +246,57 @@ public SqlCommand queryFor(@NonNull ModelSchema modelSchema,
 
         rawQuery.append(";");
         final String queryString = rawQuery.toString();
-        return new SqlCommand(table.getName(), queryString, columns, bindings);
+        return new SqlCommand(table.getName(), queryString, bindings);
+    }
+
+    @NonNull
```
jamesonwilliams, line 252:
```suggestion
    @NonNull
    @Override
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
```diff
@@ -380,17 +406,29 @@ public SqlCommand deleteFor(@NonNull ModelSchema modelSchema,
...
                 preparedDeleteStatement,
-                Collections.emptyList(),
-                sqlPredicate.getBindings(),
-                compiledDeleteStatement
+                sqlPredicate.getBindings() // WHERE clause
         );
     }
 
+    // extract model field values to save in database
+    private List<Object> extractFieldValues(@NonNull Model model) throws DataStoreException {
```
jamesonwilliams, line 416: We have a number of variants on this same basic concept across the DataStore code base, it feels like. In future work, we could try to bring some of them back together.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -172,7 +172,7 @@ public synchronized void initialize(
         Objects.requireNonNull(context);
         Objects.requireNonNull(onSuccess);
         Objects.requireNonNull(onError);
-        this.threadPool = Executors.newCachedThreadPool();
+        this.threadPool = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
jamesonwilliams, line 175: > Reduced SQLiteStorageAdapter thread pool size to equal number of available processors

Seems like a reasonable bet for compute-bound work. But, ours is probably IO-bound. Skipping a whole lot of pedantry, I think it comes down to:

- Can we ever have more pending tasks than available processors?

Setting a cap on the thread pool sounds like a reasonable idea. Maybe the cap should be an order of magnitude higher, though.

For reference, `Schedulers.io()` in RxJava is actually _unbounded_, and has exactly the problem you called out:
> Note that this scheduler may create an unbounded number of worker threads that can result in system slowdowns or OutOfMemoryError.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -235,10 +229,28 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
                         processor.onComplete();
                     }
                 })
-                // Flatten the PaginatedResult objects into a stream of ModelWithMetadata objects.
-                .concatMapIterable(PaginatedResult::getItems)
-                // Stop after fetching the maximum configured records to sync.
-                .take(dataStoreConfigurationProvider.getConfiguration().getSyncMaxRecords());
+                // If it's a SerializedModel, add the ModelSchema, since it isn't added during deserialization.
+                .map(paginatedResult -> Flowable.fromIterable(paginatedResult)
```
jamesonwilliams, line 233: Ah I see, it's a `Flowable` of `Lists` since there can be 0...n pages for a particular model, where each page is a list?

Then on line 145, you flatten each page back into a single Flowable, right?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -235,10 +229,28 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
...
+                        .map(modelWithMetadata -> hydrateSchemaIfNeeded(modelWithMetadata, schema))
+                        .toList()
+                        .blockingGet()
+                )
+                .takeUntil(items -> recordsFetched.accumulateAndGet(items.size(), Integer::sum) >= syncMaxRecords);
```
jamesonwilliams, line 238: Were we not actually enforcing this logic anywhere, before?

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SqlCommandTest.java
```diff
@@ -244,6 +249,20 @@ public void queryWithSortBy() throws DataStoreException {
...
+     * @throws DataStoreException From {@link SQLCommandFactory#existsFor(ModelSchema, QueryPredicate)}
+     */
+    @Test
+    public void existsFor() throws DataStoreException {
+        final ModelSchema personSchema = getPersonModelSchema();
+        String personId = RandomString.string();
+        final SqlCommand sqlCommand = sqlCommandFactory.existsFor(personSchema, Person.ID.eq(personId));
+        assertEquals("SELECT EXISTS(SELECT 1 FROM `Person` WHERE id = ?);",
+            sqlCommand.sqlStatement());
+        assertEquals(Arrays.asList(personId), sqlCommand.getBindings());
```
jamesonwilliams, line 263:
```suggestion
        assertEquals(Collections.singletonList(personId), sqlCommand.getBindings());
```

jamesonwilliams (commented):
Some nice improvements here!

Does it make sense to split any out to their own PRs?

Love to see the `SQLiteStorageAdapter` becoming shorter! :-D

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandFactory.java
richardmcclellan: Previously, the `QueryPredicate` was derived from a `Model` instance before being passed in.  This PR moves that logic into this method, so this class now derives the `QueryPredicate` from the `Model` instance.  

And yes, it does look like all the Javadoc in this interface is missing some params, so I will update them all.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteCommandFactory.java
richardmcclellan: Agreed, we could DRY up this logic somehow.

This is also an expensive operation (~5 ms for one Blog model instance).   When an an item is received from the backend, the following transformations occur: 
```
String -> JsonObject -> Java model -> Map<String, Object> (for saving to SQLite)
```

If we switch to using `SerializedModel` for everything internally, I believe it would improve performance quite a bit, since we would only need to do this:

```
String -> JsonObject -> Map<String, Object>
```

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
richardmcclellan: > Maybe the cap should be an order of magnitude higher, though.

My main goal in capping the thread pool is just to avoid an `OutOfMemoryError` or `CursorWindowAllocationException`, which I think we can probably accomplish by capping it to ~100.   On my Nexus 5x test device, `availableProcessors()` is 6.  I'd be okay with raising it an order of magnitude.  But why do you suggest that?  What number do you think would make sense?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: Yep, that's right.  I had intended to batch each "page" of results into one single transaction, which is why this code has been refactored in this way.  I didn't quite get there, in this PR, but plan to do so in the future.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
richardmcclellan: We were.  Previously, it was `.take(syncMaxRecords)`, because it was flattened into a `Flowable<ModelWithMetadata<T>>` in this method, first.  Since it's returning a `Flowable<List<ModelWithMetadata<T>>>` now so that we can do batch saves in the future, I had to use `takeUntil` and increment a counter manually.

richardmcclellan:
> Does it make sense to split any out to their own PRs?

In retrospect, it probably could have been split up.  Not sure if it's worth doing now though :/.  Other than that, I've responded to all of the outstanding comments.    I just had one question for you about what we should use for the thread pool size.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
jamesonwilliams: I think we're on the same page about this, but just for completeness: Software threads can be scheduled on/off of physical cores. The reason compute-bound work is usually capped to `nprocs` is because compute-bound work usually can _actually make use of_ that many cores. IO-bound work is different, it does a little work on the CPU, then waits around. Meanwhile, the kernel will schedule some other interesting work for the CPI, while the IO-bound thread sleeps. That's why IO thread pools are usually unbounded -- on average, an IO thread isn't actually doing anything and shouldn't be scheduled.

I think we probably use a handful of threads to _orchestrate_ a CRUD op on DataStore right now. What would happen if I try to save three unrelated models at once -- a `Bike`, a `HousePlant`, and a `Tuba`? We may block 6 threads before doing anything with the Tuba.

I thought maybe (nproc * 10) - (nproc * 100) would be a sweetspot. We could guard against the memory exhaustion bugs while also not serializing things too heavily.

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
richardmcclellan: Thanks for the clarification!  That all makes sense to me.  I'll update it nproc * 20, if that sounds good to you!

jamesonwilliams (approved):

jamesonwilliams (commented):
Can we get a `SQLCommandProcessorTest.java`, perchance?

rjuliano:
Looks like the unit test is failing with a couple of checkstyle issues

jamesonwilliams (approved):

richardmcclellan:
Not sure why integration tests are failing, but they are passing locally via `gradlew cAT`, so I think this PR is ready for final review and merge.

jamesonwilliams (approved):
Great work!

PR #1167, opened 2021-02-11
chore(devmenu) set dev menu disabled by default

richardmcclellan:
Disable the dev menu by default.

Customers can still enable and use the dev menu with the follow code:
```
val config = AmplifyConfiguration.builder(ctx)
    .devMenuEnabled(true)
    .build()
Amplify.configure(config, ctx)
```


core/src/main/java/com/amplifyframework/core/AmplifyConfiguration.java
```diff
@@ -65,7 +65,7 @@
     @SuppressWarnings("WeakerAccess") // These are created and accessed as public API
     public AmplifyConfiguration(@NonNull Map<String, CategoryConfiguration> configs) {
         // Dev menu is enabled by default in debug mode
```
jamesonwilliams, line 67:
```suggestion
        // Dev menu is disabled by default.
```

changxu0306 (approved):

jamesonwilliams (approved):

PR #1168, opened 2021-02-13
fix: Delete relationships fix for flutter

Amplifiyer:
*Description of changes:* Model instances are of always of type `SerializedModel` when the requests are coming from flutter and so always fail to retrieve modelSchema when querying based on instance type. This PR replaces the use of `getModelSchemaForModelInstance` with `getModelSchemaForModelClass`

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
@@ -30,6 +30,7 @@ import com.amplifyframework.core.model.query.predicate.QueryField;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.DataStoreException;
+import com.amplifyframework.datastore.appsync.SerializedModel;
 import com.amplifyframework.datastore.storage.sqlite.adapter.SQLiteTable;
 import com.amplifyframework.logging.Logger;
 import com.amplifyframework.util.Empty;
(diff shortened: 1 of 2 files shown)


jamesonwilliams (approved):

richardmcclellan (approved):
This change looks good, but we should make sure to add a unit test to prevent future regressions as well.  Given the urgency for this change though, I'm okay with adding the test as a separate PR.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteModelTree.java
```diff
@@ -95,10 +96,18 @@
         for (Map.Entry<ModelSchema, Set<String>> entry : modelMap.entrySet()) {
             ModelSchema schema = entry.getKey();
             for (String id : entry.getValue()) {
-                // Create dummy model instance using just the ID and model type
-                String dummyJson = gson.toJson(Collections.singletonMap("id", id));
-                Model dummyItem = gson.fromJson(dummyJson, schema.getModelClass());
-                descendants.add(dummyItem);
+                if (root.iterator().next().getClass() == SerializedModel.class) {
```
richardmcclellan, line 99: Calling `next()` on the `iterator()` is going to increment the iterator, which I don't think you want to do here, right?  

To fix this, when we call `root.iterator.next()` on line 88 above, we need to save the result to a instance variable, and then use it here.

Amplifiyer: You are right, will update and push a new PR. Thanks for catching this.

raphkim: You may not even need this `if` statement at all. The returned item from this method is used as a method of publishing to remote only, so you can always return a `List<SerializedModel>`.

PR #1177, opened 2021-02-17
[DO NOT MERGE] Multi-auth POC

rjuliano:
*Issue #, if available:*

*Description of changes:*
This POC introduces the concept of an auth mode chain to the API plugin. An auth mode chain is essentially a linked list of auth modes (user pools, oidc, api key, iam). An auth chain is constructed from the `@auth` rules specified in the graphql schema.

Given the following schema:
```
type Post 
@model 
@auth(rules: [
    { allow: owner },
    { allow: public operations:[read] }
  ]
)
{
  id: ID!
  name: String!
  description: String
  status: String 
}
```

The auth mode chain for the `Post` model would look like:

```
+-------------------------------------------------------+                                        
| Model name: Post                                      |                                        
|    +-------------------------------------------------+|                                        
|    |Operation:READ                                   ||                                        
|    |+------------+       +------------+              ||                                        
|    || User pools |------>|  API Key   |              ||                                        
|    |+------------+       +------------+              ||                                        
|    +-------------------------------------------------+|                                        
|    +-------------------------------------------------+|                                        
|    |Operation:CREATE                                 ||                                        
|    |+------------+                                   ||                                        
|    ||User pools  |                                   ||                                        
|    |+------------+                                   ||                                        
|    +-------------------------------------------------+|                                        
|    +-------------------------------------------------+|                                        
|    |Operation:UPDATE                                 ||                                        
|    |+------------+                                   ||                                        
|    ||User pools  |                                   ||                                        
|    |+------------+                                   ||                                        
|    +-------------------------------------------------+|                                        
|    +-------------------------------------------------+|                                        
|    |Operation:DELETE                                 ||                                        
|    |+------------+                                   ||                                        
|    ||User pools  |                                   ||                                        
|    |+------------+                                   ||                                        
|    +-------------------------------------------------+|                                        
+-------------------------------------------------------+                                        
                                                            
```
The order of auth modes, follows some basic rules for this POC:
- If a model has a rule with owner or group strategy, the associated provider (user pools or OIDC) is added to the list
- If a model has a rule with private strategy, the associated provider (user pools or IAM) is added to the end of the list (if not already present)
- If a model has a rule with public strategy, the associated provider (API Key or user pools) is added to the end of the list (if not already present)

As with any linked list, it has the following properties:
- A current pointer, which is initially set to the first item in the list
- A `next()` method which advances the current pointer to the next mode (if available)
- A `reset()` method which sets the current pointer to be the first element in the list.

When the API plugin receives a GraphQL request, it retrieves the chain using the model name and the operation associated with the request. The request is then signed using the current auth mode of the chain. If an auth failure occurs, the `next()` method is called and the request is retried with the next mode. If it's the end of the list, an authorization failure bubbles up.

An optimization should also be made to this in order to avoid requests that have no chance of succeeding. For example, if user is not signed in, there's no sense in trying to fulfill an owner or group rule. 

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
@@ -24,6 +24,8 @@ import com.amplifyframework.core.model.ModelOperation;
 import com.amplifyframework.core.model.ModelSchema;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -33,7 +35,16 @@ import java.util.Map;
  */
 public class AuthProviderChainRepository {
     private static final String TAG = "AuthProviderChainRepository";
+    //TODO: Use enums for keys. Right now we have a couple different enums, so we may need to consolidate.
     private Map<String, Map<String, AuthProviderChain>> authProviderChains;
+
+    /**
+     * Convenience constructor if the caller want to pass in a map.
+     * @param modelSchemas
+     */
+    public AuthProviderChainRepository(Map<String, ModelSchema> modelSchemas) {
+        this(new ArrayList<ModelSchema>(modelSchemas.values()));
+    }
 
     /**
      * Default constructor.
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
@@ -15,17 +15,16 @@
 
 package com.amplifyframework.api.aws;
 
-import android.util.Log;
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.model.AuthRule;
 import com.amplifyframework.core.model.AuthRuleProvider;
 import com.amplifyframework.core.model.ModelOperation;
 import com.amplifyframework.core.model.ModelSchema;
+import com.amplifyframework.logging.Logger;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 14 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -102,6 +116,32 @@ public AWSApiPlugin(@NonNull ApiAuthProviders apiAuthProvider) {
         this.gqlApis = new HashSet<>();
         this.executorService = Executors.newCachedThreadPool();
         this.requestDecorator = new AuthRuleRequestDecorator(authProvider);
+        this.authProviderChains = null;
+        this.modelProvider = null;
```
jamesonwilliams, line 120: AWS API Plugin is already a beastly entity. We want to be thinking of ways to fan logic out of this facade class. Even today it has overgrown its Single Responsibility. Is there any way we can avoid putting these new actors right into the top level plugin class?

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
```diff
@@ -25,23 +27,51 @@
      * Owner authorization specifies whether a user can access or operate against an object.  To use OWNER, the API
      * must have Cognito User Pool configured.
      */
-    OWNER,
+    OWNER(AuthRuleProvider.AMAZON_COGNITO_USER_POOLS, List.of(AuthRuleProvider.AMAZON_COGNITO_USER_POOLS,
```
jamesonwilliams, line 30: Hm. Is there desugaring for `List.of`? That's a pretty modern construct.

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Cache of auth provider chains by model/operation.
+ */
+public class AuthProviderChainRepository {
+    private static final String TAG = "AuthProviderChainRepository";
+    private Map<String, Map<String, AuthProviderChain>> authProviderChains;
```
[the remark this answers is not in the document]
rjuliano: Yeah...I'll definitely take a look at that. I wanted to get something together just to prove out the concept, but there certainly is some refactoring/cleanup to be done here.

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthProviderChainRepository.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    static class AuthProviderChain {
+        private final String modelName;
+        private final String operation;
+        private final List<AuthRuleProvider> providers;
+        private AuthProviderChainNode head;
+        private AuthProviderChainNode tail;
+        private AuthProviderChainNode current;
```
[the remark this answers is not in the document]
rjuliano: I can definitely do that. I'll address it as part of refactoring the map of maps

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
rjuliano: Ahhh yes...I had that feeling as well, but since the purpose here was to prove out the concept itself, I went ahead with it. That being said, I have a couple ideas to clean this up moving forward.

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
rjuliano: It didn't bark, so I figured it was OK 🤣  I'll verify though.

rjuliano: Take that back...I just got bit by it.

PR #1184, opened 2021-02-19
fix(rest-api) Expose HTTP headers in RestResponse

richardmcclellan:

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/operation/AWSRestOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/operation/AWSRestOperation.java
@@ -16,6 +16,7 @@
 package com.amplifyframework.api.aws.operation;
 
 import android.annotation.SuppressLint;
+import android.text.TextUtils;
 import androidx.annotation.NonNull;
 
 import com.amplifyframework.AmplifyException;
@@ -28,6 +29,9 @@ import com.amplifyframework.core.Consumer;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import okhttp3.Call;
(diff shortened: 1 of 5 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/operation/AWSRestOperation.java
```diff
@@ -113,13 +116,20 @@ public void onResponse(@NonNull Call call,
                                @NonNull Response response) throws IOException {
             final ResponseBody responseBody = response.body();
             final int statusCode = response.code();
-
+            final Map<String, String> headersMap = new HashMap<>();
+            final Map<String, List<String>> headersMultiMap = response.headers().toMultimap();
+            for (String key : headersMultiMap.keySet()) {
+                List<String> value = headersMultiMap.get(key);
+                if (value.size() > 0) {
+                    headersMap.put(key, value.get(0));
```
jamesonwilliams, line 124: Well, I'll be derned. RFC2616 (for HTTP/1.1) [allows multiple values per key](https://stackoverflow.com/a/24502264/695787).

It might be easy enough to transform this to something compliant with the RFC. I probably _wouldn't_ worry about this, if it weren't just a quick update to your existing code:

```suggestion
                    headersMap.put(key, TextUtils.join(",", value));
```

jamesonwilliams (approved):

aws-api/src/main/java/com/amplifyframework/api/aws/operation/AWSRestOperation.java
richardmcclellan: Yep, I considered returning a `Map<String, List<String>>`.   from `RestResponse.getHeaders()`.  But then I thought, Amplify is _opinionated_, and I believe our opinion should be that you **shouldn't** use multiple headers with the same key.

That said, I actually this suggestion!  It allows us to return a simple `Map<String, String>` for almost all use cases, but doesn't block customers from accessing all the headers if they happen to have multiple.

richardmcclellan: I've made this change, and also updated the unit test to verify it.

richardmcclellan:
Closing until we actually get a customer request for this.  We should add adding features that customers won't ever need.

kjones:
Not saying I need this yet but we do have REST APIs that return 201 or 202 with a Location header.

richardmcclellan:
> Not saying I need this yet but we do have REST APIs that return 201 or 202 with a Location header.

Thanks for the comment @kjones!  I think this is actually a reasonable thing to add.  We already support customizing the HTTP request headers, so it seems logical to expose the HTTP response headers.  I'll go ahead and merge this in.

jamesonwilliams (approved):

jamesonwilliams (approved):

PR #1195, opened 2021-02-25
chore: add workflow to sync release branch

rjuliano:
*Issue #, if available:*

*Description of changes:*
Adding a Github workflow triggered on pushes to the main branch. This workflow simply merges the changes from main into releases/next and pushes. This ensures the release branch always has all the changes from main. 

This is important because we will have another workflow that will be triggered when commits are pushed to releases/next. That will be in a separate PR.

Note on the token being used: I haven't setup the secret for the repo yet, so I'm using the auto-generated Github token. That token has limited permissions. It should allow us to do the merge and push, but won't trigger other workflows.

### DIFF
--- a/.github/workflows/sync_release_branch.yml
+++ b/.github/workflows/sync_release_branch.yml
@@ -1,0 +1,27 @@
+name: Commits pushed to main
+on:
+  push:
+    branches:
+      - main
+jobs:
+  update_release_branch:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Update git # Apparently ubuntu-latest has an older version of git that causes some issues with scripts.
+        run: |
+          sudo add-apt-repository -y ppa:git-core/ppa
+          sudo apt-get update
+          sudo apt-get install git -y
+      - uses: actions/checkout@v2
+        with:
+          ref: releases/next
+          token: ${{ github.token }} # This will need to change to a token belonging to a service account.
+          fetch-depth: 0
+      - name: Merge changes from main
+        run: |
+          cd $GITHUB_WORKSPACE
+          git fetch --tags
+          git status
+          git merge origin/main
+          git push origin releases/next


.github/workflows/sync_release_branch.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          ref: releases/next
+          token: ${{ github.token }} # This will need to change to a token belonging to a service account.
+          fetch-depth: 0
+      - name: Merge changes from main
+        run: |
+          cd $GITHUB_WORKSPACE
+          git fetch --tags
+          git status
+          git merge origin/main
```
jamesonwilliams, line 26: What do we do if the merge fails?

rjuliano: I think in case of conflict, this workflow would just fail and we would have to resort to manually merging the changes from main to releases/next. That's part of the reason why I ended up creating a separate workflow to sync the branches, instead of kicking off the entire versioning process from a single workflow. 

Ideally, the only changes made to the releases/next branch will be the updates to gradle.properties and changelog, so the risk of conflicting changes should be fairly low, but definitely something to keep an eye on and adjust.

richardmcclellan: Could we prevent conflicts altogether by doing something like:
```
git checkout main
git branch -D release
git checkout -b release
git push -f release
```

richardmcclellan:
> This is important because we will have another workflow that will be triggered when commits are pushed to releases/next. That will be in a separate PR.

I'm not sure about this strategy.  Won't the workflow being added in this PR ultimately trigger this second workflow you are proposing?  I don't think we want to release to Maven on every merge to main.

What if we try to simplify this a bit and instead just use one branch (`main`), and have two workflows:
1. A manually triggered workflow (we would literally go to https://github.com/aws-amplify/amplify-android/actions and click a button) that increments the version, commits the change with a `release:` prefix for the commit message, and creates a PR.  
2. A workflow triggered by any new commit to `main` that has a `release:` as the commit prefix.  This workflow would then kick off the release to Maven.

rjuliano:
> > This is important because we will have another workflow that will be triggered when commits are pushed to releases/next. That will be in a separate PR.
> 
> I'm not sure about this strategy. Won't the workflow being added in this PR ultimately trigger this second workflow you are proposing? I don't think we want to release to Maven on every merge to main.

Releases to maven would only happen when the PR for the new release gets merged into main. The workflow triggered when pushing to changes into release/next would only:
- Merge changes from main
- Re-run version logic calculation
- Update gradle.properties and changelog as needed and commit those changes to into releases/next. Which in turn would update the PR for the upcoming release. The action for the PR creation is not configured yet since I want to test the version bump logic first.

> 
> What if we try to simplify this a bit and instead just use one branch (`main`), and have two workflows:
> 
> 1. A manually triggered workflow (we would literally go to https://github.com/aws-amplify/amplify-android/actions and click a button) that increments the version, commits the change with a `release:` prefix for the commit message, and creates a PR.

We can definitely do that to start with. I believe there's a workflow_dispatch event I can add to the workflow which makes the button visible under actions. I'll take a look.

> 2. A workflow triggered by any new commit to `main` that has a `release:` as the commit prefix.  This workflow would then kick off the release to Maven.

That's the plan with regards to publishing. When the release PR is approved, we'll create a tag which in turn will trigger publishing to maven.





We can definitely go the manual route to start with.

rjuliano:
We'll initially start with a simpler approach using a manually triggered workflow. See #1196

PR #1196, opened 2021-02-25
chore: manually triggered workflow to start release process

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR adds the following:
- A manually triggered workflow that kicks off the release process.
- Fastlane configuration files. Fastfile is the most relevant one, particularly the `create_next_release_pr` method.

## release_pr.yml
This workflow will be triggered manually. As workflow input, we pass in the branch name which defaults to main. Basic flow is:

```
- Checkout main
- Install Ruby 
- Install gems
- Configure git options  (execute configure_git_options lane)
- Create a new branch called releases/<year><month><day>.
- Create a PR for the release (execute create_next_release_pr lane)
```

## Creating the PR (Fastlane create_next_release_pr lane (Fastfile))
The steps for this part of the process are:

```
- Calculate next release version
- Update gradle.properties and CHANGELOG.md
- Commit and push those changes
- Create the pull request.
```

I have the same setup in one of my repos. Here's [one of the workflow runs](https://github.com/rjuliano/rjuliano.github.io/actions/runs/641151337) and the [resulting PR](https://github.com/rjuliano/rjuliano.github.io/pull/6)


.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,58 @@
...
+        cd scripts
+        gem install bundler
+        bundle install
+    - name: Configure git options
+      run: |
+        cd scripts
+        bundle exec fastlane android configure_git_options git_user_email:${{ github.event.inputs.git_user_email }} git_user_name:${{ github.event.inputs.git_user_name }}
+    - name: Create/checkout a branch for the release
+      run: |
+        branch_name=releases/`date "+%Y%m%d"`
```
richardmcclellan, line 46: Two issues I see with this naming strategy:
 - If we try to do two releases on the same day, this may fail because it will try to create a branch that already exists.
 - We may end up with a bunch of stale branches on the repo, since this creates a new one for each release.

Could we instead just keep it simple, and always use the same branch name (e.g `release`), and just always do a `git branch -D release` before trying to create it?

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,192 @@
...
+        "signing.inMemoryKey": ConfigValueReader.from_environment_variable('signing.inMemoryKey'),
+        "signing.password": ConfigValueReader.from_environment_variable('signing.password'),
+        "signing.keyId": ConfigValueReader.from_environment_variable('signing.keyId'),
+        "signing.secretKeyRingFile": ConfigValueReader.from_environment_variable('signing.secretKeyRingFile')
+      }
+    )
+  end
+
+  desc "Publish release to maven without overriding gradle properties."
+  private_lane :publish_to_maven_no_overrides do |options|
```
richardmcclellan, line 154: We probably don't need this one, right?  We can just use the approach above, with environment variables?

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,192 @@
...
+  desc "Checks out a tag and runs the uploadArchives gradle task."
+  lane :push_tag_to_maven do |options|
+    UI.message("Pushing release #{options[:release_tag]} to maven")
+    sh("git", "checkout", options[:release_tag])
+    publish_to_maven
+  end
+
+  desc "Update changelog"
+  private_lane :update_change_log do |options|
+    changelog_body = changelog_from_git_commits(
```
richardmcclellan, line 59: Can we filter the commits so that only commits prefixed with `fix` or `feat` get added to the `CHANGELOG`?

richardmcclellan (commented):
I think ideally, we should have two required reviewers for the release PRs.  Otherwise, the person who manually triggers the workflow can just approve and merge it.   I don't think Github lets you configure the required reviewers per _PR_, but you can configure it per _branch_.

So, what if the create PR workflow is something like this:
```
git checkout main
git checkout -b release-pr
// Update gradle.properties and CHANGELOG
git commit -am "chore(release): x.x.x"
git push -f
create PR, with `release` as the base branch.
```

This approach also probably makes it a little more clear what triggers the publish artifacts workflow (any commit to the `release` branch).  Note, that this will result in the release PRs containing all of the changes since the previous release.  I think that's okay though.   

Alternatively, if we want to stick to a single branch strategy, we could make the release PR a little more manual (we would create it ourselves, with `main` as the base branch, and then one required approver would be enough, since we can't approve our own PR), but we could still just run the fastlane command locally to actually make the changes (increment version and generate `CHANGELOG`).

I don't really feel too strongly one way or the other about the above two ideas, but let me know your thoughts!

.github/workflows/release_pr.yml
rjuliano: I like that...I'll make the change.

scripts/Gemfile
```diff
@@ -0,0 +1,6 @@
+source "https://rubygems.org"
```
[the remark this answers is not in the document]
rjuliano: My understanding is that the Gemfile(s) need to be at the same level as the `fastlane` folder, but not inside that folder. I'm not crazy about the setup. And the `fastlane` folder has to have that name. I'll do a little more digging to see if there's a cleaner way to organize this.

scripts/fastlane/Fastfile
rjuliano: It looks like this [particular action](https://docs.fastlane.tools/actions/changelog_from_git_commits/) doesn't have any filtering options beyond including/excluding merge commits

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,192 @@
...
+    version_code = calculate_version_code(segments: segments, mode: 3)
+
+    UI.message("Updating versionName and versionCode in gradle.properties")
+
+    set_key_value(file: gradle_properties_path, key: 'VERSION_NAME', value: version)
+    set_key_value(file: gradle_properties_path, key: 'VERSION_CODE', value: version_code.to_s)
+
+  end
+
+  # Calculate the Android version code with one of 3 different approaches.
```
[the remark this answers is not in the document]
rjuliano: I never found anything authoritative on that. When discussing versioning in the developer guide, they don't even mention libraries. Only thing I found was a [Stackoverflow post](https://stackoverflow.com/questions/57230308/do-i-need-to-increase-version-code-of-a-library-module).

My reasoning for exploring multiple options was mainly related to trying to find a simpler way to come up with that number without creating constraints. 

Given the fact that's it's probably not even used, I'll clean up some of these other options and leave option 1 in there just in case. We can try to do a release without incrementing it and see what happens.

scripts/fastlane/Fastfile
rjuliano: Yeah...I can remove this one. Could be useful if we were to put the credentials in the gradle home directory, but we're unlikely to do that in a CI scenario.

.github/workflows/release_pr.yml
rjuliano: Branching strategy updated. Workflow will try to delete a branch called `release` before creating a new `release` branch from the latest `main`

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+    sh('git', 'fetch', '--tags')
+    sh('git', 'fetch')
+  end
+
+  lane :configure_git_options do |options|
+    sh('git', 'config', 'user.email', options[:git_user_email])
+    sh('git', 'config', 'user.name', options[:git_user_name])
+  end
+
+  lane :manage_next_release_pr do |options|
```
richardmcclellan, line 26:
```suggestion
  lane :create_next_release_pr do |options|
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+                        api_token: ENV["RELEASE_MANAGER_TOKEN"])
+  end
+
+
+  desc "Checks out a tag and runs the uploadArchives gradle task."
+  lane :push_tag_to_maven do |options|
+    UI.message("Pushing release #{options[:release_tag]} to maven")
+    sh("git", "checkout", options[:release_tag])
+    publish_to_maven
+  end
```
richardmcclellan, line 54: I don't think we should create the tag until after the `uploadArchives` command has succeeded.   That way, if `uploadArchives` fails for some reason, we don't have to worry about rolling back (deleting the tag).   And even then, we should just figure out how we can make a GitHub release, which should automatically create the tag as a part of that. 

```suggestion
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+  desc "Update changelog"
+  private_lane :update_change_log do |options|
+    changelog_body = changelog_from_git_commits(
+      between: [options[:last_release_tag], "HEAD"],
+      merge_commit_filtering: "exclude_merges",
+      pretty: "- %s"
+    )
+
+    open(change_log_path, 'a') { |f|
+      f.puts "# Release Amplify Android #{options[:next_version]}\n\n"
```
richardmcclellan, line 65: This matches the format currently used for the [aws-sdk-android CHANGELOG.md](https://github.com/aws-amplify/aws-sdk-android/blob/main/CHANGELOG.md)
```suggestion
      f.puts "## Release {options[:next_version]}\n\n"
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+    sh('git', 'commit', '-am', commit_message)
+  end
+
+  desc "Tag in git and push to GitHub"
+  private_lane :add_tag do |options|
+    next_version = options[:version]
+
+    add_git_tag(tag: next_version)
+    push_git_tags(tag: next_version)
+  end
```
richardmcclellan, line 108: Instead of just creating a tag, we should figure out how we can make a GitHub release, which should just create the tag automatically.  That can be added in a subsequent PR.

```suggestion
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+        "signing.keyId": ConfigValueReader.from_environment_variable('signing.keyId'),
+        "signing.secretKeyRingFile": ConfigValueReader.from_environment_variable('signing.secretKeyRingFile')
+      }
+    )
+  end
+
+  desc "Retrieve a parameter value from SSM."
+  private_lane :get_param_value do |options|
+    sh(
+      command: "aws ssm get-parameter --name #{options[:param_name]} --with-decryption --region us-east-1 --query Parameter.Value --output text",
```
richardmcclellan, line 131: Looks like this is duplicated in `ConfigValueReader`.   Can we just use one, or the other?

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+      log: false
+    )
+  end
+
+  desc "Cleanup all the build artifacts"
```
richardmcclellan, line 136: When will this be called?

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,157 @@
...
+  private_lane :calculate_version_code do |options|
+    segments = options[:segments]
+    prerelease = segments.length() > 3 ?  0 : segments[-1].to_i
+    # Android version code max version = 2_100_000_000
+
+    # Derive the version code from version segements and pre-release number (if applicable)
+    # With this approach, you're "limited" to 999 minor versions per major and 999 patches per minor version.
+    # Limits: Max version for 1.x would be 1.999.999 with version code 199_999_900
+    segments[0].to_i * 100_000_000 + segments[1].to_i * 100_000 + segments[2].to_i * 100 + prerelease
+  end
```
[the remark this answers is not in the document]
rjuliano: Yes...I think we've successfully released after removing that from the gradle.properties file.

.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,58 @@
...
+        required: true
+        default: 'main'
+      git_user_name:
+        description: 'Git user name'
+        required: true
+        default: 'devops'
+      git_user_email:
+        description: 'Git user name'
+        required: true
+        default: 'devops@test.com'
```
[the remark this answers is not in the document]
rjuliano: We'll go with amplify-android-dev+ghops@amazon.com

.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+    - name: Update git
+      run: |
+        sudo add-apt-repository -y ppa:git-core/ppa
+        sudo apt-get update
+        sudo apt-get install git -y
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.inputs.base_branch }}
+        token: ${{ github.token }} # This will need to change to a token belonging to a service account.
```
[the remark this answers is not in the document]
rjuliano: github.token is an internal token that is tied to the Github Workflows app installation on the current repository. It has limited scope but it's good enough to do things like checking out code. https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

In this particular case, we don't have to specify anything actually since it defaults to github.token. Initially I thought we'd have to use our own token, but we can run the checkout action with github.token

.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,60 @@
+name: Create PR for a release
+on:
+  workflow_dispatch:
+    inputs:
```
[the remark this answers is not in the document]
rjuliano: I'm removing those and setting them as environment variables

.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,60 @@
...
+        bundle exec fastlane android configure_git_options git_user_email:${{ github.event.inputs.git_user_email }} git_user_name:${{ github.event.inputs.git_user_name }}
+    - name: Create/checkout a branch for the release
+      run: |
+        branch_name=release
+        git fetch --all
+        (git branch -D $branch_name &>/dev/null) && (echo 'Existing $branch_name branch deleted') || (echo 'No existing $branch_name branch to delete.')
+        git checkout -b $branch_name
+    - name: Create PR for next release
+      env:
+        RELEASE_MANAGER_TOKEN: ${{ secrets.RELEASE_MANAGER_TOKEN }}
```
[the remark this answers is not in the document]
rjuliano: This token has elevated privileges to create the PR and trigger downstream workflows. Actions triggered by `github.token` do not trigger downstream workflows. So in this case, we have to use our own token.

scripts/fastlane/Fastfile
rjuliano: That's correct. Based on our discussion, we will use the github API to create a release. I just haven't removed the code.

scripts/fastlane/Fastfile
rjuliano: Yup...this can go as well. I'll be working on creating the release in a separate workflow.

scripts/fastlane/Fastfile
rjuliano: Yup...another relic. Removing it.

scripts/fastlane/Fastfile
rjuliano: I'm adding a call to it after the pull request is created.

scripts/fastlane/Fastfile
richardmcclellan: Here's a fastlane plugin that looks like it does exactly what we want: https://github.com/xotahal/fastlane-plugin-semantic_release/blob/master/README.md

scripts/fastlane/Fastfile
richardmcclellan: Hmm, what's the reasoning for that?  I would think the only time we need to call gradle clean is right before gradle build.   Or, even easier, I believe the [fastlane gradle command](https://docs.fastlane.tools/actions/gradle/) allows you to run multiple commands, like this:

```
gradle(
    tasks: ["clean", "build"]
)

File: scripts/fastlane/Fastfile (hunk 3 above)
Comment (rjuliano) [reply scripts/fastlane/Fastfile:51]: I think I looked at this one before and ran into an issue. Can't remember what it was, but I think it had something to do with customizing the version bump logic. I'll check and see if we can just use it for generating changelog.

File: scripts/fastlane/Fastfile (hunk 11 above)
Comment (rjuliano) [reply scripts/fastlane/Fastfile:100]: We always start from a clean state because a new runner is created every time the workflow is launched. The clean at the end is intended to get rid of any artifacts we wouldn't want to linger between the time the build ends and when Github actually destroys that runner. Probably not necessary since we don't have any sensitive information being handled by this workflow.

File: scripts/fastlane/Fastfile (hunk 3 above)
Comment (richardmcclellan) [reply scripts/fastlane/Fastfile:51]: We could also just write it ourselves, and just use [the source of that plugin](https://github.com/xotahal/fastlane-plugin-semantic_release/blob/master/lib/fastlane/plugin/semantic_release/actions/conventional_changelog.rb#L48-L173) as a reference / example.

File: scripts/fastlane/Fastfile (hunk 11 above)
Comment (rjuliano) [reply scripts/fastlane/Fastfile]: Yeah...that makes sense. We're not really running a build here. I removed it.

File: scripts/fastlane/Fastfile (hunk 3 above)
Comment (rjuliano) [reply scripts/fastlane/Fastfile]: I was able to use the changelog generated using plugin. The issue I had was that in order to generate the changelog you also have to run `analyze_commits`. That function calculates the next version number using a logic that increments the version number on a per commit basis. So you end up skipping version numbers. Not a huge deal, but it's just not what we've been doing.

So, I was able to keep the version number calculated by our action, but replaced the change log generation code to use the semantic_release plugin.

File: scripts/fastlane/Fastfile [hunk 17]
```diff
@@ -0,0 +1,128 @@
...
+    # used by analyze_commits increments the version number once for each commit that's part of the release. As a result,
+    # you end up "wasting" version numbers. For example, your last release is 1.1.2 and you have 5 bug fixes to go in the
+    # next release, the version calculated by analyze_commit would be 1.1.7. We'll just use the version number we
+    # get from calculate_next_release_version. We'll use semantic_release for generating the change log since it has
+    # a richer feature set.
+    isReleasable = analyze_commits(match: "#{last_release_tag}", codepush_friendly:[], releases: releases)
+    update_change_log(next_version: next_version, last_release_tag: last_release_tag, commit_format: 'default', debug: true)
+
+    git_add(path: [gradle_properties_path, change_log_path])
+    git_commit(path: [gradle_properties_path, change_log_path], message: "chore(release): #{product_name} - #{next_version}")
```
richardmcclellan, line 53: This seems special enough that we could give it its own conventional commit prefix.  Then, I think it would be a little more straightforward to distinguish it as a trigger - what do you think?

```suggestion
    git_commit(path: [gradle_properties_path, change_log_path], message: "release: #{product_name} - #{next_version}")
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,128 @@
...
+    push_to_git_remote(force: true)
+
+
+    current_branch = sh('git', 'branch', '--show-current')
+    repo_name_parts = options[:repo].split('/')
+    create_pull_request(base: 'main', 
+                        title: "chore(release): #{product_name} - #{next_version}",
```
richardmcclellan, line 60:
```suggestion
                        title: "release: #{product_name} - #{next_version}", 
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,128 @@
...
+
+  end
+
+
+  desc "Create the release commit."
+  private_lane :release_commit do |options|
+    next_version = options[:version]
+    commit_message = "chore: release #{next_version} [skip ci]"
+    sh('git', 'commit', '-am', commit_message)
+  end
```
richardmcclellan, line 95: Looks like this is unused, right? 

```suggestion
```

scripts/fastlane/Fastfile
```diff
@@ -0,0 +1,128 @@
...
+
+  def self.from_ssm(param_name)
+    result = `aws ssm get-parameter --name #{param_name} --with-decryption --region us-east-1 --query Parameter.Value --output text`
+    if $?.exitstatus != 0
+      raise "Failed to get value for parameter #{param_name}"
+    end
+    result
+  end
+end
```
richardmcclellan, line 128: Unused. 
```suggestion
```

scripts/fastlane/Fastfile
rjuliano: Yup...that can go.

scripts/fastlane/Fastfile
rjuliano: Correct. Since we don't need fastlane to execute the uploadArchives task, all this stuff can go.

.github/workflows/release_pr.yml
```diff
@@ -0,0 +1,50 @@
...
+        cd scripts
+        gem install bundler
+        bundle install
+    - name: Configure git options
+      run: |
+        cd scripts
+        bundle exec fastlane android configure_git_options git_user_email:$GIT_USER_EMAIL git_user_name:$GIT_USER_NAME
+    - name: Create/checkout a branch for the release
+      run: |
+        branch_name=release
```
richardmcclellan, line 36: Naming this branch `release` doesn't seem appropriate, since no release is actually happening - it's just bumping the version.

```suggestion
        branch_name=bump_version
```

richardmcclellan (approved):
This looks great @rjuliano !  This is really going to streamline our release process!  🚀

PR #1197, opened 2021-02-25
chore: enable per-module versioning by override

jamesonwilliams:
By default, modules will use the `VERSION_CODE` and `VERSION_NAME` in `amplify-android/gradle.properties`.

Individual modules may override this value, if they use their own versioning. Such modules must populate `VERSION_CODE` and `VERSION_NAME` in _their own_ `gradle.properties`, e.g. `core-kotlin/gradle.properties`. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -129,6 +129,14 @@ subprojects { project ->
 }
 
 private void configureAndroidLibrary(Project project) {
+    project.ext.VERSION_NAME = project.hasProperty('VERSION_NAME') ?
+        project.findProperty('VERSION_NAME') :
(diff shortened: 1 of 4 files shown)


rjuliano (approved):
Ship it! 🚢  From an automation standpoint, this should work with the approach we're planning to use. The automated version bump only happens in the root gradle.properties so any modules that choose to override should be safe.

richardmcclellan (approved):
Looks good.  I'll release the library this afternoon.  If you're free, I'll give you a call before I hit the release button so we can get two sets of eyes on the generated artifacts and just make sure everything looks right.

Issue #1178, opened 2021-02-18
Pass Custom Header through AWSApiPlugin


meenakshi-rai:
### Problem Statment
I am trying to pass external “accessToken” as header with each `Amplify GraphQL API` call while using `Android Amplify SDK`, and **I need help with a solution for this**.

#### Build gradle library version
```
dependencies {
    implementation 'com.amplifyframework:core:1.16.12'
    implementation 'com.amplifyframework:aws-api:1.16.12'
}
```

#### Code Used
```kotlin
val awsApiPlugin = AWSApiPlugin()
Amplify.addPlugin(awsApiPlugin)
Amplify.configure(applicationContext)
```

### Looking for something similar to iOS Solution
When I use **_the iOS counterpart_** of Android Amplify API SDK, while configuring the API, I get an option with `AWSAPIPlugin` extension to add a custom interceptor using protocol `URLRequestInterceptor`.
With the help of  this cusomt `URLRequestInterceptor` I can pass “accessToken” as header, which is automatically consumed with each `GraphQL API` call, and **I am trying to find something similar in Android**

#### iOS Library Link
https://github.com/aws-amplify/amplify-ios

#### Podfile libraries version
 ```
pod 'Amplify', '~> 1.6.0'
pod 'AmplifyPlugins/AWSAPIPlugin', '~> 1.6.0'
```
 
#### Code Used
Code example used for iOS, where I created a CustomURLRequestInterceptor implementing URLRequestInterceptor which contains code for me to add headers as required.
```swift
val apiPlugin = try AWSAPIPlugin()
try Amplify.configure()
try apiPlugin.add(interceptor: CustomURLRequestInterceptor(), for: "<Api Name>")
 
// Custom CustomURLRequestInterceptor used to pass header with each graphQL API requests
struct CustomURLRequestInterceptor: URLRequestInterceptor {
    func intercept(_ request: URLRequest) throws -> URLRequest {

        guard let mutableRequest = (request as NSURLRequest).mutableCopy() as? NSMutableURLRequest else {
            throw APIError.unknown("Could not get mutable request", "")
        }
        mutableRequest.setValue("abcdefgh-abcd-abcd-abcd-abcdefghijkl", forHTTPHeaderField: "accessToken")
        return mutableRequest as URLRequest
    }
}
```

richardmcclellan:
[...]

Issue #913, opened 2020-10-19
AaU, I want to configure the timeouts used by the AWSApiPlugin.


merveylcu:
I randomly get the following error when I send a request. (Regardless of GET, POST)
**`com.amplifyframework.api.ApiException: Received an IO exception while making the request.`**

Here's my code:

``` kotlin
fun <K, T> sendRequest(
        request: K? = null, api: APIs,
        queryMap: MutableMap<String, String>? = mutableMapOf(),
        pathParameters: MutableMap<String, String>? = mutableMapOf(),
        onSuccessAction: ((BaseResponse<T>) -> Unit)? = null
    ) {
        showIndicator()
        //Dynamic path parameters control
        var editedEndpoint = api.endpoint
        pathParameters?.forEach {
            editedEndpoint = editedEndpoint.replace("{${it.key}}", it.value)
        }

        when (api.method) {
            RequestMethod.POST -> {
                val json = gson.toJson(request)
                val bytes = json.toByteArray(StandardCharsets.UTF_8)
                
                val options = RestOptions.builder()
                    .addPath(editedEndpoint)
                    .addBody(bytes)
                    .addHeaders(
                        mutableMapOf(
                            "Content-Type" to "application/json",
                            "Accept" to "application/json"
                        )
                    )
                    .addQueryParameters(queryMap ?: mapOf())
                    .build()
                Amplify.API.post(options,
                    { response ->
                        when {
                            response.code.isSuccessful -> {

                                val hasData = response.data.asJSONObject().isNull("data").not()
                                if (hasData && api.responseType != Empty::class.java) {
                                    val typedData = gson.fromJson(
                                        response.data.asJSONObject()["data"].toString(),
                                        api.responseType
                                    )
                                    onSuccessAction?.invoke(BaseResponse(typedData) as BaseResponse<T>)
                                    hideIndicator()
                                } else {
                                    onSuccessAction?.invoke(BaseResponse(null))
                                    hideIndicator()
                                }
                            }
                            response.code.isClientError -> {
                                
                            }
                            response.code.isServiceFailure -> {
                                
                            }
                        }
                    },
                    {
                        //API Error
                    })
            }

            RequestMethod.GET -> {
                val options = RestOptions.builder()
                    .addPath(editedEndpoint)
                    .addHeaders(
                        mutableMapOf(
                            "Content-Type" to "application/x-www-form-urlencoded; charset=utf-8",
                            "Accept" to "application/json"
                        )
                    )
[...]

PR #1207, opened 2021-03-02
feat(aws-api): Support Custom OkHttp Configurations

jamesonwilliams:
While constructing the AWSApiPlugin, the customer may now apply
additional configurations to the OkHttpClient instance that is used
internally.

For example, a user may like to add custom headers to outgoing requests:

```java
AWSApiPlugin plugin = AWSApiPlugin.builder()
    .customizeClient("apiName", okHttpClientBuilder -> {
        okHttpClientBuilder.addInterceptor(chain -> {
            return chain.proceed(chain.request()
                .newBuilder()
                .addHeader("specialKey", "specialValue")
                .build();
            );
        });
    })
    .build();
```

The same facility may be used to set custom timeouts, e.g.:
```java
AWSApiPlugin plugin = AWSApiPlugin.builder()
    .customizeClient("apiName", okHttpClientBuilder -> {
        okHttpClientBuilder.connectTimeout(10, TimeUnit.SECONDS);
    })
    .build();
```

Resolves: https://github.com/aws-amplify/amplify-android/issues/1178
Resolves: https://github.com/aws-amplify/amplify-android/issues/913

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -781,7 +781,7 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
         /**
          * Apply customizations to an underlying OkHttpClient that will be used
          * for a particular API.
-         * @param forApiName The name of the API for which these customizations should apply
+         * @param forApiName The name of the API for which these customizations should apply.  This can be found in your `amplifyconfiguration.json` file.
          * @param byConfigurator A lambda that hands the user an OkHttpClient.Builder,
          *                       and enables the user to set come configurations on it.
          * @return A builder instance, to continue chaining configurations
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -781,7 +781,8 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
         /**
          * Apply customizations to an underlying OkHttpClient that will be used
          * for a particular API.
-         * @param forApiName The name of the API for which these customizations should apply.  This can be found in your `amplifyconfiguration.json` file.
+         * @param forApiName The name of the API for which these customizations should apply.
+                             This can be found in your `amplifyconfiguration.json` file.
          * @param byConfigurator A lambda that hands the user an OkHttpClient.Builder,
          *                       and enables the user to set come configurations on it.
          * @return A builder instance, to continue chaining configurations
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -39,6 +39,7 @@ import com.amplifyframework.core.Action;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.hub.HubChannel;
+import com.amplifyframework.util.Immutable;
 import com.amplifyframework.util.UserAgent;
 
 import org.json.JSONObject;
@@ -69,6 +70,7 @@ import okhttp3.Protocol;
 @SuppressWarnings("TypeParameterHidesVisibleType") // <R> shadows >com.amplifyframework.api.aws.R
 public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
     private final Map<String, ClientDetails> apiDetails;
+    private final Map<String, OkHttpConfigurator> apiConfigurators;
     private final GraphQLResponse.Factory gqlResponseFactory;
     private final ApiAuthProviders authProvider;
     private final ExecutorService executorService;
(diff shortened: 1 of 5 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -736,4 +776,78 @@ private void transitionTo(ApiEndpointStatus newStatus) {
...
+         * Add a global interceptor. The interceptor will be invoked for
+         * all API endpoints you've configured.
+         * Interceptors are applied in order. If you have multiple interceptors,
+         * the first-added interceptors get used first. Global interceptors
+         * are always applied first.
+         * @param interceptor An interceptor to apply to all API calls.
+         * @return A builder, to continue configuration of an AWSApiPlugin
+         */
+        @NonNull
+        public Builder addGlobalInterceptor(@NonNull Interceptor interceptor) {
```
richardmcclellan, line 816: Hmm, this is exposing our OkHttp dependency in the public interface, which goes against our tenets, because it will prevent us from ever swapping out `OkHttp` for another HTTP client.  On the other hand, this is just in the plugin, not the category, and I think it's _unlikely_ that we'd ever swap it out (famous last words).

Would it be possible to create a facade `Interceptor` that decouples the public API from `OkHttp`.  Not sure if that's worth the effort or not.

jamesonwilliams: Hey @richardmcclellan! I deliberated on this same thing.

iOS does include a first party interceptor type in their *core* spec, @lawmicha and I chatted and concluded it didn't need to be a core spec.

IMO plugins themselves don't need much abstraction. The primary Inversion of Control in Amplify is along the Category/Plug-in boundary.

I can try a version with a first party interface, too, though.

drochetti (commented):
I looked through the changes and overall it lgtm! It didn't raise any flags, but I would love to hear from someone who knows the codebase better than I do.

One suggestion I have is to put in the readme a code snippet of a header customization, so that we can look at the DX of these changes based on the use case we're trying to support. What that would look like?

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
jamesonwilliams: Update: I wrote an internal design doc and implemented a different solution. New solution is pushed. It addressed this feedback.

jamesonwilliams:
> I looked through the changes and overall it lgtm! It didn't raise any flags, but I would love to hear from someone who knows the codebase better than I do.
> 
> One suggestion I have is to put in the readme a code snippet of a header customization, so that we can look at the DX of these changes based on the use case we're trying to support. What that would look like?

Good idea! Will add the DX in the description.

aws-api/src/main/java/com/amplifyframework/api/aws/HeaderSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+ * interface to be usable. Those internal concerns are maintained in
+ * an internal state bundle, {@link HeaderSupplierInfo}.
+ */
+@FunctionalInterface
+public interface HeaderSupplier {
+    /**
+     * Supplies the next header value.
+     * @return A header value
+     */
+    String nextValue();
```
rjuliano, line 32: At first I wasn't sure about the `nextValue()` name in the context of supplier (in my mind, when I see supplier I thing `get()`). But I think in the context of this change, `nextValue()` is OK since this method will be called for each request issued. Just wanted to confirm that this is in fact your interpretation.

jamesonwilliams: I updated the PR, and this method no longer exists.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -78,30 +85,40 @@
...
         this.requestDecorator = new AuthRuleRequestDecorator(authProvider);
+        this.globalInterceptors = builder.globalInterceptors;
+        this.apiInterceptors = builder.apiInterceptors;
+    }
+
+    /**
+     * Begins construction of a new AWSApiPlugin instance by using a fluent builder.
+     * @return A builder to help construct an AWSApiPlugin
+     */
+    public static Builder builder() {
```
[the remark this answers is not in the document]
jamesonwilliams: Resolution: we discussed this in design review today and decided to go with the builder.

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -736,4 +753,53 @@ private void transitionTo(ApiEndpointStatus newStatus) {
...
+        public Builder apiAuthProviders(@NonNull ApiAuthProviders apiAuthProviders) {
+            Objects.requireNonNull(apiAuthProviders);
+            Builder.this.apiAuthProviders = apiAuthProviders;
+            return Builder.this;
+        }
+
+        /**
+         * Apply customizations to an underlying OkHttpClient that will be used
+         * for a particular API.
+         * @param forApiName The name of the API for which these customizations should apply
```
richardmcclellan, line 784:
```suggestion
         * @param forApiName The name of the API for which these customizations should apply.  This can be found in your `amplifyconfiguration.json` file.
```

aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -781,7 +781,7 @@ public Builder apiAuthProviders(@NonNull ApiAuthProviders apiAuthProviders) {
         /**
          * Apply customizations to an underlying OkHttpClient that will be used
          * for a particular API.
-         * @param forApiName The name of the API for which these customizations should apply
+         * @param forApiName The name of the API for which these customizations should apply.  This can be found in your `amplifyconfiguration.json` file.
```
jamesonwilliams, line 784:
```suggestion
         * @param forApiName The name of the API for which these customizations should apply.
                             This can be found in your `amplifyconfiguration.json` file.
```

richardmcclellan (approved):
This looks great!  🚢

PR #1209, opened 2021-03-04
chore(datastore): ignore OperationDisabled subscription errors

richardmcclellan:
### Use Case

AppSync doesn't support runtime filtering for subscriptions today.  This means there's no way to limit the network traffic transmitted via subscriptions, and for some applications, the amount of traffic could overwhelm a client.   Some customers wish to disable subscriptions on the backend, but allow DataStore to proceed with sync queries, and continue working, just without real-time updates.

### Solution

Customers can throw an `OperationDisabled` error in their VTL response mapping template which will be ignored by the DataStore `SubscriptionProcessor`.    Note: This is subject to be removed in the future, and should not be used without prior consultation with AWS.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
@@ -145,6 +145,13 @@ public final class AppSyncExtensions {
         CONFLICT_UNHANDLED("ConflictUnhandled"),
 
         /**
+         * This error is not for general use unless you have consulted directly with AWS.  When DataStore encounters
+         * this error, it will ignore the error and allow DataStore to continue running.  This error is subject to be
+         * deprecated/removed in the future.
+         */
+        OPERATION_DISABLED("OperationDisabled"),
+
+        /**
          * An Unauthorized error will occur if the provided credentials are not authorized for the requested operation.
          */
         UNAUTHORIZED("Unauthorized");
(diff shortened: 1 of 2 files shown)


jamesonwilliams:
Hey @richardmcclellan,

Last time we reviewed error handling in DataStore, we talked about broadcasting errors on the Hub. I believe we had discussed removing this closure from the config.

A user would be able to receive an error on the Hub, and then respond to it by calling .stop() on the DataStore if necessary, right?

richardmcclellan:
> Hey @richardmcclellan,
> 
> Last time we reviewed error handling in DataStore, we talked about broadcasting errors on the Hub. I believe we had discussed removing this closure from the config.
> 
> A user would be able to receive an error on the Hub, and then respond to it by calling .stop() on the DataStore if necessary, right?

Yes!  The Hub category is a great mechanism for notifying customers of an error.  However, there's not a way for a customer to _respond_ to or _handle_ an error from a Hub event.  This PR is proposing that we should use the `DataStoreErrorHandler` for this purpose.  Customers can now return true from the `handle(...)` method to indicate that they take responsibility for handling it, and DataStore should ignore it and continue running.  This is similar to the `DataStoreConflictHandler` where customers can return a `ConflictResolutionDecision`, to specify how DataStore should respond to the conflict.

I'm considering defining an `ErrorHandlerDecision` interface with some static `ErrorHandlerDecision.Handled` and `ErrorHandlerDecision.Unhandled` factory methods, that a customer would return instead of a `boolean`, to allow this to be extensible, though I don't know of any specific use case that would need this flexibility, currently.

jamesonwilliams (approved):

PR #1215, opened 2021-03-09
chore: update stalebot settings

rjuliano:
Setting daysUntilStale to 30

Borrowed the same setting from [amplify-js](https://github.com/aws-amplify/amplify-js/blob/main/.github/stale.yml)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -3,6 +3,13 @@ daysUntilStale: 30
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
 # Label to use when marking an issue as stale
+exemptLabels:
+  - Feature Request
+  - Good First Issue
+  - Improvement
+  - Infrastructure
+  - Pending Release
+  - RFC
 staleLabel: Closing Soon
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,6 +1,5 @@
 # Number of days of inactivity before an issue becomes stale
-# Setting this to 100 years, because we do not want to automatically mark issues as stale
-daysUntilStale: 36500
+daysUntilStale: 30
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -1,6 +1,5 @@
 # Number of days of inactivity before an issue becomes stale
-# Setting this to 100 years, because we do not want to automatically mark issues as stale
-daysUntilStale: 36500
+daysUntilStale: 30
```
jamesonwilliams, line 2: Looks like the docs, JS and CLI repos have something like this. _However_, they also include some "exempt labels." In practice, the exempt labels are most of their issues.

Refer:
 - https://github.com/aws-amplify/amplify-js/blob/be53c6187057728a9e2cf24265324fd1f470e5e0/.github/stale.yml#L2-L12
 - https://github.com/aws-amplify/amplify-cli/blob/91cbd57cdeb8210ff7ae159b16b3ea7604c7e124/.github/stale.yml#L2-L13
 - https://github.com/aws-amplify/docs/blob/7162c8103511e3899c034a880e43a7db3ea0fb4e/.github/stale.yml#L1-L13

Should we add some of our own exempt labels, so we don't accidentally close important roadmap items?

FWIW, I don't think iOS is currently using this functionality. But perhaps they should. https://github.com/aws-amplify/amplify-ios/tree/main/.github

rjuliano: That's a good call on the exempt labels. I'll come up with a list and will update the PR.

It looks like the iOS repo [is setup as a workflow](https://github.com/aws-amplify/amplify-ios/blob/main/.github/workflows/stale.yml). I'm not sure whether there's an advantage of using one versus the other.

jamesonwilliams: Ahh! Okay, cool.

jamesonwilliams (approved):

PR #1220, opened 2021-03-10
chore: remove DataStoreException from catch since try body never throws

richardmcclellan:
This PR fixes the issue causing our build to fail on the `main` branch.  It broke after https://github.com/aws-amplify/amplify-android/pull/1149/ was merged, which is strange, because the build on that PR passed.    The reason for this though, is because https://github.com/aws-amplify/amplify-android/pull/1121/ was merged in between, which refactored part of `SQLiteModelTree` such that the body of the try block was no longer ever throwing a `DataStoreException`. 


rjuliano (approved):
Ship it 🚢

PR #1225, opened 2021-03-12
fix(datastore): optimize sync queries with predicates by wrapping in an AND group

richardmcclellan:
When a filter is provided on a sync operation, AppSync will attempt to optimize the request by performing a DynamoDB query, rather than a scan.  A query is only possible if the filter contains the hash key, or the hash key and the sort key.     To perform this optimization, the [AppSync resolver requires the presence of a top level "and"](https://github.com/aws-amplify/amplify-cli/blob/018548a772cab14a1f99fb3d2413eb720142a562/packages/graphql-key-transformer/src/KeyTransformer.ts#L1105) group.   

This PR wraps the sync expression filter with an AND group if the predicate provided by the customer is just a predicate operation, rather than a group.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
@@ -115,7 +115,7 @@ final class AppSyncRequestFactory {
                     // provided syncPredicate is already a QueryPredicateGroup, this is not needed.  If the provided
                     // group is of type AND, the optimization will occur.  If the top level group is OR or NOT, the
                     // optimization is not possible anyway.
-                    syncPredicate = new QueryPredicateGroup(QueryPredicateGroup.Type.AND, Arrays.asList(syncPredicate));
+                    syncPredicate = QueryPredicateGroup.andOf(syncPredicate);
                 }
                 builder.variable("filter", filterType, parsePredicate(syncPredicate));
             }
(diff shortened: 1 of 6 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
```diff
@@ -108,7 +108,16 @@ private AppSyncRequestFactory() {}
...
                 String filterType = "Model" + Casing.capitalizeFirst(modelSchema.getName()) + "FilterInput";
-                builder.variable("filter", filterType, parsePredicate(predicate));
+                QueryPredicate syncPredicate = predicate;
+                if (syncPredicate instanceof QueryPredicateOperation) {
+                    // When a filter is provided, wrap it with a predicate group of type AND.  By doing this, it enables
+                    // AppSync to optimize the request by performing a DynamoDB query instead of a scan.  If the
+                    // provided syncPredicate is already a QueryPredicateGroup, this is not needed.  If the provided
+                    // group is of type AND, the optimization will occur.  If the top level group is OR or NOT, the
+                    // optimization is not possible anyway.
+                    syncPredicate = new QueryPredicateGroup(QueryPredicateGroup.Type.AND, Arrays.asList(syncPredicate));
```
jamesonwilliams, line 118: This might look better wrapped into a factory method:

```java 
QueryPredicateGroup.andOf(syncPredicate)
```
It would prevent the access escalation of the constructor, too.

jamesonwilliams (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java
richardmcclellan: I had considered something like that, but couldn't come up with a good name for the method itself.  I think `andOf` works though - I'll do that!

jamesonwilliams (approved):

jamesonwilliams (approved):

PR #1234, opened 2021-03-17
chore(ci): create test report dir before execution

rjuliano:
*Issue #, if available:*

*Description of changes:*
To troubleshoot integration tests, we need to ensure that the reports are exported correctly after execution. In certain instances, tests are failing and the reports are not getting generated. It's not clear why this is happening, but by creating the report output directory before running the build, we can at least allow the post-build event to run successfully.


richardmcclellan (approved):
🎉

PR #1238, opened 2021-03-17
feat(api):allows callers to specify auth mode

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR would enable customers to specify which auth mode to use on a per-request basis when interacting with the API plugin. 

A field called `authorizationType` was added to the `AppSyncGraphQLRequest` object and associated builder. If set, the authorization type from the request will be used when building the HTTP request.

In order to do that, the request has to be signed at the time when the HTTP request is created rather than using the interceptor. This is done by creating an `ApiRequestSigner` interface and implementations for each supported auth type.

A class called `ApiRequestSignerFactory` was created. An instance of it is added to the `ClientDetails` instance for each API. The API plugin passes the factory instance to the `AppSyncGraphQLOperation` which in turn signs the request inside the `start()` method.


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -663,18 +666,21 @@ private RestOperation createRestOperation(
...
          * It associates a http client with its dedicated endpoint.
          */
         ClientDetails(
-                final ApiConfiguration apiConfiguration,
-                final OkHttpClient okHttpClient,
-                final SubscriptionEndpoint subscriptionEndpoint) {
+            final ApiConfiguration apiConfiguration,
+            final OkHttpClient okHttpClient,
+            final SubscriptionEndpoint subscriptionEndpoint,
+            ApiRequestSignerFactory apiRequestSignerFactory) {
```
richardmcclellan, line 679:
```suggestion
            final ApiRequestSignerFactory apiRequestSignerFactory) {
```

aws-api/src/main/java/com/amplifyframework/api/aws/auth/JWTTokenApiRequestSigner.java
```diff
@@ -0,0 +1,44 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.api.aws.auth;
+
+import com.amazonaws.Request;
+
+import java.util.function.Supplier;
```
jamesonwilliams, line 20: The `java.util.function.*` are from Java 8, and potentially may be desugared by core library desugaring?

A 100% safe alternative would just be to add an inner interface:
```java
public class JWTTokenApiRequestSigner extends ApiRequestSigner {
    public JWTTokenApiRequestSigner(TokenSupplier tokenSupplier) {
        ...
    }
    ...
    @FunctionalInterface public interface TokenSupplier {
        public String supply();
    }
    ...
}
```

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSigner.java
```diff
@@ -0,0 +1,137 @@
...
+            requestBody = null;
+        }
+
+        okReqBuilder.method(req.method(), requestBody);
+
+        //continue with chain.
+        return okReqBuilder.build();
+    }
+
+    // Extracts query string parameters from a URL.
```
richardmcclellan, line 114: I realize this is existing code that you are just moving around, but it would be cleaner to use the `Uri` class in Android as [described here](https://stackoverflow.com/a/40466295/282279), instead of manually splitting the url string.

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSigner.java
```diff
@@ -0,0 +1,137 @@
...
+            bodyBytes = "".getBytes();
+        }
+        dr.setContent(new ByteArrayInputStream(bodyBytes));
+
+        //set the query string parameters
+        dr.setParameters(splitQuery(req.url().url()));
+
+        addAuthHeader(dr);
+        //dr.addHeader(AUTHORIZATION, cognitoUserPoolsAuthProvider.getLatestAuthToken());
+        //dr.addHeader(AUTHORIZATION, oidcAuthProvider.getLatestAuthToken());
```
richardmcclellan, line 89:
```suggestion
```

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSigner.java
```diff
@@ -0,0 +1,137 @@
...
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okio.Buffer;
+
+/**
+ * Abstract class that defines some basic functionality with regards to signing AppSync requests.
+ * Implementations of this class should implement the {@link ApiRequestSigner#addAuthHeader(com.amazonaws.Request)}
+ * method.
+ */
+public abstract class ApiRequestSigner {
```
jamesonwilliams, line 45: It seems like the `JWTTokenApiRequestSigner` instance of this type isn't actually a signer.

Maybe the base type should just be:

```java
public interface RequestDecorator {
    public <T> Request<T> decorate(Request<T> request);
}
```

Then, the signing stuff can be a final (closed) utility class. I don't think you need inheritance.

Each individual implementation of the request decorator could get the signing values from the signer, and then populate them into the header using their specific header keys.

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSignerFactory.java
```diff
@@ -0,0 +1,128 @@
...
+            apiAuthProviders.getApiKeyAuthProvider() != null) {
+            return new ApiKeyApiRequestSigner(apiAuthProviders.getApiKeyAuthProvider());
+        } else if (AuthorizationType.AWS_IAM.equals(authorizationType) &&
+            apiAuthProviders.getAWSCredentialsProvider() != null) {
+            AppSyncV4Signer appSyncV4Signer = new AppSyncV4Signer(region);
+            return new IamApiRequestSigner(appSyncV4Signer, apiAuthProviders.getAWSCredentialsProvider());
+        } else {
+            return NO_OP_REQUEST_SIGNER;
+        }
+
```
jamesonwilliams, line 126:
```suggestion
```

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSignerFactory.java
```diff
@@ -0,0 +1,128 @@
...
+                        return apiAuthProviders.getCognitoUserPoolsAuthProvider().getLatestAuthToken();
+                    } catch (ApiException apiException) {
+                        LOG.error("Failed to retrieve token from CognitoUserPoolsAuthProvider", apiException);
+                        return null;
+                    }
+                }
+            });
+        } else if (AuthorizationType.OPENID_CONNECT.equals(authorizationType) &&
+            apiAuthProviders.getOidcAuthProvider() != null) {
+            return new JWTTokenApiRequestSigner(new Supplier<String>() {
```
jamesonwilliams, line 105: Should be able to use a lambda expression here (and above)

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSignerFactory.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    public ApiRequestSignerFactory(@NonNull ApiAuthProviders apiAuthProviders,
+                                   @NonNull AuthorizationType defaultAuthorizationType,
+                                   @NonNull String region) {
+        Objects.requireNonNull(apiAuthProviders);
+        Objects.requireNonNull(defaultAuthorizationType);
+        Objects.requireNonNull(region);
+        this.apiAuthProviders = apiAuthProviders;
+        this.defaultAuthorizationType = defaultAuthorizationType;
+        this.region = region;
```
jamesonwilliams, line 63: For short lines, you can write this more compactly -- `Objects.requireNonNull` returns the non-null value:
```suggestion
        this.apiAuthProviders = Objects.requireNonNull(apiAuthProviders);
        this.defaultAuthorizationType = Objects.requireNonNull(defaultAuthorizationType);
        this.region = Objects.requireNonNull(region);
```

jamesonwilliams (commented):
Feels like we're being held back by some of our existing debt in the SigV4 signing code, even before you go into make the intended change.

Maybe it makes sense to refactor that a bit, before adding in the auth mode changes, itself?

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSignerFactory.java
```diff
@@ -0,0 +1,128 @@
...
+        // the request's authorization type is null
+        if (!(graphQLRequest instanceof AppSyncGraphQLRequest<?>) ||
+            ((AppSyncGraphQLRequest<?>) graphQLRequest).getAuthorizationType() == null) {
+            // Use the default auth type.
+            return forAuthType(defaultAuthorizationType);
+        } else {
+            // Honor the auth type from the request.
+            return forAuthType(((AppSyncGraphQLRequest<?>) graphQLRequest).getAuthorizationType());
+        }
+    }
```
richardmcclellan, line 82: Perhaps slightly cleaner to do something like this?
```suggestion
        AuthorizationType authType = defaultAuthorizationType;
        if(graphQLRequest instanceof AppSyncGraphQLRequest<?>
            && ((AppSyncGraphQLRequest<?>) graphQLRequest).getAuthorizationType() != null) {
            authType = ((AppSyncGraphQLRequest<?>) graphQLRequest).getAuthorizationType()
        }
        return forAuthType(authType);
    }
```

richardmcclellan (commented):
Great idea breaking this down into smaller chunks like this, it's much easier to review!  Overall direction looks good, just a few small comments!

aws-api/src/main/java/com/amplifyframework/api/aws/auth/JWTTokenApiRequestSigner.java
rjuliano: That's right...forgot about that. Every time I use `Supplier` I think to myself "why don't I see more of it in the codebase?" 🤦  🤣

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSigner.java
rjuliano: I like that. As you mentioned, there's a bit of an identity crisis with the sigv4-related code. It probably makes sense to untangle some that.

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestSigner.java
rjuliano: Makes sense. I'll take a look at that. There are some other changes that need to be made in this code, so I'll be working on that today.

rjuliano:
Need to make sure we have integration test coverage for this.

rjuliano:
Ugh...started peeling the layers and found some stuff I needed to address.

1. `AuthorizationType` is actually used for REST API as well, so putting it in the `aws-api-appsync` package was no bueno. I moved it core. Not totally convinced it belongs there since it's AWS-specific.  Thoughts?
2. Since REST and GraphQL APIs share an HTTP client, the REST calls started to fail because of the removal of the interceptor. So I put a similar pattern in place for the REST APIs. If the incoming request has the authorization type, we use that, otherwise we use default auth.

richardmcclellan:
> Ugh...started peeling the layers and found some stuff I needed to address.
> 
> 1. `AuthorizationType` is actually used for REST API as well, so putting it in the `aws-api-appsync` package was no bueno. I moved it core. Not totally convinced it belongs there since it's AWS-specific.  Thoughts?

Hmm, no bueno just because of the name, right?  Or is there actually a dependency issue?  Perhaps we just need to rename `aws-api-appsync` to something else, like `aws-api-core`?

Or, crazy idea - what if we just make `aws-api` a dependency of `aws-datastore`, and then everything in `aws-api-appsync` can just move to `aws-api`?

rjuliano:
> > Ugh...started peeling the layers and found some stuff I needed to address.
> > 
> > 1. `AuthorizationType` is actually used for REST API as well, so putting it in the `aws-api-appsync` package was no bueno. I moved it core. Not totally convinced it belongs there since it's AWS-specific.  Thoughts?
> 
> Hmm, no bueno just because of the name, right? Or is there actually a dependency issue? Perhaps we just need to rename `aws-api-appsync` to something else, like `aws-api-core`?
> 
> Or, crazy idea alert - what if we just make `aws-api` a dependency of `aws-datastore`, and then everything in `aws-api-appsync` can just move to `aws-api`?

Correct. The naming just felt a bit odd at that point. Going to think about the alternatives you mentioned and see what I can come up with.

jamesonwilliams:
> > Ugh...started peeling the layers and found some stuff I needed to address.
> 
> > 
> 
> > 1. `AuthorizationType` is actually used for REST API as well, so putting it in the `aws-api-appsync` package was no bueno. I moved it core. Not totally convinced it belongs there since it's AWS-specific.  Thoughts?
> 
> 
> 
> Hmm, no bueno just because of the name, right?  Or is there actually a dependency issue?  Perhaps we just need to rename `aws-api-appsync` to something else, like `aws-api-core`?
> 
> 
> 
> Or, crazy idea alert - what if we just make `aws-api` a dependency of `aws-datastore`, and then everything in `aws-api-appsync` can just move to `aws-api`?
> 
> 
> 
> 

I'd encourage you guys to think of this in terms of three autonomous components: data modeling, AppSync, and Auth.

aws-api-appsync should really be like an AppSync client. I would have called it just appsync.

Both API and DataStore use AppSync. Most of the meat and potatoes should live in the AppSync client, though. The API and DataStore plugins should just be thin integrations.

In a perfect world, @jpignata's org would own their clients, and Amplify's plugins would just integrate them.

Amplify's AppSync SDK module has some auth concerns and it should own the AutheoizationType enum since this is codified in the AppSync service docs.

rjuliano:
> I'd encourage you guys to think of this in terms of three autonomous components: data modeling, AppSync, and Auth.
> 
> aws-api-appsync should really be like an AppSync client. I would have called it just appsync.
> 
> Both API and DataStore use AppSync. Most of the meat and potatoes should live in the AppSync client, though. The API and DataStore plugins should just be thin integrations.

Definitely agree with that. We do still have quite a bit of AppSync-specific stuff in the API plugin. Moving those things to `aws-api-appsync` would make a lot of sense. 

> 
> In a perfect world, @jpignata's org would own their clients, and Amplify's plugins would just integrate them.
> 
> aws-api-appsync should have _some_ auth concerns and it should own the AuthorizationType enum since that concept is codified in the AppSync service docs.

That was my first instinct as well. However, API gateway also uses the same authorization types AppSync (at least that's how it gets presented through the Amplify docs). Maybe it makes sense to break up apigateway-specific things into its own package (including a construct to represent authorization types). I wasn't keen on have separate enums for each, but I think as long as they both match the values added to the config files by the CLI, we should be OK. At that point, `aws-api` (the plugin) will make use of `aws-appsync` and `aws-apigateway` (the clients) to interact with the backends.

rjuliano:
So I made a couple more tweaks to clean some stuff up. Essentially, I'm leaving REST-related logic alone for the most part. The feature that allows users to specify an auth mode applies only to GraphQL requests. That behavior is consistent with JS.

I'd like to break up a lot of the logic of this plugin into separate modules by moving AppSync-related stuff to `aws-api-appsync` and move REST-related stuff to its own module as well. That's a bigger task than this PR and I'm not sure how high it is on the priority list right now.

jamesonwilliams (approved):

PR #1249, opened 2021-03-24
fix: Add support for readonly fields such as createdAt, updatedAt

richardmcclellan:
### Problem
Currently the CLI autogenerates `createdAt` and `updatedAt` fields on every @model type, as a convenience for customers.  
Customers can also specify custom field names.  To generate `createdOn` and `updatedOn`, a customer can define their type like this:

```
type Post @model(timestamps:{createdAt: "createdOn", updatedAt:"updatedOn"} {
  id: ID!
}
```

These fields are not currently included on the code generated models yet, so customers can't really use them.   Codegen is being updated to add these fields (https://github.com/aws-amplify/amplify-codegen/pull/114).  However, these fields need to be treated differently than existing fields - that is, they should not be included in the "input" variable used for mutations.    If they are included, AppSync returns an error.  Essentially, they are read-only.

### Solution
Since these fields can be customized to any value, we can't just update the library to omit `createdAt` and `updatedAt` from the mutation input.  Instead, codegen is going to annotate these fields as read only.  This PR adds an `isReadOnly` property to the `@ModelField` annotation, which defaults to `false`.  Codegen will set this field to true for the timestamp fields.  This will ensure that they still get requested in the selection set, and read in the response, but they just won't be able to be modified.

The code generated fields will look like this:
```
@ModelField(targetType = "AWSDateTime", isReadOnly = true)
private Temporal.DateTime createdAt;

@ModelField(targetType = "AWSDateTime", isReadOnly = true)
private Temporal.DateTime updatedAt;
``` 


testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/BlogOwner.java
```diff
@@ -24,6 +25,9 @@
   private final @ModelField(targetType="ID", isRequired = true) String id;
   private final @ModelField(targetType="Blog") @HasOne(associatedWith = "owner", type = Blog.class) Blog blog = null;
   private final @ModelField(targetType="String") String wea;
+  private @ModelField(targetType = "AWSDateTime", isReadOnly = true) Temporal.DateTime createdAt;
```
jamesonwilliams, line 28: Should these two field be final?

jamesonwilliams (approved):
Great documentation in the PR description and a clean PR overall. Ship it!

testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/BlogOwner.java
richardmcclellan: I don't think they can be, because they are not initialized in the constructor.  The only time they will actually be set is through Gson deserialization from the API response.

AaronZyLee (commented):
From the sample code, I summary the changes for java modelgen as following:

1. Add timestamp as a new class field(private but not final)
2. Add `isReadOnly` in `ModelField` annotation
3. Add getter method for timestamp fields
4. Add relevant changes to `equals` and `hasCode`

Just want to confirm the changes and I will include them in the codegen PR.

richardmcclellan:
> From the sample code, I summary the changes for java modelgen as following:
> 
> 1. Add timestamp as a new class field(private but not final)
> 2. Add `isReadOnly` in `ModelField` annotation
> 3. Add getter method for timestamp fields
> 4. Add relevant changes to `equals` and `hasCode`
> 
> Just want to confirm the changes and I will include them in the codegen PR.

Yes, that is all correct, except also update `toString` as well.  The `Person.java` class in my PR didn't have a `toString` method at all before, but I've just pushed an update to add it.

rjuliano (approved):
LGTM!!! 🚢 

I checked the integration tests and I believe they're failing because of the issue with mobile client.

PR #1252, opened 2021-03-25
fix(aws-api): double-encode plus in url path segments

raphkim:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/446

*Description of changes:*
Double-encode "+" in path segments. `URLDecoder` does not behave as expected with `"+"` symbols in URL paths (it gets converted to `" "`). To prevent this from happening, `"+"` symbol is encoded again before being decoded finally. See https://developer.android.com/reference/java/net/URLDecoder for details.


rjuliano (approved):
Ship it 🚢

aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
```diff
@@ -75,7 +75,8 @@ public static URL createURL(
         }
 
         try {
-            return new URL(URLDecoder.decode(builder.build().url().toString(), "UTF-8"));
+            String encodedUrl = builder.build().url().toString();
+            return new URL(URLDecoder.decode(encodePluses(encodedUrl), "UTF-8"));
```
richardmcclellan, line 79: This seems like a hack.   I'd also imagine that "+" is not the only special character that is having this problem.    

How does iOS handle this scenario?  Can we do something similar?  

It sounds like we expect the customer to pass a path which is _not_ encoded (based on https://github.com/aws-amplify/amplify-flutter/issues/446).  If that's the case, then why are we trying to decode it in the first place?

rjuliano: That's a fair point. I looked through the iOS codebase and didn't see anything so I'm guessing it just passes the url as is. Or the HTTP client somehow handles it internally, but I would think that is less likely.

raphkim: This is a special case with Java's `URLDecoder`, which is why iOS is not having an issue. See https://bugs.openjdk.java.net/browse/JDK-8179507.

richardmcclellan: Interesting!  So it sounds like `URLDecoder`/`URLEncoder` is actually meant for HTML forms, not URL strings.  I think we should be using [`Uri.encode`](https://developer.android.com/reference/android/net/Uri#encode(java.lang.String))/[`Uri.decode`](https://developer.android.com/reference/android/net/Uri#decode(java.lang.String)).    

Ref: https://stackoverflow.com/a/16909821/282279

raphkim: `Uri.decode` has the same problem with `android.util.Log`, where it would not be mocked for the unit tests. That was my only reason to not use it and it is minor enough that I don't mind switching back to it if necessary. Is there any good reason to switch the implementation at the cost of requiring `RobolectricRunner` on our unit tests though?

richardmcclellan: > Although, is there any good reason to switch the implementation at the cost of requiring RobolectricRunner on our unit tests though?

I think yes because `URLDecoder` is meant for HTML forms, not URL strings, so we shouldn't be using it here in the first place.   The advantage of switching to `Uri.decode` is that our code will be cleaner, easier to maintain, and more reliable, which I think outweighs the downside of our tests running slightly slower due to `RoboelectricRunner`.

gpanshu (approved):

sktimalsina (approved):

PR #1253, opened 2021-03-25
fix(datastore) Make PendingMutationConverter work for SerializedModel

### LINKED ISSUE #1199: DataStore synch problem during offline item saving
Hi, this issue is related to #823 which has been closed, but actually is still a problem and is easy to recreate by using [the main Android tutorial](https://docs.amplify.aws/start/getting-started/setup/q/integration/android).

The problem occurs while saving an item during an offline condition, stop the program for any reason and start the program again: this way the local and remote items have a mismatch.

I'm using the latest version of Amplify Framework (1.16.13).

My environment:
 - Android Studio 4.1.2
 - Target API 30, minSdkVersion 26

The scenario related to #823 is neither A nor B: if you apply scenario B everything works, BUT, if you stop the program after each save, the issue is there (local data and cloud data mismatch).

So the steps to recreate the issue are:
 1.  Set the device in online mode (wifi or mobile data active), run the program (same one of the tutorial) from Android Studio, save an item (Item 1) inside `onCreate()` method.
 2.  Query the saved item inside the program (`Amplify.DataStore.query()`): check the local Item 1 is there.
 3.  Query the saved item inside DynamoDB or by using AppSync Console command:
       ```graphql
     query GetTodos { listTodos { items { id name priority description } } }
     ```
     Check the remote Item 1 is there also.
      The Item 1 is correctly stored both locally and remotely.
 4.  **Important**: stop the program (Red stop button in Android Studio).
 5. Set the device in offline mode (disable wifi and mobile data), run again the program from Android Studio, save a second item (Item 2) inside OnCreate() method: this item will be saved locally, not in the cloud.
 6. Query the saved item inside the program (Amplify.DataStore.query()): check the local Item 2 is there.
 7.  **Important**: stop the program again (Red stop button in Android Studio).
 8. Set the device in online mode again (wifi or mobile data active), run again the program from Android Studio, save a third item (Item 3) inside OnCreate() method.

Right now we expect that during step 8, the framework will save remotely both Item 2 (which was not saved in step 5 due to the lack of mobile connection) and Item 3, but only Item 3 is saved. If you query locally you'll get all three items. If you query remotely (DynamoDB and AppSync console) you get only Item 1 and 3, Item 2 is lost forever, no matter how many connections or DataStore API calls you make. If you do an `Amplify.DataStore.clear()` after that, Item 2 is gone forever.

This is definitely something the Framework should handle.

Let me know if you need more info to recreate the issue.

richardmcclellan:
There were two bugs introduced by https://github.com/aws-amplify/amplify-android/pull/1110, which changed the update mutation behavior to only send fields that had _changed_.  To do that, we convert the Java models to `SerializedModel` objects, because it allows us to represent the absence of a field, since the field is stored as a `Map`:

1. Items saved while offline (and thus persisted into the `PendingRecord` SQLite table) were not persisted correctly, and the mutations ended up getting sent to AppSync with all of the fields set to null, except for "id". (https://github.com/aws-amplify/amplify-android/issues/1199).  I've fixed this by updating the `PendingMutationConverter` to actually work for `SerializedModel` objects.

2. Mutations for models with children weren't created properly.    `ModelConverter.toMap` (added in the previously mentioned PR) just used Gson to convert a Java model to a Map.  This is not sufficient for serializing models with children, because the association fields just the child's ID needs to be serialized, not the whole child object.  

Fixes #1199

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
@@ -22,6 +22,7 @@ import com.amplifyframework.core.model.ModelField;
 import com.amplifyframework.core.model.ModelSchema;
 
 import java.lang.reflect.Field;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -45,18 +46,25 @@ public final class ModelConverter {
         for (ModelField modelField : schema.getFields().values()) {
             String fieldName = modelField.getName();
             try {
+
                 final ModelAssociation association = schema.getAssociations().get(fieldName);
                 if (association == null) {
-                    if (instance instanceof SerializedModel) {
-                        Map<String, Object> serializedData = ((SerializedModel) instance).getSerializedData();
-                        if (serializedData.containsKey(modelField.getName())) {
-                            result.put(fieldName, serializedData.get(modelField.getName()));
-                        }
-                    } else {
-                        result.put(fieldName, extractFieldValue(modelField, instance));
+                    if (instance instanceof SerializedModel
+                            && !((SerializedModel) instance).getSerializedData().containsKey(modelField.getName())) {
+                        // Skip fields that are not set, so that they are not set to null in the request.
+                        continue;
                     }
+                    result.put(fieldName, extractFieldValue(modelField, instance));
                 } else if (association.isOwner()) {
-                    result.put(association.getTargetName(), extractAssociateId(modelField, instance));
+                    Object associateId = extractAssociateId(modelField, instance);
+                    if (associateId == null) {
+                        // Skip fields that are not set, so that they are not set to null in the request.
+                        continue;
+                    }
+                    result.put(fieldName, SerializedModel.builder()
+                        .serializedData(Collections.singletonMap("id", associateId))
+                        .modelSchema(null)
+                        .build());
                 }
                 // Ignore if field is associated, but is not a "belongsTo" relationship
             } catch (Exception exception) {
--- a/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
@@ -15,13 +15,13 @@
 
 package com.amplifyframework.datastore.appsync;
 
+import com.amplifyframework.AmplifyException;
 import com.amplifyframework.core.model.Model;
-import com.amplifyframework.util.GsonFactory;
-import com.amplifyframework.util.GsonObjectConverter;
+import com.amplifyframework.core.model.ModelAssociation;
+import com.amplifyframework.core.model.ModelField;
+import com.amplifyframework.core.model.ModelSchema;
 
-import com.google.gson.Gson;
-import com.google.gson.JsonElement;
-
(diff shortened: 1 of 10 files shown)


aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
```diff
@@ -34,33 +34,64 @@ private ModelConverter() {}
...
-            return ((SerializedModel) model).getSerializedData();
+        return result;
+    }
+
+    private static Object extractAssociateId(ModelField modelField, Model instance)
+            throws NoSuchFieldException, IllegalAccessException {
+        final Object fieldValue = extractFieldValue(modelField, instance);
+        if (modelField.isModel() && fieldValue instanceof Model) {
+            return ((Model) fieldValue).getId();
+        } else if (modelField.isModel() && fieldValue instanceof Map) {
```
jamesonwilliams, line 77: Is this a valid case to consider?

```java
} else if (modelField.isModel() && fieldValue instanceof SerializedModel) {
...
}
```

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessorTest.java
```diff
@@ -122,7 +124,8 @@ public void rawQueryReturnsResults() throws AmplifyException {
         if (cursor.moveToFirst()) {
             do {
                 Map<String, Object> map = converter.buildMapForModel(cursor);
-                results.add(ModelConverter.fromMap(map, BlogOwner.class));
+                String jsonString = gson.toJson(map);
+                results.add(gson.fromJson(jsonString, BlogOwner.class));
```
jamesonwilliams, line 128: Don't we have a higher-level utility somewhere that does the object->json->model double conversion, without calling `Gson` directly?

jamesonwilliams (approved):

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
richardmcclellan: No, at least not right now.  Right now, we don't ever construct trees with `SerializedModel`s.   A `SerializedModel` only stores the `id` of its children, not any of the child data.   I think we will need to do that in the future, to fully move from Java models to SerializedModel's, internally, but that's a bigger change.

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/SQLCommandProcessorTest.java
richardmcclellan: Previously, I had added a `ModelConverter::fromMap` method, that did just that.   It was used here, and in the SQLiteStorageAdapter.   I actually removed it in this PR because `fromMap`, which uses Gson, isn't congruent with `toMap`, which uses the `ModelSchema`.    Using Gson does work for this particular use case, but I'm not sure it works for all use cases.

rjuliano (approved):
LGTM 🚢

richardmcclellan:
I did some more testing with this PR, and mutations with children are failing, so I'm working on a fix now.

jamesonwilliams (approved):

rjuliano (approved):
Ship it 🚢

PR #1254, opened 2021-03-26
fix(aws-api): use default cognito provider if needed.

rjuliano:
*Issue #, if available:*

*Description of changes:*
When making the change for #1238 a regression was introduced for cases where the API plugin was initialized without passing in an Cognito auth provider. [The previously used interceptor had logic to default](https://github.com/aws-amplify/amplify-android/blob/b0a71c80a2a0c416fa1f28f22e6f43d7114b7de7/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncSigV4SignerInterceptorFactory.java#L106) to using the `DefaultCognitoUserPoolsAuthProvider` if one was not available via `ApiAuthProviders`

This logic is now replicated here. 

Also, modified the logic to throw exceptions when the necessary provider for the requested auth mode is not available.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestDecoratorFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestDecoratorFactory.java
@@ -18,12 +18,13 @@ package com.amplifyframework.api.aws.auth;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
-import com.amplifyframework.AmplifyException;
 import com.amplifyframework.api.ApiException;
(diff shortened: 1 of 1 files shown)


richardmcclellan (approved):
I tested this change with my sample app, and it solves my issue!  Looks like there is still a unit test failure to be looked at though.

rjuliano:
> I tested this change with my sample app, and it solves my issue! Looks like there is still a unit test failure to be looked at though.

Taking a look. Unit tests passed when I ran locally, which is strange

rjuliano:
> I tested this change with my sample app, and it solves my issue! Looks like there is still a unit test failure to be looked at though.

Ok...succeeded on retry. Integration test failures are related to the SDK issue

PR #1260, opened 2021-03-26
chore: fix integration tests involving model comparison

raphkim:
*Issue #, if available:*
extension of #1259 

*Description of changes:*
Created utility class to easily override model properties and fixed another test suite with timestamp problems.

### DIFF
--- a/testutils/src/main/java/com/amplifyframework/testutils/ModelOverride.java
+++ b/testutils/src/main/java/com/amplifyframework/testutils/ModelOverride.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.testutils;
+
+import com.amplifyframework.core.model.Model;
+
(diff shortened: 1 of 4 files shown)


testutils/src/main/java/com/amplifyframework/testutils/ModelOverride.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Sets the field value of a model to new value.
+     * @param instance model instance
+     * @param fieldName name of the field being written to
+     * @param value new value to override field value with
+     * @param <T> type of model
+     * @throws NoSuchFieldException if fieldName is not a field of given instance
+     * @throws IllegalAccessException if object's field is not accessible
+     */
+    public static <T extends Model> void setField(T instance, String fieldName, Object value)
```
jamesonwilliams, line 37: `<T>` is never used. You should probably name this utility class like `FieldSetter`. Although -- there are already some utilities like this under `com.amplfiyframework.util` in (I believe) the `core` package. There is a `FieldFinder`, at least. `FieldFinder.setField(...)` would not at all be an outrageous addition.

jamesonwilliams (changes requested):

testutils/src/main/java/com/amplifyframework/testutils/ModelOverride.java
raphkim: I didn't want to include this in `core` because I didn't want this functionality to be customer facing. Will rename as you suggested though.

raphkim: `<T>` reinforces that the first parameter to this method is an extension of `Model`, although I imagine this rule is _probably_ not very important. Might be nice to restrict usage of methods like these though anyways.

jamesonwilliams: Oops sorry, I misread. `T` clearly is used in the method signature. But yea, any `Object` could theoretically suffice for the first param here.

jamesonwilliams (approved):

richardmcclellan (commented):
Thanks for taking core of this @raphkim!  Looks good 👍

PR #1261, opened 2021-03-26
chore(release): version bump to 1.17.2

raphkim:
*Issue #, if available:*

*Description of changes:*
bump version

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,12 @@ dependencies section:
 ```groovy
 dependencies {
     // Only specify modules that provide functionality your app will use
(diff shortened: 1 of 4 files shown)


changxu0306 (approved):

richardmcclellan (approved):
There is one `feat` commit since the last release (adding authMode as a field on the `AppSyncGraphQlRequest`), which technically would bump us to 1.18.0.  However, I don't think that change is actually customer facing, since we didn't enable customers to add `authMode` via the model factory classes, yet.  So, I think 1.17.2 makes sense here.

PR #1271, opened 2021-04-06
chore(build): add buildspec for publishing to maven

rjuliano:
*Issue #, if available:*

*Description of changes:*
Create a CodeBuild definition (buildspec file) that executes the `uploadArchives` gradle task. This build definition will run inside a CodeBuild project, which will be defined in the `amplify-ci-support` repo in a separate PR.

The buildspec is fairly straightforward. The one callout is that it pulls in any sensitive information from Secrets Manager by using the `secrets-manager` section and makes them available via environment variables. These environment variables are prefixed with `ORG_GRADLE_PROJECT_` which automatically sets gradle project properties. This ensures that at no point, that information is passed in via the command line. 

See [CodeBuild docs](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#build-spec.env.secrets-manager) for more information on secret handling inside the build. 

The trigger for this build will be setup as part of the `amplify-ci-support` PR. The intent is to trigger it every time there is a a merge into `main` from `release` whose commit message starts with `chore(release)`. We'll also be able to trigger it manually via the console or CLI by specifying the commit ID of the merge.

The change in `build.gradle` is due to the fact that the `signing` project settings have a period in them (`signing.keyId` for example) and cannot be represented directly as an environment variable. If a project variable with the name `signingKeyId` exists, it sets the `signing.*` project variables from the `signing*` variables read from secrets manager. For example, `signingKeyId` becomes `signing.keyId`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,8 +132,8 @@ private void configureAndroidLibrary(Project project) {
 
     if (project.hasProperty('signingKeyId')) {
         System.out.println("Getting signing info from protected source.")
-        project.ext.'signing.keyId' = findProperty('signingKeyId').trim()   
-        project.ext.'signing.password' = findProperty('signingPassword').trim()
+        project.ext.'signing.keyId' = findProperty('signingKeyId') 
+        project.ext.'signing.password' = findProperty('signingPassword')
         project.ext.'signing.secretKeyRingFile' = findProperty('signingSecretKeyRingFile')
         project.ext.'signing.inMemoryKey' = findProperty('signingInMemoryKey')
     }
--- a/build.gradle
+++ b/build.gradle
@@ -130,6 +130,14 @@ private void configureAndroidLibrary(Project project) {
         project.findProperty('VERSION_NAME') :
         rootProject.findProperty('VERSION_NAME')
 
+    if (project.hasProperty('signingKeyId')) {
+        System.out.println("Getting signing info from protected source.")
+        project.ext.'signing.keyId' = findProperty('signingKeyId').trim()   
+        project.ext.'signing.password' = findProperty('signingPassword').trim()
+        project.ext.'signing.secretKeyRingFile' = findProperty('signingSecretKeyRingFile')
+        project.ext.'signing.inMemoryKey' = findProperty('signingInMemoryKey')
+    }
+
     project.android {
         buildToolsVersion rootProject.ext.buildToolsVersion
         compileSdkVersion rootProject.ext.compileSdkVersion
--- a/build.gradle
+++ b/build.gradle
@@ -134,7 +134,6 @@ private void configureAndroidLibrary(Project project) {
         System.out.println("Getting signing info from protected source.")
         project.ext.'signing.keyId' = findProperty('signingKeyId') 
         project.ext.'signing.password' = findProperty('signingPassword')
-        project.ext.'signing.secretKeyRingFile' = findProperty('signingSecretKeyRingFile')
         project.ext.'signing.inMemoryKey' = findProperty('signingInMemoryKey')
     }
 
--- a/scripts/maven-release-publisher.yml
+++ b/scripts/maven-release-publisher.yml
@@ -25,9 +25,10 @@ phases:
       - |
         # List all available gradle tasks, grep for the uploadArchive tasks, and then use cut to strip the
         # task description and just return the name of the task, one for each module (e.g. aws-api:uploadArchives)
+        JAVA_HOME=$JDK_8_HOME ./gradlew clean build
         for task_name in $(./gradlew tasks --all | grep uploadArchives | cut -d " " -f 1); do
           echo "Gradle task $task_name"
-          JAVA_HOME=$JDK_8_HOME ./gradlew clean build $task_name;
+          JAVA_HOME=$JDK_8_HOME ./gradlew $task_name;
         done
     finally:
       - echo 'Build phase completed.'
--- a/scripts/maven-release-publisher.yml
+++ b/scripts/maven-release-publisher.yml
@@ -7,7 +7,6 @@ env:
     ORG_GRADLE_PROJECT_signingPassword:  awsmobilesdk/android/signing:password
     ORG_GRADLE_PROJECT_signingKeyId:  awsmobilesdk/android/signing:keyId
     ORG_GRADLE_PROJECT_signingInMemoryKey: awsmobilesdk/android/signing:inMemoryKey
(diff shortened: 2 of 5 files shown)


scripts/maven-release-publisher.yml
```diff
@@ -0,0 +1,37 @@
...
+    commands:
+      - echo 'Pre-build phase starting'
+    finally:
+      - echo 'Pre-build phase completed.'
+  build:
+    commands:
+      - echo 'Build phase starting.'
+      - |
+        for item in aws-* core core-kotlin rxbindings; do 
+          task_name=":$item:uploadArchives"
```
richardmcclellan, line 27:
```suggestion
        # List all available gradle tasks, grep for the uploadArchive tasks, and then use cut to strip the
        # task description and just return the name of the task, one for each module (e.g. aws-api:uploadArchives)
        for task_name in $(./gradlew tasks --all | grep uploadArchives | cut -d " " -f 1); do
```

Two benefits to this:
 - If we add/remove modules in the future, we won't have to update this script.
 - This logic can be used in any of our repos, not just `amplify-android`

rjuliano: I like it. I thought about just doing `./gradlew uploadArchives`, but I didn't test to see if there are any undesirable effects from doing that.

build.gradle
```diff
@@ -130,6 +130,14 @@ private void configureAndroidLibrary(Project project) {
         project.findProperty('VERSION_NAME') :
         rootProject.findProperty('VERSION_NAME')
 
+    if (project.hasProperty('signingKeyId')) {
+        System.out.println("Getting signing info from protected source.")
+        project.ext.'signing.keyId' = findProperty('signingKeyId').trim()   
+        project.ext.'signing.password' = findProperty('signingPassword').trim()
```
richardmcclellan, line 136: Why do you need to `trim()` the value?

rjuliano: Actually, we don't. I was troubleshooting an issue locally and just forgot to remove it

scripts/maven-release-publisher.yml
```diff
@@ -0,0 +1,38 @@
+version: 0.2
+env:
+  shell: /bin/sh
+  secrets-manager:
+    ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: amplify/android/sonatype:username
+    ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD: amplify/android/sonatype:password
+    ORG_GRADLE_PROJECT_signingPassword:  amplify/android/signing:password
+    ORG_GRADLE_PROJECT_signingKeyId:  amplify/android/signing:keyId
+    ORG_GRADLE_PROJECT_signingInMemoryKey: amplify/android/signing:inMemoryKey
+    ORG_GRADLE_PROJECT_signingSecretKeyRingFile:  amplify/android/signing:secretKeyRingFile
```
richardmcclellan, line 10: None of these values are really specific to `amplify` - these credentials are used for all of our repos.    Perhaps make this `awsmobilesdk`?

```suggestion
    ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: awsmobilesdk/android/sonatype:username
    ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD: awsmobilesdk/android/sonatype:password
    ORG_GRADLE_PROJECT_signingPassword:  awsmobilesdk/android/signing:password
    ORG_GRADLE_PROJECT_signingKeyId:  awsmobilesdk/android/signing:keyId
    ORG_GRADLE_PROJECT_signingInMemoryKey: awsmobilesdk/android/signing:inMemoryKey
    ORG_GRADLE_PROJECT_signingSecretKeyRingFile:  awsmobilesdk/android/signing:secretKeyRingFile
```

richardmcclellan, line 10: If the `inMemoryKey` is working, can we get rid of the `secretKeyRingFile`?  Isn't it intended to be used _in place of_ `secretKeyRingFile`?

rjuliano: That's a good point. I'm committing the change.

rjuliano: Yes...I left it in there for now, but we can get rid of it once we get it working with the inMemoryKey

scripts/maven-release-publisher.yml
```diff
@@ -0,0 +1,38 @@
...
+      - echo 'Pre-build phase completed.'
+  build:
+    commands:
+      - echo 'Build phase starting.'
+      - |
+        # List all available gradle tasks, grep for the uploadArchive tasks, and then use cut to strip the
+        # task description and just return the name of the task, one for each module (e.g. aws-api:uploadArchives)
+        for task_name in $(./gradlew tasks --all | grep uploadArchives | cut -d " " -f 1); do
+          echo "Gradle task $task_name"
+          JAVA_HOME=$JDK_8_HOME ./gradlew $task_name;
```
richardmcclellan, line 30: We need to clean and build before we upload, right?

```suggestion
          JAVA_HOME=$JDK_8_HOME ./gradlew clean build $task_name;
```

rjuliano: I thought the uploadArtifacts had a dependency on build, but this makes it more explicit

scripts/maven-release-publisher.yml
```diff
@@ -0,0 +1,38 @@
...
+  build:
+    commands:
+      - echo 'Build phase starting.'
+      - |
+        # List all available gradle tasks, grep for the uploadArchive tasks, and then use cut to strip the
+        # task description and just return the name of the task, one for each module (e.g. aws-api:uploadArchives)
+        for task_name in $(./gradlew tasks --all | grep uploadArchives | cut -d " " -f 1); do
+          echo "Gradle task $task_name"
+          JAVA_HOME=$JDK_8_HOME ./gradlew clean build $task_name;
+        done
```
richardmcclellan, line 31: Actually, `clean build` should be done for all modules, not inside the for loop.  Otherwise, it's going to cause dependency problems.

```suggestion
        JAVA_HOME=$JDK_8_HOME ./gradlew clean build
        for task_name in $(./gradlew tasks --all | grep uploadArchives | cut -d " " -f 1); do
          echo "Gradle task $task_name"
          JAVA_HOME=$JDK_8_HOME ./gradlew $task_name;
        done
```

richardmcclellan (approved):
🚀

PR #1274, opened 2021-04-08
fix: throws AlreadyConfiguredException when configured == true

haverchuck:
*Description of changes:*
fix: throws AlreadyConfiguredException when addPlugin called after configuration.

### DIFF
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -37,7 +37,6 @@ import com.amplifyframework.util.Immutable;
 import com.amplifyframework.util.UserAgent;
 
 import java.util.LinkedHashMap;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
@@ -192,11 +191,7 @@ public final class Amplify {
 
         synchronized (CONFIGURATION_LOCK) {
             if (CONFIGURATION_LOCK.get()) {
-                final String updateString = registryUpdateType.name().toLowerCase(Locale.US);
-                throw new AmplifyException(
-                    "The client tried to " + updateString + " a plugin after calling configure().",
-                        "Plugins may not be added or removed after configure(...) is called."
-                );
+                throw new AlreadyConfiguredException();
             }
 
             if (Empty.check(plugin.getPluginKey())) {
--- a/core/src/main/java/com/amplifyframework/core/Amplify.java
+++ b/core/src/main/java/com/amplifyframework/core/Amplify.java
@@ -136,7 +136,7 @@ public final class Amplify {
 
         synchronized (CONFIGURATION_LOCK) {
             if (CONFIGURATION_LOCK.get()) {
-                throw new AlreadyConfiguredException();
+                throw new AlreadyConfiguredException("Remove the duplicate call to `Amplify.configure()`.");
             }
 
             // Configure User-Agent utility
@@ -191,7 +191,7 @@ public final class Amplify {
 
         synchronized (CONFIGURATION_LOCK) {
             if (CONFIGURATION_LOCK.get()) {
-                throw new AlreadyConfiguredException();
+                throw new AlreadyConfiguredException("Do not add plugins after calling `Amplify.configure()`.");
             }
 
             if (Empty.check(plugin.getPluginKey())) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/amplifyframework/core/Amplify.java
```diff
@@ -192,11 +191,7 @@ private static void beginInitialization(@NonNull Category<? extends Plugin<?>> c
 
         synchronized (CONFIGURATION_LOCK) {
             if (CONFIGURATION_LOCK.get()) {
-                final String updateString = registryUpdateType.name().toLowerCase(Locale.US);
-                throw new AmplifyException(
-                    "The client tried to " + updateString + " a plugin after calling configure().",
-                        "Plugins may not be added or removed after configure(...) is called."
-                );
+                throw new AlreadyConfiguredException();
```
richardmcclellan, line 194: The recovery suggestion ("Remove the duplicate call to `Amplify.configure()`") doesn't make sense for this scenario.  Can you update `AlreadyConfiguredException` to take a recovery suggestion as a constructor parameter?    Then, pass a custom recovery suggestion like this:

```suggestion
                throw new AlreadyConfiguredException("Plugins may not be added or removed after configure(...) is called.");
```

core/src/test/java/com/amplifyframework/core/AmplifyTest.java
```diff
@@ -94,4 +95,31 @@ public void reconfigurationAttemptRaisesException() throws AmplifyException {
...
         );
     }
+
+    /**
+     * It is an error to call {@link Amplify#addPlugin(Plugin)}} after configuration.
+     * @throws Amplify.AlreadyConfiguredException
+     *  NOTE: The name of this method must result in it running last, according to {@link FixMethodOrder}
+     *
+     */
+
```
richardmcclellan, line 105:
```suggestion
```

core/src/main/java/com/amplifyframework/core/Amplify.java
haverchuck: @richardmcclellan Done!

core/src/test/java/com/amplifyframework/core/AmplifyTest.java
```diff
@@ -90,8 +91,32 @@ public void reconfigurationAttemptRaisesException() throws AmplifyException {
...
         );
     }
+
+    /**
+     * It is an error to call {@link Amplify#addPlugin(Plugin)}} after configuration.
+     * @throws Amplify.AlreadyConfiguredException
+     *  NOTE: The name of this method must result in it running last, according to {@link FixMethodOrder}
+     */
+    @Test
+    public void secondaryAddPluginAfterConfigurationRaisesException() throws Amplify.AlreadyConfiguredException {
```
richardmcclellan, line 105: Hmm, I don't see any calls to `Amplify.configure()` in this test.  Don't you need to do that before calling `Amplify.addPlugin(...)` since that is the scenario being tested?

haverchuck: I believe that the configuration status is retained from the previous test, `reconfigurationAttemptRaisesException()`.

richardmcclellan: Ahh, that makes sense.  I think this is an unfortunate side effect of the fact that the `Amplify` facade is just not designed in a very testable manner.  

Is there a way we could implement an `@After` method which resets the configuration?   I think it would need to use reflection to set the [static `CONFIGURATION_LOCK` property](https://github.com/aws-amplify/amplify-android/blob/682c9e778adac74c2261bce4de4866c5e255a5eb/core/src/main/java/com/amplifyframework/core/Amplify.java#L83) of `Amplify` back to false.  

Something like 
```
@After
public void cleanup() {
    Field configurationLockField = Field.class.getField("CONFIGURATION_LOCK");

    // the field is `private`, so make it accessible
    configurationLockField.setAccessible(true);

    // remove the `final` modifier so we can change the value.
    Field modifiersField = Field.class.getDeclaredField("modifiers"); 
    modifiersField.setAccessible(true);
    modifiersField.setInt(configurationLockField, configurationLockField.getModifiers() & ~Modifier.FINAL);

   // Now actually reset the value to false.
   configurationLockField.set(null, new AtomicBoolean(false));
}
```

If all that ends up being too complicated or not working for some reason, I think we could merge this as is, but it would be good to at least attempt a better approach here.

richardmcclellan (approved):

core/src/test/java/com/amplifyframework/core/AmplifyTest.java
richardmcclellan: Chatted offline and determined we weren't able to get this reflection approach to work, so I've approved this PR.

PR #1277, opened 2021-04-08
chore: update release number in the README.md file

rjuliano:
*Issue #, if available:*

*Description of changes:*
Update the version in the README.md file when we generate a release.


richardmcclellan (commented):
This is great.  We also usually update `rxbindings/README.md` - could we add that in?

I wonder if there's a way we could make this logic generic enough to work for all 3 of our repos, though that may be an over optimization.

rjuliano:
> This is great. We also usually update `rxbindings/README.md` - could we add that in?
Yup...it was easy enough to do that.

> I wonder if there's a way we could make this logic generic enough to work for all 3 of our repos, though that may be an over optimization.
I changed the `update_docs` function to take two additional parameters: `file_path` and `maven_group_name`. `file_path` makes it so we can pass any file we may want to update. `maven_group_name` will come in as a command line parameters (for this repo, it's `com.amplifyframework`)

scripts/fastlane/Fastfile
```diff
@@ -26,6 +28,7 @@ platform :android do |options|
   lane :create_next_release_pr do |options|
     # This will be used in the commit and PR title
     product_name = options[:product_name]
+    maven_group_name = options[:maven_group_name]
```
richardmcclellan, line 31: Looks like the package name is defined [in the Appfile, here](https://github.com/aws-amplify/amplify-android/blob/main/scripts/fastlane/Appfile#L2), so another option might be to grab the value from there using: 

```
maven_group_name = CredentialsManager::AppfileConfig.try_fetch_value(:package_name)
```

Ref: https://docs.fastlane.tools/advanced/Appfile/

I'm not sure if that's better or not, but it's another option.

rjuliano: Let me try it. It would be one less thing to pass via the command line.

richardmcclellan (approved):

scripts/fastlane/Fastfile
rjuliano: That worked! Just updated the PR

Issue #838, opened 2020-09-16
Delete mutation fails if schema includes sort keys


KayIlory:
My schema includes an `@key`, and one of the fields is of type `AWSDateTime!`:

```graphql
type TestResult @model
@key(fields:["id", "date", "createDate"])
{
    id: ID!
    userID: ID!
    testResultTypeID: ID!
    testResultType: [TestResultType] @connection(keyName: "byTestResultType", fields: ["testResultTypeID"])
    testResultImage: [TestResultImage] @connection(keyName: "byTestResult", fields: ["id"])
    date: AWSDateTime!
    notes: String
    createDate: AWSDateTime!
}
```
When I do:
```java
Amplify.API.mutate(
    ModelMutation.delete(testResult),
    response -> { /* logging */ },
    error -> { /* logging */ }
);
```

I get an error in the returned GraphQL response:
> GraphQLResponse.Error{message='Variable 'input' has coerced Null value for NonNull type 'AWSDateTime!'', locations='[GraphQLLocation{line='1', column='27'}]', path='null', extensions='null'}

However, if I _comment out the `@key` line_:

```graphql
type TestResult @model
# @key(fields:["id", "date", "createDate"])
{
    id: ID!
    userID: ID!
    testResultTypeID: ID!
    testResultType: [TestResultType] @connection(keyName: "byTestResultType", fields: ["testResultTypeID"])
    testResultImage: [TestResultImage] @connection(keyName: "byTestResult", fields: ["id"])
    date: AWSDateTime!
    notes: String
    createDate: AWSDateTime!
}
```
The same mutation will work.

--------------

The `DeleteTestResultInput` generated for this `@model` is:
```graphql
input DeleteTestResultInput {
  id: ID!
  date: AWSDateTime!
  createDate: AWSDateTime!
}
```
And the problem is the presence of the `AWSDateTime!` field(s).

However, this one (without them) will work:

```graphql
input DeleteTestResultInput {
  id: ID!
}
```

hellosagar:
yeah we need to issue to be fixed ASAP

KayIlory:
![image](https://user-images.githubusercontent.com/51476926/93446580-451f1000-f8c9-11ea-9a58-734a8dadd9e1.png)
![image](https://user-images.githubusercontent.com/51476926/93446617-4ea87800-f8c9-11ea-910f-e3efdb880e40.png)

[...]

PR #1292, opened 2021-04-21
fix(datastore,api): Update and delete mutations now work when custom primary key is defined

richardmcclellan:
Fixes delete mutations (https://github.com/aws-amplify/amplify-android/issues/838) and update mutations when a custom primary key is defined (e.g. `@key(fields: ["hashKey", "sortKey"])`) on a `@model`.  Previously, the library assumes that the primary key is `id`, and there is no sort key defined.  Now, we look at the `@Index` annotations defined on the codegenerated model, and then make sure we include all of the `fields` that make up the primary key in the `input` object for both `delete` and `update` mutations.

Note that while this PR fixes the issues with update/delete mutations, the following issues related to custom primary key are still a problem, that we can address later:
 - The SQLite primary key is still hardcoded to `id`.  I'm not sure if there are actually any use cases affected by this though.  All of the @key defined indexes (primary and secondaries) are added as SQLite indexes.   Changing this has bigger implications as well - we will need to add migration logic to wipe the tables whenever upgrading the library version.
 -  Relationship logic (`@connection`) still assumes that `id` is the primary key.  


aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/ModelConverter.java
```diff
@@ -45,41 +45,33 @@ private ModelConverter() {}
...
             }
+            // Ignore if field is associated, but is not a "belongsTo" relationship
         }
         return result;
     }
 
-    private static Object extractAssociateId(ModelField modelField, Model instance)
-            throws NoSuchFieldException, IllegalAccessException {
-        final Object fieldValue = extractFieldValue(modelField, instance);
+    private static Object extractAssociateId(ModelField modelField, Model instance, ModelSchema schema)
```
richardmcclellan, line 72: No functional changes were made in this file.  I just did a little refactoring to make `extractAssociateId` and `extractFieldValue` throw an `AmplifyException` instead of `NoSuchFieldException, and `IllegalAccessException`.

These changes were made in preparation for a larger simplification that I plan to do in a separate PR, to DRY up some of this logic which is duplicated in `ModelConverter`, `AppSyncRequestFactory`, and `AppSyncGraphQLRequestFactory`

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/SerializedModel.java
```diff
@@ -80,7 +81,8 @@ private SerializedModel(
         Map<String, Object> originalMap = ModelConverter.toMap(original, modelSchema);
         Map<String, Object> patchMap = new HashMap<>();
         for (String key : updatedMap.keySet()) {
-            if ("id".equals(key) || !ObjectsCompat.equals(originalMap.get(key), updatedMap.get(key))) {
+            List<String> primaryIndexFields = modelSchema.getPrimaryIndexFields();
```
richardmcclellan, line 84: This is the "fix" so that update mutations work correctly.

core/src/main/java/com/amplifyframework/core/model/ModelSchema.java
```diff
@@ -119,8 +120,14 @@ public static ModelSchema fromModelClass(@NonNull Class<? extends Model> clazz)
             }
 
             for (Annotation annotation : clazz.getAnnotations()) {
-                ModelIndex modelIndex = createModelIndex(annotation);
-                if (modelIndex != null) {
+                if (annotation.annotationType().isAssignableFrom(Indexes.class)) {
```
richardmcclellan, line 123: Previously, there was a bug here.  The logic was simply to iterate all of the `@Index` annotations of the class.  That works when only one `@Index` is defined.   However, when there is more than one `@Index`, it actually wraps _all_ the `@Index` annotations in a `@Indexes` annotation.  This was not accounted for.  

I've updated the logic to look for handle the single `@Index` scenario as well as the multiple `@Index` wrapped in `@Indeces` scenario.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -187,26 +187,26 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api) {
     @NonNull
     @Override
     public <T extends Model> Cancelable delete(
```
raphkim, line 189: these are public API changes. will this change affect any customers? Should we follow proper deprecation steps?

raphkim: Also, is it ever necessary to pass in both model instance _and_ model schema? I believe the original method signature was set up this way because we were more reserved about allowing model instances in the parameter early on in the development. In my understanding, that is no longer necessary because we now support `SerializedModel`. So we won't need `ModelSchema` anymore because we can directly pass in the model instance right?

richardmcclellan: > these are public API changes. will this change affect any customers? Should we follow proper deprecation steps?

I guess technically they are `public`, but this class is definitely not intended to be used by customers.  This is just the interface that DataStore uses to interact with API, so I don't think we need to be concerned.

> Also, is it ever necessary to pass in both model instance and model schema?

Hmm, there's a lot to that question, but I think we do still need both.  I'm not entirely sure why, because we should be able to derive `ModelSchema` from a generated `Model`, or a `SerializedModel`.  I vaguely remember suggesting this same thing a while back, but there was some reason I'm forgetting.  Ultimately it seems like we should strive for that.   But for now, this pattern of passing both schema and model extends throughout DataStore, much more than the scope of this PR.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.java
```diff
@@ -206,7 +206,7 @@ private AppSyncGraphQLRequestFactory() {}
                     "Input!"; // CreateTodoInput
 
             if (MutationType.DELETE.equals(type)) {
-                builder.variable("input", inputType, Collections.singletonMap("id", model.getId()));
```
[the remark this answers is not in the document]
richardmcclellan: Absolutely!  This PR does not address relationships (`@connection) at all.  For customers using `@connection`, custom primary keys are still not supported, but we plan to add that support soon.  This PR just fixes update and delete mutations, when a custom primary key is defined.

raphkim (approved):

PR #1300, opened 2021-04-27
fix: signed in hub event is now fired after currentUser is set, instead of before

richardmcclellan:
If you call `Amplify.Auth.getCurrentUser()` immediately after receiving the `AuthChannelEventName.SIGNED_IN` event from `HubChannel.AUTH`, it will return null.  This is because the hub event is getting fired before the sign in is fully completed. 

This PR updates logic so that the Hub event doesn't fire until after the current user is set. 


changxu0306 (approved):
🚢

Issue #1279, opened 2021-04-09
Unable to send user end point info from Android AWS SDK in App to Amazon pinpoint


Murali-CS:
We have setup the amplify project with pinpoint and authorization categories. We are not able to send user end point information to pinpoint from mobile App. Only standard attributes and events from the App are getting sent. We are not able to send the complete end point information. Additionally, we are able to send some custom attributes and global event with attributes for those global events. However, we want to have complete end point information sent with user attributes included. We are not able to do it yet.

We are using mobile Android App to authorize the user using Cognito and then sending the pinpoint events from the mobile App screens. 

We also included code in the App to send some **custom attributes**. We are able to see the custom attributes under the custom attributes section under the Endpoint Attributes drop down of pinpoint console.

We also included code in the App to send some **global event with attributes**. We are able to see the same global event with its attributes under the events and events attributes drop down of pinpoint console.

Regarding the user attributes, we are expecting the **user attributes** to be available from Cognito user pool data sharing operation. For this we have the Cognito user pool data sharing option turned ON. We also have a corresponding identity pool for the user pool. However none of the user attributes are shown under the user attributes section under the Endpoint Attributes drop down of pinpoint console.

Please help

poojamat:
@Murali-CS need some more info to figure out your use case, are you trying to send information that you have specified about the user to Amazon Pinpoint. If so do you make themplify.Analytics.identifyUser(userId, profile) method call as per the documentation below:

https://docs.amplify.aws/lib/analytics/identifyuser/q/platform/android

Murali-CS:
yes we are making the amplify.Analytics.identifyUser(userId, profile) method call as in the documentation

PR #1306, opened 2021-05-05
fix(analytics): allow user attributes in identifyUser

rjuliano:
*Issue #, if available: #1279

*Description of changes:*
Currently, the analytics plugin only allows custom attributes to be passed in when calling `identifyUser`. These custom attributes are related to the endpoint. When [creating audiences](https://docs.aws.amazon.com/pinpoint/latest/developerguide/audience-define-user.html) via the AWS console, users build audiences based on endpoint attributes that are divided into two categories: custom and user. As it stands, all the attributes passed to `identifyUser` are considered custom.  However, there are instances where a customer may want certain attributes to be listed under "User attributes" in the Pinpoint console. 

~~To accomplish this, this PR adds a `userAttributes` property to the `UserProfile` instance that is passed in as a parameter to the `identifyUser` API. Inside `identifyUser`, we then add those attributes to the `EndpointProfileUser` instance that is sent to Pinpoint to associate a user with an endpoint.~~

This PR adds a subclass of `UserProfile` with the addition of the `userAttributes` field. The class is called `AWSPinpointUserProfile`. This prevents us from leaking a concept specific to Pinpoint in the category definition. The one drawback of this solution is that we do have to open `UserProfile` for extension. That being said, we do have other classes in core that are open for extension.

For reference [JS](https://docs.amplify.aws/lib/analytics/getting-started/q/platform/js#update-endpoint) separates those into `attributes` for custom attributes, and `userAttributes` for user user attributes.

Please let me know if there are concerns with this approach.


richardmcclellan (approved):
Changes look good overall, could you add a unit test that specifically tests the use case that this customer is looking for?

Also, does this change need any docs updates?

rjuliano:
> Changes look good overall, could you add a unit test that specifically tests the use case that this customer is looking for?
> 
> Also, does this change need any docs updates?

I updated one of the integration tests in the latest commit.

It will definitely need doc updates. I also need to check the iOS codebase to see if they support this. If not, we'll have to coordinate.

richardmcclellan (approved):

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/models/AWSPinpointUserProfile.java
```diff
@@ -0,0 +1,189 @@
...
+        private AnalyticsProperties customProperties;
+        private AnalyticsProperties userAttributes;
+
+        /**
+         * Configures the name to be used in the next-built UserProfile.
+         * @param name User's name
+         * @return Current builder instance, for method chaining
+         */
+        @NonNull
+        public Builder name(@NonNull final String name) {
```
richardmcclellan, line 108: We should be able to avoid having to reimplement all of these methods from `UserProfile.Builder` by making `UserProfile.Builder` generic.  That is, each builder method should return an object of type `<B>`, instead of a concrete `UserProfile.Builder`.  See `StorageOptions.Builder` / `StorageUploadOptions.Builder` for an example of this.

rjuliano: Just updated this. Not sure why, but every time I look at this Builder pattern, it takes me a half an hour to wrap my head around it 🤣

richardmcclellan:
LGTM overall, but let's just make sure we do a quick review with iOS and confirm that they will implement a similar approach.

richardmcclellan (approved):

PR #1307, opened 2021-05-05
chore: fix datastore integration test timeout by using sync expressions to limit the sync

richardmcclellan:

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/BasicCloudSyncInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/BasicCloudSyncInstrumentationTest.java
@@ -28,7 +28,6 @@ import com.amplifyframework.core.AmplifyConfiguration;
 import com.amplifyframework.core.category.CategoryConfiguration;
 import com.amplifyframework.core.category.CategoryType;
 import com.amplifyframework.core.model.ModelSchema;
-import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.core.model.temporal.Temporal;
 import com.amplifyframework.datastore.appsync.AppSyncClient;
 import com.amplifyframework.datastore.appsync.ModelMetadata;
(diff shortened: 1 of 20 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/BasicCloudSyncInstrumentationTest.java
```diff
@@ -60,16 +67,8 @@
...
-    "Over time, this test will create a large DynamoDB table. Even if we delete the content " +
-    "through the AppSyncClient utility, the database will have lots of tombstone'd rows. " +
-    "These entries will be synced, the next time this test runs, and the DataStore initializes. " +
-    "After several runs, that sync will grow large and timeout the test, before the test can " +
-    "run any business logic. A manual workaround exists, by running this cleanup script: " +
-    "https://gist.github.com/jamesonwilliams/c76169676cb99c51d997ef0817eb9278#quikscript-to-clear-appsync-tables"
-)
 public final class BasicCloudSyncInstrumentationTest {
-    private static final int TIMEOUT_SECONDS = 10;
+    private static final int TIMEOUT_SECONDS = 30;
```
richardmcclellan, line 71: Even with the sync expression filtering, it still takes about 9 seconds to complete the sync.  Raising this to 30 seconds just to avoid any transient timeout issues.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/BasicCloudSyncInstrumentationTest.java
```diff
@@ -103,17 +101,31 @@ public void setup() throws AmplifyException {
         api = SynchronousApi.delegatingTo(apiCategory);
         appSync = SynchronousAppSync.using(AppSyncClient.via(apiCategory));
 
+        long tenMinutesAgo = new Date().getTime() - TimeUnit.MINUTES.toMillis(10);
+        Temporal.DateTime tenMinutesAgoDateTime = new Temporal.DateTime(new Date(tenMinutesAgo), 0);
         DataStoreCategory dataStoreCategory = DataStoreCategoryConfigurator.begin()
             .api(apiCategory)
-            .clearDatabase(true)
```
richardmcclellan, line 108: This deletes the database file before each test run.  This was causing the second test to fail each time, and I'm not entirely sure why.  Instead, I am now calling datastore.clear() after each test, in the `teardown()` method.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/BasicCloudSyncInstrumentationTest.java
```diff
@@ -103,17 +100,35 @@ public void setup() throws AmplifyException {
...
             .context(context)
             .modelProvider(AmplifyModelProvider.getInstance())
             .resourceId(configResourceId)
             .timeout(TIMEOUT_SECONDS, TimeUnit.SECONDS)
+            .dataStoreConfiguration(DataStoreConfiguration.builder()
+                .syncExpression(BlogOwner.class, () -> BlogOwner.CREATED_AT.gt(tenMinutesAgoDateTime))
```
rjuliano, line 112: I assume that existing records in DynamoDB that don't have createdAt are excluded from the result right?

richardmcclellan: Great question.  Generally, I think the answer is yes - a null value would probably not match a greater than filter.  However, I don't think it matters because all the records should actually have values for `createdAt`.  This field has been implicitly created by AppSync for a while now.  This PR just makes it an explicitly defined field on the model, so that it can be used in a sync expression.

raphkim (approved):
🚢

rjuliano (approved):
Ship it 🚢

testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/AmplifyModelProvider.java
```diff
@@ -1,19 +1,19 @@
 package com.amplifyframework.testmodels.commentsblog;
 
+import com.amplifyframework.util.Immutable;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelProvider;
-import com.amplifyframework.util.Immutable;
```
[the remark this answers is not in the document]
richardmcclellan: Something must have changed in how codegen generates the imports, I guess?    The updates to this test models in this PR are exactly what the latest codegen spits out.

PR #1308, opened 2021-05-06
chore: add builders for DS and SubscriptionOperation

rjuliano:
*Issue #, if available:*

*Description of changes:*
In preparation for some of the upcoming multiauth work. 

Added a builder for the DataStore plugin since we're going to be adding a new parameter. Rather than keep the constructor growing, I figured I might as well refactor if. No change in functionality. Users will have the option to use the builder or continue to use the constructors, which are marked as deprecated, but are still around. 

Simplified the builder for SubscriptionOperation. Rather than the multi-interface approach, I'm just following the same pattern we do for all the other builders throughout our codebase. No change in functionality

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
@@ -101,6 +101,30 @@ public final class AWSDataStorePlugin extends DataStorePlugin<Void> {
         this.userProvidedConfiguration = userProvidedConfiguration;
     }
 
+    private AWSDataStorePlugin(@NonNull Builder builder) throws DataStoreException {
+        ModelSchemaRegistry modelSchemaRegistry = builder.modelSchemaRegistry == null ?
+            ModelSchemaRegistry.instance() :
+            builder.modelSchemaRegistry;
+        ModelProvider modelProvider = builder.modelProvider == null ?
+            ModelProviderLocator.locate() :
+            builder.modelProvider;
+
+        ApiCategory api = builder.apiCategory == null ? Amplify.API : builder.apiCategory;
+        this.userProvidedConfiguration = builder.dataStoreConfiguration;
+        this.sqliteStorageAdapter = SQLiteStorageAdapter.forModels(modelSchemaRegistry, modelProvider);
+        this.categoryInitializationsPending = new CountDownLatch(1);
+
+        // Used to interrogate plugins, to understand if sync should be automatically turned on
+        this.orchestrator = new Orchestrator(
+            modelProvider,
+            modelSchemaRegistry,
+            sqliteStorageAdapter,
+            AppSyncClient.via(api),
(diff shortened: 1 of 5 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -551,4 +583,71 @@ public void observe(
...
+    /**
+     * Builder object for the DataStore plugin.
+     */
+    public static final class Builder {
+        private DataStoreConfiguration dataStoreConfiguration;
+        private ModelProvider modelProvider;
+        private ModelSchemaRegistry modelSchemaRegistry;
+        private ApiCategory apiCategory;
+
+        private Builder() {}
```
raphkim, line 605:
```suggestion
        private Builder() {
            modelProvider = ModelProviderLocator.locate();
            modelSchemaRegistry = ModelSchemaRegistry.instance();
            apiCategory = Amplify.API;
        }
```

It would be good to assign default values for them so that `AWSDataStorePlugin.builder().build()` returns what we currently do for `new AWSDataStorePlugin()`.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -115,7 +139,9 @@ private AWSDataStorePlugin(
      * If remote synchronization is enabled, it will be performed via {@link Amplify#API}.
      *
      * @throws DataStoreException If it is not possible to access the code-generated model provider
+     * @deprecated Use {@link Builder} instead.
      */
+    @Deprecated
```
raphkim, line 144: i dont think we should deprecate them. Builder methods should provide the extra knobs when in need, but `new AWSDataStorePlugin()` constructor should still be available for basic use-case customers. It should return a default plugin (`AWSDataStorePlugin.builder().build()`) instance.

raphkim (changes requested):

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Moving the ModelProviderLocation in here meant also adding a `throws DataStoreException` to the constructor. That wasn't too bad, but for whatever reason, calling the locate method from here was yielding some inconsistent results. 

With that said, those same defaults are applied in the constructor that takes the builder, so the end result of calling `AWSDataStorePlugin.builder().build()` is equivalent to calling `new AWSDataStorePlugin()`

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
rjuliano: Removed the deprecation decorator.

raphkim (approved):

PR #1309, opened 2021-05-07
chore(api):tweaks to the api init process

rjuliano:
*Issue #, if available:*

*Description of changes:*
These are some minor changes needed for the multi-auth work that can be done in isolation to keep the PRs small.

- `AWSApiPluginConfigurationReader`: read API key from config file if available. Right now it only reads if the default auth type is API_KEY which won't work if API_KEY is configured as an additional auth type.
- `ApiRequestDecoratorFactory`: make the `forAuthType` method public so it can be used by the classes outside of the package.


aws-api/src/main/java/com/amplifyframework/api/aws/ApiAuthProviders.java
```diff
@@ -74,6 +77,28 @@ public CognitoUserPoolsAuthProvider getCognitoUserPoolsAuthProvider() {
         return this.cognitoUserPoolsAuthProvider;
     }
 
+    /**
+     * Returns a list of the auth providers currently configured.
+     * @param apiConfiguration A reference to the API configuration.
+     * @return a list of {@link AuthorizationType}
+     */
+    public List<AuthorizationType> getAvailableAuthorizationTypes(ApiConfiguration apiConfiguration) {
```
raphkim, line 86: If the order doesn't matter, `Set` is probably the way to go here.

aws-api/src/test/java/com/amplifyframework/api/aws/ApiAuthProvidersTest.java
```diff
@@ -0,0 +1,210 @@
...
+        // Create an empty ApiAuthProviders instance.
+        ApiAuthProviders withoutProvider = ApiAuthProviders.noProviderOverrides();
+
+        List<AuthorizationType> expectedAuthTypesApiKeyOnly = Collections.singletonList(AuthorizationType.API_KEY);
+
+        // No API Key provider and no API key in the config should result in 0 auth types being returned.
+        List<AuthorizationType> actualAuthTypes = withoutProvider.getAvailableAuthorizationTypes(BASIC_API_CONFIG);
+        assertEquals(0, actualAuthTypes.size());
+
+        // No API Key provider but there's an API key in the config should result in 1 auth type (API_KEY)
```
raphkim, line 97: Can you explain this behavior? 
Hypothetically, if a user 
1. starts out with API key mode for testing, 
2. manually switches to a more secure auth mode such as user pools, 
3. and then forgets to remove the key from config, 

then are we treating it as if the customer set up a multi-auth? Shouldn't the customer be required to be explicit regarding which modes of auth types to be available for the plugin? In above scenario, wouldn't it be possible for the customer to not realize that API key mode is being used unintentionally by the plugin for certain operations that should not have been authorized?

aws-api/src/main/java/com/amplifyframework/api/aws/ApiAuthProviders.java
rjuliano: You're correct. The order here is not super relevant so I should be able to change to a set here.

aws-api/src/test/java/com/amplifyframework/api/aws/ApiAuthProvidersTest.java
rjuliano: It's possible. However, this covers the use case that's probably more common where the API key is in the config but the customer does not set the apiKeyProvider in the ApiAuthProviders passed in to the constructor (or the builder now). I think that in most of our docs, the examples don't show the API plugin getting initialized with an explicitly provided ApiAuthProviders object.

Also, having multiple providers available doesn't necessarily mean they're using multi-auth. In non-multiauth scenarios, each request will only be attempted with one auth type, but that auth type can be any of the ones configured in ApiAuthProviders. 

For example, if the API is setup with user pools as default and API key as additional, any API requests where no auth type is specified will use the default. If an API call is made where the caller specifies the auth type (API key for instance), that request will use that auth type if the provider is configured.

aws-api/src/main/java/com/amplifyframework/api/aws/ApiAuthProviders.java
rjuliano: Done.

raphkim (approved):

PR #1314, opened 2021-05-11
chore(core): move AuthorizationType to core

rjuliano:
*Issue #, if available:*

*Description of changes:*
Part of the multi-auth project. Moving the AuthorizationType enum to core so we can use it in AuthRule. This change was initially in #1304 but I'm breaking those changes down into separate PRs


raphkim (approved):
Is it possible that this change may break some customers who use `AuthorizationType` in their code?

rjuliano:
> Is it possible that this change may break some customers who use `AuthorizationType` in their code?

It's possible I suppose since it is public. One option would be to preserve the package name and still move it to core, but that feels hacky. 

Let me see if there's a different way to tackle this. The primary reason for wanting to move this was to make it accessible to the AuthRule and ModelSchema classes in core, but maybe I can move that around.

rjuliano:
Closed in favor of #1320

Issue #1286, opened 2021-04-15
Rapid update of a newly created record breaks Sync Process


sbaxter:
I expect to be able to create a record in the DataStore, immediately update the record, and have AppSync publish the changes without fail; but I am presently experiencing an issue where quickly updated records break the AppSync process.

The Exception encountered reads:
```
DataStoreException{message=Wanted 1 metadata for item with id = eec4e87b-fd3c-47e0-9785-d1e38021feba, but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}
```

Once this issue is encountered in the application, all subsequent syncing fails and we lose the ability to maintain current data in the DynamoDB tables. There is no impact to the local Datastore, as records can continually be created and updated. There is no indication that this is happening, other than by viewing the logs.

It is possible that I am implementing the update using a Bad Pattern and am open to critique ;). 

Observed on Amplify Android 1.17.1 and 1.17.3. Pixel 4a ADB connected device. Android Studio 4.1.2. (Update: Android 10, API 29.)

* [Here is some sample code](https://gist.github.com/sbaxter/fa9d9f8c3e46efbca9f59f061561c082#file-bug-java) to reproduce the issue I am experiencing.
* [Here is the sample output](https://gist.github.com/sbaxter/fa9d9f8c3e46efbca9f59f061561c082#file-but-output-log) I gathered from running the above code.
* [Here is the schema.graphql](https://gist.github.com/sbaxter/fa9d9f8c3e46efbca9f59f061561c082#file-schema-json) I used in testing this issue.
* [Here is the relevant documentation](https://docs.amplify.aws/lib/datastore/data-access/q/platform/js#create-and-update) I used in writing the example.

Update: I tested with both Auto merge and optimistic concurrency. I do not believe it is a back end issue, though. Also, it seems instituting a Thread.sleep of ≈50-60ms between create and update is the sweet spot to let the local transactions to play out without issue.

joekiller:
This seems similar to https://github.com/aws-amplify/amplify-android/issues/1055 and https://github.com/aws-amplify/amplify-android/issues/812

joekiller:
The logs indicate that our update to the object is being accepted by the Datastore itself but then rejected by the Syncengine because we have a mutation in the outbox but we expect a mutation in the outbox as we mutated it after saving it and getting the result back. 

Support suggested we query the object and then mutate however the race condition still exists.

This really caught us off guard because it works fine offline.

[...]

PR #1319, opened 2021-05-14
fix(datastore): save metadata when merging even if mutation outbox has pending item

richardmcclellan:
Fixes https://github.com/aws-amplify/amplify-android/issues/1286


aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -190,12 +190,14 @@ public AWSDataStorePlugin(@NonNull ModelProvider modelProvider) {
...
-    AWSDataStorePlugin(@NonNull ModelProvider modelProvider, @NonNull ApiCategory api) {
+    AWSDataStorePlugin(@NonNull ModelProvider modelProvider,
+                       @NonNull ApiCategory api,
+                       @Nullable DataStoreConfiguration dataStoreConfiguration) {
         this(
             Objects.requireNonNull(modelProvider),
             ModelSchemaRegistry.instance(),
             Objects.requireNonNull(api),
-            null
+            dataStoreConfiguration
```
rjuliano, line 200: nit: maybe use the builder here instead.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PendingMutation.java
```diff
@@ -246,7 +246,7 @@ public int hashCode() {
     public String toString() {
         return "PendingMutation{" +
             "mutatedItem=" + mutatedItem +
-            ", modelSchema=" + modelSchema +
+//            ", modelSchema=" + modelSchema +
```
rjuliano, line 249:
```suggestion
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
```diff
@@ -105,10 +99,17 @@
...
-                .flatMapCompletable(shouldMerge ->
-                        (isDelete ? delete(model, changeTypeConsumer) : save(model, changeTypeConsumer))
-                                .andThen(save(metadata, NoOpConsumer.create()))
-                )
+                .flatMapCompletable(shouldMerge -> {
+                    Completable firstStep;
+                    if (mutationOutbox.hasPendingMutation(model.getId())) {
+                        LOG.info("Mutation outbox has pending mutation for " + model.getId()
+                            + ". Saving the metadata, but not model itself.");
+                        firstStep = Completable.complete();
```
rjuliano, line 107: What happens to any changes made to the model?

rjuliano (commented):
Left a couple of minor comments and also a question regarding the merge logic

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.java
richardmcclellan: They get dropped.  The item in the outbox wins over any model coming from the API.   However, we do still need to update the metadata, so that when the item in the outbox _does_ get processed, it doesn't get rejected because the `_version` is too old.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
richardmcclellan: It turns out when I rebased with main, I no longer have changes in this file.

rjuliano (approved):
Looks good. 🚢

PR #1320, opened 2021-05-16
refactor:add enum to represent auth rule provider

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR replaces #1314, which was moving `AuthorizationType` to core. This PR does the following instead:

- Adds an enum (`AuthStrategy.Provider`) to represent the `provider` field of an @auth rule.
- Adds an `authProvider` field to the `AuthRule` model. This field transforms the string value from the `AuthRule` annotation interface into one of enum values. If the `AuthRule` annotation has an empty string for its `provider` field, we use the default provider for the strategy.
- Creates a method on the `AuthorizationType` enum to translate an `AuthStrategy.Provider` into an `AuthorizationType`.

**NOTE** I made the `Provider` enum nested to `AuthStrategy` to make their association clearer. I though creating an `AuthProvider` enum would probably lead to more confusion since there is already an enum called `AuthProvider` in core which is used for social sign-in providers.

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
@@ -89,13 +89,13 @@ public enum AuthorizationType {
         }
         AuthStrategy.Provider authRuleProvider = AuthStrategy.Provider.valueOf(providerName);
         switch (authRuleProvider) {
-            case userPools:
+            case USER_POOLS:
                 return AMAZON_COGNITO_USER_POOLS;
-            case oidc:
+            case OIDC:
                 return OPENID_CONNECT;
-            case iam:
+            case IAM:
                 return AWS_IAM;
-            case apiKey:
+            case API_KEY:
                 return API_KEY;
             default:
                 throw new IllegalArgumentException("No such authorization type: " + providerName);
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
@@ -80,6 +80,7 @@ public enum AuthorizationType {
      * Look up an AuthorizationType by inspecting an AuthRule annotation.
      * @param authRuleAnnotation The annotation obtained from a model.
      * @return The AuthorizationType for the provider
+     * @throws IllegalArgumentException if AuthRule's provider does not match an authorization type
      */
     public static AuthorizationType from(AuthRule authRuleAnnotation) {
         String providerName = authRuleAnnotation.provider();
@@ -101,4 +102,3 @@ public enum AuthorizationType {
         }
     }
 }
-
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
@@ -15,6 +15,10 @@
 
 package com.amplifyframework.api.aws;
 
+import com.amplifyframework.core.model.AuthStrategy;
+import com.amplifyframework.core.model.annotations.AuthRule;
+import com.amplifyframework.util.Empty;
+
 /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
```diff
@@ -25,23 +25,62 @@
...
+    }
+
+    /**
+     * Represents the the value of the provider field of the @auth directive.
+     */
+    public enum Provider {
+        /**
+         * The userPools provider of the @auth rule directive.
+         */
+        userPools,
```
raphkim, line 69: nit: as much as I hate how ugly it becomes, the Java convention is to make them all caps since they are constants

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
```diff
@@ -54,6 +55,9 @@
      */
     public AuthRule(com.amplifyframework.core.model.annotations.AuthRule authRule) {
         this.authStrategy = authRule.allow();
+        this.authProvider = Empty.check(authRule.provider()) ?
+                                        authStrategy.getDefaultAuthProvider() :
+                                        AuthStrategy.Provider.valueOf(authRule.provider());
```
raphkim, line 60: looks like you already do some null checking in the getter later so maybe this should just be `authRule.provider()`?

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthorizationType.java
```diff
@@ -71,5 +75,30 @@ public static AuthorizationType from(String name) {
 
         throw new IllegalArgumentException("No such authorization type: " + name);
     }
+
+    /**
+     * Look up an AuthorizationType by inspecting an AuthRule annotation.
+     * @param authRuleAnnotation The annotation obtained from a model.
+     * @return The AuthorizationType for the provider
```
raphkim, line 82:
```suggestion
     * @return The AuthorizationType for the provider
     * @throws IllegalArgumentException if AuthRule's provider does not match an authorization type
```
We should document this behavior since it's not clear what this method returns in such case.

core/src/main/java/com/amplifyframework/core/model/AuthRule.java
rjuliano: Actually, I have to do something at this level because I need to translate from a String to an enum item. I'm going to tweak this so the null check only happens in one place though.

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
rjuliano: Yeah...I was torn on this one simply because I wanted to keep it similar to the values that they are actually representing, but I agree with you. Probably more idiomatic to use upper case here.

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
```diff
@@ -25,23 +29,94 @@
...
+        /**
+         * Retrieve the enum value for the given name.
+         * @param authDirectiveProviderName String value of the provider attribute of the @auth directive.
+         * @return The associated {@link Provider} item associated with the name.
+         * @throws IllegalArgumentException If the value provided in the parameter does not
+         * have a matching element in the enum.
+         */
+        public static Provider from(@NonNull String authDirectiveProviderName) {
+            Objects.requireNonNull(authDirectiveProviderName);
+            for (Provider provider : Provider.values()) {
```
changxu0306, line 113: Where does `Provider` come from?

rjuliano: `Provider` in this case is the enum that this function belongs to. `Provider.values()` returns a list of the enum items to iterate through. This function is essentially converts a string value in the appropriate enum item to represent the auth rule provider, by iterating through those values.

changxu0306: Got it!

changxu0306 (approved):
LGTM🚢

PR #1330, opened 2021-05-21
fix(datastore): add syncExpression method to configuration builder that takes a ModelSchema

richardmcclellan:
Customers today can configure DataStore syncExpressions like this: 
```
Amplify.addPlugin(new AWSDataStorePlugin(DataStoreConfiguration.builder()
    .syncExpression(Post.class, () -> Post.RATING.gt(5))
    .build()));
```

Support for sync expressions is now being added to Flutter, but in Flutter, there are no generated Java model classes.  We need to provide an alternative way of specifying sync expressions.    This PR adds a new method to the `DataStoreConfiguration.Builder` so that sync expressions can be declared using a `ModelSchema`, instead of a `Class<? extends Model>`.    This will enable Flutter to do something like this:

```
ModelSchema postModelSchema = ...
Amplify.addPlugin(new AWSDataStorePlugin(DataStoreConfiguration.builder()
    .syncExpression(postModelSchema, () -> Post.RATING.gt(5))
    .build()));
```

This is very similar to the API provided today by iOS: ([reference](https://docs.amplify.aws/lib/datastore/sync/q/platform/ios#selectively-syncing-a-subset-of-your-data)) , so this PR closes this parity gap.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
@@ -341,6 +341,23 @@ public final class DataStoreConfiguration {
             return Builder.this;
         }
 
+        /**
+         * Sets a sync expression for a particular model to filter which data is synced locally.  The expression
+         * is evaluated each time DataStore is started.  The QueryPredicate is applied on both sync and subscriptions.
+         * @param modelName the name of the model for which the filter applies
+         * @param syncExpression DataStoreSyncExpression that should be used to filter the data that is synced.
+         * @return Current builder
+         */
+        @NonNull
+        public Builder syncExpression(@NonNull String modelName,
+                                      @NonNull DataStoreSyncExpression syncExpression) {
+            this.syncExpressions.put(
+                    Objects.requireNonNull(modelName),
+                    Objects.requireNonNull(syncExpression)
+            );
+            return Builder.this;
+        }
+
         private void populateSettingsFromJson() throws DataStoreException {
             if (pluginJson == null) {
                 return;
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
@@ -17,11 +17,14 @@ package com.amplifyframework.datastore;
 
 import androidx.annotation.NonNull;
 
+import com.amplifyframework.AmplifyException;
 import com.amplifyframework.core.Consumer;
 import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.datastore.DataStoreConfiguration.ConfigKey;
 import com.amplifyframework.datastore.DataStoreConflictHandler.AlwaysApplyRemoteHandler;
 import com.amplifyframework.testmodels.commentsblog.BlogOwner;
+import com.amplifyframework.testmodels.commentsblog.Post;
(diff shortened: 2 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -341,6 +342,23 @@ public Builder syncExpression(@NonNull Class<? extends Model> modelClass,
...
+         * @return Current builder
+         */
+        @NonNull
+        public Builder syncExpression(@NonNull ModelSchema modelSchema,
+                                      @NonNull DataStoreSyncExpression syncExpression) {
+            this.syncExpressions.put(
+                    Objects.requireNonNull(modelSchema).getName(),
+                    Objects.requireNonNull(syncExpression)
+            );
+            return Builder.this;
```
raphkim, line 359: if we are only storing the name of model anyways, then why not take in the model name as parameter?

raphkim: ^also this way, `Builder#syncExpression(Model, DataStoreSyncExpression)` can be implemented as
```
return syncExpression(Objects.requireNonNull(model).getSimpleName(), syncExpression);
```
which would _kind of_ reduce duplicate-ish code

raphkim (commented):
minor suggestion you can feel free to skip. looks good otherwise!

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
cshfang: Fwiw, the `query` call from Flutter also just takes `modelName` as the param directly. So there may be an argument to be made that it's more consistent to do that here as well?

richardmcclellan: I considered this as well.  The main reason I went with `ModelSchema` was for parity with iOS (see [iOS sync expression docs](https://docs.amplify.aws/lib/datastore/sync/q/platform/ios#selectively-syncing-a-subset-of-your-data)) . 

Thinking about it further though, it's still not actually in parity with iOS, since iOS takes an array of `SyncExpression` objects, which wrap a `ModelSchema` and a `QueryPredicate`.

I'll go ahead and switch to use a `String` instead!

aws-datastore/src/test/java/com/amplifyframework/datastore/DataStoreConfigurationTest.java
```diff
@@ -91,23 +96,27 @@ public void testDefaultOverriddenFromConfiguration() throws JSONException, DataS
...
+        DataStoreSyncExpression ownerSyncExpression = () -> BlogOwner.ID.beginsWith(RandomString.string());
+        DataStoreSyncExpression postSyncExpression = () -> Post.ID.beginsWith(RandomString.string());
         DataStoreConfiguration configObject = DataStoreConfiguration
             .builder()
             .syncMaxRecords(expectedSyncMaxRecords)
             .conflictHandler(dummyConflictHandler)
             .errorHandler(errorHandler)
-            .syncExpression(BlogOwner.class, syncExpression)
+            .syncExpression(BlogOwner.class, ownerSyncExpression)
+            .syncExpression("Post", postSyncExpression)
```
raphkim, line 119: what happens if the entered string is not a model? where does that fail? is it ignored in runtime?

richardmcclellan: There would be no failure - it would just be ignored.

raphkim (approved):

PR #1332, opened 2021-05-21
test: Multiauth instrumentation tests

rjuliano:
*Issue #, if available:*

*Description of changes:*
These are the instrumentation tests for the multiauth implementation from #1331 

A few things to note here:
1 - The models added to the `testmodels` package were generated using a version of `amplify-codegen` that includes [PR 168](https://github.com/aws-amplify/amplify-codegen/pull/168) from the amplify-codegen repo. The new annotation parameter field was added to the `AuthRule` annotation class in PR #1313. That change is merged, but I don't believe it's been included in a release though.

2 - I have configured the instrumentation tests to use [Android Test Orchestrator](https://developer.android.com/training/testing/junit-runner#using-android-test-orchestrator). No modification to existing tests is necessary. This came in handy especially since there are a lot of scenarios to be tested and manually handling cleanup each time was flaky to say the least.

3 - The test itself uses a parameterized approach where each model type is configured with different combinations of auth rules. Each scenario includes the following parameters: 
- model type (class)
- does the scenario require a signed in user (true or false)
- What's the auth type that will succeed for the given scenario.


rjuliano:
Closing in favor of #1349

Issue #1263, opened 2021-03-29
Update User Attributes


vivekathreya1:
Hello,

I am trying to update user attributes with the following code. 
 
```kotlin
val attributes = listOf(
    AuthUserAttribute(AuthUserAttributeKey.familyName(), "lastname"),
    AuthUserAttribute(AuthUserAttributeKey.givenName(), "firstname")
)

try {
    val result = Amplify.Auth.updateUserAttributes(attributes)
    println(result)
} catch (error: Exception) {
    Timber.e(error)
}
```

Though the method returns success, but I get this error

```console
 java.lang.IllegalStateException: Already resumed
        at kotlin.coroutines.SafeContinuation.resumeWith(SafeContinuationJvm.kt:44)
        at com.amplifyframework.kotlin.auth.KotlinAuthFacade$updateUserAttributes$2$2.accept(KotlinAuthFacade.kt:246)
        at com.amplifyframework.kotlin.auth.KotlinAuthFacade$updateUserAttributes$2$2.accept(KotlinAuthFacade.kt:41)
        at com.amplifyframework.auth.cognito.AWSCognitoAuthPlugin$17.onError(AWSCognitoAuthPlugin.java:840)
        at com.amazonaws.mobile.client.internal.InternalCallback.call(InternalCallback.java:77)
        at com.amazonaws.mobile.client.internal.InternalCallback.access$000(InternalCallback.java:34)
        at com.amazonaws.mobile.client.internal.InternalCallback$1.run(InternalCallback.java:103)
        at java.lang.Thread.run(Thread.java:764)
```

I don't get this error when the single-valued `updateUserAttribute(...)` is called.

Thanks,
Vivek

jamesonwilliams:
Looks like the AWSMobileClient is trying to invoke its error callback, but has already previously invoked its success callback.

I'm hesitant that this is a bug with the `core-kotlin` module itself, since that code is fairly cut-and-dry:

https://github.com/aws-amplify/amplify-android/blob/7dc76ee8ee635b217fed015a00fb93bfe6c846ee/core-kotlin/src/main/java/com/amplifyframework/kotlin/auth/KotlinAuthFacade.kt#L240-L249

Not immediately obvious to me from reading the SDK code how that could happen, though.

[...]

PR #1339, opened 2021-05-26
fix(Auth): prevent multiple invocations of success callback for updateUserAttributes

Jordan-Nelson:
*Issue #, if available:* #1263

*Description of changes:*
- Move the onSuccess callback outside of the for loop so that it is only invoked one time


richardmcclellan (approved):
Nice work!  🎉 

I believe this should fix https://github.com/aws-amplify/amplify-android/issues/1263

richardmcclellan (approved):

PR #1346, opened 2021-05-30
feat: add supporting types for multi-auth

rjuliano:
*Issue #, if available:*

*Description of changes:*
I screwed up #1342 when switching between branches. This is that same PR but against `main` instead of `rjuiano/multiauth/main`.

This PR does the following:

- **aws-api-appsync**
  * Adds `AuthModeStrategyType` that represents the auth mode strategies supported by the API plugin. Current values are MULTIAUTH and DEFAULT.
  * Adds an `authModeStrategyType` field to the `AppSyncGraphQLRequest` object. This will effectively serve as the flag that indicates whether an API request should be handled in multi-auth mode or not. Not setting this field when constructing the request means using the DEFAULT strategy which is the current behavior.
- **aws-datastore**
  * Adds a field called `authModeStrategy` (to be consistent with the JS implementation) of type `AuthModeStrategyType`. Customers will be able to set this value via the `AWSDataStore.Builder` when constructing the plugin. Again, not setting this field means using the DEFAULT strategy which is the current behavior. The sync engine will use the value in this field to set the `authModeStrategyType` field of the `AppSyncGraphQLRequest`
  * Modified the `AppSyncRequestFactory` so the sync engine can set the `authModeStrategyType` field when constructing API requests. **NOTE** this factory can only be used by the sync engine due to its visibility (package private). Essentially, we're limiting MULTIAUTH requests to be made only from the sync engine.
  * Added the necessary parameters to `AppSyncClient` so it can be used when constructing API requests.
- **core**
  * Add a method called `hasModelLevelRules` to model schema. This is a convenience method that will be used when determining whether or not the default auth type should be used for models that don't specify any rules.


aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthModeStrategyType.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raphkim, line 2: 2021

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -82,13 +83,16 @@
     // overrides provided via the userProvidedConfiguration
     private DataStoreConfiguration pluginConfiguration;
 
+    private final AuthModeStrategyType authModeStrategy;
```
raphkim, line 86: why do we need this instance variable for the datastore plugin? shouldn't it reside in the API plugin so that datastore uses whichever strategy that the API category uses?

rjuliano: TLDR; From a customer's perspective, multiauth is a DataStore feature. To enable multiauth, the customer has to set that option during creation of the DataStore plugin. That setting controls whether the sync engine submits the API requests with the multiauth flag or not.

A little background...
Initially, the intent was to have the feature be at the API level. However, after some discussion it was decided that multiauth was a DataStore feature mainly due to the nature of the interaction between DataStore and API. When invoking API directly, developers have the option to pass in an authorization type so it didn't make sense to have that option available. 

As far as implementation goes, it made more sense (at least for native) to implement the bulk of the feature on the API side mainly because of the complexity it would have added to the sync engine. We would have had to potentially create a wrapper operation object to keep track of retries and then try to decide whether to bubble up the error or try a different auth type. 

To limit the use of that feature to just the sync engine, only the AppSync request factory under DataStore sets that parameter on the request. The implementation is such that we can make it available for direct API calls as well in the future, but as part of this launch, the feature only applies to DataStore.

raphkim: hmm interesting. i can't say i agree with this, but it seems that we aren't really at a stage to redesign the approach

raphkim (approved):

aws-api-appsync/src/main/java/com/amplifyframework/api/aws/AuthModeStrategyType.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Represents that different auth strategies supported by the client that
+ * interfaces with the AppSync backend.
+ */
+public enum AuthModeStrategyType {
+    /**
+     * Uses the default authorization type from the API configuration
+     * unless the incoming request specifies one.
+     */
+    DEFAULT("default"),
```
raphkim, line 27: would `SINGLEAUTH` make sense here?

I'm imagining a switch on this enum type and it looks a bit weird because `default` is a keyword:

```
switch (authModeStrategyType) {
    case MULTIAUTH:
        break;
    case DEFAULT:
        break;
    default:
        // wait another default?!?
}
```

PR #1349, opened 2021-06-01
feat(datastore): Multiauth integration tests

rjuliano:
*Issue #, if available:*

*Description of changes:* 

Replacement PR for #1332

These are the instrumentation tests for the multiauth implementation from #1347 

A few things to note here:
1 - The models added to the `testmodels` package were generated using a version of `amplify-codegen` that includes [PR 168](https://github.com/aws-amplify/amplify-codegen/pull/168) from the amplify-codegen repo. The new annotation parameter field was added to the `AuthRule` annotation class in PR #1313. That change is merged, but I don't believe it's been included in a release though.

2 - I have configured the instrumentation tests to use [Android Test Orchestrator](https://developer.android.com/training/testing/junit-runner#using-android-test-orchestrator). No modification to existing tests is necessary. This came in handy especially since there are a lot of scenarios to be tested and manually handling cleanup each time was flaky to say the least.

3 - The test itself uses a parameterized approach where each model type is configured with different combinations of auth rules. Each scenario includes the following parameters: 
- model type (class)
- does the scenario require a signed in user (true or false)
- What's the auth type that will succeed for the given scenario.


build.gradle
```diff
@@ -169,10 +173,18 @@ private void configureAndroidLibrary(Project project) {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
+
+        // Needed when running integration tests. The oauth2 library uses relies on two
+        // dependencies (Apache's httpcore and httpclient), both of which include
+        // META-INF/DEPENDENCIES. Tried a couple other options to no avail.
+        packagingOptions {
+            exclude 'META-INF/DEPENDENCIES'
+        }
```
raphkim, line 182: can we not scope it just to the modules that use oauth2?

raphkim (approved):
So we don't yet have any tests for multiauth and this is just one of the prerequisites for testing it right?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
```diff
@@ -0,0 +1,1036 @@
...
+        try {
+            Constructor<?> constructor = modelType.getDeclaredConstructors()[0];
+            constructor.setAccessible(true);
+            String recordDetail = "IntegTest-" + modelType.getSimpleName() + " " + RandomString.string();
+            // Most constructor for the multiauth test models have 2 parameters.
+            if (constructor.getParameterCount() == 2) {
+                return (Model) constructor.newInstance(modelId, recordDetail);
+            }
+            // For the one exception, we just pass a null for the extra parameter.
+            return (Model) constructor.newInstance(modelId, recordDetail, null);
```
raphkim, line 906: This doesn't quite sit right with me.
Assuming that they all have the format of
```
model {
  id: ID!
  name: String!
}
```

then perhaps you can do something along the line of:
```
Map<String, String> model = new HashMap<>();
model.put("id", id);
model.put("name", name);
String serialized = gson.toJson(model);
return gson.fromJson(serialized, modelType);
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
```diff
@@ -0,0 +1,1036 @@
...
+        if (signInToCognito) {
+            auth.signOut(AuthSignOutOptions.builder().build());
+        }
+    }
+
+    /**
+     * Create a instance of the model using the private constructor via reflection.
+     * @return A instance of the model being tested.
+     */
+    private Model createRecord(Class<?> modelType, String modelId) {
```
raphkim, line 896: we can probably make this more strongly typed with something like
```suggestion
    private <T extends Model> T createRecord(Class<T> modelType, String modelId) {
```

raphkim:
I feel like adding some failing scenarios to test error handling would be a good idea too

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/TestStorageAdapter.java
```diff
@@ -32,9 +32,23 @@
...
+     * For testing purposes only.
+     * @param schemaRegistry The schema registry for the adapter.
+     * @param modelProvider The model provider with the desired models.
+     * @param databaseName The name of the database file.
+     * @return An instance of SQLiteStorageAdapter backed by the database file name specified.
+     */
+    public static SQLiteStorageAdapter create(ModelSchemaRegistry schemaRegistry,
+                                              ModelProvider modelProvider,
+                                               String databaseName) {
+        return SQLiteStorageAdapter.forModels(schemaRegistry, modelProvider, databaseName);
```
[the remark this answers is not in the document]
rjuliano: The intent there is to keep `SQLiteStorageAdapter#forModels` package-private since we may not necessarily want to expose that method.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/TestStorageAdapter.java
```diff
@@ -89,6 +103,6 @@ static void cleanup(@NonNull SynchronousStorageAdapter synchronousStorageAdapter
 
     private static void deleteDatabase() {
         ApplicationProvider.getApplicationContext()
-            .deleteDatabase(SQLiteStorageAdapter.DATABASE_NAME);
+            .deleteDatabase(SQLiteStorageAdapter.DEFAULT_DATABASE_NAME);
```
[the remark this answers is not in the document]
rjuliano: Correct. This just deletes the database with the default name.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -657,12 +660,26 @@ public Builder modelSchemaRegistry(ModelSchemaRegistry modelSchemaRegistry) {
             return this;
         }
 
-        @VisibleForTesting
```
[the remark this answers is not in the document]
rjuliano: No particular reason, other than I'm not a huge fan of having those scattered throughout the code, but it probably makes sense to keep it here.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
rjuliano: that's a good idea. That works nicely and I can remove all of the reflection hack.

raphkim (approved):

Issue #1350, opened 2021-05-19
API - IAM Authorization mode - not working 


jodafm:
**Describe the bug**
 I am trying to configure the Credentials Provider in the **amplifyconfiguration** file and use a GraphQL API with IAM authType, but the sdk is returning an error like this:
<img width="980" alt="Screen_Shot_2021-05-17_at_11 12 16_AM" src="https://user-images.githubusercontent.com/8147056/118848873-66873000-b88c-11eb-9ab6-93ee8ddc5a6a.png">

So, in order to fix it, I must add the credentials provider in the APIPlugin configuration.

```kotling
val authProviders = ApiAuthProviders.builder()
      .awsCredentialsProvider(AWSMobileClient.getInstance())
      .build()

  val apiPlugin = AWSApiPlugin.builder()
      .apiAuthProviders(authProviders)
      .build()

  Amplify.addPlugin(apiPlugin)
```

**To Reproduce**
Configure a **awsconfiguration** file with cognito credentials provider and a API using a IAM auth mode.

```json
const amplifyconfig = '''{
   "UserAgent":"aws-amplify-cli/2.0",
   "Version":"1.0",
   "auth":{
      "plugins":{
         "awsCognitoAuthPlugin":{
            "IdentityManager":{
               "Default":{
                  
               }
            },
            "CredentialsProvider":{
               "CognitoIdentity":{
                  "Default":{
                     "PoolId":"us-east-1:xxxxx-831d-473e-8018-529c1c6c2291",
                     "Region":"us-east-1"
                  }
               }
            },
            "CognitoUserPool":{
               "Default":{
                  "PoolId":"us-east-xxxxxx",
                  "AppClientId":"7ofvhkxxxxxxxxxq7fmcsa1ntth",
                  "Region":"us-east-1"
               }
            },
            "Auth":{
               "Default":{
                  "authenticationFlowType":"USER_SRP_AUTH"
               }
            }
         }
      }
   },
   "analytics":{
      "plugins":{
         "awsPinpointAnalyticsPlugin":{
            "pinpointAnalytics":{
               "appId":"c6ba148f8c9a41178e04a72dxxxxxxx",
               "region":"us-east-1"
            },
            "pinpointTargeting":{
               "region":"us-east-1"
            }
         }
      }
   },
    "api": {
        "plugins": {
            "awsAPIPlugin": {
                "publicApi": {
                    "endpointType": "GraphQL",
[...]

PR #1351, opened 2021-06-01
fix(api): default to mobile client for iam auth mode

raphkim:
*Issue #, if available:*
#1350 

*Description of changes:*
`ApiRequestDecoratorFactory` would throw if no `AWSCredentialsProvider` was explicitly provided when configuring the API plugin, but the intended behavior was to use `AWSMobileClient` by default. This additional logic is added in this PR.

(Also changed to switch statements for slight performance boost :) )


rjuliano (approved):
LGTM 🚢

sdhuka (approved):

Issue #1102, opened 2021-01-12
errorMessage field availability on AuthException objects via Amplify calls/AuthException child types as error cause


vanceanderson41:
1. **Description**:  I use some APIs associated with the `AWSCognitoAuthPlugin` and am unable to easily get useful error messages from the root error objects returned by `Amplify` with which to to surface to my users.
2. The version of Amplify Android you're using: `1.6.8`
3. The **code** you're using to call Amplify Android:

```java
// Dependencies: Amplify/Cognito
def amplifyVersion = '1.6.8'
implementation "com.amplifyframework:core:$amplifyVersion"
implementation "com.amplifyframework:aws-auth-cognito:$amplifyVersion"
implementation "com.amplifyframework:rxbindings:$amplifyVersion"

// Configuration in Application class:
Amplify.addPlugin(AWSCognitoAuthPlugin())
Amplify.configure(this

// For resetting a user pool user's password:
RxAmplify.Auth.resetPassword(username)
    .doOnSubscribe { _resetPassword.postValue(Resource.Loading()) }
    .subscribeBy(
        onSuccess = {
            userRepository.saveUsername(username)
            _resetPassword.postValue(Resource.Success(username))
        },
        onError = {
            (it as AmplifyException).let {
                _resetPassword.postValue(Resource.Error(it.message))
            }
        }
    )

// For confirming a user pool user's password reset
RxAmplify.Auth.confirmResetPassword(newPassword, code)
    .doOnSubscribe { _confirmResetPassword.postValue(Resource.Loading()) }
    .subscribeBy(
        onComplete = {
            _confirmResetPassword.postValue(Resource.Success(newPassword))
        },
        onError = {
            (it as? AmplifyException)?.let {
                Timber.e(it)
                _confirmResetPassword.postValue(Resource.Error(it.message))
            }
        }
    )

// For updating a user pool user's password who already has an active session
RxAmplify.Auth.updatePassword(oldPassword, newPassword)
    .doOnSubscribe { _updatePassword.postValue(Resource.Loading()) }
    .subscribeBy(
        onComplete = {
            _updatePassword.postValue(Resource.Success(Any()))
        },
        onError = {
            (it as AmplifyException).let {
                _updatePassword.postValue(Resource.Error(it.message))
            }
        }
    )
```
4. Any relevant [logs](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#getting-more-output):
```
[...]

PR #1370, opened 2021-06-10
fix(auth): throw correct auth exception for code mismatch

div5yesh:
*Issue #, if available:*  #1102 

*Description of changes:*
Throw AuthException with correct message and recovery suggestion mapped to Cognito exception instead of generic exception.

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
@@ -33,9 +33,9 @@ import com.amazonaws.services.cognitoidentityprovider.model.ResourceNotFoundExce
 import com.amazonaws.services.cognitoidentityprovider.model.SoftwareTokenMFANotFoundException;
 import com.amazonaws.services.cognitoidentityprovider.model.TooManyFailedAttemptsException;
 import com.amazonaws.services.cognitoidentityprovider.model.TooManyRequestsException;
-import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
+import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
 
 /**
  * Convert AWS Cognito Exceptions to AuthExceptions.
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
@@ -32,6 +32,7 @@ import com.amazonaws.services.cognitoidentityprovider.model.TooManyFailedAttempt
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
 import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
+import com.amazonaws.services.cognitoidentityprovider.model.MFAMethodNotFoundException;
 
 /**
  * Convert AWS Cognito Exceptions to AuthExceptions.
@@ -89,6 +90,10 @@ public final class CognitoAuthExceptionConverter {
 
         if (error instanceof LimitExceededException) {
             return new AuthException.LimitExceededException(error);
+        }
+
+        if (error instanceof MFAMethodNotFoundException) {
+            return new AuthException.MFAMethodNotFoundException(error);
         }
 
         if (error instanceof ResourceNotFoundException) {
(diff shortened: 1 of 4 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -32,6 +32,7 @@
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
 import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
+import com.amazonaws.services.cognitoidentityprovider.model.MFAMethodNotFoundException;
```
changxu0306, line 35: This line should be moved to line 29 to fit the import rule.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
```diff
@@ -32,6 +32,7 @@
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
 import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
 import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
```
changxu0306, line 34:
```suggestion
import com.amazonaws.services.cognitoidentityprovider.model.UsernameExistsException;
import com.amazonaws.services.cognitoidentityprovider.model.UserNotConfirmedException;
import com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException;
```

changxu0306 (changes requested):
Overall it looks good. But I think similar changes should also be made in the `onError()` callback for `fetchAuthSession()`, `resetPassword()`, `confirmResetPassword()`, `updatePassword()`, `fetchUserAttributes()`, etc. in `AWSCognitoAuthPlugin`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
changxu0306: Move line 34 to line before line 32 to fit the import rule.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/util/CognitoAuthExceptionConverter.java
changxu0306: Since you have added `MFAMethodNotFoundException`, I found that there are more Exception types under `com.amazonaws.services.cognitoidentityprovider.model`, such as `EnableSoftwareTokenMFAException`, etc. I'm not sure whether those should be added as well. Could you please reach out to iOS to see whether they have them or not?

div5yesh: Yes that's a good idea. These are the exceptions iOS has: https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Constants/AuthPluginErrorConstants.swift

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -449,6 +502,24 @@ public FailedAttemptsLimitExceededException(Throwable cause) {
         }
     }
 
+    /**
+     *
+     */
+    public static class TooManyRequestsException extends AuthException {
+        private static final long serialVersionUID = 1L;
+        private static final String MESSAGE = "";
```
changxu0306, line 510: The message should not be empty. 
```suggestion
        private static final String MESSAGE = "Failed since the user made too many requests.";
```

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -396,6 +396,41 @@ public LimitExceededException(Throwable cause) {
...
+         * Default message/recovery suggestion with a cause.
+         * @param cause The original error.
+         */
+        public MFAMethodNotFoundException(Throwable cause) {
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     *
```
changxu0306, line 417: This shouldn't be empty.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -449,6 +502,24 @@ public FailedAttemptsLimitExceededException(Throwable cause) {
         }
     }
 
+    /**
+     *
```
changxu0306, line 506: This shouldn't be empty.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -396,6 +396,41 @@ public LimitExceededException(Throwable cause) {
...
+            super(MESSAGE, cause, RECOVERY_SUGGESTION);
+        }
+    }
+
+    /**
+     *
+     */
+    public static class NotAuthorizedException extends AuthException {
+        private static final long serialVersionUID = 1L;
+        private static final String MESSAGE = "";
```
changxu0306, line 421: The message should not be empty.
```suggestion
        private static final String MESSAGE = "Failed since user is not authorized.";
```

changxu0306 (changes requested):

changxu0306 (approved):
Ship it 🚢

Issue #1324, opened 2021-05-18
SyncProcessor Retry Logic


richardmcclellan:
In the DataStore sync processor, if a transient error occurs with any sync request, the entire sync process is stopped.  To increase reliability, every page request should be retried with a jittered exponential backoff strategy. 

Ref: `amplify-js` uses [this `jitteredExponentialRetry` method](https://github.com/aws-amplify/amplify-js/blob/df95ea3724eb6406f64b03f25086cd3e8644cb5f/packages/core/src/Util/Retry.ts#L74-L78) in the [SyncProcessor](https://github.com/aws-amplify/amplify-js/blob/dedd5641dfcfce209433088fe9570874cd810997/packages/datastore/src/sync/processors/sync.ts#L201-L278).

PR #1374, opened 2021-06-14
fix(datastore): added retry logic for SyncProcessor

poojamat:
https://github.com/aws-amplify/amplify-android/issues/1324

### DIFF
--- a/aws-analytics-pinpoint/build.gradle
+++ b/aws-analytics-pinpoint/build.gradle
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
 
 dependencies {
     implementation project(path: ':core')
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
@@ -142,6 +142,9 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
                     return;
                 }
             }
+            if(response.code() >= 500) {
+                onFailure.accept(new ApiException.ServerErrorException("OkHttp client request failed.", "Server Error"));
+            }
 
             try {
                 onResponse.accept(wrapResponse(jsonResponse));
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
@@ -142,8 +142,8 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
                     return;
                 }
             }
-            if(response.code() >= 500) {
-                onFailure.accept(new ApiException.ServerErrorException("OkHttp client request failed.", "Server Error"));
+            if(response.code() >= 400) {
+                onFailure.accept(new ApiException.ApiIrrecoverableException("OkHttp client request failed.", "Irrecoverable error"));
             }
 
             try {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnSuccessConsumer.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnSuccessConsumer.kt
@@ -1,0 +1,28 @@
+package com.amplifyframework.datastore.syncengine
+
+import com.amplifyframework.api.graphql.GraphQLResponse
+import com.amplifyframework.api.graphql.PaginatedResult
+import com.amplifyframework.core.Consumer
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.datastore.DataStoreException
+import com.amplifyframework.datastore.ErrorType
+import com.amplifyframework.datastore.appsync.ModelWithMetadata
+import io.reactivex.rxjava3.core.SingleEmitter
+
+
+class OnSuccessConsumer<T : Model>(private val emitter: SingleEmitter<PaginatedResult<ModelWithMetadata<T>>>) :
+    Consumer<GraphQLResponse<PaginatedResult<ModelWithMetadata<T>>>> {
+
+    override fun accept(result: GraphQLResponse<PaginatedResult<ModelWithMetadata<T>>>) {
+        if (result.hasErrors()) {
+            emitter.onError(DataStoreException(
+                    "A model sync failed: $result.errors",
+                    "Check your schema."))
+        } else if (!result.hasData()) {
+            emitter.onError(DataStoreException(
+                    "Empty response from AppSync.", "Report to AWS team.", ErrorType.IRRECOVERABLE_ERROR))
+        } else {
+            emitter.onSuccess(result.data)
+        }
+    }
+}
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
@@ -1,0 +1,52 @@
+package com.amplifyframework.datastore.syncengine
+
+import com.amplifyframework.api.graphql.GraphQLRequest
+import com.amplifyframework.api.graphql.GraphQLResponse
+import com.amplifyframework.api.graphql.PaginatedResult
+import com.amplifyframework.core.Amplify
+import com.amplifyframework.core.Consumer
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.datastore.AmplifyDisposables
+import com.amplifyframework.datastore.DataStoreException
+import com.amplifyframework.datastore.appsync.AppSync
+import com.amplifyframework.datastore.appsync.ModelWithMetadata
+import io.reactivex.rxjava3.core.SingleEmitter
+import java.util.*
+import java.util.concurrent.TimeUnit
+
+class RetryHandler() {
+
+    private var numberOfAttempts = 0
+    private val maxExponent = 8
+    private val jitterFactor = 100
+    private val LOG = Amplify.Logging.forNamespace("amplify:aws-datastore")
+
+    fun <T : Model> retry(
+        emitter: SingleEmitter<PaginatedResult<ModelWithMetadata<T>>>,
+        appSync: AppSync,
+        request: GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>>,
+        onResponse: Consumer<GraphQLResponse<PaginatedResult<ModelWithMetadata<T>>>>,
+        onFailure: Consumer<DataStoreException>
+    ) {
+        numberOfAttempts++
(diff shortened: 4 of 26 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -229,6 +231,7 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
                         processor.onComplete();
                     }
                 })
+                .retry(retryStrategy::retryHandler)
```
richardmcclellan, line 234: If an error occurs syncing page 3 for a given model, will this just retry page 3 with exponential backoff, or is it going to start over with page 1?  Ideally, we want the former.    I think you may need to call `.retry` on the `Single` returned by the `syncPage` method instead, to accomplish this behavior.

Also, we should make sure that we are only retrying retryable errors (non 4xx errors).

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -142,6 +142,9 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
                     return;
                 }
             }
+            if(response.code() >= 400) {
```
rjuliano, line 145: When you merge the latest commit from main, make sure to add this block in the `MultiAuthAppSyncOperation` class as well.

richardmcclellan: Would it be possible to centralize this logic into one place that can be referenced from `AppSyncGraphQLOperation` and `MultiAuthAppSyncOperation`?

aws-analytics-pinpoint/build.gradle
```diff
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
```
richardmcclellan, line 19: Ooh....Kotlin!  Before we commit to adding Kotlin to the library (no pun intended 😆 ) - 
 - Does adding this plugin have any impact on library build/compile time?
 - Does adding this plugin have any impact on library aar size?
 - Will this have any impact on customers, who are only using Java in their application?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -83,8 +84,9 @@ private SyncProcessor(Builder builder) {
         this.dataStoreConfigurationProvider = builder.dataStoreConfigurationProvider;
         this.queryPredicateProvider = builder.queryPredicateProvider;
         this.modelNames =
-            ForEach.inCollection(modelProvider.modelSchemas().values(), ModelSchema::getName)
-                .toArray(new String[0]);
+                ForEach.inCollection(modelProvider.modelSchemas().values(), ModelSchema::getName)
+                        .toArray(new String[0]);
```
richardmcclellan, line 88:
```suggestion
            ForEach.inCollection(modelProvider.modelSchemas().values(), ModelSchema::getName)
                .toArray(new String[0]);
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -257,42 +260,37 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
      * Fetches one page for a sync.
      * @param request GraphQLRequest object for the sync, obtained from {@link AppSync#buildSyncRequest}, or from
      *                response.getData().getRequestForNextResult() for subsequent requests.
-     * @param <T> The type of model to sync.
+     * @param <T>     The type of model to sync.
```
richardmcclellan, line 263:
```suggestion
     * @param <T> The type of model to sync.
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -332,14 +330,14 @@ public DataStoreConfigurationProviderStep merger(@NonNull Merger merger) {
         @NonNull
         @Override
         public QueryPredicateProviderStep dataStoreConfigurationProvider(
-            DataStoreConfigurationProvider dataStoreConfigurationProvider) {
+                DataStoreConfigurationProvider dataStoreConfigurationProvider) {
```
richardmcclellan, line 333:
```suggestion
            DataStoreConfigurationProvider dataStoreConfigurationProvider) {
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -91,7 +92,7 @@
     private static final long OP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(2);
     private static final long BASE_SYNC_INTERVAL_MINUTES = TimeUnit.DAYS.toMinutes(1);
     private static final List<String> SYSTEM_MODEL_NAMES =
-        ForEach.inCollection(SystemModelsProviderFactory.create().models(), Class::getSimpleName);
+            ForEach.inCollection(SystemModelsProviderFactory.create().models(), Class::getSimpleName);
```
richardmcclellan, line 95:
```suggestion
        ForEach.inCollection(SystemModelsProviderFactory.create().models(), Class::getSimpleName);
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -100,6 +101,8 @@
     private SyncProcessor syncProcessor;
     private int errorHandlerCallCount;
     private int modelCount;
+    RetryHandler retryHandler;
```
richardmcclellan, line 104: Should this be private?

```suggestion
    private RetryHandler retryHandler;
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -133,26 +137,27 @@ private void initSyncProcessor(int syncMaxRecords) throws AmplifyException {
...
-            .errorHandler(dataStoreException -> errorHandlerCallCount++)
-            .syncExpression(BlogOwner.class, () -> BlogOwner.NAME.beginsWith("J"))
-            .build();
+                .builder()
+                .syncInterval(BASE_SYNC_INTERVAL_MINUTES, TimeUnit.MINUTES)
+                .syncMaxRecords(syncMaxRecords)
+                .syncPageSize(1_000)
+                .errorHandler(dataStoreException -> errorHandlerCallCount++)
+                .syncExpression(BlogOwner.class, () -> BlogOwner.NAME.beginsWith("J"))
+                .build();
```
richardmcclellan, line 146:
```suggestion
            .builder()
            .syncInterval(BASE_SYNC_INTERVAL_MINUTES, TimeUnit.MINUTES)
            .syncMaxRecords(syncMaxRecords)
            .syncPageSize(1_000)
            .errorHandler(dataStoreException -> errorHandlerCallCount++)
            .syncExpression(BlogOwner.class, () -> BlogOwner.NAME.beginsWith("J"))
            .build();
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -133,26 +137,27 @@ private void initSyncProcessor(int syncMaxRecords) throws AmplifyException {
...
-            .build();
+                .modelProvider(modelProvider)
+                .modelSchemaRegistry(modelSchemaRegistry)
+                .syncTimeRegistry(syncTimeRegistry)
+                .appSync(appSync)
+                .merger(merger)
+                .dataStoreConfigurationProvider(dataStoreConfigurationProvider)
+                .queryPredicateProvider(queryPredicateProvider)
+                .retryHandler(retryHandler)
+                .build();
```
richardmcclellan, line 160:
```suggestion
            .modelProvider(modelProvider)
            .modelSchemaRegistry(modelSchemaRegistry)
            .syncTimeRegistry(syncTimeRegistry)
            .appSync(appSync)
            .merger(merger)
            .dataStoreConfigurationProvider(dataStoreConfigurationProvider)
            .queryPredicateProvider(queryPredicateProvider)
            .retryHandler(retryHandler)
            .build();
```

gradle.properties
```diff
@@ -12,7 +12,7 @@ org.gradle.jvmargs=-Xmx4g
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 org.gradle.parallel=true
 
-VERSION_NAME=1.17.8
+VERSION_NAME=1.21.0
```
richardmcclellan, line 15: Hmm, why is this change in this PR?

poojamat: This is just for local testing, accidentally checked it in, will revert it.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnSuccessConsumer.kt
```diff
@@ -0,0 +1,28 @@
+package com.amplifyframework.datastore.syncengine
```
richardmcclellan, line 1: On the subject of Kotlin vs Java, we need to be thoughtful in which one we choose.  For this `OnSuccessConsumer` (and `OnErrorConsumer`), since it is so closely tied to the `SyncProcessor`, I think we should be using the same language (Java).  One reason for this is that it makes refactoring difficult.  For example, if we wanted to move a chunk of code from `SyncProcessor` to `OnSuccessConsumer`, or vice versa, we can no longer just copy and paste, like we could if they were written in the same language, so it will likely slow us down a bit.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
```diff
@@ -0,0 +1,52 @@
+package com.amplifyframework.datastore.syncengine
```
richardmcclellan, line 1: Currently it looks like this is a pretty specific retry implementation - it only handles retrying an AppSync sync request, rather than just any general request or operation.   Similar to the `OnSuccessConsumer`, since it is so closely related to the `SyncProcessor`, I think it should probably be the same language as the `SyncProcessor` (Java).

That said, it would be awesome to see this be a little more generic. Is there a way we could adapt it to take any generic RxJava `Single`?  Then, we could potentially use it in other places in the app.  In that case, it would be a standalone utility that I think we could justify using Kotlin for.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
poojamat: Created a task for this one.

PR #1376, opened 2021-06-15
fix(datastore): fix subscription timeout period not increasing

HuiSF:
**Issue #, if available:**

Flutter open issues: [#615](https://github.com/aws-amplify/amplify-flutter/issues/615)

**Description of changes:**

`amplify-flutter` relies on `ModelProvider.modelSchemas` to leverage `amplify-android` functionality. The original subscription timeout period incrementing mechanism doesn't cover `modelSchemas`, thus, we saw subscription timeout issues when attempting subscribe more than 5 models in Flutter.

Changed to use  `modelSchemas` method to get the number of models. This change should be safe, as `ModelProvider` has the [default method](https://github.com/aws-amplify/amplify-android/blob/891ae1c5fdd5bf008315f82c343ed1868739d635/core/src/main/java/com/amplifyframework/core/model/ModelProvider.java#L49) that ensures the coherence of `models` and `modelSchemas`.

----


richardmcclellan (approved):
LGTM! 🚀

PR #1377, opened 2021-06-16
chore: remove deprecated WRITE_EXTERNAL_STORAGE permission

richardmcclellan:
Resolves https://github.com/aws-amplify/amplify-android/issues/896

`WRITE_EXTERNAL_STORAGE` was deprecated in Android 10.  It was used for accessing files on an external disc, or other locations outside of the app sandbox.  As far as I know, we don't need it, and I'm not sure why it was added in the first place.  If we did need it, any customer building an app that targets Android 10 (released almost 2 years ago) or above would have reported an issue to us about it, and that hasn't happened.
 
### DIFF
--- a/aws-datastore/src/androidTest/AndroidManifest.xml
+++ b/aws-datastore/src/androidTest/AndroidManifest.xml
@@ -17,5 +17,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.amplifyframework.datastore">
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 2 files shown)


raphkim (approved):
Do we use the correct storage scope in the actual code to reflect the new Android 10 requirements?

richardmcclellan:
> Do we use the correct storage scope in the actual code to reflect the new Android 10 requirements?

I'm not completely sure.  However, having the permission in the manifest file is effectively a no-op if your target SDK is Android 10 or above since it is deprecated, so removing it won't break anything.

Issue #1288, opened 2021-04-19
Multiple offline update on same item not all synced


mathieuedet:
I'm developing Android app with offline support. In online use-case, no problem.
Unfortunately, when I test an offline scenario, I encountering a bug during multiple updates on same item.

After each updates, I log local object and all fields values are ok (last edition values with new editions)

Example with fake data : 

Original : 

```
{
"name": "value1",
"nickname": "mat",
"age": 27,
"city": "Paris"
}
```

Update name and city offline :
```
{
"name": "value2",
"nickname": "mat",
"age": 27,
"city": "London"
}
```

Update age offline too :
```
{
"name": "value2",
"nickname": "mat",
"age": 28,
"city": "London"
}
```

When I request my object from Datastore, I get updated object from my local version

When connection is established, I use `Datastore.start` to sync data.

However, pushed object contains only last update :

```
{
"name": "value1",
"nickname": "mat"
"age": 28,
"city": "London"
}
```

Problem occurs on 1.17.2 and 1.17.3 versions but **not on 1.6.9**. I cannot use 1.6.9 because of some bugs on this version.

Thanks

shantanu-csedu:
For my case, every column reset to null except the ID. This bug was introduced in version 1.17.2 and still exists in 1.17.5.

whisller:
@richardmcclellan any update on this issue? We're experiencing same problem.

whisller:
It seems that fix merged by @sdhuka is working on our side. Thanks everyone!

richardmcclellan:
This issue has been resolved with the [Amplify Android 1.19.0](https://github.com/aws-amplify/amplify-android/releases/tag/release_v1.19.0) release!

shantanu-csedu:
Issue still exists in version 1.19.0. It's less frequent but still happening

richardmcclellan:
@shantanu-csedu can you please provide specific steps to reproduce?

shantanu-csedu:
Steps to reproduce the bugs:
1. Force stop app and open (so that amplify sync trigger by query or manually)
2. Create a row and update at least once before sync started (while amplify sync is checking network states and it takes around 20 seconds)
3. After sync ends, newly created row data will be empty except the primary key

Note: it will happen after the sync started but it's hard to reproduce.

richardmcclellan:
[...]

PR #1379, opened 2021-06-16
fix(datastore): merge incoming mutation with existing update mutation

sdhuka:
*Issue #1288 *

*Description of changes:*
Issue: Consecutive offline update overwrites existing update in the queue, resulting in losing the changes in the previous update.
Fix: For consecutive updates, merge the fields unchanged from existing update into incoming update.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
@@ -23,6 +23,7 @@ import com.amplifyframework.AmplifyException;
 import com.amplifyframework.core.Amplify;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelSchema;
+import com.amplifyframework.core.model.SerializedModel;
 import com.amplifyframework.core.model.query.Where;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.core.model.query.predicate.QueryPredicates;
@@ -33,7 +34,9 @@ import com.amplifyframework.datastore.storage.StorageItemChange;
 import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.logging.Logger;
 
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Semaphore;
@@ -332,10 +335,18 @@ final class PersistentMutationOutbox implements MutationOutbox {
                     if (QueryPredicates.all().equals(incoming.getPredicate())) {
                         // If the incoming update does not have a condition, we want to delete any
                         // existing mutations for the modelId before saving the incoming one.
-                        return removeNotLocking(existing.getMutationId()).andThen(saveIncomingAndNotify());
+                        try {
+                            @SuppressWarnings("unchecked")
+                            PendingMutation<T> mergedPendingMutation = (PendingMutation<T>) merge(incoming, existing);
+                            return removeNotLocking(existing.getMutationId())
+                                    .andThen(saveIncomingAndNotify(mergedPendingMutation));
+                        } catch (Exception exception) {
+                            LOG.error("Failed to merge, skip merging. exceptions: " + exception);
+                            return removeNotLocking(existing.getMutationId()).andThen(saveIncomingAndNotify(incoming));
+                        }
                     } else {
                         // If it has a condition, we want to just add it to the queue
-                        return saveIncomingAndNotify();
+                        return saveIncomingAndNotify(incoming);
                     }
                 case DELETE:
                     // Incoming update after a delete -> throw exception
@@ -381,7 +392,7 @@ final class PersistentMutationOutbox implements MutationOutbox {
                 .andThen(notifyContentAvailable());
         }
 
-        private Completable saveIncomingAndNotify() {
+        private Completable saveIncomingAndNotify(PendingMutation<T> incoming) {
             return save(incoming)
                 .andThen(notifyContentAvailable());
         }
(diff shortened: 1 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -332,10 +335,18 @@ private Completable handleIncomingUpdate() {
                     if (QueryPredicates.all().equals(incoming.getPredicate())) {
                         // If the incoming update does not have a condition, we want to delete any
                         // existing mutations for the modelId before saving the incoming one.
-                        return removeNotLocking(existing.getMutationId()).andThen(saveIncomingAndNotify());
+                        try {
+                            @SuppressWarnings("unchecked")
+                            PendingMutation<T> mergedPendingMutation = (PendingMutation<T>) merge(incoming, existing);
+                            return removeNotLocking(existing.getMutationId())
+                                    .andThen(saveIncomingAndNotify(mergedPendingMutation));
+                        } catch (Exception exception) {
```
richardmcclellan, line 343: Generally it's best only to catch specific exceptions, rather than this top level exception.  Can you enumerate the specific ones you are expecting here instead?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -414,5 +425,25 @@ private Completable unexpectedMutationScenario() {
                 "Please report at https://github.com/aws-amplify/amplify-android/issues."
             ));
         }
+
+        private PendingMutation<Model> merge(PendingMutation<T> incoming, PendingMutation<T> existing)
+                throws AmplifyException {
+            SerializedModel incomingSerializedModel = SerializedModel.create(incoming.getMutatedItem(),
```
richardmcclellan, line 431: Instead of creating a new `SerializedModel` instance (e.g. `SerializedModel.create(...)`, we should just use the object itself, if  `incoming.getMutatedItem() instanceof SerializedModel`.    If it's not, fallback to existing behavior (overwrite instead of merge)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -414,5 +425,25 @@ private Completable unexpectedMutationScenario() {
                 "Please report at https://github.com/aws-amplify/amplify-android/issues."
             ));
         }
+
+        private PendingMutation<Model> merge(PendingMutation<T> incoming, PendingMutation<T> existing)
```
richardmcclellan, line 429: Maybe this could be a utility method in the `PendingMutation` class?  I'm just thinking we should try to avoid `PersistentMutationOutbox` from getting too long.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -414,5 +425,25 @@ private Completable unexpectedMutationScenario() {
...
+                throws AmplifyException {
+            SerializedModel incomingSerializedModel = SerializedModel.create(incoming.getMutatedItem(),
+                                                                            incoming.getModelSchema());
+            SerializedModel existingSerializedModel = SerializedModel.create(existing.getMutatedItem(),
+                                                                            existing.getModelSchema());
+            Map<String, Object> mergedSerializedData = new HashMap<>();
+            for (String key : incomingSerializedModel.getSerializedData().keySet()) {
+                Object value = incomingSerializedModel.getSerializedData().get(key) != null ?
```
richardmcclellan, line 437: If the value is null, that is actually a valid value that the customer may want to set this field to.  What we should be checking here instead, is if the key _exists_.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -321,6 +322,7 @@ private Completable handleIncomingCreate() {
          * Determine which action to take when the incoming mutation type is {@linkplain PendingMutation.Type#UPDATE}.
          * @return A completable with the actions needed to resolve the conflict
          */
+        @SuppressWarnings("unchecked")
```
richardmcclellan, line 325: Is this because of the cast to `SerializedModel`?    If so, can you just add a comment indicating that, like:

```suggestion
        @SuppressWarnings("unchecked") // cast to SerializedModel
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -330,12 +332,27 @@ private Completable handleIncomingUpdate() {
...
+                                    (SerializedModel) incoming.getMutatedItem(),
+                                    (SerializedModel) existing.getMutatedItem(),
+                                    incoming.getModelSchema());
+                            PendingMutation<T> mergedPendingMutation = (PendingMutation<T>) PendingMutation.update(
+                                    mergedSerializedModel,
+                                    incoming.getModelSchema());
+                            return removeNotLocking(existing.getMutationId())
+                                    .andThen(saveIncomingAndNotify(mergedPendingMutation));
+                        } else {
+                            // If the incoming update does not have a condition, we want to delete any
```
richardmcclellan, line 349: I think this comment makes more sense up above where it was before.  Can you move it back?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
sdhuka: cast SerializedModel to Model

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -321,6 +322,7 @@ private Completable handleIncomingCreate() {
          * Determine which action to take when the incoming mutation type is {@linkplain PendingMutation.Type#UPDATE}.
          * @return A completable with the actions needed to resolve the conflict
          */
+        @SuppressWarnings("unchecked") // cast SerializedModel to Model
```
richardmcclellan, line 325:
```suggestion
        @SuppressWarnings("unchecked") // cast to SerializedModel
```

sdhuka: As discussed offline, moved the annotation inline

richardmcclellan (approved):
LGTM 🚀

whisller:
@sdhuka thanks for this patch! @richardmcclellan thanks for review! 
Great to see it going out so quickly :)

PR #1385, opened 2021-06-22
fix(datastore): explicitly include id field for update mutations, to support types with custom primary keys

richardmcclellan:
Currently, when you do a `DataStore.save(…)`, we query for the item in SQLite, and compute the difference in fields.  Then, when the mutation gets sent to AppSync, it only sends the fields that have changed, as well as the partition key and all sort keys, since those are always required.   

If the model being saved defines a custom primary key other than `id`, an error occurs on [this line](https://github.com/aws-amplify/amplify-android/blob/release_v1.18.0/aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/SerializedModel.java#L90) in the method that computes the difference of fields.    Essentially, since `id` is not part of the primary key, it isn’t getting included, but `SerializedModel` actually still requires a hardcoded `id` field. 

This PR resolves the issue by always including the `id` field on update mutations.  Once we add complete support for custom primary keys though, this won't be necessary anymore.

### DIFF
--- a/testmodels/src/main/java/com/amplifyframework/testmodels/ecommerce/AmplifyModelProvider.java
+++ b/testmodels/src/main/java/com/amplifyframework/testmodels/ecommerce/AmplifyModelProvider.java
@@ -1,0 +1,60 @@
+package com.amplifyframework.testmodels.ecommerce;
+
+import com.amplifyframework.core.model.Model;
+import com.amplifyframework.core.model.ModelProvider;
+import com.amplifyframework.testmodels.commentsblog.Author;
+import com.amplifyframework.testmodels.commentsblog.Blog;
+import com.amplifyframework.testmodels.commentsblog.BlogOwner;
+import com.amplifyframework.testmodels.commentsblog.Comment;
+import com.amplifyframework.testmodels.commentsblog.Post;
+import com.amplifyframework.testmodels.commentsblog.PostAuthorJoin;
+import com.amplifyframework.util.Immutable;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ *  Contains the set of model classes that implement {@link Model}
+ * interface.
+ */
+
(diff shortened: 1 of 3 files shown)


testmodels/src/main/java/com/amplifyframework/testmodels/ecommerce/AmplifyModelProvider.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+public final class AmplifyModelProvider implements ModelProvider {
+  private static final String AMPLIFY_MODEL_VERSION = "291b773087e72b546d24c87e137d7b91";
+  private static AmplifyModelProvider amplifyGeneratedModelInstance;
+  private AmplifyModelProvider() {
+    
+  }
+  
+  public static AmplifyModelProvider getInstance() {
```
poojamat, line 29: Can we make all the new classes Kotlin?

poojamat: Also does this class need to be singleton?

richardmcclellan: > Can we make all the new classes Kotlin?

Whenever we are ready to add a new category, absolutely - I think we should write it entirely in Kotlin.   For existing categories, I think we might be able to do it incrementally as well, as long as we put some thought into where to draw the boundary between languages, simply for our own sanity.    We actually discussed this as a team before here (and everyone is generally for it): https://github.com/aws-amplify/amplify-android/discussions/743

This class is actually code generated though, so we probably don't want to make any changes to it in this PR. 

> Also does this class need to be singleton?

Hmm, can you clarify your question?  I believe it is a singleton, right?    Either way, we probably don't want to make changes to this class since it's code generated.

poojamat (approved):

PR #1395, opened 2021-06-29
[POC] - accept user attributes via custom property prefix

rjuliano:
*Issue #, if available:*

*Description of changes:*
An alternative to #1306. This prevents a plugin implementation detail (the distinction between custom and user attributes) from being leaked via the category interface. 


richardmcclellan (commented):
Interesting solution!  Previously, I had approved https://github.com/aws-amplify/amplify-android/pull/1306, but you do have a good point that it leaks a plugin implementation detail into the category interface, which is not good.  

I'm curious - do you know if JS/iOS/Flutter supports this feature, and if so, how do they handle it?  We should make sure to do a design review with them before finalizing this change.

I'm not crazy about the approach of asking customers to add a `user:` prefix - it feels confusing / non-intuitive.

As a 3rd option, could we instead create an `AWSUserProfile` object in the plugin module, which extends `UserProfile`, and contains a `userAttributes` method?   We use this pattern of inheritance for adding plugin specific properties in a few other places as well.  For example, `AWSS3StorageUploadFileOptions` extends `StorageUploadFileOptions` to add `serverSideEncryption`, which is AWS specific.  This suggestion is probably more similar to https://github.com/aws-amplify/amplify-android/pull/1306 than https://github.com/aws-amplify/amplify-android/pull/1395

rjuliano:
> I'm curious - do you know if JS/iOS/Flutter supports this feature, and if so, how do they handle it? We should make sure to do a design review with them before finalizing this change.

[JS supports it](https://docs.amplify.aws/lib/analytics/getting-started/q/platform/js#update-endpoint) by accepting `attributes` and `userAttributes` as part of the payload that gets passed in to `updateEndpoint`. For native, the category interface doesn't expose `updateEndpoint` directly; instead, we have the `identifyUser` API that calls `updateEndpoint` internally.

> I'm not crazy about the approach of asking customers to add a `user:` prefix - it feels confusing / non-intuitive.

Me neither.

> As a 3rd option, could we instead create an `AWSUserProfile` object in the plugin module, which extends `UserProfile`, and contains a `userAttributes` method? We use this pattern of inheritance for adding plugin specific properties in a few other places as well. For example, `AWSS3StorageUploadFileOptions` extends `StorageUploadFileOptions` to add `serverSideEncryption`, which is AWS specific. This suggestion is probably more similar to #1306 than #1395

I like that. I'll take a stab at it and will update #1306 with that approach. I don't see why it shouldn't work.

richardmcclellan:
Closing in favor of https://github.com/aws-amplify/amplify-android/pull/1306

PR #1397, opened 2021-06-30
fix: Revert multiauth

richardmcclellan:
I identified some issues with the recent multiauth changes - inside `SubscriptionEndpoint` , there is [some new logic](https://github.com/aws-amplify/amplify-android/blob/5eace8336ab4691cf8257af892dc5f5a87eaf1a2/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java#L420-L434) that checks if there are GraphQL errors.  If there are, then we are now dispatching them to the error callback, whereas previously we were dispatching them to the response callback.   By doing so, downstream consumers no longer have a way to get details about the specific error type.   As a result, the logic for ignoring certain errors in the `SubscriptionProcessor` and `MutationProcessor` no longer works (e.g. `Unauthorized` or `OperationDisabled` errors).  I believe this also affects customers using `Amplify.API.subscribe` functionality.

I tried to think of a way we could just fix this now, but I don't think there's an easy change to make. It will require some thought since we need to implement multiauth using a bit of a different approach.  In the short term, we should just revert the changes.


rjuliano (approved):
LGTM. We can keep an eye on two potential fixes #1398 and #1399 before we revert

rjuliano (approved):
Meant to approve it

richardmcclellan:
Closing in favor of https://github.com/aws-amplify/amplify-android/pull/1399

PR #1398, opened 2021-07-01
fix(api): only handle auth errors as exception

rjuliano:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -427,11 +427,8 @@ final class SubscriptionEndpoint {
                     );
                     return;
                 }
-                // Otherwise, just dispatch as an ApiException
-                dispatchError(new ApiException(
-                    "Error in subscription response: " + response.getErrors(),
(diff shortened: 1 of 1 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -427,11 +427,8 @@ void dispatchNextMessage(String message) {
                     );
                     return;
                 }
-                // Otherwise, just dispatch as an ApiException
-                dispatchError(new ApiException(
-                    "Error in subscription response: " + response.getErrors(),
-                    AmplifyException.TODO_RECOVERY_SUGGESTION)
-                );
+                // Otherwise, just dispatch it as a response so callers can deal with the error.
+                onNextItem.accept(response);
```
richardmcclellan, line 431: This will probably resolve part of the problem, but not all of it.    This should fix the issue where `OperationDisabled` errors should be ignored.  However, we also have logic that ignores errors if `errorType=Unauthorized`, that still won't work because of lines 423-426.    

We could solve that by refactoring that logic to check for `ApiAuthException` instead, but even if we did, this change could still potentially break customers who are checking for an Unauthorized error.

PR #1399, opened 2021-07-01
fix(api): move error handling to multi-auth operation

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR does the following:

- Moves the code that checks for error codes in the AppSync response into the multiauth operation classes.
- Moved AppSyncExtensions to aws-api-appsync.
- Check for both `Unauthorized` and `UnauthorizedException` error codes returned from AppSync. We are only checking for `Unauthorized` right now but AppSync is returning `UnauthorizedException`. Not sure if that was a service side change.


aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
```diff
@@ -151,6 +164,16 @@ public synchronized void cancel() {
         }
     }
 
+    private boolean hasAuthRelatedErrors(GraphQLResponse<T> response) {
+        for (GraphQLResponse.Error error : response.getErrors()) {
+            if (error.getExtensions() != null &&
+                UNAUTHORIZED_EXCEPTION.equals(error.getExtensions().get("errorType"))) {
+                return true;
+            }
```
richardmcclellan, line 172: Minor suggestion for improvement that is a little bit more type safe, not necessarily needed for this PR: 

```suggestion
                if (!Empty.check(error.getExtensions())) {
                    AppSyncExtensions extensions = new AppSyncExtensions(error.getExtensions());
                    return AppSyncErrorType.UNAUTHORIZED.equals(extensions.getErrorType());
                }
```

Note this will change will require moving AppSyncExtensions from `aws-datastore` to `aws-api-appsync` first though.

rjuliano: I'm testing that. If AppSyncExtensions doesn't have any other baggage, it should be an easy enough move.

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
```diff
@@ -116,7 +116,20 @@ private void dispatchRequest() {
...
+                        if (hasAuthRelatedErrors(response)) {
+                            executorService.submit(this::dispatchRequest);
+                            return;
+                        }
+                        // Otherwise, we just want to dispatch it as a next item and
+                        // let callers deal with the errors.
+                        onNextItem.accept(response);
+                    } else {
+                        onNextItem.accept(response);
+                    }
```
richardmcclellan, line 131: Same logic, just a little simplified - 

```suggestion
                    if (response.hasErrors() && hasAuthRelatedErrors(response)) {
                        // If there are auth-related errors, dispatch an ApiAuthException
                        executorService.submit(this::dispatchRequest);
                    } else {
                        // Otherwise, we just want to dispatch it as a next item and
                        // let callers deal with the errors.
                        onNextItem.accept(response);
                    }
```

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
```diff
@@ -116,7 +116,20 @@ private void dispatchRequest() {
                     MutiAuthSubscriptionOperation.this.subscriptionId = subscriptionId;
                     onSubscriptionStart.accept(subscriptionId);
                 },
-                onNextItem,
+                response -> {
+                    if (response.hasErrors()) {
+                        // If there are auth-related errors, dispatch an ApiAuthException
+                        if (hasAuthRelatedErrors(response)) {
+                            executorService.submit(this::dispatchRequest);
```
richardmcclellan, line 123: With this approach, downstream callers (DataStore) are going to get a `ApiAuthException` instead of a `GraphQLResponse` with `errorType=Unauthorized`, right, if using multiauth, right?  I think it would be a little bit better if we could make `MultiAuthSubscriptionOperation` behave more congruently with the regular `SubscriptionOperation`.  

I think this would actually be pretty easy - we would just only call`this::dispatchRequest` here if `authTypes.hasNext()` is true, otherwise, call `onNextItem.accept(response)`

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
```diff
@@ -116,7 +116,20 @@ private void dispatchRequest() {
...
+                            return;
+                        }
+                        // Otherwise, we just want to dispatch it as a next item and
+                        // let callers deal with the errors.
+                        onNextItem.accept(response);
+                    } else {
+                        onNextItem.accept(response);
+                    }
+                },
```
richardmcclellan, line 132: Now that the error handling logic is in the response callback, can the similar logic (`executorService.submit(this::dispatchRequest);`) be removed from the error callback?

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
rjuliano: Just checked this one in. Moving that file didn't cause any side effects

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
rjuliano: Good call on this one...Just committed this change as well.

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
rjuliano: `ApiAuthException` can still be thrown when `SubscriptionEndpoint` calls `createHeadersForSubscription` in the authorizer. Specifically if they're trying to use OIDC without having passed in a provider at configuration time.

aws-api/src/main/java/com/amplifyframework/api/aws/MutiAuthSubscriptionOperation.java
rjuliano: I'm testing this locally right now. This would basically serve to short-circuit the last submission to the executor where we would exit anyways because `authTypes.hasNext()` would be false (line 90).

richardmcclellan: I imagine we might need a similar change to any other callers of the `dispatchRequest` method as well.

richardmcclellan:
I just tested this branch and confirmed the "ignore `OperationDisabled` error" behavior is working again, so that's good!   

I did notice there may be one more problem to fix though - I believe we need to make the same changes that you've already made in `MultiAuthSubscriptionOperation` in `MultiAuthAppSyncGraphQLOperation` class.  That is - don't convert a `GraphQLResponse` that `hasErrors()` to an `ApiException`.

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
```diff
@@ -129,6 +129,15 @@ public String toString() {
             '}';
     }
 
+    /**
+     * Checks whether the current extension data indicates an authorization related error.
+     * @return True if it is authorization related, false otherwise.
+     */
+    public boolean isAuthorizationError() {
+        return AppSyncExtensions.AppSyncErrorType.UNAUTHORIZED.equals(this.getErrorType()) ||
+               AppSyncExtensions.AppSyncErrorType.UNAUTHORIZED_EXCEPTION.equals(this.getErrorType());
```
richardmcclellan, line 138: Huh, I didn't realize AppSync returned both of these :/. That's kind of annoying.  I guess we should handle both though.

aws-api-appsync/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java
```diff
@@ -129,6 +129,15 @@ public String toString() {
             '}';
     }
 
+    /**
+     * Checks whether the current extension data indicates an authorization related error.
+     * @return True if it is authorization related, false otherwise.
+     */
+    public boolean isAuthorizationError() {
```
richardmcclellan, line 136: This is kind of a public facing class, so naming is important.  Perhaps name this method `isUnauthorizedErrorType`?

richardmcclellan (approved):
:shipit:

PR #1411, opened 2021-07-13
feat(auth): Add signin api for passwordless custom auth

sdhuka:
*Issue #, if available:*

*Description of changes:*
- Add api for passwordless custom-auth signin

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
@@ -117,6 +117,7 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
     private static final long SECONDS_BEFORE_TIMEOUT = 10;
     private static final String COGNITO_USER_ID_ATTRIBUTE = "sub";
     private static final String MOBILE_CLIENT_TOKEN_KEY = "token";
+    private static final String EMPTY_STRING = "";
     private String userId;
     private AWSMobileClient awsMobileClient;
     private AuthChannelEventName lastEvent;
@@ -407,6 +408,15 @@ public final class AWSCognitoAuthPlugin extends AuthPlugin<AWSMobileClient> {
             @NonNull final Consumer<AuthException> onException
     ) {
         signIn(username, password, AuthSignInOptions.defaults(), onSuccess, onException);
+    }
+
+    @Override
+    public void signIn(
+            @Nullable String username,
+            @NonNull Consumer<AuthSignInResult> onSuccess,
+            @NonNull Consumer<AuthException> onException
+    ) {
+        signIn(username, EMPTY_STRING, AuthSignInOptions.defaults(), onSuccess, onException);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


richardmcclellan (commented):
Has this been discussed with iOS/JS/Flutter, and approved in an API review?  If not, we should do that.

richardmcclellan:
> Has this been discussed with iOS/JS/Flutter, and approved in an API review? If not, we should do that.

We discussed offline that this API sort of already exists on iOS, because the password parameter on the signIn is optional (default value is nil), so this change is basically just closing a parity gap. 

It seems like a reasonable addition - let's just get @royjit to review / approve as well.

core/src/main/java/com/amplifyframework/auth/AuthCategory.java
```diff
@@ -113,6 +115,15 @@ public void signIn(
         getSelectedPlugin().signIn(username, password, onSuccess, onError);
     }
 
+    @Override
+    public void signIn(
+            @Nullable String username,
+            @NonNull Consumer<AuthSignInResult> onSuccess,
+            @NonNull Consumer<AuthException> onError
+    ) {
+        getSelectedPlugin().signIn(username, EMPTY_STRING, onSuccess, onError);
```
royjit, line 124:
```suggestion
        getSelectedPlugin().signIn(username, onSuccess, onError);
```

core/src/main/java/com/amplifyframework/auth/AuthCategory.java
```diff
@@ -45,6 +45,8 @@
  */
 public final class AuthCategory extends Category<AuthPlugin<?>> implements AuthCategoryBehavior {
 
+    private static final String EMPTY_STRING = "";
```
royjit, line 48: Remove from here, we are already doing this in the plugin.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -117,6 +117,7 @@
     private static final long SECONDS_BEFORE_TIMEOUT = 10;
     private static final String COGNITO_USER_ID_ATTRIBUTE = "sub";
     private static final String MOBILE_CLIENT_TOKEN_KEY = "token";
+    private static final String EMPTY_STRING = "";
```
richardmcclellan, line 120: Per https://github.com/aws-amplify/amplify-android/pull/763#issuecomment-684967095, the current way to specify no password is to pass null, instead of empty string.    Is that incorrect?

palpatim:
Nit: This is not a `chore`--it updates the public API, and so would be a `feat`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
raphkim: Saijad noticed that the lambda that is generated by CLI to handle custom auth will only work with empty string (and break with `null`)

richardmcclellan:
As discussed internally, we are going to hold off on this change for now until we can design a complete solution that will work for all platforms, and addresses some other missing use cases as well.

In the interim, customers can sign in with passwordless auth by passing an empty string for the password, like this:

```
Amplify.Auth.signIn(
    "username",
    "",
    result -> Log.i("AuthQuickstart", result.isSignInComplete() ? "Sign in succeeded" : "Sign in not complete"),
    error -> Log.e("AuthQuickstart", error.toString())
);
```

PR #1412, opened 2021-07-13
feat: add support for AWS_LAMBDA auth type

richardmcclellan:
Today, [AppSync supports four authorization types](https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html) - `API_KEY`, `AWS_IAM`, `OPENID_CONNECT`, and `AMAZON_COGNITO_USER_POOLS`.    This PR adds support for a 5th type: `AWS_LAMBDA`.  Customers will be able to create their own AWS Lambda function, which will determine whether to authorize each request based on an authorization token that the customer provides to Amplify.
 
The developer experience is very similar to `OPENID_CONNECT`.  Today, customers can provide their own OIDC token to Amplify by implementing an `OIDCTokenProvider`, and providing it to the `AWSApiPlugin` like this:

```
Amplify.addPlugin(new AWSApiPlugin(ApiAuthProviders.builder()
    .oidcAuthProvider(() -> "token-from-your-oidc-provider")
    .build()));
```

For AWS Lambda custom auth, developers can provide their own token by implementing a `CustomAuthProvider` like this:

```
Amplify.addPlugin(new AWSApiPlugin(ApiAuthProviders.builder()
    .functionAuthProvider(() -> "function-token")
    .build()));
```

Additionally, this PR includes some refactoring to DRY up some of our request signing logic.  In the beginning, all requests for `REST` and `GraphQL` were signed using the `AppSyncSigV4SignerInterceptor`, which was a static interceptor added to the `OkHttpClient`.  When multi-auth came along, we needed something more flexible - that would allow us to specify an auth type at the type the Request was being built, so the `ApiRequestDecoratorFactory` was created, and replaced `AppSyncSigV4SignerInterceptor` for all GraphQL operations.     The REST operations still rely on the old `AppSyncSigV4SignerInterceptor` though.

This PR refactors the REST operations to use the `ApiRequestDecoratorFactory`, and removes the `AppSyncSigV4SignerInterceptor` entirely.  


aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestDecoratorFactory.java
```diff
@@ -146,25 +144,50 @@ public RequestDecorator forAuthType(@NonNull AuthorizationType authorizationType
...
+                if (apiAuthProviders.getCustomAuthProvider() == null) {
+                    throw new ApiAuthException("Attempting to use AWS_LAMBDA authorization " +
+                            "without a provider implemented.",
+                            "Configure a CustomAuthProvider when initializing the API plugin.");
+                }
+                final String customToken;
+                try {
+                    customToken = apiAuthProviders.getCustomAuthProvider().getLatestAuthToken();
+                } catch (ApiException exception) {
+                    throw new ApiAuthException("Failed to retrieve auth token from OIDC provider.",
```
rjuliano, line 158: Probably need a different error message here for the custom provider.

aws-api/src/main/java/com/amplifyframework/api/aws/auth/ApiRequestDecoratorFactory.java
```diff
@@ -146,25 +144,50 @@ public RequestDecorator forAuthType(@NonNull AuthorizationType authorizationType
...
                 } else if (apiKey != null) {
                     return new ApiKeyRequestDecorator(() -> apiKey);
                 } else {
                     throw new ApiAuthException("Attempting to use API_KEY authorization without " +
-                                                                "an API key provider or an API key in the config file",
-                                                            "Verify that an API key is in the config file or an " +
-                                                                "ApiKeyAuthProvider is setup during the API " +
-                                                                "plugin initialization.");
+                                                "an API key provider or an API key in the config file",
+                                            "Verify that an API key is in the config file or an " +
```
rjuliano, line 171:
```suggestion
                                                "Verify that an API key is in the config file or an " +
```

nit: alignment fix.

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
```diff
@@ -75,28 +75,33 @@ public int getPriority() {
...
-     * Represents the the value of the provider field of the @auth directive.
+     * Represents the value of the provider field of the @auth directive.
      */
     public enum Provider {
         /**
-         * The userPools provider of the @auth rule directive.
+         * The custom provider of the @auth rule directive.
          */
-        USER_POOLS("userPools", 1),
+        CUSTOM("custom", 1),
```
rjuliano, line 84: This is probably a good call in terms of ranking higher than everything else. Do you know if the other platforms are following suit?

rjuliano (approved):
Left a couple minor comments, but overall I think this looks good. 🚢

core/src/main/java/com/amplifyframework/core/model/AuthStrategy.java
richardmcclellan: Yes, the other platforms are doing the same!   We decided on this order in the design review because they should be ordered by "most specific" to "least specific".    Since you can authorize based on specific attributes of a user in an AWS Lambda function, `CUSTOM` is the most specific.

rjuliano (approved):
Ship it 🚀

poojamat (approved):

PR #1414, opened 2021-07-15
Retry generic

poojamat:
*Issue #, if available:*

*Description of changes:*
Rough PR for geneaicizing the retry logic.

### DIFF
--- a/aws-analytics-pinpoint/build.gradle
+++ b/aws-analytics-pinpoint/build.gradle
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
 
 dependencies {
     implementation project(path: ':core')
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
@@ -142,6 +142,9 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
                     return;
                 }
             }
+            if(response.code() >= 500) {
+                onFailure.accept(new ApiException.ServerErrorException("OkHttp client request failed.", "Server Error"));
+            }
 
             try {
                 onResponse.accept(wrapResponse(jsonResponse));
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
@@ -142,8 +142,8 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
                     return;
                 }
             }
-            if(response.code() >= 500) {
-                onFailure.accept(new ApiException.ServerErrorException("OkHttp client request failed.", "Server Error"));
+            if(response.code() >= 400) {
+                onFailure.accept(new ApiException.ApiIrrecoverableException("OkHttp client request failed.", "Irrecoverable error"));
             }
 
             try {
--- a/aws-datastore/build.gradle
+++ b/aws-datastore/build.gradle
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
 
 dependencies {
     implementation project(':core')
@@ -35,7 +36,10 @@ dependencies {
     testImplementation dependency.mockito
     testImplementation dependency.robolectric
     testImplementation dependency.androidx.test.core
+    testImplementation 'org.mockito:mockito-core:3.8.0'
+    testImplementation 'io.mockk:mockk:1.11.0'
 
+    androidTestImplementation 'org.mockito:mockito-android:3.8.0'
     androidTestImplementation project(path: ':testmodels')
     androidTestImplementation project(path: ':testutils')
     androidTestImplementation project(path: ':aws-api')
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
@@ -35,6 +35,7 @@ import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.core.model.query.predicate.QueryPredicates;
 import com.amplifyframework.datastore.DataStoreException;
+import com.amplifyframework.datastore.ErrorType;
 
 /**
  * An implementation of the {@link AppSync} client interface.
@@ -97,10 +98,19 @@ public final class AppSyncClient implements AppSync {
             }
         };
         final Consumer<ApiException> failureConsumer =
-            failure -> onFailure.accept(new DataStoreException(
-                "Failure performing sync query to AppSync.",
-                failure, AmplifyException.TODO_RECOVERY_SUGGESTION
-            ));
+            failure -> {
+            if(failure instanceof ApiException.ApiIrrecoverableException){
+                onFailure.accept(new DataStoreException(
+                        "Failure performing sync query to AppSync.",
+                        failure, AmplifyException.TODO_RECOVERY_SUGGESTION,
+                        ErrorType.IRRECOVERABLE_ERROR
+                ));
+            } else {
+                onFailure.accept(new DataStoreException(
+                        "Failure performing sync query to AppSync.",
+                        failure, AmplifyException.TODO_RECOVERY_SUGGESTION
+                ));
+            } };
 
         final Cancelable cancelable = api.query(request, responseConsumer, failureConsumer);
         if (cancelable != null) {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnErrorConsumer.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnErrorConsumer.kt
@@ -1,0 +1,30 @@
+package com.amplifyframework.datastore.syncengine
+
+import com.amplifyframework.api.graphql.GraphQLRequest
+import com.amplifyframework.api.graphql.GraphQLResponse
+import com.amplifyframework.api.graphql.PaginatedResult
+import com.amplifyframework.core.Consumer
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.datastore.DataStoreException
+import com.amplifyframework.datastore.ErrorType
+import com.amplifyframework.datastore.appsync.AppSync
+import com.amplifyframework.datastore.appsync.ModelWithMetadata
+import io.reactivex.rxjava3.core.SingleEmitter
+
+class OnErrorConsumer<T: Model>( val emitter: SingleEmitter<PaginatedResult<ModelWithMetadata<T>>>,
+                                 val appSync: AppSync,
+                                 val request: GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>>,
+        private val onResponse: Consumer<GraphQLResponse<PaginatedResult<ModelWithMetadata<T>>>>,
+        private val retryHandler: RetryHandler): Consumer<DataStoreException> {
+
+
+    override fun accept(result: DataStoreException) {
+        if (result.type.equals( ErrorType.IRRECOVERABLE_ERROR)) {
+            emitter.onError(
+               result
+            )
+        } else {
+            retryHandler.retry(emitter,appSync,request, onResponse, this)
+        }
+    }
+}
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -120,7 +120,7 @@ public final class Orchestrator {
             .merger(merger)
             .dataStoreConfigurationProvider(dataStoreConfigurationProvider)
             .queryPredicateProvider(queryPredicateProvider)
-            .retryStrategy(RetryStrategy.RX_INTERRUPTIBLE_WITH_BACKOFF)
+            .retryHandler(new RetryHandler())
             .build();
         this.subscriptionProcessor = SubscriptionProcessor.builder()
                 .appSync(appSync)
@@ -293,7 +293,6 @@ public final class Orchestrator {
 
     /**
      * Start syncing models to and from a remote API.
-     * @return A Completable that succeeds when API sync is enabled.
      */
     private void startApiSync() {
         LOG.info("Setting currentState to SYNC_VIA_API");
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
@@ -120,6 +120,7 @@ public final class Orchestrator {
             .merger(merger)
             .dataStoreConfigurationProvider(dataStoreConfigurationProvider)
             .queryPredicateProvider(queryPredicateProvider)
+            .retryStrategy(RetryStrategy.RX_INTERRUPTIBLE_WITH_BACKOFF)
             .build();
         this.subscriptionProcessor = SubscriptionProcessor.builder()
                 .appSync(appSync)
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
@@ -1,0 +1,52 @@
+package com.amplifyframework.datastore.syncengine
+
+import com.amplifyframework.api.graphql.GraphQLRequest
+import com.amplifyframework.api.graphql.GraphQLResponse
+import com.amplifyframework.api.graphql.PaginatedResult
+import com.amplifyframework.core.Amplify
+import com.amplifyframework.core.Consumer
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.datastore.AmplifyDisposables
+import com.amplifyframework.datastore.DataStoreException
+import com.amplifyframework.datastore.appsync.AppSync
+import com.amplifyframework.datastore.appsync.ModelWithMetadata
+import io.reactivex.rxjava3.core.SingleEmitter
+import java.util.*
+import java.util.concurrent.TimeUnit
+
+class RetryHandler() {
+
+    private var numberOfAttempts = 0
+    private val maxExponent = 8
+    private val jitterFactor = 100
+    private val LOG = Amplify.Logging.forNamespace("amplify:aws-datastore")
+
+    fun <T : Model> retry(
+        emitter: SingleEmitter<PaginatedResult<ModelWithMetadata<T>>>,
+        appSync: AppSync,
+        request: GraphQLRequest<PaginatedResult<ModelWithMetadata<T>>>,
+        onResponse: Consumer<GraphQLResponse<PaginatedResult<ModelWithMetadata<T>>>>,
+        onFailure: Consumer<DataStoreException>
+    ) {
+        numberOfAttempts++
+        val jitteredDelay = jitteredDelay()
+        Timer().schedule(object : TimerTask() {
+            override fun run() {
+                LOG.verbose("Retrying attempt number:$numberOfAttempts jittered delay:$jitteredDelay")
+                val cancelable = appSync.sync(request, onResponse, onFailure)
+                emitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable))
+                cancel()
+            }
+        }, jitteredDelay)
+    }
+
+
+    private fun jitteredDelay(): Long {
+        val waitTimeSeconds: Long = java.lang.Double.valueOf(
+            Math.pow(2.0, (numberOfAttempts % maxExponent).toDouble())
+                    + jitterFactor * Math.random()
+        ).toLong()
+        return TimeUnit.MILLISECONDS.toMillis(waitTimeSeconds)
+    }
+
+}
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
@@ -330,7 +330,6 @@ final class SyncProcessor {
         @NonNull
         @Override
         public QueryPredicateProviderStep dataStoreConfigurationProvider(
-            DataStoreConfigurationProvider dataStoreConfigurationProvider) {
                 DataStoreConfigurationProvider dataStoreConfigurationProvider) {
             this.dataStoreConfigurationProvider = dataStoreConfigurationProvider;
             return Builder.this;
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
@@ -73,6 +73,7 @@ final class SyncProcessor {
     private final DataStoreConfigurationProvider dataStoreConfigurationProvider;
     private final String[] modelNames;
     private final QueryPredicateProvider queryPredicateProvider;
+    private RetryStrategy.RxRetryStrategy retryStrategy;
 
     private SyncProcessor(Builder builder) {
         this.modelProvider = builder.modelProvider;
@@ -85,6 +86,7 @@ final class SyncProcessor {
         this.modelNames =
             ForEach.inCollection(modelProvider.modelSchemas().values(), ModelSchema::getName)
                 .toArray(new String[0]);
+        this.retryStrategy = builder.retryStrategy;
     }
 
     /**
@@ -229,6 +231,7 @@ final class SyncProcessor {
                         processor.onComplete();
                     }
                 })
+                .retry(retryStrategy::retryHandler)
                 // If it's a SerializedModel, add the ModelSchema, since it isn't added during deserialization.
                 .map(paginatedResult -> Flowable.fromIterable(paginatedResult)
                         .map(modelWithMetadata -> hydrateSchemaIfNeeded(modelWithMetadata, schema))
@@ -285,7 +288,7 @@ final class SyncProcessor {
      */
     public static final class Builder implements ModelProviderStep, ModelSchemaRegistryStep,
             SyncTimeRegistryStep, AppSyncStep, MergerStep, DataStoreConfigurationProviderStep,
-            QueryPredicateProviderStep, BuildStep {
+            QueryPredicateProviderStep, RetryStrategyStep, BuildStep {
         private ModelProvider modelProvider;
         private ModelSchemaRegistry modelSchemaRegistry;
         private SyncTimeRegistry syncTimeRegistry;
@@ -293,6 +296,7 @@ final class SyncProcessor {
         private Merger merger;
         private DataStoreConfigurationProvider dataStoreConfigurationProvider;
         private QueryPredicateProvider queryPredicateProvider;
+        private RetryStrategy.RxRetryStrategy retryStrategy;
 
         @NonNull
         @Override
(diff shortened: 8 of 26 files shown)


settings.gradle
```diff
@@ -32,3 +32,7 @@ include ':testmodels'
 include ':core-kotlin'
 include ':rxbindings'
 include ':aws-api-appsync'
+
+
+include ':sample-onCall'
+project(":sample-onCall").projectDir=new File(rootDir, "../OnCall/app")
```
richardmcclellan, line 38:
```suggestion
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
```diff
@@ -8,45 +8,23 @@ import com.amplifyframework.core.Consumer
...
-                    + jitterFactor * Math.random()
-        ).toLong()
-        return TimeUnit.MILLISECONDS.toMillis(waitTimeSeconds)
+    override fun execute() {
+        val cancelable = appSync.sync(request, onResponse, onFailure)
+        emitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable))
     }
+}
 
+interface RetryCallbackInterface<T: Model>{
```
richardmcclellan, line 28: We have an existing interface called `Action`, that you could probably use instead of creating this new one.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/OnErrorConsumer.kt
```diff
@@ -24,7 +24,7 @@ class OnErrorConsumer<T: Model>( val emitter: SingleEmitter<PaginatedResult<Mode
                result
             )
         } else {
-            retryHandler.retry(emitter,appSync,request, onResponse, this)
+            retryHandler.retry(RetryHandler(emitter, appSync, request, onResponse, this))
```
richardmcclellan, line 27: Since the `RetryHandler` class is really only two lines of code, maybe just put the implementation inline, like this?

```suggestion
            retryHandler.retry({
               val cancelable = appSync.sync(request, onResponse, this)
               emitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable))
        })
```

poojamat: Would this work with Java? I need to convet this class back to java.

poojamat: I can create an inline implementation of the interface in Java.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+
+
+        fun <T> retry(single: Single<T>, skipExceptions: List<Class<out Throwable>?>): Single<T> {
+            return Single.create { emitter -> call(single, emitter, 0, skipExceptions) }
+        }
+
+        private fun <T> call(single: Single<T>, emitter: SingleEmitter<T>, delayInSeconds: Long, skipExceptions: List<Class<out Throwable>?> ) {
+           single.delaySubscription(delayInSeconds, TimeUnit.SECONDS)
+                .subscribe({
+                            t: T -> emitter.onSuccess(t)
```
richardmcclellan, line 25: I believe this can be simplified to just use the implied `it` variable.  This is the recommendation from the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#lambda-parameters).

```suggestion
                            emitter.onSuccess(it)
```

Same thing with the error handler below (use implied `it`)

richardmcclellan: Actually, I think you can simplify it even further, and make this just `emitter::onSuccess`

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+
+
+class RequestRetry(private val maxExponent: Int = 8,
+                    private val jitterFactor: Int = 100,
+                    private val maxAttempts: Int = 3) {
+
+    private var numberOfAttempts = 0
+
+
+        fun <T> retry(single: Single<T>, skipExceptions: List<Class<out Throwable>?>): Single<T> {
```
richardmcclellan, line 18: This line should be indented 4 spaces, not 8.  There's some other weird indentation in this file too, so it would be good to clean it up a bit.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+                        emitter.onError(error)
+                    } else {
+                        call(single, emitter, jitteredDelay(),skipExceptions )
+                    }
+                }
+        }
+
+
+    fun jitteredDelay(): Long {
+        val waitTimeSeconds: Long = java.lang.Double.valueOf(
```
richardmcclellan, line 38: Can you import `java.lang.Double` at the top so that you don't have to use the fully qualified package name here?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import com.amplifyframework.AmplifyException
+
+import io.reactivex.rxjava3.core.Single
+import io.reactivex.rxjava3.core.SingleEmitter
+import java.util.concurrent.TimeUnit
+import kotlin.math.pow
+
+
+class RequestRetry(private val maxExponent: Int = 8,
```
richardmcclellan, line 11: Maybe name this class `RetryHandler` instead?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+            return Single.create { emitter -> call(single, emitter, 0, skipExceptions) }
+        }
+
+        private fun <T> call(single: Single<T>, emitter: SingleEmitter<T>, delayInSeconds: Long, skipExceptions: List<Class<out Throwable>?> ) {
+           single.delaySubscription(delayInSeconds, TimeUnit.SECONDS)
+                .subscribe({
+                            t: T -> emitter.onSuccess(t)
+                }) { error ->
+                    numberOfAttempts++
+                    if (numberOfAttempts > maxAttempts || (error.javaClass in skipExceptions)) {
```
richardmcclellan, line 28: FYI, we have a `ErrorInspector` utility class that checks to see if a `Throwable` is a certain type, or if any nested `cause` is of that type.  I'm not entirely sure if you need to check causes, but I'm guessing you might, since you are probably going to have a `DataStoreException` here, which has a cause that is an `NonRetryableException`.

core/src/main/java/com/amplifyframework/datastore/ErrorType.kt
```diff
@@ -0,0 +1,5 @@
+package com.amplifyframework.datastore
+
+ enum class ErrorType {
```
richardmcclellan, line 3: Is this needed?  I don't think it's used anymore

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -118,10 +119,16 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api,
...
+            failure -> {
+            if(failure instanceof ApiException.ApiIrrecoverableException){
+                onFailure.accept(new DataStoreException.IrRecoverableException(
+                        "Failure performing sync query to AppSync.", AmplifyException.TODO_RECOVERY_SUGGESTION));
+            } else {
+                onFailure.accept(new DataStoreException(
+                        "Failure performing sync query to AppSync.",
+                        failure, AmplifyException.TODO_RECOVERY_SUGGESTION
+                ));
+            } };
```
richardmcclellan, line 131: I don't think you need a `DataStoreException.IrRecoverableException` class.  Instead, don't make any changes here - just wrap the `ApiException` with a `DataStoreException` like it was.  Then, in the `RetryHandler` use `ErrorInspector.contains(error, ApiException.ApiIrrecoverableException.class)` to determine if it should be retried or not.    The `ErrorInspector` utility traverses the nested causes and checks each one to see if it matches the type you are looking for.

core/src/main/java/com/amplifyframework/api/ApiException.java
```diff
@@ -52,6 +53,13 @@ public ApiException(
         super(message, recoverySuggestion);
     }
 
+    public static final class ApiIrrecoverableException extends ApiException {
```
richardmcclellan, line 56: Maybe call this `NonRetryableException` instead?

aws-datastore/build.gradle
```diff
@@ -35,7 +36,10 @@ dependencies {
     testImplementation dependency.mockito
     testImplementation dependency.robolectric
     testImplementation dependency.androidx.test.core
+    testImplementation 'org.mockito:mockito-core:3.8.0'
```
richardmcclellan, line 39: Can you define these in the root `build.gradle`, and then reference the variables from here?  Similar to how all the other dependencies are defined.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -142,6 +142,9 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
                     return;
                 }
             }
+            if(response.code() >= 400) {
```
richardmcclellan, line 145: 5xx level errors should be retried.  

```suggestion
            if(response.code() >= 400 && response.code() <= 499) {
```

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
```diff
@@ -0,0 +1,45 @@
...
+
+class RequestRetry(private val maxExponent: Int = 8,
+                    private val jitterFactor: Int = 100,
+                    private val maxAttempts: Int = 3) {
+
+    private var numberOfAttempts = 0
+
+
+        fun <T> retry(single: Single<T>, skipExceptions: List<Class<out Throwable>?>): Single<T> {
+            return Single.create { emitter -> call(single, emitter, 0, skipExceptions) }
```
richardmcclellan, line 19: As this class is currently designed, you will need to create an instance of `RequestRetry` for each `Single` that you want to retry.   This is because `numberOfAttempts` is only set to 0 when the class is instantiated, and is never reset after that.  This is not a great design because it might accidentally get used to retry more than one thing (you actually are doing this in the `SyncProcessor` currently).   I think you can solve this by removing the `numberOfAttempts` variable, and just adding a `numAttempts` to the `call(...)` method.  The first invocation will set it to `MAX_ATTEMPTS`, like this:
```suggestion
            return Single.create { emitter -> call(single, emitter, 0, MAX_ATTEMPTS, skipExceptions) }
```

Then, inside the `call` method, when you recursively call it again, decrement `numAttempts` by 1.

aws-datastore/build.gradle
```diff
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
```
richardmcclellan, line 19: Please make sure to add `ktlint` and ensure it is enabled.  Refer to core-kotlin/build.gradle to see how.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
poojamat: Done

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
poojamat: Done

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+            }
+    }
+
+
+    fun jitteredDelay(): Long {
+        val waitTimeSeconds: Long =
+            2.0.pow((numberOfAttempts % maxExponent)).toLong()
+        +jitterFactor * Math.random()
+
+        return TimeUnit.MILLISECONDS.toMillis(waitTimeSeconds)
```
richardmcclellan, line 48: I think this is converting milliseconds to milliseconds, which is basically not doing anything.  I'd suggest making sure all variable and method names include the units in the name itself, for clarity.  So, change `jitteredDelay()` to `jitteredDelayMillis()` or jitteredDelaySeconds()`, depending on what it is returning.

core/src/main/java/com/amplifyframework/datastore/ErrorType.kt
poojamat: Done

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
poojamat: Great catch, thanks!

aws-analytics-pinpoint/build.gradle
```diff
@@ -16,6 +16,7 @@
 apply plugin: 'com.android.library'
 apply from: rootProject.file("configuration/checkstyle.gradle")
 apply from: rootProject.file("configuration/publishing.gradle")
+apply plugin: 'kotlin-android'
```
frimfram, line 19: Although I am a big fan of Kotlin, I do agree with others that we probably need more planning before including Kotlin on amplify-android.  There are many developer customers out there with various app environments with varying degrees of Kotlin integration in their apps, some of which may not tolerate sudden inclusion of Kotlin plugin in their app.  So I'm thinking that we probably need to coordinate with PM to do at least some pre-announcement of upcoming Kotlin inclusion and increase major version number.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RequestRetry.kt
poojamat: Done.

aws-analytics-pinpoint/build.gradle
poojamat: Are you pointing out to Kotlin plugin in this gradle or any kotlin class?

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -142,6 +142,9 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
                     return;
                 }
             }
+            if(response.code() >= 400 && response.code() <= 499) {
+                onFailure.accept(new ApiException.NonRetryableException("OkHttp client request failed.", "Irrecoverable error"));
+            }
```
rjuliano, line 147: Probably need to add this to `MultiAuthAppSyncGraphQLOperation`. Looking at that code, I think it would probably have to be right before [this else statement](https://github.com/aws-amplify/amplify-android/blob/05715041669c97d4e39f3c5c1d75045e771f151e/aws-api/src/main/java/com/amplifyframework/api/aws/MultiAuthAppSyncGraphQLOperation.java#L182-L184)

rjuliano: Actually, we may even want to do it as a follow up PR given the added complexity of the multiauth logic.

settings.gradle
```diff
@@ -32,3 +32,7 @@ include ':testmodels'
 include ':core-kotlin'
 include ':rxbindings'
 include ':aws-api-appsync'
+
+
+include ':sample-onCall'
```
rjuliano, line 37: This may have already been discussed, but I didn't see a resolution in the conversation below. Are we going to include a sample project?

rjuliano (commented):
Left a couple of minor comments regarding formatting (not sure if it may just be how the Kolin linter works).

Let me know what you think about the comment regarding multiauth.

PM 8/04/2021: Converted Kotlin files to java and added a task for multiauth handling.

settings.gradle
poojamat: My bad, will remove it.

rjuliano (approved):
LGTM 🚀

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
poojamat: Created a task for it.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -142,6 +143,11 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
                     return;
                 }
             }
+            if (response.code() >= START_OF_CLIENT_ERROR_CODE && response.code() <= END_OF_CLIENT_ERROR_CODE) {
+                onFailure.accept(new ApiException
+                        .NonRetryableException("OkHttp client request failed.", "Irrecoverable error")
+                );
+            }
```
frimfram, line 150: Shouldn't the client have access to jsonResponse?  Maybe we can call onFailure instead of throwing exception?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
```diff
@@ -118,10 +118,11 @@ public static AppSyncClient via(@NonNull GraphQLBehavior api,
             }
         };
         final Consumer<ApiException> failureConsumer =
-            failure -> onFailure.accept(new DataStoreException(
-                "Failure performing sync query to AppSync.",
-                failure, AmplifyException.TODO_RECOVERY_SUGGESTION
-            ));
+            failure -> {
+                onFailure.accept(new DataStoreException(
```
frimfram, line 122: The Java lambda style guideline usually recommend not using {} for single line.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -264,15 +265,12 @@ private void transitionToApiSync() throws DataStoreException {
...
-            boolean subscribed = mutationOutbox.load()
+            mutationOutbox.load()
                 .andThen(Completable.create(emitter -> {
                     storageObserver.startObservingStorageChanges(emitter::onComplete);
                     LOG.info("Setting currentState to LOCAL_ONLY");
                     currentState.set(State.LOCAL_ONLY);
-                })).blockingAwait(LOCAL_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-            if (!subscribed) {
-                throw new TimeoutException("Timed out while preparing local-only mode.");
-            }
```
frimfram, line 275: Again, we need to be careful about throwing exceptions since they can lead to app crashes.  When would this be invoked?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
```diff
@@ -292,7 +290,6 @@ private void stopObservingStorageChanges() {
 
     /**
      * Start syncing models to and from a remote API.
-     * @return A Completable that succeeds when API sync is enabled.
```
frimfram, line 295: is this to be deleted?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
```diff
@@ -0,0 +1,104 @@
...
+            List<Class<? extends Throwable>> skipExceptions) {
+        single.delaySubscription(delayInSeconds, TimeUnit.SECONDS)
+                .subscribe(emitter::onSuccess, error -> {
+                    if (attemptsLeft == 0 || ErrorInspector.contains(error, skipExceptions)) {
+                        emitter.onError(error);
+                    } else {
+                        call(single, emitter, jitteredDelaySec(attemptsLeft),
+                                attemptsLeft - 1, skipExceptions);
+                    }
+                });
```
frimfram, line 88: How do you make sure that the flow was not canceled/disposed (which can lead to client app crashes)?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param attemptsLeft number of attempts left.
+     * @return delay in seconds.
+     */
+    long jitteredDelaySec(int attemptsLeft) {
+        int numAttempt = maxAttempts - (maxAttempts - attemptsLeft);
+        double waitTimeSeconds =
+                Math.pow(2, ((numAttempt) % maxExponent))
+                        + jitterFactor * Math.random();
+        return (long) waitTimeSeconds;
+    }
```
frimfram, line 103: Have you thought about also adding max time?  (in addition to max tries)?  I can see cases that the total retry time can be unreasonably high and in that case we should stop retrying.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
poojamat: The exception and the since the timeout are removed from here  , it was thrown earlier.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
poojamat: We are converting the client error code family to non retryable exception which is passed in to on failure call back. The exception is not thrown here.

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncClient.java
poojamat: Good catch.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
poojamat: Thanks for pointing this out. The single passed in this retry method is added to the [setDisposable on the emitter.](https://github.com/aws-amplify/amplify-android/pull/1414/files/a8f2d4d2be382af53d1045b95a25f57c531c09e6#:~:text=emitter.setDisposable(AmplifyDisposables.fromCancelable(cancelable))%3B)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java
poojamat: I wonder why is this change showing up. I did not make this change and this change doesn't show up on android studio either. Moreover this method is void so its not valid anyways.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
poojamat: Good idea. I added the max delay based on [Java script version of retry](https://github.com/aws-amplify/amplify-js/blob/df95ea3724eb6406f64b03f25086cd3e8644cb5f/packages/core/src/Util/Retry.ts#:~:text=function%20jitteredBackoff(maxDelayMs,args%2C%20jitteredBackoff(maxDelayMs))%3B) the max time is 300s. I feel that's too long. Any suggestion?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
```diff
@@ -0,0 +1,109 @@
...
+    private <T> void call(
+            Single<T> single,
+            SingleEmitter<T> emitter,
+            Long delayInSeconds,
+            int attemptsLeft,
+            List<Class<? extends Throwable>> skipExceptions) {
+        single.delaySubscription(delayInSeconds, TimeUnit.SECONDS)
+                .subscribe(emitter::onSuccess, error -> {
+                    if (attemptsLeft == 0 || ErrorInspector.contains(error, skipExceptions)) {
+                        emitter.onError(error);
```
frimfram, line 88: I believe this is the potential crash point - invoking onError here when the subscription is already canceled will lead to crash.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -221,7 +224,7 @@ private SyncTime filterOutOldSyncTimes(SyncTime lastSyncTime) throws DataStoreEx
                 BehaviorProcessor.createDefault(
                         appSync.buildSyncRequest(schema, lastSyncTimeAsLong, syncPageSize, predicate));
 
-        return processor.concatMap(request -> syncPage(request).toFlowable())
+        return processor.concatMap(request -> syncPageWithRetry(request).toFlowable())
```
frimfram, line 227: Have you thought of adding a feature flag to turn on/off retry?  It would be useful to quickly disable the feature or test this in subset of users.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
poojamat: So when the subscription cancels, this entire single gets disposed. As demonstrated by the test above. If you have time I can show you how.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
poojamat: Done.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/RetryHandler.java
poojamat: Added isDisposed check on the emitter before retrying and emitting error.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
poojamat: The android amplify implementation doesn't support feature flags at this time but we can make enable retry configurable by passing it as a property on the builder in the constructor.

poojamat: Added syncRetry enable as a property in builder which can be configured through constructor.

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -93,14 +95,16 @@ private AWSDataStorePlugin(
         this.sqliteStorageAdapter = SQLiteStorageAdapter.forModels(modelSchemaRegistry, modelProvider);
         this.categoryInitializationsPending = new CountDownLatch(1);
         this.authModeStrategy = AuthModeStrategyType.DEFAULT;
+        this.isSyncRetryEnabled = false;
```
rjuliano, line 98: Wouldn't you want to set this to `userProvidedConfiguration.doSyncRetry`?

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -129,7 +134,8 @@ private AWSDataStorePlugin(@NonNull Builder builder) throws DataStoreException {
             sqliteStorageAdapter,
             AppSyncClient.via(api, this.authModeStrategy),
             () -> pluginConfiguration,
-            () -> api.getPlugins().isEmpty() ? Orchestrator.State.LOCAL_ONLY : Orchestrator.State.SYNC_VIA_API
+            () -> api.getPlugins().isEmpty() ? Orchestrator.State.LOCAL_ONLY : Orchestrator.State.SYNC_VIA_API,
+                isSyncRetryEnabled
```
rjuliano, line 138:
```suggestion
            isSyncRetryEnabled
```

rjuliano (commented):
Left a couple of comments

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
poojamat: Updated.

rjuliano (approved):
LGTM 🚢

PR #1422, opened 2021-07-21
feat(auth): add options to resendSignUpCode

Jordan-Nelson:
*Issue #, if available:* n/a

*Description of changes:*
- Add ResendSignUpCodeOptions/AWSCognitoResendSignUpCodeOptions to enable passing metadata in resendSignUpCode()


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/options/AWSCognitoAuthResendSignUpCodeOptions.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return a builder object.
+     */
+    @NonNull
+    public static CognitoBuilder builder() {
+        return new CognitoBuilder();
+    }
+
+    @Override
+    public int hashCode() {
+        return ObjectsCompat.hash(
```
richardmcclellan, line 64: `hashCode`, `equals` should have calls to `super()`, right?

Jordan-Nelson: I don't think so. AuthResendSignUpCodeOptions is an abstract class and doesn't implement hashCode/equals. This is the pattern used for other similar options. Let me know if I am missing something.

rxbindings/src/main/java/com/amplifyframework/rx/RxAuthBinding.java
```diff
@@ -82,6 +83,12 @@
             delegate.confirmSignUp(username, confirmationCode, onResult, onError));
     }
 
+    @Override
+    public Single<AuthSignUpResult> resendSignUpCode(
+            @NonNull String username, @NonNull AuthResendSignUpCodeOptions options) {
+        return toSingle((onResult, onError) -> delegate.resendSignUpCode(username, onResult, onError));
```
Jordan-Nelson, line 89:
```suggestion
        return toSingle((onResult, onError) -> delegate.resendSignUpCode(username, options, onResult, onError));
```

frimfram:
Can you please elaborate on what type of testings you have done so that this PR does not introduce regressions?  I see some unit tests are done but we are looking for more testing.

Jordan-Nelson:
Hi @frimfram - For context, I am adding this for use in amplify-flutter.

I have tested an end to end resend sign up code flow from a flutter app with the changes in https://github.com/aws-amplify/amplify-flutter/pull/738 and confirmed that I can see the metadata in a lambda trigger.

If there is something specific you are looking for from a testing perspective, please let me know.

frimfram:
Please make sure to test edge cases.  These are some:
- Backend error handling (list all possible error cases and test each)
- Client error such as the phone having no network or slow network.
- some other edge cases

Jordan-Nelson:
@frimfram - I have tested the following edge cases. I first tested them with amplify android version 1.21.0 to verify the existing behavior and then tested the same scenarios against this PR (with default options and with options + metadata).

If there are other scenarios you would like to see tested, please let me know.


| Scenario                  	        | Expected Result (based on v1.21.0 behavior)           	        | PR w/o metadata 	| PR w/ metadata 	|
|---------------------------	|-------------------------------------------------------	|-----------------	|-------------	|
| No Network                	        | AuthException, cause = Unable to execute HTTP request 	|  ✅             	        |  ✅            	|
| Non-existent Username     	| UserNotFoundException                                 	                 |  ✅             	        |  ✅            	|
| No Username                         | InvalidParameterException                             	                 |  ✅             	        |  ✅            	|
| Already confirmed user    	| InvalidParameterException                             	                 |  ✅             	        |  ✅            	|
| 5+ Attempts in a row      	| LimitExceededException                                	                 |  ✅             	        |  ✅            	|
| Failure in lambda trigger 	| AuthException, cause = UserLambdaValidationException  |  ✅             	        |  ✅            	|

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
```diff
@@ -350,10 +352,17 @@ public void confirmSignUp(
     @Override
     public void resendSignUpCode(
             @NonNull String username,
+            @NonNull AuthResendSignUpCodeOptions options,
```
frimfram, line 355: Isn't it possible to introduce another overloaded function instead of inserting new NonNull param in a public function?  If someone was calling this function, then they would also need to change as a result of this change.  For public SDK's it's best to minimize breaking changes.

core-kotlin/src/main/java/com/amplifyframework/kotlin/auth/Auth.kt
```diff
@@ -86,11 +87,16 @@ interface Auth {
...
+     *                If not provided, default options will be used
      * @return A sign-up result; if the code is requested, typically the result will
      *         include a next step requiring confirmation of the re-sent code.
      */
     @Throws(AuthException::class)
-    suspend fun resendSignUpCode(username: String): AuthSignUpResult
+    suspend fun resendSignUpCode(
+        username: String,
+        options: AuthResendSignUpCodeOptions = AuthResendSignUpCodeOptions.defaults()
+    ): AuthSignUpResult
```
frimfram, line 99: Same concern as the above - can you think of a way to introduce this in a backward compatible way?

frimfram (changes requested):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java
Jordan-Nelson: This PR does introduce a new overloaded function. The diff here is a possibly a little misleading since it shows `resendSignUpCode` being modified. `resendSignUpCode` definied on like 381 below has the same method signature as before, and it calls this newly added method with a default option. This will not be a breaking change.

core-kotlin/src/main/java/com/amplifyframework/kotlin/auth/Auth.kt
Jordan-Nelson: The newly added parameter has a default option. This is backwards compatible.

raphkim (approved):

frimfram (approved):

PR #1428, opened 2021-07-23
feat(api) add CUSTOM case to AuthStrategy

richardmcclellan:
The CLI does not support the `AWS_LAMBDA` auth type yet, but will soon.  When it does, the code-generated models will have this annotation:  

`@ModelConfig(authRules = { @AuthRule(allow = AuthStrategy.CUSTOM, provider = "function") })`

The following table shows the allowed combinations of authorization strategies and providers.

|  | owner | groups | public | private | custom |
| --- | --- | --- | --- | --- | --- |
| userPools | X | X |  | X |  |
| oidc | X | X |  |  |  |
| apiKey |  |  | X |  |  |
| iam |  |  | X | X |  |
| function |  |  |  |  | X | 

This PR adds the CUSTOM case to the `AuthStrategy` enum, which the code generated models will rely on.


frimfram (approved):
LGTM

Issue #1434, opened 2021-07-27
Cannot create query predicated with null value


HuiSF:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line


```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 6.7.1
------------------------------------------------------------

Build time:   2020-11-16 17:09:24 UTC
Revision:     2972ff02f3210d2ceed2f1ea880f026acfbab5c0

Kotlin:       1.3.72
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.8 compiled on May 10 2020
JVM:          1.8.0_292 (AdoptOpenJDK 25.292-b10)
OS:           Mac OS X 10.16 x86_64

```

</details>

### Please include any relevant guides or documentation you're referencing

Original customer issue https://github.com/aws-amplify/amplify-flutter/issues/511

### Describe the bug

When query data from datastore using null value, amplify-android throws exception and query fails.

Example:
```java
Amplify.DataStore.query(Post.class,
        Where.matches(Post.PUBLISHED.eq(null)),
        queryMatches -> {
            if (queryMatches.hasNext()) {
                Log.i("MyAmplifyApp", "Successful query, found posts.");
            } else {
                Log.i("MyAmplifyApp", "Successful query, but no posts.");
            }
        },
        error -> Log.e("MyAmplifyApp", "Error retrieving posts", error)
);
```

Schema: 
```graphql
type Post @model @key(name: "byBlog", fields: ["blogID"]) {
  id: ID!
  title: String!
  rating: Int!
  created: AWSDateTime
  blogID: ID!
  blog: Blog @connection(fields: ["blogID"])
  comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
  state: PostState
  published: Boolean
}
```
SQLite table and data
<img width="970" alt="Screen Shot 2021-07-27 at 9 46 57 AM" src="https://user-images.githubusercontent.com/10602282/127194869-e705c4fe-5a83-4f66-89d6-0b854be0b75f.png">

Exception thrown from amplify-android

```
[...]

PR #1435, opened 2021-07-29
fix: handle null values for predicates

rjuliano:
*Issue #, if available:* #1434

*Description of changes:*
Predicate logic was not handling null values for equal and not equal operations.

This PR addresses this bug bug checking for null values when the operator is EQUAL or NOT EQUAL and constructs the SQL statement accordingly.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
@@ -187,6 +187,21 @@ public final class SQLPredicate {
                         .append("?");
             case EQUAL:
             case NOT_EQUAL:
+                Object operatorValue = getOperatorValue(op);
+                if (operatorValue != null) {
+                    addBinding(operatorValue);
+                    return builder.append(column)
+                        .append(SqlKeyword.DELIMITER)
+                        .append(SqlKeyword.fromQueryOperator(op.type()))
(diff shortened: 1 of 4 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/adapter/SQLPredicate.java
```diff
@@ -187,6 +187,21 @@ private StringBuilder parsePredicateOperation(QueryPredicateOperation<?> operati
...
             case EQUAL:
             case NOT_EQUAL:
+                Object operatorValue = getOperatorValue(op);
+                if (operatorValue != null) {
+                    addBinding(operatorValue);
+                    return builder.append(column)
+                        .append(SqlKeyword.DELIMITER)
+                        .append(SqlKeyword.fromQueryOperator(op.type()))
+                        .append(SqlKeyword.DELIMITER)
+                        .append("?");
```
frimfram, line 197: Since there is no "break" statement, you don't need to repeat this code here and just check for special case using if statement.  Codepath the does not fall into if statement will continue executing lines at 209.

rjuliano: I like it!!! Making that change right now

poojamat (approved):

poojamat (approved):

rjuliano:
Integration test failing for `createWaitThenUpdate10TimesWithPredicate`. Need to investigate

PR #1440, opened 2021-08-04
fix(api): expose selectionSet in request builder

raphkim:
*Description of changes:*
Selection set could not be specified in `AppSyncGraphQLRequest` could not be set by the builder because its setter was never exposed. Added public `AppSyncGraphQLRequest.Builder#selectionSet(SelectionSet)` method.


sdhuka (approved):
LGTM!

Issue #1273, opened 2021-04-07
App Crash when trying to save item to DataStore offline


rommank:
If I go offline and try to save an item the app crashes.

aws-datastore: 1.17.2

```
    private fun addDevice() {
        val device = Device.builder()...
            .build()

        Amplify.DataStore.save(device,
            {},
            {
                Log.e("addDevice", "AddDevice error: ", it)
            }
        )
    }
```

```
FATAL EXCEPTION: OkHttp https://fnguccyr6nhmtdnniyxq6xpzte.appsync-api.eu-central-1.amazonaws.com/...
    Process: net.escalibre.myapplication, PID: 29980
    io.reactivex.rxjava3.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | DataStoreException{message=Failure during mutation., cause=ApiException{message=OkHttp client request failed., cause=java.net.SocketTimeoutException: timeout, recoverySuggestion=See attached exception for more details.}, recoverySuggestion=Check details.}
        at io.reactivex.rxjava3.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
        at io.reactivex.rxjava3.internal.operators.single.SingleCreate$Emitter.onError(SingleCreate.java:82)
        at com.amplifyframework.datastore.syncengine.-$$Lambda$H2kP2N9w3X0gYQg1QzSwv0s9uXc.accept(Unknown Source:4)
        at com.amplifyframework.datastore.appsync.AppSyncClient.lambda$mutation$5(AppSyncClient.java:328)
        at com.amplifyframework.datastore.appsync.-$$Lambda$AppSyncClient$FfgHx1XvivjmTcZhtf9BjbrAq_M.accept(Unknown Source:4)
        at com.amplifyframework.api.aws.AppSyncGraphQLOperation$OkHttpCallback.onFailure(AppSyncGraphQLOperation.java:156)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:525)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
[...]

PR #1441, opened 2021-08-04
fix: check for canceled call to fix RxJava crash

frimfram:
*Issue #, if available:*
#1273 

*Description of changes:*
Fix crash due to attempting to invoke client of already canceled call.
- added  isCanceled() check in start to guard against calls that were canceled but still in queue
- added isCanceled() check in failure callback (for calls that hit 20sec socket timeout)

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
@@ -86,8 +86,8 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
 
     @Override
     public void start() {
-        // No-op if start() is called post-execution
-        if (ongoingCall != null && ongoingCall.isExecuted()) {
+        // No-op if start() is called post-execution or canceled
+        if (ongoingCall != null && (ongoingCall.isExecuted() || ongoingCall.isCanceled())) {
             return;
         }
 
@@ -153,9 +153,11 @@ public final class AppSyncGraphQLOperation<R> extends GraphQLOperation<R> {
 
         @Override
         public void onFailure(@NonNull Call call, @NonNull IOException exception) {
-            onFailure.accept(new ApiException(
-                "OkHttp client request failed.", exception, "See attached exception for more details."
-            ));
+            if (!call.isCanceled()) {
+                onFailure.accept(new ApiException(
+                        "OkHttp client request failed.", exception, "See attached exception for more details."
+                ));
+            }
         }
     }
 


aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -86,8 +86,8 @@ private AppSyncGraphQLOperation(
 
     @Override
     public void start() {
-        // No-op if start() is called post-execution
-        if (ongoingCall != null && ongoingCall.isExecuted()) {
+        // No-op if start() is called post-execution or canceled
+        if (ongoingCall != null && ((ongoingCall.isExecuted()) || ongoingCall.isCanceled())) {
```
raphkim, line 90:
```suggestion
        if (ongoingCall != null && (ongoingCall.isExecuted() || ongoingCall.isCanceled())) {
```
nitpicking, but there was an extra set of parentheses there :)

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
```diff
@@ -153,9 +153,11 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
...
         @Override
         public void onFailure(@NonNull Call call, @NonNull IOException exception) {
-            onFailure.accept(new ApiException(
-                "OkHttp client request failed.", exception, "See attached exception for more details."
-            ));
+            if (!call.isCanceled()) {
+                onFailure.accept(new ApiException(
+                        "OkHttp client request failed.", exception, "See attached exception for more details."
+                ));
+            }
```
raphkim, line 160: what is the expected behavior if the call is canceled? Will there be an appropriate callback?

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
sdhuka, line 90: Is this added for precaution as the crash logs points that crash originates from onFailure.

sdhuka (approved):
LGTM except one clarification.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
frimfram: The callback after the canceled call leads to Crash - ( UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow )

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
frimfram: Thanks, will fix.

frimfram: This line is not in the crash log, but isCanceled should also be checked here since the canceled call should not be started...

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLOperation.java
raphkim: discussed offline that canceled event is already handled at higher level

raphkim (approved):
LGTM 🚢

PR #1452, opened 2021-08-16
feat(geo): add Maps API

raphkim:
* Added two APIs to Geo interface and implemented them:
  * `getAvailableMaps()`
  * `getDefaultMap()`

Dependent on #1451 


aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.json.JSONObject
+
+/**
+ * A plugin for the Geo category to interact with Amazon Location Service.
+ */
+class AWSLocationGeoPlugin(
+    private val userConfiguration: GeoConfiguration? = null // for programmatically overriding amplifyconfiguration.json
+): GeoCategoryPlugin<AmazonLocationClient?>() {
+    companion object {
+        private const val GEO_PLUGIN_KEY = "awsLocationGeoPlugin"
```
rjuliano, line 35: For constants it looks like you have to use `@JvmField` annotation.

https://developer.android.com/kotlin/interop#companion_constants

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
```diff
@@ -0,0 +1,83 @@
...
+    override fun getVersion(): String {
+        return BuildConfig.VERSION_NAME
+    }
+
+    override fun getAvailableMaps(
+        onResult: Consumer<Collection<MapStyle>>,
+        onError: Consumer<GeoException>
+    ) {
+        try {
+            onResult.accept(pluginConfiguration.maps!!.items)
```
rjuliano, line 67: Can `pluginConfiguration.maps` be null? 

You may be able to do something like:

```kotlin
pluginConfiguration.maps?.let { onResult.accept(it.items) }
```

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
```diff
@@ -0,0 +1,83 @@
...
+            onError.accept(Errors.mapsError(error))
+        }
+    }
+
+    override fun getDefaultMap(
+        onResult: Consumer<MapStyle>,
+        onError: Consumer<GeoException>
+    ) {
+        try {
+            onResult.accept(pluginConfiguration.maps!!.default)
```
rjuliano, line 78: Same here

rjuliano (approved):
A couple minor comments for your review. Other than that, it looks good 🚢

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
raphkim: it's a private constant, so it won't be exposed

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
raphkim: yep! i imagine some people may choose not to use the maps API and not configure it to begin with

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
raphkim: same as above

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
raphkim: Using `let` will simply return null rather than triggering an error as it should, which is why I simply used `!!` instead.

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
rjuliano: Doh!!! 🤣

PR #1479, opened 2021-09-03
announce which schema is failing to sync

joekiller:
*Issue #, if available:* N/A

*Description of changes:* Announce which schema is failing to sync

We noticed our datastore "silently failing over time" and observed that one sync processor wasn't ever connecting due to the sync query taking too long. The datastore would go into local mode and allow transactions however because they were never syncing and never were going to sync, our application chugged along until around 11000 - 12000 PersistentRecords were jammed in our mutationOutbox. At that point the android out of memory process would kick in and kill our application. The mutationOutbox would then attempt to load the 10k+ records and take forever causing the dreaded, "Failed to start DataStore." forever if you are lucky and have only like 3000 pending mutations. If you are unlucky and have accrued 11000 well the application will never start because the outbox will exceed the processed memory.

`java.net.SocketTimeoutException: timeout` and the bland `Initial cloud sync failed.` errors
would have made more sense if we hadn't had 23 different schemas. (Don't start about that). Once I added this little extra information to the output it was obvious WHICH sync was failing.

I also observed that the okHttpClient auth defaults to timeout after 10 seconds. We needed to update our Api configuration to match the AppSync api maximum of 30 seconds. That allowed our sync query to work for now.

```java
Amplify.addPlugin(AWSApiPlugin.builder().configureClient("AppSyncAPI", okHttpBuilder -> {
      okHttpBuilder.readTimeout(30, TimeUnit.SECONDS).writeTimeout(30, TimeUnit.SECONDS);
  }).build());
```

I was tempted to just set the [default api endpoint OkHttpBuilder](https://github.com/aws-amplify/amplify-android/blob/567ac2895e07619a1c47c0d0227febb874bce787/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java#L139) to have it's read and write timeout as 30 in this PR but perhaps 10 seconds is good b/c it means your access pattern is likely unsustainable.

We were also waiting for SYNC_QUERIES_READY instead of READY because of "READY" in the past not ever coming in. I think in hindsight this was due to the timeout however that wasn't apparent. Anyway, SYNC_QUERIES_READY is kinda dangerous and can lead to the app working for a long time and then getting crushed by the stacked up mutations.


joekiller:
I'll fix the test

joekiller:
Hi I cannot actually see the error or run the integration tests due to not being part of the access list.

jamesonwilliams (approved):

raphkim (approved):
LGTM! 🚢 
Looks like the tests were failing from internal issues unrelated to this PR. I will merge it in once the tests are passing again.

Issue #1231, opened 2021-03-15
Can't cancel subscription (AppSync)


AlbertoSinigaglia:
Hi, I'm trying to close a subscription inside it's handler:
``` 
var sub : GraphQLOperation<Player>? = null
sub = Amplify.API.subscribe(
        ModelSubscription.onCreate(Model::class.java),
        { Log.i("ApiQuickStart", "Subscription established") },
        { onCreated ->
            if (something) {
                ...
                sub?.cancel()
            } 
        },
        { onFailure -> Log.e("ApiQuickStart", "Subscription failed", onFailure) },
        { Log.i("ApiQuickStart", "Subscription completed") }
)
```
However, no matter what I try, when i get to that line, I receive this exception:
```
E/ApiQuickStart: Subscription failed
    ApiException{message=Subscription completion not acknowledged., cause=null, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}
        at com.amplifyframework.api.aws.SubscriptionEndpoint$Subscription.awaitSubscriptionCompleted(SubscriptionEndpoint.java:376)
        at com.amplifyframework.api.aws.SubscriptionEndpoint.releaseSubscription(SubscriptionEndpoint.java:258)
        at com.amplifyframework.api.aws.SubscriptionOperation.cancel(SubscriptionOperation.java:105)
        ...
        at com.amplifyframework.api.aws.SubscriptionEndpoint$Subscription.dispatchNextMessage(SubscriptionEndpoint.java:392)
        at com.amplifyframework.api.aws.SubscriptionEndpoint.notifySubscriptionData(SubscriptionEndpoint.java:226)
        at com.amplifyframework.api.aws.SubscriptionEndpoint.access$700(SubscriptionEndpoint.java:60)
        at com.amplifyframework.api.aws.SubscriptionEndpoint$AmplifyWebSocketListener.processJsonMessage(SubscriptionEndpoint.java:573)
        at com.amplifyframework.api.aws.SubscriptionEndpoint$AmplifyWebSocketListener.onMessage(SubscriptionEndpoint.java:476)
        at okhttp3.internal.ws.RealWebSocket.onReadMessage(RealWebSocket.kt:333)
        at okhttp3.internal.ws.WebSocketReader.readMessageFrame(WebSocketReader.kt:245)
        at okhttp3.internal.ws.WebSocketReader.processNextFrame(WebSocketReader.kt:106)
[...]

PR #1482, opened 2021-09-03
fix(api): prevent thread blocking on subx cancellation

rjuliano:
*Issue #, if available:*
#1231 

*Description of changes:*
I believe the behavior outlined in the above issue is being caused by the fact that cancelling a subscription on the same thread receiving a message causes that thread to be blocked while we wait for the cancellation confirmation from the server. To prevent this behavior, I'm just executing the releaseSubscription call in a separate thread of the executor.

This only happens when trying to close the subscription from inside the onNext callback of the subscribe call. The workaround is to make the call in a separate thread, but I guess we can avoid this scenario altogether by having that in the library.


rjuliano:
Integration test is failing on createWaitThenUpdateMultipleTimes. Need to investigate

sdhuka (approved):

eeatonaws (approved):

PR #1497, opened 2021-09-15
feat(geo): add convenience method for setting map style

raphkim:
Added `setStyle` convenience method that sets map style on maplibre map instance using correct thread types for different tasks.

Previously, user would have to:
1. launch the map getter (`MapView#getMapAsync()`) from the _main UI thread_, 
2. fetch style from the worker thread (`Geo#getMapStyleDescriptor()`) from the _worker thread_,
3. and then apply style on the map (`MapboxMap#setStyle()`) from the _main UI thread_

This resulted in making one simple step of loading a style on the map quite complicated. This PR adds a service executor to the plugin to enforce that network requests are made from a worker thread, and adds a convenience method that behaves like MapLibre's `setStyle()` method.

### DIFF
--- a/aws-geo-location/build.gradle
+++ b/aws-geo-location/build.gradle
@@ -21,7 +21,6 @@ apply from: rootProject.file("configuration/publishing.gradle")
 dependencies {
     implementation project(':core')
     implementation dependency.aws.location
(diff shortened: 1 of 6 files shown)


rjuliano (approved):
LGTM

Issue #923, opened 2020-10-23
Updating tables on model schema change fails for models with foreign keys


saltonmassally:
Amplify refuses to initialize when I upgrade my schema. Digging deeper it appears to be because the delete statement in `com.amplifyframework.datastore.storage.sqlite.SQLiteStorageHelper#dropAllTables` fails on trying to delete models with foreign keys. In my case, it could not find the associated table that the foreign key points to. I got around this by issuing `sqliteDatabase.execSQL("PRAGMA foreign_keys = OFF;");` before the try block and turning it back on after.

PR #1501, opened 2021-09-17
Fix for issue with foreign keys on schema upgrade delete

poojamat:
*Issue #923  , if available:*

*Description of changes:*
Customer was having problem with deleting the tables with foreign key. This fix involves turning foreign key check before deleting and turning it back on after. 


raphkim (approved):
I think a more elegant solution would have been to [topologically sort](https://en.wikipedia.org/wiki/Topological_sorting) the tables by relationship to delete children  models before parents... but this works too since the goal is to just drop every table anyways. LGTM!

PR #1506, opened 2021-09-21
chore: add sync expression to multi-auth test

rjuliano:
*Issue #, if available:*

*Description of changes:*
Some of the multi-auth tests were failing sporadically likely due to the amount of data being sync'd. The sync expression should prevent that.

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
@@ -39,7 +39,6 @@ import com.amplifyframework.core.model.ModelSchema;
 import com.amplifyframework.core.model.SchemaRegistry;
 import com.amplifyframework.core.model.SerializedModel;
 import com.amplifyframework.core.model.query.Where;
-import com.amplifyframework.core.model.query.predicate.QueryPredicate;
 import com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter;
 import com.amplifyframework.datastore.storage.sqlite.TestStorageAdapter;
 import com.amplifyframework.hub.HubChannel;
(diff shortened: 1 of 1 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
```diff
@@ -807,12 +810,18 @@ private void configure(Class<? extends Model> modelType,
...
+                                                    Log.e(tag,
+                                                          "DataStore error handler received an error.",
+                                                          exception);
+                                                })
+                                                .syncExpression(modelSchema.getName(), new DataStoreSyncExpression() {
+                                                    @Override
+                                                    public QueryPredicate resolvePredicate() {
+                                                        return Where.id("FAKE_ID").getQueryPredicate();
+                                                    }
+                                                })
```
raphkim, line 823: can't we use lambda here?
```
.syncExpression(modelSchema.getName(), () -> Where.id("FAKE_ID").getQueryPredicate())
```

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
```diff
@@ -807,12 +810,18 @@ private void configure(Class<? extends Model> modelType,
...
-                                                                    Log.e(tag,
-                                                                          "DataStore error handler received an error.",
-                                                                          exception);
-                                                                })
-                                                                .build();
+                                                .errorHandler(exception -> {
+                                                    Log.e(tag,
+                                                          "DataStore error handler received an error.",
+                                                          exception);
+                                                })
```
raphkim, line 817: we don't need `{}` for single liners :)
```
.errorHandler(error -> Log.e(tag, message, error))
```

but even better, wouldn't a Hub publish be more useful here for testing?

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
raphkim: `Where.id()` helper method also has a bug with ambiguous column names IIRC. I guess it's fine unless we start writing multiauth tests that has model relationship

raphkim (approved):
Few comments here and there, but looks good otherwise.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
rjuliano: Yeah...I think for the purposes of this test, we're using a single table for each test case so we should be OK

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/MultiAuthSyncEngineInstrumentationTest.java
rjuliano: Updated it to be a lambda expression. I only added this in here to make sure we could pick up on any exceptions thrown during the test. We can explore some alternatives as part of a different PR though.

PR #1513, opened 2021-09-28
fix(datastore): Ensure not to parse SerializedCustomType if value is …

HuiSF:
…null

*Issue #, if available:*

*Description of changes:*

Missed `null` value check before invoking `CustomType` field value serialization/deserialization logic.
This PR is for adding the necessary `null` value check.

----

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
@@ -875,7 +875,7 @@ public final class SQLiteStorageAdapter implements LocalStorageAdapter {
                 continue;
             }
 
-            if (field.isCustomType()) {
+            if (field.isCustomType() && entry.getValue() != null) {
                 if (field.isArray()) {
                     @SuppressWarnings("unchecked")
                     List<Map<String, Object>> listItems = (List<Map<String, Object>>) entry.getValue();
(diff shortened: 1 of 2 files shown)


poojamat (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java
```diff
@@ -875,7 +875,7 @@ private SerializedCustomType createSerializedCustomType(
                 continue;
             }
 
-            if (field.isCustomType()) {
+            if (field.isCustomType() && entry.getValue() != null) {
```
raphkim, line 878: What is the field was defined as `nullable`? Shouldn't the serialized custom type still have the field included in that case?

HuiSF: Ahh yes, thanks for catching it. Will update.

HuiSF: Thanks for looking!
If `field.isCustomType() && entry.getValue() != null` is false it should fallback to line 895, which should be able to handle the nullable field case.

raphkim: I must have missed that :)

raphkim (approved):
LGTM! 🚢

PR #1561, opened 2021-11-02
add AmplifyMapView to encapsulate the map initialization logic

drochetti:
**Notes:**

- turn Adapter into an instantiable type to facilitate testing / dependency injection
- reduce the boilerplate for MapView to work

*Issue #, if available:*

*Description of changes:*


maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/http/AWSRequestSignerInterceptor.kt
```diff
@@ -0,0 +1,90 @@
...
+        this.header(name, value)
+    }
+    return this.url(urlBuilder.build())
+}
+
+/**
+ * Interceptor that can authorize requests using AWS Signature V4 signer.
+ */
+internal class AWSRequestSignerInterceptor(
+    private val plugin: AWSLocationGeoPlugin
```
raphkim, line 50: Wouldn't it make more sense to pass the service name directly rather than passing an entire plugin?

raphkim (approved):
Few comments, but LGTM!

maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/AmplifyMapLibreAdapter.kt
```diff
@@ -16,70 +16,45 @@
...
- * Entry point to initialize MapLibre instance and configure it to be
- * authorized to make requests directly to Amazon Location Service.
- */
-object AmplifyMapLibreAdapter {
-    private const val COGNITO_AUTH_PLUGIN_KEY = "awsCognitoAuthPlugin"
-    private const val GEO_SERVICE_NAME = "geo"
+private const val LOCATION_GEO_PLUGIN_KEY = "awsLocationGeoPlugin"
 
-    private val log = Amplify.Logging.forNamespace("amplify:maplibre-adapter")
+class AmplifyMapLibreAdapter(
```
raphkim, line 36: I understand the benefits of having the adapter be instantiable, but my only concern is that `Mapbox` itself is a singleton. Individually instantiable adapter might erroneously lead developers to think that they can have multiple differently configured adapters when in fact each one will be using the exact same copy of Mapbox instance.

drochetti: Developers never instantiate the adapter directly, the MapView component itself does. I think the open question here is if the adapter should even be exposed directly.

This loose dependency between an adapter that needs to be setup before the component is initialized is error-prone and unnecessary. My proposed solution fixes that.

What we can do is to make the constructor internal and expose a static singleton method.

maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/http/AWSRequestSignerInterceptor.kt
drochetti: The signer comes from the plugin itself, in this refactor the dependency is resolved from the Geo category passed to the MapView, which defaults to `Amplify.Geo`.

The interceptor cannot work without the correct plugin, so I think that creating an explicit dependency between both not only favors the actual signer logic (we get whichever signer implementation the SDK provides, even if it changes in the future) but also removes a logic that can fail, where you pass a service name as a string, but your setup is not correct. So although you can sign URLs, the MapView won't work because the plugin hasn't been correctly initialized.

drochetti:
This PR will be closed for now since we will keep the MapView code in the `feature/amplify-mapview` branch until it reaches stability. Future PRs will be diffed against that branch

PR #1562, opened 2021-11-04
fix(api): GraphQL operation is now launched from a new thread

raphkim:
*Issue #, if available:*

*Description of changes:*
`AppSyncGraphQLOperation` used to make a network call from the thread it was called in. This caused network request for getting request credentials to fail when called from main thread. This was addressed by dispatching request from a new thread.


gpanshu (approved):

div5yesh (approved):
LGTM

PR #1564, opened 2021-11-07
feat(api): Add support for custom GraphQL endpoints.

sktimalsina:
*Issue #, if available:*

*Description of changes:*
- Support custom declared graphql endpoints in configuration.
- Use region defined in configuration instead of parsing from endpoint URL.


aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
```diff
@@ -0,0 +1,66 @@
...
+     * Standard domain type composed of randomly generated 26 char subdomain.
+     */
+    STANDARD,
+
+    /**
+     * Custom domain defined by user. It must be "https" and have path "graphql" in URL.
+     */
+    CUSTOM;
+
+    private static final String STANDARD_ENDPOINT_REGEX =
```
raphkim, line 40: is the endpoint format spec documented somewhere?

sdhuka (approved):
LGTM!

aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
```diff
@@ -0,0 +1,66 @@
...
+            "^https:\\/\\/\\w{26}\\.appsync\\-api\\.\\w{2}(?:(?:\\-\\w{2,})+)\\-\\d\\.amazonaws.com\\/graphql$";
+
+    /**
+     * Get Domain type based on defined endpoint.
+     * @param endpoint Endpoint defined in config.
+     * @return {@link DomainType} based on supplied endpoint.
+     */
+    static DomainType from(String endpoint) {
```
raphkim, line 48: why not make this method throw `IllegalArgumentException` if format is invalid and remove `INVALID` enum value?

aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
```diff
@@ -0,0 +1,66 @@
...
+        if (!(endpoint.startsWith("https://") && endpoint.endsWith("/graphql"))) {
+            return INVALID;
+        }
```
lawmicha, line 51: iOS isn't doing any schema or path checks, this seems like safe validation to add though. should other platforms be doing this? and would either of these two validation checks be changed in the future? What if the endpoint comes in as `wss://` already? what if the path is ever changed to "/graphql2" by AppSync?

aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
sktimalsina: This makes sense now as it is eventually thrown in only place it is accessed.

aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
sktimalsina: These validations has been confirmed with the appsync. I agree with you argument that the path might be changed and even though that would mean the requirements will be changed, having this validation removed will simply mean that it might fail in other locations.
Endpoints should not be sent as `wss://` to this method as per the definition and scope of this particular method. If there is any such usecase, this method can be called by replacing it with `https://` before passing as a parameter.

aws-api/src/main/java/com/amplifyframework/api/aws/DomainType.java
sktimalsina: Endpoint and protocol for appsync service is outlined here:
https://docs.aws.amazon.com/general/latest/gr/appsync.html
The regex of generating 26 alphanumeric character unique id portion is what is used by appsync team and used here as is.

raphkim: Can we have that link directly commented in code for easy reference?

raphkim (approved):
LGTM!

Issue #1268, opened 2021-04-01
DataStore SQLiteException: Ambiguous Column Name


infernal008:
I have a schema like below:

_I trimmed down most unecessary fields at it seems too verbose_

```
type CategoryObject @model {
  id: ID!
  transactions: [TransactionObject] @connection(keyName: "byCategory", fields: ["id"])
}

type WalletObject @model {
  id: ID!
  transactionsFrom: [TransactionObject] @connection(keyName: "byWalletFrom", fields: ["id"])
  transactionsTo: [TransactionObject] @connection(keyName: "byWalletTo", fields: ["id"])
}

type TransactionObject @model 
  @key(name: "byCategory", fields: ["categoryId"])
  @key(name: "byWalletFrom", fields: ["walletFromId"])
  @key(name: "byWalletTo", fields: ["walletToId"]){
  id: ID!
  categoryId: ID!
  category: CategoryObject @connection(fields: ["categoryId"])
  walletFromId: ID!
  walletFrom: WalletObject @connection(fields: ["walletFromId"])
  walletToId: ID!
  walletTo: WalletObject @connection(fields: ["walletToId"])
}
```
Basically:

- A TransactionObject has one CategoryObject
- A CategoryObject can be used by different TransactionObjects
- A TransactionObject has one WalletObject for walletFrom (in business logic: what wallet was used for the transaction)
- A TransactionObject can have a WalletObject for walletTo (in business logic: in what wallet will the transaction go, optional)
- A WalletObject can be used as either WalletFrom or WalletTo by TransactionObject (based on business logic)

I was able to provision this schema to my AWS instances via the Amplify CLI and was able to check if the models and relationships are working through the AppSync Console. It seems fine, I was able to create objects for all three models and relationships seem fine on query.

However, upon running the command await Amplify.DataStore.query(TransactionObject.classType) locally, I am blocked by a SQLiteException for Ambiguous Column Name:

```
[...]

Issue #1572, opened 2021-11-14
DataStoreException{message=Error in querying the model >> ambiguous column name: User.id


iamAliRaza:
Hi Guyz, I am facing issue while fetching the list of comments from Comment table. I am not sure but i think it's happening because of model relationships.  **Need help !!**

Look at these 3 models. Comment Model has one to one relation with User. And post also has one to one relation with a user.
Comment also has a one to many relation with Post.

Where i am wrong please let me know??

```
type User @model  @auth(rules: [{ allow: private }]) {
  id: ID!
  email: String!
  username: String
  phoneNumber:String
  sex:String
  state:String
  city:String
  zipcode:String
  interest:String
  status:String
  userImage:String
  professions:[String]
  danceStyles:[String]
}

type Post @model @auth(rules: [{ allow: private }]){
    id:ID!
    postImage:String
    caption:String
    date:String
    startTime:String
    endTime:String
    cost:Int
    paymentLink:String
    location : String
    about : String
    likeList : [String]
    bookmarkList : [String]
    postType : PostType
    categories : [String]
    danceStyles : [String]
    postUser : User @connection
    comments: [Comment] @connection(keyName: "byPost", fields: ["id"])
}


type Comment @model @auth(rules: [{ allow: private }])
@key(name: "byPost", fields: ["postID", "comment"]){
    id : ID!
    comment: String!
    postID: ID!
    post: Post! @connection(fields: ["postID"])
    commentedUser : User @connection
    likeList : [String]
}
```

Error that i am getting (while querying Comment model):

> ambiguous column name: User.id
> 
[...]

PR #1603, opened 2021-12-10
fix(datastore): Alias table names and make column aliases unique in query

eeatonaws:
*Issue #, if available:* #1268

*Description of changes:* Aliases tables and makes the column aliases unique when there are multiple foreign keys of the same model type to address the SQLite ambiguous column exception that occurs.


eeatonaws:
The fix in this PR will work when there are multiple fields of the same model type on a single model (e.g., Call model with Person caller and Person callee fields as in the test added by this PR). For a schema like the one in [issue 1572](https://github.com/aws-amplify/amplify-android/issues/1572), where there will be multiple models of the same type (User in the case of that issue) in the query, this fix will result in the ambiguous column exception no longer being thrown, but the same value will be returned for each model of the same type (the same user would be returned for the user who commented and the user who posted in that issue), which is incorrect. I am working on extending the idea in this PR to handle this case.

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapterQueryTest.java
```diff
@@ -64,21 +68,39 @@ public static void enableStrictMode() {
...
         TestStorageAdapter.cleanup();
-        this.adapter = TestStorageAdapter.create(AmplifyModelProvider.getInstance());
+        this.blogAdapter = TestStorageAdapter.create(AmplifyModelProvider.getInstance());
+    }
+
+    /**
+     * Remove any old database files, and then re-provision a new storage adapter,
+     * that is able to store the Phone Call family of models.
+     */
+    public void setupForCallModel() {
```
sktimalsina, line 84: Is `blogAdapter` and `callAdapter` being used exclusively in each tests? Or in other words, are they used in same test method?
If not, we can avoid having these two instances of `SynchronousStorageAdapter` and multiple renamings. Cheapest way to do it is to call `tearDown` as first call in `setupForCallModel`
Then initialize `adapter` with the `phonecall.AmplifyModelProvider`.
With that, `teardownForCallModel` can also be removed as it will be same as `tearDown`

sktimalsina (approved):

raphkim (approved):

Issue #1606, opened 2021-12-16
aws-api + aws-auth-cognito instantiation failure after updating library versions


djweber:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication, REST API

### Gradle script dependencies

<details>
Old:

```kotlin
object Amplify {
    private const val version = "1.17.5"

    const val api = "com.amplifyframework:aws-api:$version"
    const val cognito = "com.amplifyframework:aws-auth-cognito:$version"

    object Core {
        const val kotlin = "com.amplifyframework:core-kotlin:0.3.0"
    }
}
```
New:
```kotlin
object Amplify {
    private const val version = "1.31.0"

    const val api = "com.amplifyframework:aws-api:$version"
    const val cognito = "com.amplifyframework:aws-auth-cognito:$version"

    object Core {
        const val kotlin = "com.amplifyframework:core-kotlin:0.15.0"
    }
}
```
</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 7.0.2
------------------------------------------------------------

Build time:   2021-05-14 12:02:31 UTC
Revision:     1ef1b260d39daacbf9357f9d8594a8a743e2152e

Kotlin:       1.4.31
Groovy:       3.0.7
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.10 (JetBrains s.r.o 11.0.10+0-b96-7249189)
OS:           Windows 10 10.0 amd64

```

</details>

### Describe the bug

Hello,

I am currently using the Amplify SDKs for API Gateway and Cognito, with Kotlin coroutine support. The Android app is only using `AMAZON_COGNITO_USER_POOLS` for API authorization (we are not using identity pools), and was using an older version of these libs:

```kotlin
object Amplify {
    private const val version = "1.17.5"

    const val api = "com.amplifyframework:aws-api:$version"
    const val cognito = "com.amplifyframework:aws-auth-cognito:$version"

    object Core {
        const val kotlin = "com.amplifyframework:core-kotlin:0.3.0"
    }
}
```

We then updated to the latest versions for each:

```kotlin
object Amplify {
    private const val version = "1.31.0"

[...]

PR #1623, opened 2022-01-06
fix(api): create RequestDecorator in interceptor

eeatonaws:
*Issue #, if available:* #1606

*Description of changes:* When configuring the AWSApiPlugin for REST with `AMAZON_COGNITO_USER_POOLS` as the authorization type, configuration fails due to trying to fetch auth tokens when creating the `RequestDecorator`. This PR moves the creation of the `RequestDecorator` inside the interceptor so the auth tokens aren't retrieved when configuring the AWSApiPlugin.


sktimalsina (approved):

sdhuka (commented):
Were you able to test the change? my concern is that fetchAuth can still fail when the call is intercepted.

sdhuka (approved):

eeatonaws:
With the change, I am able to successfully configure the AWSApiPlugin and make API calls (post, get, etc.).

Issue #1667, opened 2022-02-22
App Crashes When Saving Item to DataStore


rayshi67:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line
def amplify_version = "1.32.0"

implementation "com.amplifyframework:core:$amplify_version"
implementation "com.amplifyframework:aws-api:$amplify_version"
implementation "com.amplifyframework:aws-datastore:$amplify_version"
implementation "com.amplifyframework:aws-auth-cognito:$amplify_version"
implementation "com.amplifyframework:aws-storage-s3:$amplify_version"

```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 7.2
------------------------------------------------------------

Build time:   2021-08-17 09:59:03 UTC
Revision:     a773786b58bb28710e3dc96c4d1a7063628952ad

Kotlin:       1.5.21
Groovy:       3.0.8
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.8 (Oracle Corporation 11.0.8+10-LTS)
OS:           Mac OS X 10.15.7 x86_64


```

</details>

### Describe the bug

My app saves the current location coordinates to datastore every five seconds.

It works as expected for a few minutes. And the saved items get synchronised to the backend successfully. However the app then crashes with exceptions thrown from datastore mutations.

The Android tablet has reasonable internet connection through WI-FI.

Seems the issue #1273 is similar to the above error that I am getting, except the Android tablets used in the testing all have good WI-FI connections. 

### Code Snippet

```java
// Put your code below this line.
Amplify.DataStore.query(
       Taximeter::class.java,
        Where.id(taximeterId),
        { matches ->
                if (matches.hasNext()) {
                       val taximeter = matches.next()
                       val edited = taximeter.copyOfBuilder()
                              .currentLat(location.lat!!)
                              .currentLng(location.lng!!)
                              .build()

                       Amplify.DataStore.save(edited,
                           { Timber.i("Updated latest location to meter") },
                            { Timber.e("Failed to save latest location to meter", it) }
                       )
                 }
          },
[...]

PR #1688, opened 2022-03-25
Connectivity crash fix

poojamat:
*Issue #, if available:*
[1667](https://github.com/aws-amplify/amplify-android/issues/1667)
*Description of changes:*
Fixing the crash on connectivity error caused because rx subscription is already disposed.

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
@@ -316,7 +316,11 @@ final class MutationProcessor {
             @NonNull PublicationStrategy<T> publicationStrategy) {
         return Single
             .<GraphQLResponse<ModelWithMetadata<T>>>create(subscriber ->
-                publicationStrategy.publish(mutation.getMutatedItem(), subscriber::onSuccess, subscriber::onError)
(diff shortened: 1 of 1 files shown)


gpanshu (approved):

sdhuka (approved):

sktimalsina (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java
```diff
@@ -316,7 +316,11 @@ void stopDrainingMutationOutbox() {
             @NonNull PublicationStrategy<T> publicationStrategy) {
         return Single
             .<GraphQLResponse<ModelWithMetadata<T>>>create(subscriber ->
-                publicationStrategy.publish(mutation.getMutatedItem(), subscriber::onSuccess, subscriber::onError)
+                publicationStrategy.publish(mutation.getMutatedItem(), subscriber::onSuccess, (exception) -> {
+                    if (!subscriber.isDisposed()) {
```
alharris-at, line 320: Is there anywhere else in our pubsub model we may want to apply a similar strategy or checking for subscriber state? Or do all published events pass through this `publishWithStrategy` method?

poojamat: Good point, all the publish events pass through this method.

PR #1692, opened 2022-04-04
feat(maplibre-adapter): add clustering to map view

eeatonaws:
*Description of changes:* Adds the ability to enable clustering on the map and customize various properties of the clusters (e.g., cluster color). Also fixes an issue where calling MapLibreView.setStyle(...) would remove the symbol on click listener, meaning the popup with place info would not show up when clicking on a place on the map.

Clustering enabled with the default clustering options:
Screen Shot 2022-04-06 at 3 22 01 PM

Screen Shot 2022-04-06 at 3 22 40 PM

Screen Shot 2022-04-06 at 3 23 05 PM

### DIFF
--- a/maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/Credentials.kt
+++ b/maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/Credentials.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.geo.maplibre
+
+import android.content.Context
+import com.amplifyframework.core.Resources
+
+/**
+ * Utility to load Cognito user credentials from a json document.
+ * The document must be formatted as following:
+ *
+ * <pre>
+ * {
+ *   "username": "username123",
+ *   "password": "kool@pass22"
+ * }
+ * </pre>
+ */
+object Credentials {
+    private const val DEFAULT_ID = "credentials" // credentials.json
+    private var credentials: Pair<String, String>? = null
+
+    fun load(context: Context, identifier: String = DEFAULT_ID): Pair<String, String> {
+        if (credentials == null) {
+            credentials = synchronized(this) {
+                val json = Resources.readJsonResource(context, identifier)
+                val username = json.getString("username")
+                val password = json.getString("password")
+                username to password
+            }
+        }
+        return credentials!!
+    }
+}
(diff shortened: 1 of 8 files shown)


maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/view/MapLibreView.kt
```diff
@@ -72,10 +83,15 @@ class MapLibreView
     }
 
     lateinit var symbolManager: SymbolManager
+    internal lateinit var symbolOnClickListener: (Symbol) -> Boolean
 
     var defaultPlaceIcon = R.drawable.place
     var defaultPlaceActiveIcon = R.drawable.place_active
 
+    private var shouldCluster = true
```
poojamat, line 91: Could these be lateinit?

maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/view/MapLibreView.kt
```diff
@@ -154,16 +171,119 @@ class MapLibreView
...
-                this.symbolManager = SymbolManager(this, map, it, null, null).apply {
-                    iconAllowOverlap = true
-                    iconIgnorePlacement = true
+
+                // Clear the current symbols from the map since a new SymbolManager will be created
+                if (this::symbolManager.isInitialized) {
+                    this.symbolManager.deleteAll()
+                }
+
+                removeClusterLayers(it)
```
poojamat, line 180: If the clusterlayers are already set should be remove them and recreate or just leave them alone?

maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/view/MapLibreView.kt
eeatonaws: These 3 properties are initialized to their default values, which I think makes the code a bit cleaner compared to using lateinit.

maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/view/MapLibreView.kt
eeatonaws: If clustering is enabled with different clustering options, then we'll need to update the properties for the cluster circle and number layers, but adding a layer to the style that has the same ID as an existing layer on the style throws an error, so I removed the old cluster circle and number layers (which also needs to be done when disabling clustering).

poojamat (approved):
LGTM

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/Credentials.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.content.Context
+import com.amplifyframework.core.Resources
+
+/**
+ * Utility to load Cognito user credentials from a json document.
+ * The document must be formatted as following:
+ *
+ * <pre>
+ * {
+ *   "username": "username123",
```
gpanshu, line 27: Question: In the documentation should we add values as "YOUR_AWS_USERNAME" instead of "username123" to remove ambiguity ?

gpanshu (approved):

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/Credentials.kt
eeatonaws: The value isn't the AWS username, it's just a username/password for testing purposes that's defined in the integration test resources (no need for anyone running the tests to change the username or password used for these tests).

drochetti (commented):
Left a few comments regarding the UI experience, other than that it looks great. Thanks for adding clustering support, really! :)

Issue #1599, opened 2021-12-09
Crash caused by NullPointerException when parsing GraphQLResponse.Error with null/missing message


millsjustin:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

GraphQL API

### Gradle script dependencies

<details>

```groovy
implementation 'com.amplifyframework:core:1.24.1'
implementation 'com.amplifyframework:aws-api:1.24.1'
```

</details>

### Environment information

<details>

```text
------------------------------------------------------------
Gradle 6.8.2
------------------------------------------------------------

Build time:   2021-02-05 12:53:00 UTC
Revision:     b9bd4a5c6026ac52f690eaf2829ee26563cad426

Kotlin:       1.4.20
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.10 (JetBrains s.r.o. 11.0.10+0-b96-7281165)
OS:           Mac OS X 11.6.1 x86_64
```

</details>

### Describe the bug

Yesterday during the US-EAST-1 outage I was seeing app crashes caused by a `NullPointerException` thrown while deserializing `GraphQLResponse.Error`. The full stacktrace is:

```
Fatal Exception: java.lang.NullPointerException
       at java.util.Objects.requireNonNull(Objects.java:220)
       at com.amplifyframework.api.graphql.GraphQLResponse$Error.<init>(GraphQLResponse.java:138)
       at com.amplifyframework.api.graphql.GsonResponseAdapters$ErrorDeserializer.deserialize(GsonResponseAdapters.java:216)
       at com.amplifyframework.api.graphql.GsonResponseAdapters$ErrorDeserializer.deserialize(GsonResponseAdapters.java:156)
       at com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
       at com.google.gson.Gson.fromJson(Gson.java:932)
       at com.google.gson.Gson.fromJson(Gson.java:1003)
       at com.google.gson.internal.bind.TreeTypeAdapter$GsonContextImpl.deserialize(TreeTypeAdapter.java:162)
[...]

PR #1700, opened 2022-04-08
Provide default message for GraphQLResponse.Error when null/missing

sktimalsina:
*Issue #, if available:*
#1599 

*Description of changes:*
com.amplifyframework.api.graphql.GraphQLResponse.Error#Error expects message parameter to be NonNull.
In case, the MESSAGE_KEY (="message") is not passed in error JSON, it is not initialized.

This change adds a default string in such exceptional cases.

Note: This is pulled from [fork](https://github.com/millsjustin/amplify-android/tree/jm/provide-default-message-when-null)

- [X] Added Unit Tests

- [X] No

### DIFF
--- a/aws-api-appsync/src/main/java/com/amplifyframework/api/graphql/GsonResponseAdapters.java
+++ b/aws-api-appsync/src/main/java/com/amplifyframework/api/graphql/GsonResponseAdapters.java
@@ -201,6 +201,10 @@ public final class GsonResponseAdapters {
                 }
             }
 
+            if (message == null) {
+                message = "Message was null or missing while deserializing error";
+            }
+
             // Merge nonSpecifiedData into extensions, but don't overwrite anything already there.
             for (String key : nonSpecifiedData.keySet()) {
                 if (!extensionsJson.has(key)) {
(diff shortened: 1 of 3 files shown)


aws-api-appsync/src/main/java/com/amplifyframework/api/graphql/GsonResponseAdapters.java
```diff
@@ -201,6 +201,10 @@ private JsonElement skipQueryLevel(JsonElement jsonData) throws JsonParseExcepti
                 }
             }
 
+            if (message == null) {
```
gpanshu, line 204: Should we also check for when the message is not null but empty?

gpanshu (approved):

aws-api-appsync/src/main/java/com/amplifyframework/api/graphql/GsonResponseAdapters.java
sktimalsina: We might be overwriting an intentional empty message in that case. 
This particular null check is to satisfy `NonNull` expectation of Error object and avoid NPE

PR #1735, opened 2022-05-13
Fix for adding data back to the delete request if it contains data as that is allowed for the Delete api

gpanshu:
*Issue #, if available:* 1729

*Description of changes:*
The Delete api does allow for a request to be passed in and therefore this check was removed to accommodate that.

- [ ] Added Unit Tests

- [ x] No

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
@@ -710,14 +710,19 @@ public final class AWSApiPlugin extends ApiPlugin<Map<String, OkHttpClient>> {
             case GET:
             case DELETE:
                 if (options.hasData()) {
-                    throw new ApiException("HTTP method does not support data object! " + type,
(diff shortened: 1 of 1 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java
```diff
@@ -710,14 +710,19 @@ private RestOperation createRestOperation(
...
+                            options.getData(),
+                            options.getHeaders(),
+                            options.getQueryParameters());
+                } else {
+                    operationRequest = new RestOperationRequest(
+                            type,
+                            options.getPath(),
+                            options.getHeaders(),
+                            options.getQueryParameters());
                 }
```
eeatonaws, line 725: With these changes, a GET or HEAD request would now be allowed to have a data object, which shouldn't be allowed. The original code under the DELETE case (that didn't allow a data object) can be added under `case GET:` to prevent a data object being added for GET or HEAD.

eeatonaws (approved):

PR #1759, opened 2022-05-27
fix: start transfer service as foreground for >= API26

sdhuka:
*Issue #, if available: #amplify-flutter/issues/1635*

*Description of changes:*
- Starting service while app is in background is not allowed starting API 26. This change is to start transfer service as a foreground service for API >= 26 with a default notification.

- Manually tested the changes.


HuiSF (dismissed):

aws-storage-s3/src/main/res/values/strings.xml
```diff
@@ -14,4 +14,8 @@
    permissions and limitations under the License.
 -->
 
-<resources />
+<resources>
+    <string name="amplify_notification_channel_id">Transfer Channel</string>
+    <string name="amplify_notification_channel_name">Amplify Transfer Channel</string>
```
tylerjroach, line 19: Lets consider naming amplify_storage_transfer_notification_channel_name, so that if other categories require a notification channel, we aren't using a generic name. This also gives the customer the opportunity to override this specific channel name.

aws-storage-s3/src/main/res/drawable/ic_notification_test.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
tylerjroach, line 1: We should rename to "amplify_storage_transfer_notification_icon.xml" so that we are properly namespaced to Amplify. This prevents collision with other libraries, and allows the customer a clear path to override this drawable with a different image.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.java
```diff
@@ -262,4 +275,32 @@ private void startServiceIfNotAlreadyStarted() {
     public AmazonS3Client getClient() {
         return client;
     }
+
+    private Notification createDefaultNotification() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            createChannel();
+        }
+        int appIcon = context.getApplicationContext().getApplicationInfo().icon > 0 ?
```
tylerjroach, line 283: `context.getApplicationContext().getApplicationInfo().icon` will not look right in a lot of cases due to the way notifications are flattened into a solid color. Instead, I would propose providing a default icon named "amplify_storage_transfer_notification_icon", and then make it clear in our docs that the customer can provide a drawable under the same name to override our own. Ours would be a fallback if no override is provided.

tylerjroach (changes requested):
I've commented in a few areas on setting up an Amplify namespace. I would propose a common practice of amplify_{category}_{friendly name} for any resources we provide. This presents multiple advantages to us.
1. It prevents Amplify from colliding with resources found in other libraries.
2. It allows us to easily document resources values the customer can override with their own

tylerjroach (approved):

HuiSF (dismissed):

tylerjroach (approved):

tjleing (approved):

PR #1793, opened 2022-06-22
fix(datastore): Add typename to selection set

lawmicha:
*Issue #, if available:*

*Description of changes:*
When using DataStore from iOS, the mutations sent from the Android DataStore client is missing the `__typename` in the selection set. Since the DataStore client on iOS is subscribed to these mutations, the subscription event received by the iOS client fails to decode a mutation response that is missing `__typename`. This metadata field required for iOS to decode properly into the Model types. Adding the `__typename` to the selection set fixes only Models without any associations. Models with associations may have other decoding issues, that is out of scope for this change. 

- [ ] Added Unit Tests

- [ ] No

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -30,5 +30,6 @@ POM_DEVELOPER_ORGANIZATION=Amazon Web Services
 POM_DEVELOPER_ORGANIZATION_URL=http://aws.amazon.com
 
 android.useAndroidX=true
-android.enableJetifier=false
+android.enableJetifier=true
+android.jetifier.blacklist=bcprov-jdk15on
 
--- a/aws-datastore/build.gradle
+++ b/aws-datastore/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     testImplementation dependency.robolectric
     testImplementation dependency.androidx.test.core
     testImplementation dependency.mockk
+    testImplementation project(path: ':aws-datastore')
 
     androidTestImplementation dependency.mockito
     androidTestImplementation project(path: ':testmodels')


build.gradle
```diff
@@ -125,7 +125,7 @@ ext {
         mockito: 'org.mockito:mockito-core:3.1.0',
         mockk: 'io.mockk:mockk:1.10.6',
         mockwebserver: 'com.squareup.okhttp3:mockwebserver:4.9.1',
-        robolectric: 'org.robolectric:robolectric:4.5.1',
+        robolectric: 'org.robolectric:robolectric:4.7.1',
```
lawmicha, line 128: I had to upgrade this to at least 4.7.1 to get the unit tests running successfully on M1 Mac.

When I tried to run the unit tests, the ones that had a dependency on SQLite would fail with the error:

```
Failed to initialize the local storage adapter for the DataStore plugin.
AmplifyException{message=Failed to initialize the local storage adapter for the DataStore plugin., cause=java.lang.RuntimeException: DataStoreException{message=Error in initializing the SQLiteStorageAdapter, cause=java.lang.RuntimeException: com.almworks.sqlite4java.SQLiteException: [-91] cannot load library: java.lang.UnsatisfiedLinkError: /private/var/folders/76/l6xl3gl11clcj6yd5f0k7ddc0000gr/T/1655928563383-0/libsqlite4java.dylib: dlopen(/private/var/folders/76/l6xl3gl11clcj6yd5f0k7ddc0000gr/T/1655928563383-0/libsqlite4java.dylib, 0x0001): tried: '/private/var/folders/76/l6xl3gl11clcj6yd5f0k7ddc0000gr/T/1655928563383-0/libsqlite4java.dylib' (fat file, but missing compatible architecture (have 'i386,x86_64', need 'arm64e')), recoverySuggestion=See attached exception}, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}
```

Looking around, it looks like this was fixed in at least [robolectric 4.7](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.7). Related issue: https://github.com/robolectric/robolectric/issues/6311 . We can also consider going to the current 4.8.1. 4.8 had a lot of changes https://github.com/robolectric/robolectric/releases

gradle.properties
```diff
@@ -30,5 +30,5 @@ POM_DEVELOPER_ORGANIZATION=Amazon Web Services
 POM_DEVELOPER_ORGANIZATION_URL=http://aws.amazon.com
 
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
```
lawmicha, line 33: This i'm not too sure about.. once I upgraded to roboeletric 4.7.1, i got this error:

```

```

Related issue: https://github.com/robolectric/robolectric/issues/6521

What's the implications of disabling Jetifer? what should be the correct change?

mikepschneider: From stackoverflow:
```
Jetifier helps in migrating an android project to AndroidX.

Jetifier helps in making 3rd party library compatible with your AndroidX project.

If you're not using any 3rd party library in your android project, then you don't need to use Jetifier 
because Jetifier helps only in "making 3rd party library" compatible with your project in 
case you're migrating to AndroidX. So to make your project use Jetifier, you write below 
code in gradle.properties.

android.enableJetifier=true
android.useAndroidX=true

If you're not migrating to AndroidX and using the previous way of using Android libraries
(e.g: com.android.support), then these 3rd party libraries need not be made compatible 
as these 3rd party libraries are already compatible with the previous way of using Android libraries. 
In that case, you don't need to use Jetifier, So now, you don't need to write lines given above or 
you can simply write as below to change the values in future to move to androidx:-

```

So I guess the critical question is if we have any 3rd party libraries that are pointing to the old location.  Given that all of the tests pass when this is turned off, that would indicate to me we don't.  I will try to look back to where it was turned on and see if there was a specific reason for it.

mikepschneider: Jetifier was enabled in this release: https://github.com/aws-amplify/amplify-android/commit/e4eb992564e47bc34d4f76940d2880ec343f8fcd 

That was over 3 years ago.  A lot of time for our dependencies to be updated.   The fact that all of the tests pass is a very good sign that this is ok.

mikepschneider (approved):
🚢

gradle.properties
mikepschneider: I found a limited fix that allows us to block just the offending library.  This should be much safer than turning off jetifier altogether.  Pushed to the branch.

lawmicha: thanks for looking into this, it sounds like the impact of disabling it would only impact the amplify developers writing tests and not to the developers using the library, so given the existing tests are passing, it seems like a low risk change to make. The disallow list also makes sense and is even better then

dpilch (approved):

lawmicha:
I believe @poojamat  or @mikepschneider  incorporated the `__typename` as part of the CPK feature changes and this PR is no longer needed

PR #1851, opened 2022-07-26
feat(auth): call confirm reset password use case from api

sktimalsina:

*Issue #, if available:*

*Description of changes:*

- [x] Added Unit Tests

- [x] No

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
@@ -32,6 +32,7 @@ import com.amplifyframework.auth.AuthUser
 import com.amplifyframework.auth.AuthUserAttribute
 import com.amplifyframework.auth.AuthUserAttributeKey
 import com.amplifyframework.auth.cognito.helpers.JWTParser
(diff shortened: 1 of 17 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -862,7 +863,26 @@ internal class RealAWSCognitoAuthPlugin(
...
+                    username,
+                    newPassword,
+                    confirmationCode,
+                    options,
+                    onSuccess,
+                    onError
+                )
+            }
+        } catch (ex: Exception) {
+            onError.accept(AuthException.InvalidUserPoolConfigurationException(ex))
```
banji180, line 884: There are possibilities of multiple exceptions being thrown here, I'd expect the onError to propagate the appropriate error to the caller.

sktimalsina: Ideally, we would reach this catch block only if `requireNotNull` is not satisfied for `appClientId` or `cognitoIdentityProviderClient`. Which should not occur if all the configuration is correct.

All other exceptions (like ones returned by identity provider and remote API calls) are caught and boxed in onError passed to the use case and returned there.

Issue #1868, opened 2022-08-02
TransferService causes ForegroundServiceDidNotStartInTimeException


yaroslav-v:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

Authentication, Storage

### Gradle script dependencies

<details>

```groovy
implementation "com.amplifyframework:core:1.36.4"
implementation "com.amplifyframework:aws-storage-s3:1.36.4"
implementation "com.amplifyframework:aws-auth-cognito:1.36.4"


```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 7.0.2
------------------------------------------------------------

Build time:   2021-05-14 12:02:31 UTC
Revision:     1ef1b260d39daacbf9357f9d8594a8a743e2152e

Kotlin:       1.4.31
Groovy:       3.0.7
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          1.8.0_301 (Oracle Corporation 25.301-b09)
OS:           Mac OS X 10.16 x86_64


```

</details>

### Describe the bug

Hi!

One of the latest versions 1.36.4 causes fatal exceptions. The cause is obvious from the crash description - **Context.startForegroundService() did not then call Service.startForeground() ... com.amazonaws.mobileconnectors.s3.transferutility.TransferService**

```
ActivityThread.java line 2147
android.app.ActivityThread.throwRemoteServiceException

Fatal Exception: android.app.ForegroundServiceDidNotStartInTimeException
Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{2637b1f u0 com.example.package/com.amazonaws.mobileconnectors.s3.transferutility.TransferService}
```

We've observed the crash on Android 9, 11, 12.

And the following list of devices at the moment:
Galaxy S20+ 5G
Galaxy A13 5G
Galaxy A13
Galaxy S10
moto g stylus 5G (2022)
Lenovo Tab M10 (FHD)
Galaxy A12

We've reverted back to one the previous versions for now, so actually there might be more of them. The data is obtained from Firebase crash reports.

### Reproduction steps (if applicable)

Use amplifyframework v1.36.4

### Code Snippet

```java
// Initialization in Application class

[...]

PR #1875, opened 2022-08-05
fix(category): Remove startForegroundService in favor of binding service

tylerjroach:

*Issue #, if available:*
[1868](https://github.com/aws-amplify/amplify-android/issues/1868)

*Description of changes:*
`startForegroundService` requires that `startForeground` is called within a specific time period, which is nearly possible to guarantee. If any app using Amplify experiences an ANR (even if ANR is not from Amplify), theres a chance that we will crash with an exception that `startForeground` was not called.

This PR is using a strategy to bind to the service and call `startForeground`, which is a workaround that doesn't encounter the time limit imposed from `startForegroundService`

- [ ] Added Unit Tests

- [ ] No


tylerjroach:
Do not merge. I will be adding additional tests to make sure AmplifyTansferService is not running on test start, as well as writing an additional test that stress tests a large count of uploads/downloads at once.

tylerjroach:
Stress test added and tests updated.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageTransferStressTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import junit.framework.Assert.assertFalse
+import junit.framework.Assert.assertTrue
+import kotlin.random.Random
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+
+class AWSS3StorageTransferStressTest {
```
sdhuka, line 43: Should we delete the files in the teardown? or replace the files with static list so that we don't grow the bucket size.

sdhuka (approved):
I have tested this and works as expected. Couple of minor comments to address.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageTransferStressTest.kt
tylerjroach: I specifically used non-random names so the bucket wouldn't grow in size. We overwrite the existing keys with new uploads on each run.

aws-storage-s3/src/main/AndroidManifest.xml
```diff
@@ -21,5 +21,6 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
     <application>
         <service android:name="com.amazonaws.mobileconnectors.s3.transferutility.TransferService" android:enabled="true" />
```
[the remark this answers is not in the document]
tylerjroach: @sdhuka I left this in place in case the escape hatch is being used, and the customer interacts with sdk directly. They may have previously relied on us importing the service, rather than adding to their own manifest, so it felt like a breaking change if we were to remove.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StorageTransferStressTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.amplifyframework.storage.s3.test.R
+import com.amplifyframework.testutils.FileAssert
+import com.amplifyframework.testutils.random.RandomTempFile
+import com.amplifyframework.testutils.sync.SynchronousMobileClient
+import com.amplifyframework.testutils.sync.SynchronousStorage
+import java.io.File
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import junit.framework.Assert.assertFalse
+import junit.framework.Assert.assertTrue
```
eeatonaws, line 37: `junit.framework.Assert` is deprecated, suggest using `org.junit.Assert` instead.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AmplifyTransferService.kt
```diff
@@ -0,0 +1,242 @@
...
+    private val binder = LocalBinder()
+
+    /**
+     * registers a BroadcastReceiver to receive network status change events. It
+     * will update transfer records in database directly.
+     */
+    private var transferNetworkLossHandler: TransferNetworkLossHandler? = null
+
+    /**
+     * A flag indicates whether or not the receiver has is started the first time.
```
eeatonaws, line 54:
```suggestion
     * A flag indicates whether or not the receiver has started the first time.
```

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AmplifyTransferService.kt
```diff
@@ -0,0 +1,242 @@
...
+     */
+    private var isReceiverNotRegistered = true
```
eeatonaws, line 56: I think it would be clearer to change this variable to isReceiverRegistered, so it's a bit easier to understand what !isReceiverNotRegistered means.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AmplifyTransferService.kt
```diff
@@ -0,0 +1,242 @@
...
+            log.verbose("AmplifyTransferService unbind check running")
+            if (!TransferStatusUpdaterAccessor.hasActiveTransfer(applicationContext)) {
+                try {
+                    log.verbose("Removing AmplifyTransferService from foreground and unbinding")
+                    stopForegroundAndUnbind(applicationContext)
+                } catch (e: Exception) {
+                    log.error("Error in moving the service out of the foreground state: $e")
+                }
+            } else {
+                log.verbose("Transfers info progress, rescheduling unbind check")
```
eeatonaws, line 112: "Transfers in progress" rather than "Transfers info progress"?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AmplifyTransferService.kt
```diff
@@ -0,0 +1,242 @@
...
+                        boundService = null
+                    }
+                }
+            }
+
+            boundServiceConnection?.let {
+                context.bindService(Intent(context, AmplifyTransferService::class.java), it, Context.BIND_AUTO_CREATE)
+            }
+
+            // A new call to to bind will restart counter, removing potential to unbind service too early
```
eeatonaws, line 189:
```suggestion
            // A new call to bind will restart counter, removing potential to unbind service too early
```

eeatonaws (changes requested):
Left a few comments to address. I still need to test the changes in a sample app.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AmplifyTransferService.kt
tylerjroach: agree. changed. Some of this I had moved from TransferService and tried to keep as close in appearance as possible, but makes sense to make improvements like this.

eeatonaws (approved):
LGTM! I tested the changes in a sample app and it works as expected.

sdhuka (approved):

PR #1876, opened 2022-08-07
fix(datastore): memory issues, add disableSync and disableSubscription

### LINKED ISSUE #1869: Memory Pressure Induced Via Mechanisms in PersistentMutationOutbox and SubscriptionProcessor Consistently Lead To Android OutOfMemory
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

DataStore

### Gradle script dependencies

<details>
Note that `1.35.4.1-tz` is 1.35.4 with #1743 and #1755 applied

```groovy
// Put output below this line
"com.amplifyframework:aws-datastore:1.35.4.1-tz"
"com.amplifyframework:aws-api:1.35.4.1-tz"
"com.amplifyframework:aws-api-appsync:1.35.4.1-tz"
"com.amplifyframework:aws-auth-cognito:1.35.4.1-tz"
```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 7.0.2
------------------------------------------------------------

Build time:   2021-05-14 12:02:31 UTC
Revision:     1ef1b260d39daacbf9357f9d8594a8a743e2152e

Kotlin:       1.4.31
Groovy:       3.0.7
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.16 (Oracle Corporation 11.0.16+8)
OS:           Linux 5.18.15-arch1-1 amd64
```

</details>

### Please include any relevant guides or documentation you're referencing

https://github.com/aws-amplify/amplify-android/discussions/1849

### Describe the bug

Per discussion #1849, I am filing this bug regarding excessive Android Datastore Memory Consumption.

# Overview:
The memory consumption in 2 is amplified by the bug in 1 and the source of excessive consumption is in 2 and 3.

  1. [MutationProcessor's core Observable] logic will cease all transmission operations given _any_ exception.
  2. PendingMutations within [Node] objects stored within the [mutationMap HashMap of the MutationQueue] when offline transactions accumulate.
  3. The [unbound SubscriptionProcessor ReplaySubject] accumulates infinitely with a long running online application.
 

# 1. MutationProcessor's Core Observable Ceases All Transmission Given An Exception
 The [MutationProcessor's core Observable] logic is flawed because it will cease all attempts to transmit messages whenever the Observable encounters _any_ exception and will not restart without unrelated events triggering another invocation of [startDrainingMutationOutbox] by the Orchestrator itself. I believe this logic is the source of many years of frustration by developers who cite in nearly all major issues, "I got X exception and then application stopped transmitting data to the backend". Furthermore, the chance of the Observable being halted is extremely high due to the an arbitrary [blockingAwait in drainMutationOutbox] which throws an exception after 10 seconds. Given one bad transmission on an otherwise fine connection and the engine is dead. The mutation outbox draining will not resume without the application transitioning between online and offline or being restarted. A final recommendation could be given in PR #1845 as my team and I finish our analysis however the analysis is delayed due to the creation of this issue. I am leaning towards my suggestion to [have the observable resume draining the mutation outbox on error] which would make the warnings due to unrecoverable Exceptions more pronounced due to effectively an infinite loop of retries. However, being that many of the exceptions that would cause the warnings are transient in nature allowing the Observer to resume would make the transmission of the outbox much more resilient.

## 1a. Remove arbitrary blockingAwait timeout
 Regarding exception sources within processing, it seems better to remove the [blockingAwait in drainMutationOutbox] all together as it introduces an arbitrary unhandled situation where a pending mutation is transmitting but taking a long time, such as on a weak cellular connection. The outbox draining will cease until the entire datastore sync engine is cycled, such as what PR #1845 causes currently on exception, which may create unnecessary thrashing in the case of a network transmission time out such as: `java.net.SocketException: Socket closed` or `java.net.SocketTimeoutException: timeout`. It would be better to just let the outbox try some more. The blockingAwait is further unnecessary because, unless the application owner configures alternative timeouts [during configuration] for the client, the default [okhttp3 timeouts] seem reasonable, 10 seconds for operations and unlimited time as long as individual operations respond within 10 seconds, and I cannot think of a reason make this loop cease otherwise than when it is truly intended to via [stopObservingStorageChanges].

# 2. Pending Mutations within Mutation Map Exhaust Local Heap
As cited by @sbaxter in #1849, given enough time offline the PersistentMutationOutbox will accumulate so many PendingMutations that eventually the application will encounter an OutOfMemoryError for exceeding heap limits imposed by the operating system. There are two major origins of the unhandled heap growth and occur at two different times of operation. Both involve long term accumulation of offline records and both result in an application that must have its memory cleared for recovery.

Excessive pending mutations have a very good chance of accumulating silently due to the bug mentioned in 1 regarding the [MutationProcessor's core Observable].

Given that PendingMutations while offline are effectively unnecessary to reside in memory, it would seem reasonable to replace the outbound mutation processor with something like [Paging](https://developer.android.com/topic/libraries/architecture/paging/v3-overview). As some other issues have raised, it is also a waste of time force the datastore to block all operations while loading this map.

## 2a. Prolonged Offline Operation

The simplest example is during periods of prolonged offline operation the [mutationMap] will accumulate PendingMutations within the Node objects. As the mutations stack up the application will eventually exhaust all heap available and likely be killed by an OutOfMemoryError:

```
05-05 12:45:24.318 25011 25034 I -: WaitForGcToComplete blocked Background on HeapTrim for 2.032s
05-05 12:45:24.318 25011 26463 I -: WaitForGcToComplete blocked Alloc on HeapTrim for 2.030s
05-05 12:45:24.318 25011 15918 W -: Throwing OutOfMemoryError "Failed to allocate a 9437192 byte allocation with 3030896 free bytes and 2959KB until OOM, target footprint 201326592, growth limit 201326592" (VmSize 103637692 kB)
```

During experiments of prolonged offline running with a model decorated with user ownership `@auth` containing `{timestamp, 20-character-string}` we found that typically on a Pixel 4a the application could accumulate up to 11,000 before it was evident that the pending mutations were causing a problem. The past analysis was complicated due to the tendency for the mutationProcessor to halt given any exception, as discussed in 1 and PR #1845. With that said, we found that accumulating 11,000 to 33,000 records typically resulted in enough records to notice extreme garbage collection pauses and eventually produce OutOfMemoryErrors.

As mentioned in Discussion #1849 the accumulation of memory use is magnified by a factor of approximately 10x via the use of GJSON serialized objects vs JSON objects, the pending mutation object form vs the transmission object form.

## 2b. Cold Start [mutationOutbox.load()] during Orchestrator transitions after accumulating offline records

During experiments to determine the maximum number of pending mutations the application could support we discovered that while the slow heap growth of accumulation records in the [mutationMap] was tolerated by Android, rapid heap growth caused by a cold start and [mutationOutbox.load()] was not. We found that we could accumulate 3-4x more records in situation 2a than the operating system would tolerate in situation 2b. Effectively in our production scenarios, around 4,000 records could be loaded rapidly however 11,000 could not.

The [startObservingStorageChanges()] method will always be invoked regardless of if the Orchestrator is in "LOCAL_ONLY" mode, due to [transitionToLocalOnly] or if is in "SYNC_VIA_API" mode, due to [transitionToApiSync]. The only way out of this situation is to clear the local data of the application.

# 3. SubscriptionEvents Will Exhaust Local Heap
Given an application that is long running with a continuous connection, the accumulation of SubscriptionEvents will eventually exhaust the local heap of the application.

While trying to figure out why our application was exhausting memory even when the PendingMutations were draining we discovered via taking heap dumps at periodic times that the number of SubscriptionEvents continued to grow with every PendingMutation submitted to the API. Further analysis led us to understand that even if we always transmitted the mutations, thus freeing up memory from that end, we'd still have another leak with the permanent copy of the object wrapped in a newly created incoming SubscriptionEvent. Thus even with successful draining, we would always exhaust our heap after prolonged operation with a reliable radio to transmit. Given our immutable time series design, we were able to work around this by forcing all GraphQL subscription endpoints to respond with unauthorized. Note that using the [formal `@model` subscriptions off decoration](https://docs.amplify.aws/cli-legacy/graphql-transformer/auth/#authorizing-subscriptions) resulted in an unusable datastore as it refused to start (yet another bug). The workaround was to make all subscription endpoints use a hard coded unauthorized response.

It is very likely that the [unbound SubscriptionProcessor ReplaySubject] is the culprit and according to the [rxjava3.subjects.ReplaySubject]:

> creates an empty, unbounded ReplaySubject that caches all items and the terminal event it receives.

In situations of continuous connectivity and many transactions, we observed the heap will be exhausted by the unbound structure.

In the short term it might be better to let the constructor use the [createWithSize()] method which would bind the number of updates, say 100 or 500, and keep memory within predictable bounds. In the long term it might be best to allow the maintainer of the application to choose what behavior they want, `createWithSize()`, `createWithTime()`, or `createWithTimeAndSize()`.

## 3a. Remove SubscriptionProcessor Latch
The [SubscriptionProcessor Latch] should be removed and is another example of an arbitrary timeout that results in introducing extra complexity, extra exception handling, and hurts new adopters of the platform with the assumption that the radio (wifi or cellular) will be robust at a given time. Just like the argument in 1a, the timeout introduces unnecessary complexity where the underlying okhttp3 timeouts should be sufficient.

[mutationMap]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java#L34

[Node]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java#L213-L218

[PersistentMutationOutbox]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java#L192

[unbound SubscriptionProcessor ReplaySubject]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L127

[mutationMap HashMap of the MutationQueue]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationQueue.java#L34
[subscriptions]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L129

[transitionToLocalOnly]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L221
[transitionToApiSync]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L246

[MutationProcessor's core Observable]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L91-L106

[blockingAwait in drainMutationOutbox]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L118

[startDrainingMutationOutbox]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L90

[have the observable resume draining the mutation outbox on error]: https://github.com/aws-amplify/amplify-android/pull/1845#discussion_r934041094

[during configuration]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java#L139-L146

[okhttp3 timeouts]: https://www.baeldung.com/okhttp-timeouts

[stopObservingStorageChanges]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L278-L283

[SubscriptionProcessor ReplaySubject]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L127

[rxjava3.subjects.ReplaySubject]: http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/subjects/ReplaySubject.html

[createWithSize()]: http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/subjects/ReplaySubject.html#createWithSize-int-

[SubscriptionProcessor Latch]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L121-L123

[mutationOutbox.load()]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L262

[startObservingStorageChanges()]: https://github.com/aws-amplify/amplify-android/blob/release_v1.36.5/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Orchestrator.java#L259

### Reproduction steps (if applicable)

_No response_

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### amplifyconfiguration.json

_No response_

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

### Additional information and screenshots

_No response_

joekiller:
- [X] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*
Fixes #1869 
Fixes #1849 

*Description of changes:*

## Fix memory management

  - mutation outbox will now allocate 200 slots by default
  - mutation outbox loads only first 100 mutations plus any pending mutations related to first 100 loaded
  - SubscriptionProcessor will now only hold a maximum of 100 subscription events with a max age of 10 minutes

## add options to disable Sync or Subscriptions
Give a "producer only" time series database the sync and subscription systems can be cumbersome or unnecessary.
Sometimes people just want to push data from the datastore and this gives more flexibility to the engine, makes
for faster startup in these conditions, and reduced the need to hack around these features with items such as
having the graphQL endpoint respond with empty lists in the sync or unauthorized for a subscription.

  - use datastore configuration disableSync(Model.class) and disableSubscription(Model.class) to allow toggling if model should sync or receive subscriptions
  - https://docs.amplify.aws/lib/datastore/sync/q/platform/android/ should be updated to reflect new options for Syncing data.
 
## Fix Pending Mutation Engine Halting on Error
Currently the mutation engine will halt on any given error. This update tries to resume reading updates every 30 seconds. Configurable via datastore option outboxErrorRestartDelay.

## Remove timeouts
There are various timeouts that are effectively CPU or network bound. Removing the timeouts makes the system more
likely to work on a diversity of client.

  - remove restart delay for tests as they were CPU bound and caused predicable failure when a thread goes slow
  - remove Orchestrator semaphore timeout
  - remove ITEM_PROCESSING_TIMEOUT_MS from MutationProcessor

- [X] Added Unit Tests
- [X] Tested on frontend (Add Screenshots)
image

- [ ] No
- [X] Yes (Please include a PR link for the documentation update)

### DIFF
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/operation/AWSRestOperationTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/operation/AWSRestOperationTest.java
@@ -123,7 +123,7 @@ public final class AWSRestOperationTest {
         RestOperationRequest request =
             new RestOperationRequest(HttpMethod.GET, baseUrl.uri().getPath(), emptyMap(), emptyMap());
         assertTimedOut(() ->
-            Await.<RestResponse, ApiException>result(timeToWaitForResponse, (onResult, onError) -> {
+            Await.<RestResponse, ApiException>result((onResult, onError) -> {
                 AWSRestOperation operation =
                     new AWSRestOperation(request, baseUrl.url().toString(), client, onResult, onError);
                 operation.start();
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/AppSyncClientInstrumentationTest.java
@@ -327,7 +327,7 @@ public final class AppSyncClientInstrumentationTest {
     private <T> T awaitResponseData(
             Await.ResultErrorEmitter<GraphQLResponse<T>, DataStoreException> resultErrorEmitter)
             throws DataStoreException {
-        final GraphQLResponse<T> response = Await.result(TimeUnit.SECONDS.toMillis(20), resultErrorEmitter);
+        final GraphQLResponse<T> response = Await.result(resultErrorEmitter);
         if (response.hasErrors()) {
             String firstErrorMessage = response.getErrors().get(0).getMessage();
             throw new DataStoreException("Response contained errors: " + firstErrorMessage, "Check request.");
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/SynchronousStorageAdapter.java
@@ -97,7 +97,6 @@ public final class SynchronousStorageAdapter {
     @SuppressWarnings("UnusedReturnValue")
     public List<ModelSchema> initialize(@NonNull Context context) throws DataStoreException {
         return Await.result(
-            operationTimeoutMs,
             (Consumer<List<ModelSchema>> onResult, Consumer<DataStoreException> onError) -> {
                 try {
                     asyncDelegate.initialize(context,
@@ -144,7 +143,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> void save(@NonNull T model, @NonNull QueryPredicate predicate)
             throws DataStoreException {
         Await.result(
-            operationTimeoutMs,
             (Consumer<StorageItemChange<T>> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.save(
                     model,
@@ -177,7 +175,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> DataStoreException saveExpectingError(
             @NonNull T model, @NonNull QueryPredicate predicate) {
         return Await.error(
-            operationTimeoutMs,
             (Consumer<StorageItemChange<T>> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.save(
                     model,
@@ -212,7 +209,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> List<T> query(@NonNull Class<T> modelClass, @NonNull QueryOptions options)
             throws DataStoreException {
         Iterator<T> resultIterator = Await.result(
-            operationTimeoutMs,
             (Consumer<Iterator<T>> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.query(modelClass, options, onResult, onError)
         );
@@ -268,7 +264,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> void delete(@NonNull T model, @NonNull QueryPredicate predicate)
             throws DataStoreException {
         Await.result(
-            operationTimeoutMs,
             (Consumer<StorageItemChange<T>> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.delete(
                     model,
@@ -290,7 +285,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> void delete(@NonNull Class<T> modelType, @NonNull QueryPredicate predicate)
             throws DataStoreException {
         Await.result(
-            operationTimeoutMs,
             (Consumer<Object> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.delete(
                     modelType,
@@ -328,7 +322,6 @@ public final class SynchronousStorageAdapter {
     public <T extends Model> DataStoreException deleteExpectingError(
             @NonNull T model, @NonNull QueryPredicate predicate) {
         return Await.error(
-            operationTimeoutMs,
             (Consumer<StorageItemChange<T>> onResult, Consumer<DataStoreException> onError) ->
                 asyncDelegate.delete(
                     model,
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/ModelUpgradeSQLiteInstrumentedTest.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/storage/sqlite/ModelUpgradeSQLiteInstrumentedTest.java
@@ -106,7 +106,6 @@ public final class ModelUpgradeSQLiteInstrumentedTest {
         // Initialize StorageAdapter with models
         sqliteStorageAdapter = SQLiteStorageAdapter.forModels(schemaRegistry, modelProvider);
         List<ModelSchema> firstResults = Await.result(
-            SQLITE_OPERATION_TIMEOUT_MS,
             (Consumer<List<ModelSchema>> onResult, Consumer<DataStoreException> onError) -> {
                 try {
                     sqliteStorageAdapter.initialize(context, onResult, onError,
@@ -144,7 +143,6 @@ public final class ModelUpgradeSQLiteInstrumentedTest {
 
         // Now, initialize storage adapter with the new models
         List<ModelSchema> secondResults = Await.result(
-            SQLITE_OPERATION_TIMEOUT_MS,
             (Consumer<List<ModelSchema>> onResult, Consumer<DataStoreException> onError) -> {
                 try {
                     sqliteStorageAdapter.initialize(context, onResult, onError,
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
@@ -27,9 +27,11 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -47,6 +49,7 @@ public final class DataStoreConfiguration {
     static final boolean DEFAULT_DO_SYNC_RETRY = false;
     static final int MAX_RECORDS = 1000;
     static final long MAX_TIME_SEC = 2;
+    static final int DEFAULT_OUTBOX_ERROR_RESTART_DELAY = 30;
 
     private final DataStoreErrorHandler errorHandler;
     private final DataStoreConflictHandler conflictHandler;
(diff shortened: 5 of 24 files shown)


joekiller:
Unit tests were verified to run successfully on several workstations using

```
 openjdk version "1.8.0_322"
 OpenJDK Runtime Environment (build 1.8.0_322-bre_2022_02_28_15_01-b00)
 OpenJDK 64-Bit Server VM (build 25.322-b00, mixed mode)
```

```
 openjdk 11.0.16 2022-07-19
 OpenJDK Runtime Environment Homebrew (build 11.0.16+0)
 OpenJDK 64-Bit Server VM Homebrew (build 11.0.16+0, mixed mode)
```

```
 openjdk version "1.8.0_345"
 OpenJDK Runtime Environment (build 1.8.0_345-b01)
 OpenJDK 64-Bit Server VM (build 25.345-b01, mixed mode)
```

joekiller:
A little more testing is making me very happy with this solution. In my [previous comment]  I had mentioned a theoretical maximum number of offline records of 27868 which in my stress testbed I have blown by and have recorded 172830 records, all while offline. That's a little over 6x the previous maximum. The testbed is creating approximately 167 records a minute. All memory consumption is very nominal. Needless to say, I'm very happy with this.


[previous comment]: https://github.com/aws-amplify/amplify-android/issues/1869#issuecomment-1203956750

mikepschneider:
Hello @joekiller we are evaluating this PR.  Can you please update it to resolve the conflicts?

mikepschneider:
@joekiller The changes regarding disabling sync or subscriptions per type will introduce differences across platforms that will require more consideration but we like the changes regarding memory issues and timeouts.  If you can split those into a separate PR we should be able to accept them.

joekiller:
@mikepschneider I'll work to split it per your request.

joekiller:
@mikepschneider I'm closing this for https://github.com/aws-amplify/amplify-android/pull/2217

PR #1939, opened 2022-09-07
fix(auth): fix auth states when invoking signup related APIs

gpanshu:

*Issue #, if available:*

*Description of changes:*
Allowed for signup, signin, confirmsignup, resendsignupcode to be able to function if authentication state is in signedout state.

- [x] Added Unit Tests
- [ ] Added Integration Tests

- [ ] No

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
@@ -131,19 +131,14 @@ internal class RealAWSCognitoAuthPlugin(
         onError: Consumer<AuthException>
     ) {
         authStateMachine.getCurrentState { authState ->
-            if (authState.authNState is AuthenticationState.Configured ||
-                authState.authNState is AuthenticationState.SignedOut
-            ) {
-                GlobalScope.launch {
+            when (authState.authNState) {
+                is AuthenticationState.NotConfigured -> onError.accept(
+                    AWSCognitoAuthExceptions.NotConfiguredException()
+                )
+                is AuthenticationState.SignedIn, is AuthenticationState.SignedOut -> GlobalScope.launch {
                     _signUp(username, password, options, onSuccess, onError)
                 }
-            } else {
-                onError.accept(
-                    AuthException(
-                        "Sign up failed.",
-                        "Cognito User Pool not configured. Please check amplifyconfiguration.json file."
-                    )
-                )
+                else -> onError.accept(AuthException.InvalidStateException())
             }
         }
     }
@@ -223,19 +218,14 @@ internal class RealAWSCognitoAuthPlugin(
         onError: Consumer<AuthException>
     ) {
         authStateMachine.getCurrentState { authState ->
-            if (authState.authNState is AuthenticationState.Configured ||
-                authState.authNState is AuthenticationState.SignedOut
-            ) {
-                GlobalScope.launch {
+            when (authState.authNState) {
+                is AuthenticationState.NotConfigured -> onError.accept(
+                    AWSCognitoAuthExceptions.NotConfiguredException()
+                )
+                is AuthenticationState.SignedIn, is AuthenticationState.SignedOut -> GlobalScope.launch {
                     _confirmSignUp(username, confirmationCode, options, onSuccess, onError)
                 }
-            } else {
-                onError.accept(
-                    AuthException(
-                        "Confirm sign up failed.",
-                        "Cognito User Pool not configured. Please check amplifyconfiguration.json file."
-                    )
-                )
+                else -> onError.accept(AuthException.InvalidStateException())
             }
         }
     }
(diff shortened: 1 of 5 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -127,7 +127,9 @@ internal class RealAWSCognitoAuthPlugin(
         onError: Consumer<AuthException>
     ) {
         authStateMachine.getCurrentState { authState ->
-            if (authState.authNState is AuthenticationState.Configured) {
+            if (authState.authNState is AuthenticationState.Configured ||
```
tylerjroach, line 130: I still don't think this is quite right. Ex: if were are not in a Configured or SignedOut state, we will throw an error that Amplify configuration file is problematic which really isn't the case.

tylerjroach (changes requested):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
div5yesh: yeah we should throw InvalidStateException.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/StateTransitionTests.kt
```diff
@@ -299,7 +299,13 @@ class StateTransitionTests : StateTransitionTestBase() {
         token = stateMachine.listen(
             {
                 val authState =
-                    it.takeIf { it is AuthState.Configured && it.authNState is AuthenticationState.SignedOut }
+                    it.takeIf {
+                        it is AuthState.Configured &&
+                            (
+                                it.authNState is AuthenticationState.SignedOut ||
+                                    it.authNState is AuthenticationState.Configured
```
div5yesh, line 306: Configured is not final state, so AuthN will always be in either signedIn or signedOut state.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -60,8 +60,8 @@ public AuthException(
      */
     public static class InvalidStateException extends AuthException {
         private static final long serialVersionUID = 1L;
-        private static final String MESSAGE = "Auth state reached an invalid state.";
-        private static final String RECOVERY_SUGGESTION = "Please reset auth plugin and reattempt the operation.";
+        private static final String MESSAGE = "Auth state is an invalid state, cannot process the request.";
```
sktimalsina, line 63: nit: Are customers aware of state machine? If not, how about just having something like `Auth is in an invalid state` etc? Can we also add the context on *what* the request is? if not possible here, we can define that where the exception is thrown.

core/src/main/java/com/amplifyframework/auth/AuthException.java
```diff
@@ -60,8 +60,8 @@ public AuthException(
...
+        private static final String RECOVERY_SUGGESTION = "Please reset auth plugin or reattempt the operation later.";
```
sktimalsina, line 64: nit: `later` does not sound right. How do customer know when later is?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -221,17 +218,14 @@ internal class RealAWSCognitoAuthPlugin(
         onError: Consumer<AuthException>
     ) {
         authStateMachine.getCurrentState { authState ->
-            if (authState.authNState is AuthenticationState.Configured) {
-                GlobalScope.launch {
+            when (authState.authNState) {
+                is AuthenticationState.NotConfigured -> onError.accept(
+                    AWSCognitoAuthExceptions.NotConfiguredException()
+                )
+                is AuthenticationState.SignedIn, is AuthenticationState.SignedOut -> GlobalScope.launch {
```
tylerjroach, line 225: Why are we allowing when already SignedIn? The previous working commit was SigningUp, not SignedIn. Shouldn't we be checking that?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -285,17 +279,14 @@ internal class RealAWSCognitoAuthPlugin(
...
```
tylerjroach, line 286: See similar comment on SignedIn vs SigningUp

tylerjroach (approved):

tylerjroach (changes requested):
See comments from last review

tylerjroach (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
tylerjroach: Resolving after discussing

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
tylerjroach: Resolving after discussing

core/src/main/java/com/amplifyframework/auth/AuthException.java
tylerjroach: @sktimalsina Agree with these comments, but can we fast-follow with a second PR? This PR will resolve signup's which was otherwise making work difficult.

sktimalsina (approved):

PR #1941, opened 2022-09-07
feat(auth): publish hub events for auth

sktimalsina:

*Issue #, if available:*

*Description of changes:*
- Added hub events publication for sign in/ out and delete user events

- [ ] Added Unit Tests
- [x] Added Integration Tests

_Note that delete user integration test has not been added as it will require to delete the integration test user._

- [ ] No

### DIFF
--- a/aws-auth-cognito/src/androidTest/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPluginInstrumentationTests.kt
+++ b/aws-auth-cognito/src/androidTest/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPluginInstrumentationTests.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.cognito
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.amplifyframework.auth.AuthChannelEventName
+import com.amplifyframework.auth.cognito.testutils.Credentials
+import com.amplifyframework.hub.HubChannel
+import com.amplifyframework.testutils.HubAccumulator
+import com.amplifyframework.testutils.sync.SynchronousAuth
+import java.lang.RuntimeException
+import java.util.concurrent.TimeUnit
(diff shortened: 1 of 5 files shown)


aws-auth-cognito/src/androidTest/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPluginInstrumentationTests.kt
```diff
@@ -0,0 +1,107 @@
...
+        auth?.signOut()
+        signInWithCognito()
+
+        signInAccumulator.await(10, TimeUnit.SECONDS)
+        signOutAccumulator.await(10, TimeUnit.SECONDS)
+        // if we made it this far without timeout, it means hub event was received
+    }
+
+    // This compliments the hub_Events_Are_Received_Only_Once_Per_Change test
+    @Test(expected = RuntimeException::class)
```
div5yesh, line 88: Should we also test that 2 events are expected but only one was received?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -1383,6 +1386,20 @@ internal class RealAWSCognitoAuthPlugin(
...
     }
 
+    private fun dispatchHubEvent(authState: AuthState) {
+        when {
+            authState.authNState is AuthenticationState.SignedIn -> AuthChannelEventName.SIGNED_IN
+            authState.authNState is AuthenticationState.SignedOut -> AuthChannelEventName.SIGNED_OUT
+            authState.authZState?.deleteUserState is DeleteUserState.UserDeleted -> AuthChannelEventName.USER_DELETED
+            else -> null
+        }?.takeUnless {
+            it != lastPublishedHubEventName.get()
```
div5yesh, line 1396: If current authState == new authState then [listeners](https://github.com/aws-amplify/amplify-android/blob/dev-preview/aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/StateMachine.kt#L188) are not invoked. I don't think you need to keep track of previous event. Maybe test it out once to confirm.

aws-auth-cognito/src/androidTest/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPluginInstrumentationTests.kt
sktimalsina: I think that case will be covered by this test as well and that it would not be necessary. Note that customers won't be using HubAccumulator in their production code as this is from `testutils` but just the [listeners](https://docs.amplify.aws/lib/auth/auth-events/q/platform/android/).
Let me know if you have any requirements in mind.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
sktimalsina: Yes, I added this case after observing that. Technically, Auth states will be different when either of `authN` or `authZ` changes right? So when say, `authZ` changes while `authN` stays as `SignedIn`, listeners will still be invoked.

sdhuka (approved):

PR #1954, opened 2022-09-13
fix(auth): SignIn call when already signed in should not always call onSuccess

gpanshu:
- [x ] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*

*Description of changes:*

- [x ] Added Unit Tests
- [x] Added Integration Tests

- [ x] No


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -397,9 +394,24 @@ internal class RealAWSCognitoAuthPlugin(
                     onSuccess,
                     onError
                 )
-                is AuthenticationState.SignedIn -> onSuccess.accept(
-                    AuthSignInResult(true, AuthNextSignInStep(AuthSignInStep.DONE, mapOf(), null))
-                )
+                is AuthenticationState.SignedIn -> {
```
tylerjroach, line 397: To match iOS and old version, we should not do a username check. If signed in and another sign in is attempted, fail.

tylerjroach (changes requested):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: Done

tylerjroach (approved):

sktimalsina (approved):

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewTestActivityTest.kt
```diff
@@ -70,6 +70,7 @@ class MapViewTestActivityTest {
                 activity.mapView.setStyle { style ->
                     continuation.resume(style)
                 }
+                signOutFromCognito(activity.auth)
```
div5yesh, line 73: I think we should investigate the problem with auth SM flow or test case. Signing out after every case is not the best solution since tests were passing earlier.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -1366,7 +1368,7 @@ internal class RealAWSCognitoAuthPlugin(
                 if (authState is AuthState.Configured) {
                     val (authNState, authZState) = authState
                     when {
-                        authNState is AuthenticationState.SignedOut && authZState is AuthorizationState.Configured -> {
+                        authNState is AuthenticationState.SignedOut -> {
```
div5yesh, line 1371: AuthZ needs to be in Configured state for a successful signout.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -397,9 +394,15 @@ internal class RealAWSCognitoAuthPlugin(
                     onSuccess,
                     onError
                 )
-                is AuthenticationState.SignedIn -> onSuccess.accept(
-                    AuthSignInResult(true, AuthNextSignInStep(AuthSignInStep.DONE, mapOf(), null))
-                )
+                is AuthenticationState.SignedIn -> {
+                    onError.accept(
+                        AuthException(
```
div5yesh, line 399: Should we use `SignedInException` instead?

tylerjroach: @div5yesh I believe the messaging here aligns with iOS. SignedInException messaging is close, but in this instance, we would lose messaging parity.

div5yesh: We can change the messaging in SignedInException if it doesn't affect parity for any other cases where the exception is used.

PR #1956, opened 2022-09-14
feat(analytics): Remove Analytics notification client

tjleing:
*Description of changes:*
Removing the stub notification client that was there to provide attributes when updating `EndpointProfile`.

Ran unit tests, tested locally to make sure that Pinpoint didn't require these features.

*Note:*
Forcing `optOut: ALL` since we don't support notifications here, but customers can take the escape hatch and call `updateEndpoint()` after each `identifyUser()` to set `optOut` back to `NONE` and re-enable them.  We believe that would be the case even without this forced, and it makes sure that users don't get notifications by mistake.


gpanshu (approved):

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/targeting/endpointProfile/EndpointProfile.kt
```diff
@@ -45,16 +42,6 @@ internal class EndpointProfile(
     private val metrics: MutableMap<String, Double> = ConcurrentHashMap()
     private val currentNumOfAttributesAndMetrics = AtomicInteger(0)
 
-    /**
-     * Returns whether the endpoint is opted out of notification.
-     *
-     * @return String (ALL | NONE)
-     */
-    val optOut: String
```
sdhuka, line 53: what is the default value for optOut? should we just pass "NONE" until we have notifications category?

sdhuka (approved):

PR #1996, opened 2022-09-30
feat(pinpoint): Add escape hatch and corresponding integration tests

tjleing:

Added escape hatch + version, also re-enabled the integration tests that were failing due to not being able to access the endpoint after `identifyUser`.

If we have time (not a blocker) we can investigate why the SDK seems to not be returning all of the properties of the endpoint location, and why passing attributes doesn't exactly work the way I'd expect (not passing an attribute doesn't erase it from the user profile, ~seems to maybe be a long lag time between sending attributes and seeing them back~ wait, is that because I've been running multiple tests in a row and it gets deleted??, etc.)  Not a blocker because I'm confident that this does not represent a regression from `main` (Android SDK) based on testing.

- [ ] Added Unit Tests
- [x] Added Integration Tests

- [x] No


sdhuka (approved):

sktimalsina (approved):

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointManager.kt
```diff
@@ -39,16 +39,16 @@ internal class PinpointManager constructor(
...
     val sessionClient: SessionClient
     val targetingClient: TargetingClient
+    internal val pinpointClient: PinpointClient = PinpointClient {
+        credentialsProvider = this@PinpointManager.credentialsProvider
+        region = awsPinpointConfiguration.region
+    }
 
     companion object {
-        private const val SDK_NAME = "AMPLIFY-ANDROID"
+        private const val SDK_NAME = "amplify-android"
```
sktimalsina, line 48: nit: You might check if we can pull this from `BUILD` constants.

Issue #2013, opened 2022-10-10
clientMetadata can't be passed into AWSCognitoAuthPlugin::signIn


wangf1978:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
// Put output below this line
    // Amplify core dependency
    def amplify_version = "1.37.3"
    implementation "com.amplifyframework:core:$amplify_version"
    implementation "com.amplifyframework:aws-auth-cognito:$amplify_version"
    implementation "com.amplifyframework:aws-storage-s3:$amplify_version"
    implementation "com.amplifyframework:aws-api:$amplify_version"
    implementation "com.amplifyframework:aws-datastore:$amplify_version"

```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 7.2
------------------------------------------------------------

Build time:   2021-08-17 09:59:03 UTC
Revision:     a773786b58bb28710e3dc96c4d1a7063628952ad

Kotlin:       1.5.21
Groovy:       3.0.8
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.12 (JetBrains s.r.o 11.0.12+7-b1504.28-7817840)
OS:           Windows 10 10.0 amd64
```

</details>

### Describe the bug
Now metadata is passed into the parameter 'validationData' in this function, but the parameter 'clientMetadata' is always using a fixed value Collections.emptyMap()
If so, when signing in successfully with username/password, in the Post authentication Lambda trigger of AWS Cognito, 
<img width="889" alt="image" src="https://user-images.githubusercontent.com/7458444/194892705-11d98c02-89fd-4014-9bb2-cd31791c381d.png">

clientMetadata in this trigger parameters can't have the specified key/value pairs from android client side, but it can be set by amplify-js and amplify-ios.

[...]

PR #2024, opened 2022-10-12
fix(auth): Add clientMetadata to signIn

tylerjroach:

*Issue #, if available:* [203](https://github.com/aws-amplify/amplify-android/issues/2013)

*Description of changes:*
Passes customer provided clientMetadata through to MobileClient sign in.

Pending manual verification

- [ ] Added Unit Tests
- [ ] Added Integration Tests

- [ ] No


tylerjroach:
Summary of this issue is that we were sending empty client metadata (Collections.emptyMap) to the signIn call. Instead, we were passing clientMetadata into the validationData field. iOS has this reversed, passing clientMetadata to the proper field and sending validationData of nil. I'm a bit hesitant to change what we send to validationData in case a customer became reliant on accessing the metadata this way. From the report, validationData will show in the pre authentication lambda trigger, but it is not accessible in the post authentication lambda trigger.

div5yesh (approved):

wangf1978:
@tylerjroach I have the same concern with you, if validationData is nil, it will cause another issue, the pre authentication lambda trigger can't have validationData, according to our test, although amplify-ios passed the nil value, but finally it seems it fills validationData with clientMetadata in somewhere, we didn't dig into this case in amplify-ios.
In amplify-js, there is a global clientMetadata passed from amplfy.config, metadata passed in Amplify Auth signIn will work as valididationData, but the global clientMetadata will work as clientMetdata. for this part, I am also a little confused about the implementation and lambada trigger, signIn is related with pre/post authentication lambda trigger, but these 2 lambda trigger need 2 different metadata from clientSide.

sdhuka (approved):

PR #2074, opened 2022-10-31
feat(auth): Track pinpoint endpoint id if set for auth

tylerjroach:

*Issue #, if available:*

*Description of changes:*
This ticket also fixes an issue in Pinpoint where the SharedPreference name did not match v1, which would cause an issue from v1 -> v2 transition.

- [ ] Added Unit Tests
- [ ] Added Integration Tests

- [ ] No


aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointManager.kt
```diff
@@ -46,12 +46,26 @@ internal class PinpointManager constructor(
...
 
     init {
         val pinpointDatabase = PinpointDatabase(context)
-        val sharedPrefs =
-            context.getSharedPreferences(awsPinpointConfiguration.appId, Context.MODE_PRIVATE)
+
+        /*
+        Auth plugin needs to read from Pinpoint shared preferences, but we don't currently have an architecture
+        that allows the plugins to pass data between each other. If the storage mechanism of UniqueId changes, we
+        must also refactor AuthEnvironment from the Cognito Auth Plugin.
```
gpanshu, line 63: Duplicated comment.

gpanshu (approved):
Just a couple of documentation points but either way good to go.

tylerjroach:
> Just a couple of documentation points but either way good to go.

I purposefully placed it next to prefix and shared prefs call. If either of these things change, we need to be aware of the implication. Unfortunately, with the way we are pulling data here, Android Studio is not going to catch this trough refactor commands.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/AuthConfiguration.kt
```diff
@@ -55,7 +56,10 @@ internal data class AuthConfiguration internal constructor(
                     pluginJson.optJSONObject("Auth")
                         ?.optJSONObject(configName)
                         ?.optString("authenticationFlowType")
-                )
+                ),
+                pinpointAppId = pluginJson.optJSONObject("PinpointAnalytics")
```
div5yesh, line 60: Looks like there is some disparity with iOS on how we are reading the AppId: https://github.com/aws-amplify/amplify-swift/pull/2371/files#diff-9cb850c74bd6d5694784f8790c80f355906e130d46e5af241f1d17bb3f576e45R27

I think what we have makes more sense. Please sync with iOS.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointManager.kt
```diff
@@ -46,12 +46,26 @@ internal class PinpointManager constructor(
...
+         */
+        val sharedPrefs = context.getSharedPreferences(
```
div5yesh, line 65: Do we need to migrate any legacy properties?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -33,14 +34,50 @@ import java.util.Date
...
+        Auth plugin needs to read from Pinpoint shared preferences, but we don't currently have an architecture
+        that allows the plugins to pass data between each other. We are retrieving the pinpointEndpointId by reading
+        Pinpoint preferences constructed from pinpointAppId + a shared prefs suffix. If the storage of UniqueId changes
+        in Pinpoint, we need to update here as well.
+         */
+    fun getPinpointEndpointId(): String? {
+        if (configuration.pinpointAppId == null) return null
+        if (cachedPinpointEndpointId != null) return cachedPinpointEndpointId
+
+        val pinpointPrefs = context.getSharedPreferences(
```
div5yesh, line 70: Do we need to migrate any legacy properties?

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointManager.kt
div5yesh, line 65: Should we namespace the stored values as per credential store namespaces?  So that we don't have to add migration later when we use credential store for all other categories.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -33,14 +34,50 @@ import java.util.Date
...
     val cognitoAuthService: AWSCognitoAuthServiceBehavior,
     val credentialStoreClient: StoreClientBehavior,
     private val userContextDataProvider: UserContextDataProvider? = null,
     val hostedUIClient: HostedUIClient?,
     val logger: Logger
 ) : Environment {
+
+    companion object {
+        /*
+        Auth plugin needs to read from Pinpoint shared preferences, but we don't currently have an architecture
```
div5yesh, line 48: Credential store is supposed to be the tool that allows plugins to share data.

tylerjroach: This is not a credential so we should look into possibly renaming, if that is the case.

aws-analytics-pinpoint/src/main/java/com/amplifyframework/analytics/pinpoint/PinpointManager.kt
tylerjroach: No, right now we are not migrating the file from v1 to v2. The namespace used here is intended to match exactly to v1.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
tylerjroach: We are not migrating at the moment. This is a direct read from v1.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/AuthConfiguration.kt
tylerjroach: Correct, we discovered this and are setting up a discussion. v1 read with iOS property as well but it doesn't appear that CLI event adds that field. There are a few bugs around this feature in v1 we will also be discussing. I'm moving this PR to draft until we resolve how to handle these.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -33,14 +35,57 @@ import java.util.Date
...
+        that allows the plugins to pass data between each other. We are duplicating this suffix constant because it
+        is internal to the Pinpoint class, which analytics does not pull in. If the Pinpoint suffix is updated, this
+        needs updated as well.
+         */
+        const val PINPOINT_SHARED_PREFS_SUFFIX = "515d6767-01b7-49e5-8273-c8d11b0f331d"
```
sdhuka, line 54: would be good to add this const in core so both modules use the same value.

sdhuka (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
tylerjroach: I believe we may soon need an aws-core module hold things specific to aws. It doesn't belong in Core because this is Pinpoint related and ideally, all things core should be abstract.

tjleing (approved):

PR #2113, opened 2022-11-15
fix(auth): Prevent attempting to read backed up EncryptedSharedPreferences

tylerjroach:

*Issue #, if available:*

*Description of changes:*
EncryptedSharedPreferences are backed up with the rest of SharedPreferences files depending on app backup settings. As a library, we can't control backup configurations as it would conflict with app developers config.

As a workaround, we create a UUID installationIdentifier in noBackupDir to add as a suffix to our EncryptedSharedPreferences, ensuring that file we are attempting to read was not created from a backup.

- [ ] Added Unit Tests
- [ ] Added Integration Tests

- [ ] No

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
@@ -22,36 +22,81 @@ import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV
 import androidx.security.crypto.EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
 import androidx.security.crypto.MasterKeys
+import java.io.File
+import java.util.UUID
 
 internal class EncryptedKeyValueRepository(
     private val context: Context,
     private val sharedPreferencesName: String,
 ) : KeyValueRepository {
 
-    override fun put(dataKey: String, value: String?) {
-        with(getSharedPreferences().edit()) {
-            putString(dataKey, value)
-            apply()
-        }
-    }
-
-    override fun get(dataKey: String): String? = getSharedPreferences().getString(dataKey, null)
-
-    override fun remove(dataKey: String) {
-        with(getSharedPreferences().edit()) {
-            remove(dataKey)
-            apply()
-        }
-    }
-
     @VisibleForTesting
-    fun getSharedPreferences(): SharedPreferences {
-        return EncryptedSharedPreferences.create(
-            sharedPreferencesName,
+    internal val sharedPreferences: SharedPreferences by lazy {
+        EncryptedSharedPreferences.create(
+            "$sharedPreferencesName.${getInstallationIdentifier(context, sharedPreferencesName)}",
             MasterKeys.getOrCreate(MasterKeys.AES256_GCM_SPEC),
             context,
             AES256_SIV,
             AES256_GCM
         )
+    }
+
+    @VisibleForTesting
+    internal val editor: SharedPreferences.Editor by lazy {
+        sharedPreferences.edit()
+    }
+
+    override fun put(dataKey: String, value: String?) {
+        with(editor) {
+            putString(dataKey, value)
+            apply()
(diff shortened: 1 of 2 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
```diff
@@ -22,36 +22,81 @@ import androidx.security.crypto.EncryptedSharedPreferences
...
+     * Creates a new installation identifier for the install
+     */
+    private fun createInstallationIdentifier(identifierFile: File): String {
+        val newIdentifier = UUID.randomUUID().toString()
+        try {
+            identifierFile.writeText(newIdentifier)
+        } catch (e: Exception) {
+            // Failed to write identifier to file, session will be forced to be in memory
+        }
+        return newIdentifier
```
banji180, line 100: any harm in still returning newIdentifier if line 96 throws an exception?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
```diff
@@ -22,36 +22,81 @@ import androidx.security.crypto.EncryptedSharedPreferences
...
+
+    /**
+     * Creates a new installation identifier for the install
+     */
+    private fun createInstallationIdentifier(identifierFile: File): String {
+        val newIdentifier = UUID.randomUUID().toString()
+        try {
+            identifierFile.writeText(newIdentifier)
+        } catch (e: Exception) {
+            // Failed to write identifier to file, session will be forced to be in memory
```
sktimalsina, line 98: 1. interesting! no warning for empty catch?
2. >  session will be forced to be in memory
How would this work?

sktimalsina (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
tylerjroach: @banji180 @sktimalsina You are both commenting on the same area. I should probably try to add logger in here and add a warning. This should never happen in practice as the noBackupDir app directory should alway be present and writeable. If there were some fatal error, there is really no graceful way to recover from this.

We could:
* Crash the application
* Use in memory storage (which obviously wouldn't retain credentials on a restart. 

I would want to avoid crashing the application at all costs.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
```diff
@@ -22,36 +22,81 @@ import androidx.security.crypto.EncryptedSharedPreferences
...
-            AES256_GCM
-        )
+    /**
+     * EncryptedSharedPreferences may have been backed up by the application, but will be unreadable due to the
+     * KeyStore record being lost. To prevent an unreadable EncryptedSharedPreferences, we append a suffix to the name
+     * with a UUID created in the noBackupFilesDir
+     */
+    @Synchronized
+    private fun getInstallationIdentifier(context: Context, keyValueRepoID: String): String {
+        val identifierFile = File(context.noBackupFilesDir, "$keyValueRepoID.installationIdentifier")
```
div5yesh, line 72: 🚢

div5yesh (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/EncryptedKeyValueRepository.kt
div5yesh: > Use in memory storage (which obviously wouldn't retain credentials on a restart.

Should we set identifier as empty string in case of fatal error so that app can retain credentials on a restart?

tylerjroach: I don't think so. This would mean that we would end up in the same situation as the original crash if an uninstall/reinstall occurred. Once again, this should never really happen in practice on a functioning Android device.

PR #2148, opened 2022-12-01
(feat) Add a network status listener to restart DataStore after the network comes online

mikepschneider:
Add a network status listener to restart DataStore after the network comes online. This brings Android closer in line with ios and js behavior.

- [x ] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*

*Description of changes:*

This change adds a ReachabilityMonitor that listens to the state of the network and publishes an event when the network is gained or lost.  AWSDataStorePlugin listens to these events and calls start() and stop() when the network is gained or lost. 

Unit tests, manual tests

- [ x] Added Unit Tests
- [ ] Added Integration Tests

- [ x] No


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ReachabilityMonitor.kt
```diff
@@ -0,0 +1,78 @@
...
+            val callback = getCallback(emitter)
+            context.getSystemService(ConnectivityManager::class.java).registerDefaultNetworkCallback(callback)
+        }
+        getObservable(emitter)
+            .subscribe()
+    }
+
+    internal fun getObservable(observable: ObservableOnSubscribe<Boolean>): Observable<Boolean> {
+        return Observable.create(observable)
+            .debounce(250, TimeUnit.MILLISECONDS)
```
manueliglesias, line 55: Let's use the overload that accepts an scheduler so we can simulate time moving on in unit tests (http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/schedulers/TestScheduler.html)

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -130,6 +135,7 @@ private AWSDataStorePlugin(@NonNull Builder builder) throws DataStoreException {
             SQLiteStorageAdapter.forModels(schemaRegistry, modelProvider) :
             builder.storageAdapter;
         this.categoryInitializationsPending = new CountDownLatch(1);
+        this.reachabilityMonitor = new ReachabilityMonitor();
```
manueliglesias, line 138: Why is this not available to pass in the builder if needed? (e.g. for cleaner tests)

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -255,6 +261,30 @@ public void configure(
...
+                        (Action) () -> { },
+                        ((e) -> LOG.error("Error starting datastore plugin after network event: " + e))
+                    );
+                } else {
+                    LOG.info("Network lost");
+                    start(
+                            (Action) () -> { },
+                            ((e) -> LOG.error("Error stopping datastore plugin after network event: " + e))
+                    );
+                }
```
manueliglesias, line 285: `start` is being called in both "Network gained" and "Network lost", is this correct?

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ReachabilityMonitor.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+    }
+
+    private fun publishNetworkStatusEvent(active: Boolean) {
+        Amplify.Hub.publish(
+            HubChannel.DATASTORE,
+            HubEvent.create(DataStoreChannelEventName.NETWORK_STATUS, NetworkStatusEvent(active))
+        )
+    }
```
manueliglesias, line 77: If I am reading this correctly, the reachability monitor is not subscribable itself, but just publishes to Hub, right?

My recommendation here is to make the ReachabilityMonitor observable (and make it emmit the current state as soon as a busbscriber subscribes)

manueliglesias (commented):
Can you elaborate a little on why is the ReachabilityMonitor used in the plugin and not in the orchestrator?

mikepschneider:
> Can you elaborate a little on why is the ReachabilityMonitor used in the plugin and not in the orchestrator?

I originally was calling stop/start in the Orchestrator.  But that was not the right place to do it, it caused everything to get messed up.  I realized that the Stop/Start methods of the plugin are meant to be called externally unlike Orchestrator and that's the right place to stop and start the datastore.  When I tried it out it works correctly.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ReachabilityMonitor.kt
mikepschneider: Done

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
mikepschneider: Done

manueliglesias (approved):
LGTM 👍

manueliglesias (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/AWSDataStorePlugin.java
```diff
@@ -255,6 +262,33 @@ public void configure(
             event -> InitializationStatus.SUCCEEDED.toString().equals(event.getName()),
             event -> categoryInitializationsPending.countDown()
         );
+
+        reachabilityMonitor.configure(context);
```
tylerjroach, line 266: Is configure needed, or can this just be done in the init of the ReachabilityMonitor? Side effect right now of forgetting configure is a crash, and I don't see too much of a reason not just to allow class to configure itself since it holds on to the current network status to pass to a newly attached observer.

tylerjroach (approved):

joekiller:
@mikepschneider while working on a new PR for https://github.com/aws-amplify/amplify-android/pull/1876#issuecomment-1362044930 I noticed this new functionality is causing issues with the tests and potentially can cause unneeded orchestrator thrashing.

As a user I expect the data store to not be triggered unless operated upon and not the be activated during configuration. Actual behavior: reachabilityMonitor will start the datastore when it detect network even if the datastore is supposed to be stopped, cleared, or is supposed to be inactive.

This breach of contract manifests primarily in datastore crashes during tests. It will also cause the datastore to start unintentionally. 

I understand the point is to have this closer in implementation to the iOS/js stuff however this is really brute force thrashing for the orchestrator as is and introduces denial of service whenever the network transitions while the orchestrator does all it's overhead.

joekiller:
mmm. I feel like I'm hitting it though. will look a little more.

PR #2184, opened 2022-12-15
fix(auth): Fix for when loading credentials the success/error is fired twice

gpanshu:

*Issue #, if available:*

*Description of changes:* Fix for when loading credentials the success/error is fired twice it returns an already resumed error in the continuation

- [ ] Added Unit Tests
- [ ] Added Integration Tests

- [x] No

### DIFF
--- a/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/PinpointAnalyticsInstrumentationTest.kt
+++ b/aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/PinpointAnalyticsInstrumentationTest.kt
@@ -34,18 +34,20 @@ import com.amplifyframework.auth.cognito.AWSCognitoAuthPlugin
 import com.amplifyframework.core.Amplify
 import com.amplifyframework.hub.HubChannel
 import com.amplifyframework.hub.HubEvent
+import com.amplifyframework.logging.AndroidLoggingPlugin
+import com.amplifyframework.logging.LogLevel
 import com.amplifyframework.testutils.HubAccumulator
 import com.amplifyframework.testutils.Resources
 import com.amplifyframework.testutils.Sleep
 import com.amplifyframework.testutils.sync.SynchronousAuth
-import java.util.UUID
-import java.util.concurrent.TimeUnit
 import kotlinx.coroutines.runBlocking
 import org.json.JSONException
 import org.junit.Assert
 import org.junit.Before
 import org.junit.BeforeClass
 import org.junit.Test
+import java.util.UUID
+import java.util.concurrent.TimeUnit
 
 class PinpointAnalyticsInstrumentationTest {
     @Before
@@ -388,9 +390,11 @@ class PinpointAnalyticsInstrumentationTest {
             setUniqueId()
             Amplify.Auth.addPlugin(AWSCognitoAuthPlugin() as AuthPlugin<*>)
             Amplify.addPlugin(AWSPinpointAnalyticsPlugin())
+            Amplify.Logging.addPlugin(AndroidLoggingPlugin(LogLevel.DEBUG))
             Amplify.configure(context)
             Sleep.milliseconds(COGNITO_CONFIGURATION_TIMEOUT)
             synchronousAuth = SynchronousAuth.delegatingTo(Amplify.Auth)
+
         }
 
         private fun setUniqueId() {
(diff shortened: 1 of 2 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/CredentialStoreClient.kt
```diff
@@ -56,9 +56,9 @@ internal class CredentialStoreClient(configuration: AuthConfiguration, context:
     ) {
         var capturedSuccess: Result<AmplifyCredential>? = null
         var capturedError: Exception? = null
-        val token = StateChangeListenerToken()
+        var token: StateChangeListenerToken? = StateChangeListenerToken()
         credentialStoreStateMachine.listen(
-            token,
+            token as StateChangeListenerToken,
```
tylerjroach, line 61: Maybe wrap the token call in a `?.let` so you don't have to do a force cast.

tylerjroach (commented):
I'd still like for us to try and create a rapid fire test that replicates the occasional failure here as this fix is masking the problem a bit. credentialStoreStateMachine.cancel() immediately cancels the token through a pending cancellations set.

The race condition here appears that maybe 2 results are fired in `listen` from multiple threads, and since there was no further check for if token was subscribed, the fire could happen twice.

Not sure if this is the right fix though. Other options could be setting a lock on listen callbacks (also feels a bit hacky), as the issue is most likely inside the structuring/threading of CredentialStoreStateMachine. Want to see what others think on this.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/CredentialStoreClient.kt
mattcreaser: Or a simple `!!`

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/CredentialStoreClient.kt
```diff
@@ -72,11 +72,12 @@ internal class CredentialStoreClient(configuration: AuthConfiguration, context:
...
                         val error = capturedError
-                        if (success != null) {
-                            credentialStoreStateMachine.cancel(token)
+                        if (success != null && token != null) {
+                            credentialStoreStateMachine.cancel(token as StateChangeListenerToken)
+                            token = null
                             onSuccess(success)
-                        } else if (error != null) {
-                            credentialStoreStateMachine.cancel(token)
+                        } else if (error != null && token != null) {
```
mattcreaser, line 79: Token also has to be set to null for the error case.

mattcreaser (changes requested):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/CredentialStoreClient.kt
gpanshu: Done

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/CredentialStoreClient.kt
gpanshu: Good catch. Done.

tylerjroach (dismissed):

sdhuka (approved):

sdhuka (approved):

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/PinpointAnalyticsInstrumentationTest.kt
```diff
@@ -53,11 +54,18 @@ class PinpointAnalyticsInstrumentationTest {
         val context = ApplicationProvider.getApplicationContext<Context>()
         @RawRes val resourceId = Resources.getRawResourceId(context, CREDENTIALS_RESOURCE_NAME)
         val userAndPasswordPair = readCredentialsFromResource(context, resourceId)
-        synchronousAuth.signOut()
-        synchronousAuth.signIn(
-            userAndPasswordPair!!.first,
-            userAndPasswordPair.second
-        )
+        try {
```
tylerjroach, line 57: We should not be leaking state between tests.

tylerjroach (changes requested):

aws-analytics-pinpoint/src/androidTest/java/com/amplifyframework/analytics/pinpoint/PinpointAnalyticsInstrumentationTest.kt
gpanshu: reverted as it passes locally without this change

tylerjroach (approved):

sdhuka (approved):

PR #2217, opened 2023-01-05
fix(datastore): fix memory sinks and mutation outbox halting

joekiller:
- [ x] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:* #1869

*Description of changes:*
## Fix memory management
  - Increased the default number of slots allocated by the mutation outbox to 200
  - Limited the mutation outbox to loading the first 100 mutations, as well as any pending mutations related to the first 100 loaded
  - The SubscriptionProcessor will now only hold a maximum of 100 subscription events, with a maximum age of 10 minutes

## Remove timeouts
There are various timeouts that are effectively CPU or network bound. Removing these timeouts makes the system more likely to work on a diverse range of clients.
  - Removed the restart delay for tests, as it was CPU bound and caused predictable failures when a thread slowed down
  - Removed the Orchestrator semaphore timeout

## Fix hanging mutation outbox processor
  - The mutation outbox will now resume operation after encountering uncaught exceptions, disrupted transmissions, or other issues
  - The default resumption delay is 30 seconds, or can be set using the outboxErrorRestartDelay builder option
  - Wrapped the startDrainingMutationOutbox function to further handle emitter thrashing

The changes were tested on an alternative branch. This branch is representative of those changes.

- [ ] Added Unit Tests
- [ ] Added Integration Tests

- [ x] No


joekiller:
oh man these tests... :shrug:

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -184,7 +187,7 @@ public Completable load() {
         return Completable.create(emitter -> {
             inFlightMutations.clear();
             mutationQueue.clear();
-            storage.query(PendingMutation.PersistentRecord.class, Where.matchesAll(),
+            storage.query(PendingMutation.PersistentRecord.class, Where.matchesAll().paginated(Page.firstPage()),
```
mikepschneider, line 190: Loading one page at a time will definitely help the memory pressure compared to loading the entire pending mutation table.  However an alternative which is used in the [Javascript implementation here](https://github.com/aws-amplify/amplify-js/blob/a8f4cbe3a238098a9d916b139fc5771c57149615/packages/datastore/src/sync/processors/mutation.ts#L164) is to pull the items one at a time for processing.  This would eliminate the need for `mutationQueue` altogether and would further reduce memory usage.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
```diff
@@ -195,13 +198,42 @@ public Completable load() {
...
                         }
                     }
-                    // Publish outbox status upon loading
-                    publishCurrentOutboxStatus();
                     emitter.onComplete();
                 },
                 emitter::onError
             );
-        })
+        }).andThen(Observable.fromIterable(mutationQueue.iterator())
```
mikepschneider, line 205: Can you explain the purpose of this section?

mikepschneider (changes requested):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
joekiller: in the previous section the first page of results is loaded and then pushed up. After the mutation queue is drained, if there was more than one page of pending mutations on disk then the next page needs to be loaded in to the queue.

The reason this is after the initial is to ensure that publishCurrentOutboxStatus only fires when there are no pending mutations in the queue and none left in the db.

the original design had a bug that would fire the outbox is empty after the first 100 left the queue but this prevents that situation.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
joekiller: The mutation processor intertwines with the subscription processor in ways that prevented me from straight deleting the mutationQueue in the time I had to make a solution.

I agree nuking it from orbit seems more profitable. Just haven't the margin at the moment and the changes as is have run offline and online for months straight on the testbed.

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -124,6 +127,7 @@ public static DataStoreConfiguration defaults() throws DataStoreException {
                 .doSyncRetry(DEFAULT_DO_SYNC_RETRY)
                 .observeQueryMaxTime(MAX_TIME_SEC)
                 .observeQueryMaxRecords(MAX_RECORDS)
+            .outboxErrorRestartDelay(DEFAULT_OUTBOX_ERROR_RESTART_DELAY)
```
mikepschneider, line 130: We have a large PR in progress [here](https://github.com/aws-amplify/amplify-android/pull/2201) which will catch all errors generated within the outbox, so we don't want this restart logic.

(Also side note this code adds `outboxErrorRestartDelay` to the class builder. Our process is we don't add items to the public-facing api without adding them to all of the platforms, so we would need buyin from the other platforms to accept that part, just FYI)

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/PersistentMutationOutbox.java
mikepschneider: Having some internal discussions about whether we want to accept this as a step in the right direction or wait till we can remove mutationqueue altogether.

mikepschneider (changes requested):

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
joekiller: I'd have to review the other side to understand if I'm comfortable with the efforts to make the outbox processor restart. The code that I concluded with was very effective and did not disrupt the overall orchestrator with stops and starts as the outbox processor is a passive system.

That being said I can remove those components from this PR.

joekiller:
@mikepschneider now that #2201 is merged should I cull this down to memory fixes only while y'all go for long term? Cut it to replay subject fix? I'm looking forward to checking out 2201 some with projects warming up.

The [SubscriptionProcessor ReplaySubject buffer](
https://github.com/aws-amplify/amplify-android/blob/9074d878460bd9e53506c08887515c5563950976/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java#L127) is still unbound and will exhaust memory unless addressed.  I chose to use a bounded buffer [via ReplaySubject.createWithTimeAndSize](https://github.com/aws-amplify/amplify-android/pull/2217/files#diff-147b3e61fa117662abdc6862a5878d4bf0c72040103ade4d9c0db568ee83ecd9R124)

manueliglesias:
Thank you so much for the contribution @joekiller !
I've been looking at the PR and testing it locally, I also took the liberty of updating the branch (and fixing a couple of conflicts along the way)

The PR covers a lot of ground, we would prefer to break it up into more discrete pieces.

As mentioned before, we recently merged #2201 which introduced some recoverability improvements and generally better error handling and restart. The changes from #2201 make some pieces of this PR redundant.

The piece that I really like about your PR and would love to merge in first place are the improvements on the Outbox pagination. The change you do on your PR that paginates the query that loads items into the MutationOutbox's MutationQueue , with the addition of getting only the first item (using `Where.matchesAll().paginated(Page.firstResult()`), means that mutations are loaded into memory one by one (yes, the queue and part of the outbox logic becomes superfluous), the memory concerns are gone and the behavior with regards of the sequential draining is kept the same.

Would you be willing to take this idea forward and update this PR to include just that (or drop this one in favor of a new one)?

joekiller:
@manueliglesias I can give it the chop. Were you actively working on this? I saw the merging of main into take-2

manueliglesias:
@joekiller Yeah, the merges you see was me getting this branch up to date with main, and solving a couple of merge conflicts I got while doing it.

joekiller:
@manueliglesias thanks for the response. Most of this happens on my own time, so I really do not want to set any precedent of turn around. With that said, I'll likely try to hit this up in the weekend  :)

joekiller:
@manueliglesias rebased/force pushed/updated to only address the unbound memory issues.

manueliglesias (changes requested):
Hi @joekiller 

I've been looking at this PR at multiple times and after pondering considerably I don't feel comfortable mergin it as is, while limiting the number of records to 1 when loading the PersistentMutationOutbox seems to help, it comes with other problems we need to address more hollistically (e.g. making sure that Hub events are accurate).

What we can and should totally merge as soon as possible is the change for the UnicastSubject, can you make the PR only about that change? (subscriptions memory pressure/ UnicastSubject)

Apologies for the churn, let's get it merged for the pieces we feel comfortable with and we can separately discuss the outbox OOM issue.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SubscriptionProcessor.java
```diff
@@ -124,7 +126,10 @@ synchronized void startSubscriptions() throws DataStoreException {
 
         // Need to create a new buffer so we can properly handle retries and stop/start scenarios.
         // Re-using the same buffer has some unexpected results due to the replay aspect of the subject.
-        buffer = ReplaySubject.create();
+        // We allow mutations to reside for 10 minutes and collect a maximum of 100 for predicable
+        // memory usage.
+        buffer = ReplaySubject.createWithTimeAndSize(
+            DEFAULT_MAX_AGE_IN_MINUTES, TimeUnit.MINUTES, Schedulers.io(), DEFAULT_MAX_TOTAL_SUBSCRIPTIONS);
```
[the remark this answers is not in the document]
joekiller: @manueliglesias I'm comfortable with you just making a PR. It seems I have pointed out the problems and you all have a grasp on them now. Since you have already done the work, please go ahead and just merge in your change.

manueliglesias:
Closing in favor of #2353

Thank you @joekiller

PR #2221, opened 2023-01-09
feat(geo): Submit multiple batches of locations in one WorkManager run

tjleing:

When WorkManager does work, process all events currently in the DB and put them into batches, instead of the current implementation where we stop after the first batch, which can cause the DB size to grow arbitrarily over time.

- [x] Added Unit Tests

### DIFF
--- a/aws-geo-location/src/test/java/com/amplifyframework/geo/location/tracking/UploadWorkerTest.kt
+++ b/aws-geo-location/src/test/java/com/amplifyframework/geo/location/tracking/UploadWorkerTest.kt
@@ -1,0 +1,257 @@
+/*
+ *
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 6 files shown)


aws-geo-location/src/test/java/com/amplifyframework/geo/location/tracking/UploadWorkerTest.kt
```diff
@@ -0,0 +1,257 @@
...
+        ).build()
+        val now = Instant.now()
+        val locations = (0 until 5).map {
+            LocationEntity(it.toLong(), id, tracker, now.plusMillis(it * 3000L), lat + it * .05, long)
+        }
+        coEvery {mockLocationDao.getAll()}.returns(locations)
+
+        runBlocking {
+            val result = uploadWorker.doWork()
+            assertTrue(result.outputData.size() == 0) // success!
```
mattcreaser, line 218: Using `assertTrue` isn't great since a failure simply yields a message like "expected false to be true". Let's at least use `assertEquals` until we get a better assertion library in place.

aws-geo-location/src/test/java/com/amplifyframework/geo/location/tracking/UploadWorkerTest.kt
```diff
@@ -0,0 +1,257 @@
...
+        pos.deviceID = id
+        pos.timeStamp = Date.from(Instant.now())
+        pos.tracker = tracker
+        return pos
+    }
+
+    private fun assertEq(p1: GeoPosition, p2: GeoPosition) {
+        assertEquals(p1.location.latitude, p2.location.latitude, .001)
+        assertEquals(p1.location.longitude, p2.location.longitude, .001)
+        assert(p1.deviceID == p2.deviceID)
```
mattcreaser, line 254: `assertEquals`

Issue #2282, opened 2023-02-08
signInWithSocialWebUI throws a NullPointerException on 2.2.0 but not on 1.37.10


MrBenjoo:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
[versions]
amplify = "2.2.0"

[libraries]
com.amplifyframework:core-kotlin:2.2.0
com.amplifyframework:aws-auth-cognito:2.2.0

[gradle plugin]
com.amplifyframework:amplify-tools-gradle-plugin:1.0.2
```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 7.5
------------------------------------------------------------

Build time:   2022-07-14 12:48:15 UTC
Revision:     c7db7b958189ad2b0c1472b6fe663e6d654a5103

Kotlin:       1.6.21
Groovy:       3.0.10
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          1.8.0_275 (AdoptOpenJDK 25.275-b01)
OS:           Windows 10 10.0 amd64

```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/lib/auth/social/q/platform/android/#update-androidmanifestxml

### Describe the bug

Calling ```Amplify.Auth.signInWithSocialWebUI(AuthProvider.google(), callingActivity)``` throws a nullpointerexception.

It works when using `core-kotlin:0.21.10` and `aws-auth-cognito:1.37.10`, i am able to sign in without problems. 

### Code Snippet

```kotlin
private suspend fun signInWithSocialWebUI(callingActivity: Activity) = try {
        Amplify.Auth.signInWithSocialWebUI(AuthProvider.google(), callingActivity)
        true
    } catch (exception: Exception) {
        false
    }
```

### Log output

<details>

[...]

PR #2284, opened 2023-02-09
fix(auth): fix npe in initialize fetch auth session

banji180:

*Issue #, if available:*
#2282 
*Description of changes:*
The initiatefetchauthsession throws npe when id token is null. The fix is to do the check and throw appropriate exception.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/AuthorizationCognitoActions.kt
```diff
@@ -76,12 +77,22 @@ internal object AuthorizationCognitoActions : AuthorizationActions {
                     )
                 )
                 else -> {
-                    val logins = LoginsMapProvider.CognitoUserPoolLogins(
-                        configuration.userPool.region,
-                        configuration.userPool.poolId,
-                        signedInData.cognitoUserPoolTokens.idToken!!
-                    )
-                    FetchAuthSessionEvent(FetchAuthSessionEvent.EventType.FetchIdentity(logins))
+                    signedInData.cognitoUserPoolTokens.idToken?.let {
```
div5yesh, line 80: move this as when case and handle the success case in else.

div5yesh:
Even if we guard the null value, social sign-in will continue to fail. We need to root cause the issue.

gpanshu (approved):

div5yesh (approved):

PR #2316, opened 2023-03-03
feat(auth): Add aws-core and AWSCredentialsProvider

tylerjroach:

*Issue #, if available:*

*Description of changes:*
There is currently a Kotlin SDK dependency on aws.credentials that should not exist in core. AWS specific files have been moved out of core into a new aws-core lib. While this is technically a breaking change, all aws feature libraries now include aws-core, so there shouldn't be any customer impact. If a customer pulls in any aws packages, aws-core will be pulled in.

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
```diff
@@ -26,6 +26,7 @@ import kotlin.coroutines.suspendCoroutine
 /**
  * Wrapper to provide credentials from Auth synchronously and asynchronously
  */
+@Deprecated("This interface was released with public visibility, but is not intended to be consumed.")
 open class CognitoCredentialsProvider : AuthCredentialsProvider {
```
tylerjroach, line 30: I'm having second thoughts about this. If the customer uses the Kotlin SDK directly, this CredentialsProvider allows them to easily use a pre-built one that fetches credentials via `Amplify.Auth`.

tylerjroach: I removed the deprecation. This gives customers the AWSCredentialsProvider that we intend for them to implement on their own, while keeping this as a prebuilt one that can be passed into Kotlin SDK.

aws-core/src/main/java/com/amplifyframework/auth/AWSAuthSessionInternal.kt
```diff
@@ -16,6 +16,7 @@ package com.amplifyframework.auth
 
 import com.amplifyframework.auth.result.AuthSessionResult
 
+@Deprecated("This class was released with public visibility, but is not intended to be consumed.")
```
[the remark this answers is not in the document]
palpatim: Would the deprecation message generate errors during normal build, or only if a customer consumes the deprecated symbols directly in their app code?

aws-core/src/main/java/com/amplifyframework/auth/AWSAuthSessionInternal.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
palpatim, line 2: FWIW, I think our standard is to adopt the no-year version of the copyright message. See examples at https://github.com/aws-amplify/amplify-swift/blob/main/Amplify/Amplify.swift#L1-L6, enforced by https://github.com/aws-amplify/amplify-swift/blob/main/.swiftformat#L12-L13

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentials.kt
```diff
@@ -85,3 +86,12 @@ class AWSTemporaryCredentials(
      */
     val expiration: Instant
 ) : AWSCredentials(accessKeyId, secretAccessKey)
+
+internal fun AWSCredentials.toCredentials(): Credentials {
```
palpatim, line 90: `Credentials` as a return type makes sense b/c it's from the Kotlin SDK. For the method name, I might consider something a more specific, especially since it's internal: `toSdkCredentials` or similar?

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
palpatim, line 2: Date-free copyright message, or 

```suggestion
 * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

aws-core/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
palpatim, line 3: Year-free version, or

```suggestion
   Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

aws-core/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
palpatim, line 2: Year-free version, or 
```suggestion
 * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
palpatim, line 2: Why revert the year? (I noticed that had happened above as well) Just a copy/paste artifact? In any case, use year-free version, or 

```suggestion
 * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

aws-core/src/main/java/com/amplifyframework/auth/AWSAuthSessionInternal.kt
tylerjroach: Great! I'll create a ticket for us to do in a future PR.

aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
tylerjroach: Hmm, I may have the wrong year in my auto license config, but its odd it changed correctly on others. I'll fix all of these.

aws-core/src/main/java/com/amplifyframework/auth/AWSAuthSessionInternal.kt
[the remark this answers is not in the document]
tylerjroach: I *think* its only if they directly consume, but I will confirm. We have build warnings right now that will show when building Amplify directly that we can choose to suppress or ignore ( there are quite a few files I'll need to add a suppress warning on some of these internally).

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentials.kt
tylerjroach: Sure, this is pre-existing and moved files where I felt like it belongs, but I like the suggestion.

aws-core/src/main/java/com/amplifyframework/auth/AWSAuthSessionInternal.kt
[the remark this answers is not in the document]
sdhuka: Could add a new annotation specific to amplify library? something similar to [this](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt)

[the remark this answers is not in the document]
tylerjroach: @sdhuka I decided against it since we shouldn't remove these implementations since they were once public. I don't want us to forget down the line that removal would be a breaking change because that comment block wasn't always there. Open to reconsidering options here though, especially for a class like this that has "Internal" in the name.

build.gradle.kts
```diff
@@ -99,6 +104,14 @@ subprojects {
             failOnPassedAfterRetry.set(true)
         }
     }
+
+    tasks.withType<KotlinCompile> {
```
div5yesh, line 108: 🙌

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -105,6 +107,7 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
      * Constructs the AWS S3 Storage Plugin initializing the executor service.
      */
     @SuppressWarnings("unused") // This is a public API.
+    @OptIn(markerClass = InternalApiWarning.class)
```
div5yesh, line 110: Does this mean creating plugins will throw internal API warning? Both the methods are intended for public use.

mattcreaser: The `OptIn` means that the warning won't be presented to callers of these methods.

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+ */
+interface AWSCredentialsProvider<out T : AWSCredentials> {
+
+    fun fetchAWSCredentials(
+        onSuccess: Consumer<@UnsafeVariance T>,
+        onError: Consumer<AuthException>
+    )
+}
+
+internal fun <T : AWSCredentials> convertToCredentialsProvider(
```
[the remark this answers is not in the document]
div5yesh: naming suggestion to got with Tim's comment: `convertToSdkCredentialsProvider`

div5yesh (approved):

aws-core/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=aws-core
+POM_NAME=Amplify Framework for Android - AWS Core
+POM_DESCRIPTION=Amplify Framework for Android - AWS Core components and utilties for AWS Amplify Libraries
```
eeatonaws, line 3: nit: spelling error
```suggestion
POM_DESCRIPTION=Amplify Framework for Android - AWS Core components and utilities for AWS Amplify Libraries
```

aws-geo-location/src/main/java/com/amplifyframework/geo/location/AWSLocationGeoPlugin.kt
```diff
@@ -64,6 +64,7 @@ class AWSLocationGeoPlugin(
         configuration.searchIndices!!.default
     }
 
+    @Deprecated("This credentialsProvider was released with public visibility, but is not intended to be consumed.")
```
eeatonaws, line 67: The Geo credentialsProvider is used in maplibre-adapter [here](https://github.com/aws-amplify/amplify-android/blob/main/maplibre-adapter/src/main/java/com/amplifyframework/geo/maplibre/http/AWSRequestSignerInterceptor.kt#L105). Could we use a similar strategy as in other parts of this PR for showing a warning message to customers if they use this variable while still allowing use within the library?

tylerjroach: Nice catch. Applied.

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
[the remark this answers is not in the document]
tylerjroach: opened up convertToSdkCredentialsProvider to public to allow customers an easy way to convert to SDKs credentialsprovider to be used in the escape hatch. This way a single provider could be shared.

eeatonaws (approved):

settings.gradle.kts
```diff
@@ -135,7 +135,9 @@ dependencyResolutionManagement {
     }
 }
 
+include(":annotations")
 include(":core")
+include("aws-core")
```
div5yesh, line 140:
```suggestion
include(":aws-core")
```

div5yesh (approved):
One code change, otherwise LGTM.

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+ */
+interface AWSCredentialsProvider<out T : AWSCredentials> {
+
+    fun fetchAWSCredentials(
+        onSuccess: Consumer<@UnsafeVariance T>,
+        onError: Consumer<AuthException>
+    )
+}
+
+fun <T : AWSCredentials> convertToSdkCredentialsProvider(
```
div5yesh, line 46: I dont think this needs to be publicly exposed. If the goal is to pass AWSCredentialsProvider to other components and SDK CredentialsProvider is always accessible through escape hatch, I dont see a reason for this conversion API. Maybe start with non-public and if customers need it we can exposed it in the future.

tylerjroach: @div5yesh The reason I exposed is because I don't think we want customers having to use multiple credentials providers, or at least, having to create their own AWSCredentialsProvider and SDKs CredentialsProvider.

Our suggested implementation should be to create a single AWSCredentialsProvider, and when that is needed for the Kotlin SDK escape hatch, they can use this public conversion method.

div5yesh (approved):

eeatonaws (approved):

PR #2340, opened 2023-03-15
fix(notifications): journey attributes parsing

div5yesh:

*Issue #, if available:*

*Description of changes:*
- fix journey attributes parsing based on pinpoint correct pinpoint prefix

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-push-notifications-pinpoint/src/main/java/com/amplifyframework/pushnotifications/pinpoint/EventSourceType.kt
```diff
@@ -39,6 +39,8 @@ internal class EventSourceType private constructor(
         private const val AWS_EVENT_TYPE_RECEIVED_FOREGROUND = "received_foreground"
         private const val AWS_EVENT_TYPE_RECEIVED_BACKGROUND = "received_background"
 
+        private val pinpointJourneyPrefix = PushNotificationsConstants.PINPOINT_PREFIX.dropLast(1)
```
cshfang, line 42: Nit: I would suggest not tying this key to Journey, after all, it ends up just being the string `pinpoint`. If Pinpoint were to add a new feature in the future, I would expect it to follow this pattern vs the `pinpoint.whatever` one as well.

I'd also just call it key vs prefix since it's going to be treated as the whole key string
```suggestion
        private val pinpointMetadataKey = PushNotificationsConstants.PINPOINT_PREFIX.dropLast(1)
```

cshfang (commented):
Hmm does this line also need to be looked at or will it be a part of a separate PR?
https://github.com/aws-amplify/amplify-android-staging/blob/d56b745a86cc46fbec519af0cb89c2916b107e30/aws-push-notifications-pinpoint-utils/src/main/java/com/amplifyframework/pushnotifications/pinpoint/PinpointNotificationPayload.kt#L43

aws-push-notifications-pinpoint/src/main/java/com/amplifyframework/pushnotifications/pinpoint/EventSourceType.kt
div5yesh: I went complete 360 on this and removed the dot from prefix 😅. As you pointed out, change was needed in payload too.

PR #2387, opened 2023-04-06
chore(auth): Add internal API to add to user-agent

tjleing:

*Issue #, if available:*

*Description of changes:*

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
@@ -21,6 +21,7 @@ import android.content.Intent
 import androidx.annotation.VisibleForTesting
 import com.amplifyframework.AmplifyException
 import com.amplifyframework.annotations.InternalAmplifyApi
+import com.amplifyframework.auth.AWSCognitoAuthMetadataType
 import com.amplifyframework.auth.AuthCodeDeliveryDetails
 import com.amplifyframework.auth.AuthDevice
 import com.amplifyframework.auth.AuthException
@@ -100,8 +101,8 @@ class AWSCognitoAuthPlugin : AuthPlugin<AWSCognitoAuthService>() {
     }
 
     @InternalAmplifyApi
-    fun addToUserAgent(key: String, value: String) {
-        realPlugin.addToUserAgent(key, value)
+    fun addToUserAgent(type: AWSCognitoAuthMetadataType, value: String) {
+        realPlugin.addToUserAgent(type, value)
     }
 
     private fun Exception.toAuthException(): AuthException {
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
@@ -99,6 +99,11 @@ class AWSCognitoAuthPlugin : AuthPlugin<AWSCognitoAuthService>() {
         return pluginConfigurationJSON
     }
 
+    @InternalAmplifyApi
+    fun addToUserAgent(key: String, value: String) {
+        realPlugin.addToUserAgent(key, value)
+    }
+
     private fun Exception.toAuthException(): AuthException {
         return if (this is AuthException) {
             this
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
@@ -33,7 +33,6 @@ interface AWSCognitoAuthService {
         internal fun fromConfiguration(configuration: AuthConfiguration): AWSCognitoAuthService {
             val customPairs: MutableMap<String, String> = mutableMapOf()
             val cognitoIdentityProviderClient = configuration.userPool?.let { it ->
-
                 CognitoIdentityProviderClient {
                     this.region = it.region
                     this.endpointProvider = it.endpoint?.let { endpoint ->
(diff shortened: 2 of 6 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
```diff
@@ -17,34 +17,62 @@ package com.amplifyframework.auth.cognito
...
+import aws.smithy.kotlin.runtime.client.RequestInterceptorContext
 import aws.smithy.kotlin.runtime.client.endpoints.Endpoint
 import aws.smithy.kotlin.runtime.client.endpoints.EndpointProvider
+import aws.smithy.kotlin.runtime.http.interceptors.HttpInterceptor
 import com.amplifyframework.statemachine.codegen.data.AuthConfiguration
 
 interface AWSCognitoAuthService {
     var cognitoIdentityProviderClient: CognitoIdentityProviderClient?
     var cognitoIdentityClient: CognitoIdentityClient?
+    var customUserAgentPairs: MutableMap<String, String>
```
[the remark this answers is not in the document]
mattcreaser: In fact they should all be `val`. Can we make that change? Then do this:

```kotlin
val customUserAgentPairs = mutableMapOf<String, String>()
val cognitoIdentityProviderClient =  // ...
val cognitoIdentityClient = // ...

return object : AWSCognitoAuthService {
    override val cognitoIdentityProviderClient = cognitoIdentityProviderClient
    override val cognitoIdentityClient = cognitoIdentityClient
    override val customUserAgentPairs = customUserAgentPairs
} 
```

mattcreaser (commented):
Looks good to go after these changes

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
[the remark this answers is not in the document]
tjleing: changed the map, but I believe val doesn’t work for `cognitoIdentityProviderClient` or `cognitoIdentityClient` due to the changes I made (it would’ve been possible before).  the `AWSCognitoAuthService` needs to be defined at the start of the function so that it can be referenced in the transformer, so there’s a sort of loop between the service and the two clients.  to break the loop, we create the service first and then reassign its clients

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
```diff
@@ -17,33 +17,57 @@ package com.amplifyframework.auth.cognito
...
+                    }
+                }
             }
 
             return object : AWSCognitoAuthService {
-                override var cognitoIdentityProviderClient = cognitoIdentityProviderClient
-                override var cognitoIdentityClient = cognitoIdentityClient
+                override val cognitoIdentityProviderClient = cognitoIdentityProviderClient
+                override val cognitoIdentityClient = cognitoIdentityClient
+                override val customUserAgentPairs = customPairs
```
div5yesh, line 70: Once the custom pairs are added, would any API call from cognito plugin also send these custom pairs?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
```diff
@@ -99,6 +99,11 @@ class AWSCognitoAuthPlugin : AuthPlugin<AWSCognitoAuthService>() {
         return pluginConfigurationJSON
     }
 
+    @InternalAmplifyApi
+    fun addToUserAgent(key: String, value: String) {
```
div5yesh, line 103: Is `addToUserAgent` meant to be called before each API call or once during configure?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthService.kt
tjleing: Yes, all Auth calls will include these pairs.  This is also in parity with JS

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
tjleing: Once during configure, this will add it to all future Auth calls

mattcreaser (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -165,6 +167,11 @@ internal class RealAWSCognitoAuthPlugin(
 
     fun escapeHatch() = authEnvironment.cognitoAuthService
 
+    @InternalAmplifyApi
+    fun addToUserAgent(type: AWSCognitoAuthMetadataType, value: String) {
+        authEnvironment.cognitoAuthService.customUserAgentPairs[type.key] = value
```
div5yesh, line 172: Im guessing value is just version number? If yes, then can we use `BuildConfig.VERSION_NAME` instead?

div5yesh: IMO we should hardcode `value` too.

div5yesh (dismissed):
Minor comment, else LGTM.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
mattcreaser: The version information is not available here. We can add a regex to ensure the value conforms to the expected format but it must be passed in.

div5yesh (approved):

mattcreaser (approved):

Issue #2401, opened 2023-04-19
Crash on Application start when calling method Amplify.configure


reisfernanda:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
implementation("com.squareup.okhttp3:okhttp:5.0.0-alpha.11")
implementation("com.amplifyframework:aws-auth-cognito:2.7.0")
implementation("com.amplifyframework:core-kotlin:2.7.0")

```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 7.5.1
------------------------------------------------------------

Build time:   2022-08-05 21:17:56 UTC
Revision:     d1daa0cbf1a0103000b71484e1dbfe096e095918

Kotlin:       1.6.21
Groovy:       3.0.10
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.12 (Oracle Corporation 11.0.12+8-LTS-237)
OS:           Windows 10 10.0 amd64

```

</details>

### Describe the bug

Our crash monitoring is reporting some occurrences of a crash that happens in the application initialization when calling the amplify's configure method.
I couldn't reproduce the issue in any way, but the rare users that find this problem can reproduce it frequently.

`java.lang.IllegalArgumentException - port must be in range [1, 65535]`

This issue is a duplication of https://github.com/aws-amplify/amplify-android/issues/2297 which was closed due to inactivity. I added a comment on the other issue but nobody replied.

### Reproduction steps (if applicable)

We couldn't reproduce the issue but it happens a lot for a few users according to our crash monitoring.

### Code Snippet

```java
class AppApplication : Application() {

    override fun onCreate() {
        super.onCreate()
        try {
            Amplify.addPlugin(AWSCognitoAuthPlugin())
            Amplify.configure(context) // The crash happens here in this call
        } catch (exception: AmplifyException) {
            loggingManager.logNonFatalException(exception)
        }
    }
}

```

### Log output

<details>

```
Fatal Exception: java.lang.ExceptionInInitializerError:
       at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine.<init>(OkHttpEngine.kt:27)
[...]

PR #2438, opened 2023-05-05
fix(misc): bumping kotlin sdk version to include enhanced proxy exception messages

gpanshu:

*Issue #, if available:* #2401

*Description of changes:* When using a proxy if your port is invalid it produces a crash without providing additional info from the kotlin sdk as to what the current port is. This PR will update the latest kotlin SDK which has the fix to provide more information : https://github.com/awslabs/smithy-kotlin/pull/837

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
+++ b/aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
@@ -17,6 +17,7 @@ package com.amplifyframework.auth
 
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
 import aws.smithy.kotlin.runtime.auth.awscredentials.CredentialsProvider
+import aws.smithy.kotlin.runtime.util.Attributes
 import com.amplifyframework.core.Consumer
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
@@ -48,7 +49,7 @@ fun <T : AWSCredentials> convertToSdkCredentialsProvider(
 ): CredentialsProvider {
 
     return object : CredentialsProvider {
-        override suspend fun getCredentials(): Credentials {
+        override suspend fun resolve(attributes: Attributes): Credentials {
             return suspendCoroutine { continuation ->
                 awsCredentialsProvider.fetchAWSCredentials(
                     { continuation.resume(it.toSdkCredentials()) },
--- a/aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
+++ b/aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
@@ -16,6 +16,7 @@
 package com.amplifyframework.auth
 
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
+import aws.smithy.kotlin.runtime.util.Attributes
 import com.amplifyframework.AmplifyException
 import com.amplifyframework.core.Amplify
 import com.amplifyframework.core.Consumer
@@ -31,7 +32,7 @@ open class CognitoCredentialsProvider : AuthCredentialsProvider {
     /**
      * Request [Credentials] from the provider.
      */
-    override suspend fun getCredentials(): Credentials {
+    override suspend fun resolve(attributes: Attributes): Credentials {
         return suspendCoroutine { continuation ->
             Amplify.Auth.fetchAuthSession(
                 { authSession ->
--- a/aws-pinpoint-core/src/main/java/com/amplifyframework/pinpoint/core/EventRecorder.kt
+++ b/aws-pinpoint-core/src/main/java/com/amplifyframework/pinpoint/core/EventRecorder.kt
@@ -199,10 +199,7 @@ class EventRecorder(
     }
 
     private fun isRetryableError(code: Int?): Boolean {
-        code?.let {
-            return code in 500..599
-        }
-        return false
+        return code in 500..599
     }
 
     private fun processEndpointResponse(endpointResponse: EndpointItemResponse?) {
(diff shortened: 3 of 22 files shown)


eeatonaws (changes requested):

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewTestActivityTest.kt
```diff
@@ -153,15 +154,15 @@ class MapViewTestActivityTest {
                 signInWithCognito()
             }
             val result = block()
-            rule.scenario.onActivity { signOutFromCognito() }
+            rule.scenario.onActivity { }
```
div5yesh, line 157: I think `signOut` was added to reset the state before running each test. Not all tests need signIn and tests are not guaranteed to run in fixed order.

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewStressTest.kt
```diff
@@ -77,7 +77,7 @@ class MapViewStressTest {
                 signInWithCognito()
             }
             val result = block()
-            rule.scenario.onActivity { signOutFromCognito() }
+            rule.scenario.onActivity { }
```
div5yesh, line 80: Same as my below comment.

aws-core/src/main/java/com/amplifyframework/auth/AWSCredentialsProvider.kt
```diff
@@ -48,7 +49,7 @@ fun <T : AWSCredentials> convertToSdkCredentialsProvider(
 ): CredentialsProvider {
 
     return object : CredentialsProvider {
-        override suspend fun getCredentials(): Credentials {
+        override suspend fun resolve(attributes: Attributes): Credentials {
```
div5yesh, line 52: @tylerjroach We definitely dodged a breaking change with this abstraction 🙌.

aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
```diff
@@ -31,7 +32,7 @@ open class CognitoCredentialsProvider : AuthCredentialsProvider {
     /**
      * Request [Credentials] from the provider.
      */
-    override suspend fun getCredentials(): Credentials {
+    override suspend fun resolve(attributes: Attributes): Credentials {
```
div5yesh, line 35: What can we do with the attributes? Do we need to support them for fetchAuthSession?

aws-pinpoint-core/src/main/java/com/amplifyframework/pinpoint/core/EventRecorder.kt
```diff
@@ -198,8 +198,11 @@ class EventRecorder(
         return eventIdToDelete
     }
 
-    private fun isRetryableError(code: Int): Boolean {
-        return code in 500..599
+    private fun isRetryableError(code: Int?): Boolean {
+        code?.let {
```
div5yesh, line 202: `null in 500..599` should be always false. Do we need the null check on `code`?

gpanshu: Thats what I am doing with this change, if its a null it will return false. I can make it cleaner by doing an if else. Whether or not this null guard is actually needed or not was out of scope as I simply did the kotlin upgrade and was not auditing the tests.

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewTestActivityTest.kt
gpanshu: If signout is necessary before running each test it should be part of the test instead of depending on another test for the teardown. However I did check all these tests and they are running without this extra sign out. Unless you can identify a particular test that needs this?

aws-core/src/main/java/com/amplifyframework/auth/CognitoCredentialsProvider.kt
gpanshu: Currently since we do not need the attributes I am passing `emptyAttributes()` into the resolve.

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewTestActivityTest.kt
div5yesh: I thought `signOutFromCognito()` was a private function and not another test.

div5yesh: Sometimes these tests fail when run on different devices.

aws-pinpoint-core/src/main/java/com/amplifyframework/pinpoint/core/EventRecorder.kt
div5yesh: Im saying just `return code in 500..599` would do.

gpanshu: ok yeah gotcha. done

maplibre-adapter/src/androidTest/java/com/amplifyframework/geo/maplibre/MapViewTestActivityTest.kt
gpanshu: If I keep that command in it fails 100% of the time on our device farm so its a catch 22.

eeatonaws (dismissed):

eeatonaws (approved):

div5yesh (approved):

Issue #2069, opened 2022-10-27
DataStore - ObserveQuery deleted items are showing


a-alen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

GraphQL API, DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line
    def amplify_version = "1.37.5"

    implementation "com.amplifyframework:core:$amplify_version"
    implementation "com.amplifyframework:aws-api:$amplify_version"
    implementation "com.amplifyframework:aws-auth-cognito:$amplify_version"
    implementation "com.amplifyframework:aws-datastore:$amplify_version"
    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.6'

```

</details>

### Environment information

<details>

```
# Put output below this line
distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip

```

</details>

### Describe the bug

When item is deleted with Amplify.DataStore.delete, that deleted item stays in Amplify.DataStore.observeQuery. You need to exit from app, and when you start it again ObserveQuery shows correct number of items, until next delete.

Column _deleted is not accessible, so it is not possible to filter that item.

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line
2022-10-27 20:54:22.699  5341-5409  AWS_DATASTORE_GIT       com.example.barcodelogin             I  OBSERVE STARTED
2022-10-27 20:54:22.727  5341-5417  AWS_DATASTORE_GIT       com.example.barcodelogin             I  OBSERVE number of items: 8
2022-10-27 20:54:22.755  5341-5441  AWS_DATASTORE_GIT       com.example.barcodelogin             I  QUERY number of items: 8
2022-10-27 20:54:39.407  5341-5493  AWS_DATASTORE_GIT       com.example.barcodelogin             I  OBSERVE number of items: 8
2022-10-27 20:54:39.407  5341-5493  AWS_DATASTORE_GIT       com.example.barcodelogin             I  ITEM DELETED
2022-10-27 20:54:39.416  5341-5495  AWS_DATASTORE_GIT       com.example.barcodelogin             I  QUERY number of items: 7
[...]

PR #2522, opened 2023-07-19
fix(datastore): Observe Query should not return deleted items, fixes - #2069

ankpshah:

*Issue #, if available: #2069 
 
*Description of changes: 
Title: Fix for the bug: ObserveQuery returning deleted items
Change Description:
1. Additional case for StorageItemChange Type "Delete". 
2. Updated the sequence of event processing. Consumer subscribes to the Subject before the onObservationStarted event is emitted.      
3. Datastore records are retrieved before emitting onObservationStarted event. To achieve this the consumer subscription and onObservationStarted event emission are moved to lambda function passed as argument to queryLocalData. This ensures instance variables have appropriate model records to process deletion change.
4. Update instance variable completeItemMap before invoking onQuerySnapshot.

Additional Notes:
1.  The issue was primarily due to three reasons:
          a. Missing case for Delete StorageItemType.
          b. OnObservationStarted event being emitted before consumer subscribing to subject. This caused first deleted record to appear in offline deletion scenario within observeQuery response in case of multiple deletions due to asynchronous execution. 
          c. Deletion event being processed before queryLocalData could populate instance variable completeItemList with model records.

*How did you test these changes?
Two Unit test cases added to test the change.
1. observeQueryShouldNotReturnDeletedRecord(): This test checks the deleted records are not returned by observeQuery.
2. observeQueryShouldFailIfReturnsDeletedRecord(): This is to ensure test fails if the deleted record are returned by ObserveQuery.

- [X] No

*General Checklist*
- [X] Added Unit Tests
- [ ] Added Integration Tests


aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/ObserveQueryExecutorTest.java
```diff
@@ -264,4 +263,140 @@ public void observeQueryCancelsTheOperationOnQueryError() throws DataStoreExcept
...
+        Assert.assertTrue(changeLatch.await(10, TimeUnit.SECONDS));
+    }
+
+    /***
+     * Test should Fail when ObserveQuery returns deleted record.
+     * @throws InterruptedException InterruptedException
+     * @throws DataStoreException DataStoreException
+     */
+    @Test(expected = AssertionError.class)
+    public void observeQueryShouldFailIfReturnsDeletedRecord() throws InterruptedException, DataStoreException {
```
tylerjroach, line 341: What is the purpose of this test? Why would observe query return a deleted record with the fix in place? If were forcing the wrong behavior in a test, this isn't a valid test.

tylerjroach (changes requested):

aws-datastore/src/test/java/com/amplifyframework/datastore/storage/sqlite/ObserveQueryExecutorTest.java
ankpshah: I have removed this test as it was redundant and testing the same behavior.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/ObserveQueryExecutor.java
```diff
@@ -132,42 +132,50 @@ public void observeQuery(
...
+                                    }
+                                    onObservationError.accept(new DataStoreException(
+                                            "Failed to observe items in storage adapter.",
+                                            failure,
+                                            "Inspect the failure details."
+                                    ));
+                                },
+                                onObservationComplete::call
+                        );
+                onObservationStarted.accept(this);
```
mattcreaser, line 174: 🤔 Is it strictly necessary for the `onObservationStarted` to be so late?

There is a slight behaviour change here in that if the initial query fails the caller will never see the onStarted callback, only onError (prior behaviour was to see onStarted and then onError). I'm not sure that is a problem, in fact it may be more correct than the prior implementation.

I was also looking at whether it would be possible for `onItemChange` to get invoked prior on `onObservationStarted`, it's a bit ambiguous to me but I don't think so?

tylerjroach: I had requested this change after diving in together because it was easy to replicate incorrect values and counts if mutations were made before the initial query was successful. Once we report `onObservationStarted`, my expectation is that no observations would be missed once that value is returned. This was not the case with the immediate callback.

If `completeItemMap` is empty (which is the case before its populated with initial values from a successful query, any item changes (such as a delete) will not properly be updated/removed from competeItemMap.

mattcreaser: Makes sense, thanks.

mattcreaser (approved):

tylerjroach (approved):

Issue #2481, opened 2023-06-15
SESSION_EXPIRED event not called after refresh token expires


KamilSucharski:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin, RxJava

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
implementation 'com.amplifyframework:core:2.8.5'
implementation 'com.amplifyframework:aws-auth-cognito:2.8.5'
implementation 'com.amplifyframework:rxbindings:2.8.5'
```

</details>

### Environment information

<details>

```
------------------------------------------------------------ 
Gradle 8.0
------------------------------------------------------------

Build time:   2023-02-13 13:15:21 UTC
Revision:     62ab9b7c7f884426cf79fbedcf07658b2dbe9e97

Kotlin:       1.8.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.6 (JetBrains s.r.o. 17.0.6+0-b2043.56-9586694)
OS:           Windows 10 10.0 amd64
```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/lib/auth/auth-events/q/platform/android/#subscribing-events

### Describe the bug

When the refresh token expires, the SESSION_EXPIRED Hub Event is not sent out. I traced the reason to `RealAWSCognitoPlugin.kt::_fetchAuthSession`:
```
private fun _fetchAuthSession(
	onSuccess: Consumer<AuthSession>,
	onError: Consumer<AuthException>
) {
	val token = StateChangeListenerToken()
	authStateMachine.listen(
		token,
		{ authState ->
			when (val authZState = authState.authZState) {
				is AuthorizationState.SessionEstablished -> {
					authStateMachine.cancel(token)
					onSuccess.accept(authZState.amplifyCredential.getCognitoSession())
				}
				is AuthorizationState.Error -> {
					authStateMachine.cancel(token)
					when (val error = authZState.exception) {
						is SessionError -> {
							when (error.exception) {
								is SignedOutException -> {
									onSuccess.accept(error.amplifyCredential.getCognitoSession(error.exception))
								}
								is SessionExpiredException -> {
[...]

PR #2524, opened 2023-07-20
fix(auth): Fix for sending session expired when an invalid grant exception is received

gpanshu:

*Issue #, if available:* #2481

*Description of changes:* When we get an invalid_grant exception back from cognito we do not send the session expired hub event but we should as that means your session is no longer valid. This PR fixes that. Cognito also sends the error_description which was not captured

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
@@ -103,12 +103,13 @@ internal class AuthEnvironment internal constructor(
     }
 
     suspend fun getDeviceMetadata(username: String): DeviceMetadata.Metadata? {
-        val deviceCredentials =
+        var deviceCredentials =
             credentialStoreClient.loadCredentials(CredentialType.Device(username)) as? AmplifyCredential.DeviceData
         if (deviceCredentials == null) {
             logger.warn("loadCredentials returned unexpected AmplifyCredential Type.")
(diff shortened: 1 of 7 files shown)


tjleing (dismissed):

sdhuka (approved):

mattcreaser (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -103,12 +103,13 @@ internal class AuthEnvironment internal constructor(
...
     suspend fun getDeviceMetadata(username: String): DeviceMetadata.Metadata? {
-        val deviceCredentials =
+        var deviceCredentials =
             credentialStoreClient.loadCredentials(CredentialType.Device(username)) as? AmplifyCredential.DeviceData
         if (deviceCredentials == null) {
             logger.warn("loadCredentials returned unexpected AmplifyCredential Type.")
+            deviceCredentials = AmplifyCredential.DeviceData(DeviceMetadata.Empty)
         }
-        return (deviceCredentials as AmplifyCredential.DeviceData).deviceMetadata as? DeviceMetadata.Metadata
+        return deviceCredentials.deviceMetadata as? DeviceMetadata.Metadata
```
tylerjroach, line 112: Instead of moving to var usages, and adding L110, doing this instead.

`return deviceCredentials?.deviceMetadata ?: DeviceMetadata.Empty as? DeviceMetadata.Metadata`

The as? cast may not be necessary in example used.

Issue #2261, opened 2023-01-27
DataStore conflict handle cannot handle delete mutation


HuiSF:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java, Kotlin

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line


```

</details>

### Environment information

<details>

```
# Put output below this line


```

</details>

### Use case of DataStore conflict handler:

1. I'm deleting a local model
2. There is a conflict happened during the mutation
3. I want to make the local delete mutation always win, so I use `RETRY_LOCAL` strategy

### Issue
`RETRY_LOCAL` strategy doesn't make the local delete mutation work.

### Potential cause

The `publish` function used inside the conflict handler, seems always generate an update mutation on the resolved model, and not capable to generate a delete mutation following the use case described above.

https://github.com/aws-amplify/amplify-android/blob/9074d878460bd9e53506c08887515c5563950976/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/ConflictResolver.java#L169-L176

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

stevegaunt:
@HuiSF are there any updates with this?

stevegaunt:
@HuiSF any idea when this bug will get resolved??  It's causing data inconsistency

mattcreaser:
@stevegaunt could you share a little information about the reproduction steps you're seeing this issue with?

I will make a fix here but I'd like to ensure I am covering all the circumstances.

stevegaunt:
@mattcreaser 

It comes down to timing mainly due to network inconsistencies with our users.. 
 
If a mutation is created or updated and synced to the remote database, but before amplify has received the response of the outgoing pending event to update the version locally,  we have already within the app done a delete mutation on that object..  however, it's version number is not updated to the synced version number DynamoDB will throw a conflict back to app sync and consequently to the conflict resolution function..   It looks like the received local and remote model instances doesn't include conflict detection meta fields such as _deleted 

simple step to reproduce would be
1. mutate create object and wait until it's synced.
[...]

PR #2532, opened 2023-07-26
fix(datastore): Fix retry local after a deletion

mattcreaser:

*Issue #, if available:*

#2261 

*Description of changes:*

If a conflict occurs during a deletion and a retry local strategy is used in the conflict handler then we want to invoke a delete operation. This was previously always doing an update operation, which would result in the deletion being lost.

- Followed these reproduction steps:
  - Configure DataStore with a `conflictHandler` that always returns a `retryLocal()` decision.
  - Sync a model to the device
  - Put the device into airplane mode
  - Delete the model on the device
  - In DynamoDB, increase the version number of that model
  - Turn off airplane mode on the device
 
When the device tries to publish the pending mutation it will result in a conflict because the version numbers do not match. The conflict handler will initiate a retry of the local deletion. The expected behaviour is that this retry will delete the record.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


tylerjroach (dismissed):
Nice find!

ankpshah (approved):

Issue #2458, opened 2023-05-22
Datastore, random crash when switching from online to offline use case


antoineberthelin:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin, Kotlin - Coroutines

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line


```

</details>

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

Use case : lose the network and recover the network 
Random crash on production (not easy to reproduce)
RX java crash
io.reactivex.rxjava3.exceptions.OnErrorNotImplementedException: call. java.lang.IllegalStateException: Only a single observer allowed.

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line



2023-05-22 09:04:19.601 11216-12708 amplify:aws-datastore   com.appsonic.android.whosnext        I  Started subscription processor for models: [Item, Draw, DrawBatch, Question] of types [ON_CREATE, ON_UPDATE, ON_DELETE].
2023-05-22 09:04:19.652 11216-12747 amplify:aws-datastore   com.appsonic.android.whosnext        I  Success on attempt #1
2023-05-22 09:04:19.703 11216-11216 ViewRootIm...nActivity] com.appsonic.android.whosnext        I  ViewPostIme pointer 0
2023-05-22 09:04:19.842 11216-11216 ViewRootIm...nActivity] com.appsonic.android.whosnext        I  ViewPostIme pointer 1
[...]

Issue #1010, opened 2020-11-30
Unable to acquire orchestrator lock. Transition currently in progress.


renebrandel:
While creating M:N relationships, I get an error:
> Unable to acquire orchestrator lock. Transition currently in progress.

@jamesonwilliams suggested a _work-around_ to require a DataStore. `READY` event to be set before relationship creation. We agreed the work-around shouldn't be necessary.

When I run this code:

```java
Content content = Content.builder()
    .body("Hello from Android13")
     .build();

Tag tag = Tag.builder()
    .label("Test 123123-12313")
    .build();

ContentTag contentTag = ContentTag.builder()
    .content(content)
    .tag(tag)
    .build();

Amplify.DataStore.save(content,
    savedPost -> {
        Log.i("MyAmplifyApp", "Content saved.");
        Amplify.DataStore.save(tag,
            savedEditor -> {
                Log.i("MyAmplifyApp", "Tag saved.");
                Amplify.DataStore.save(contentTag,
                    saved -> Log.i("MyAmplifyApp", "ContentTag saved."),
                    failure -> Log.e("MyAmplifyApp", "ContentTag not saved.", failure)
                );
            },
            failure -> Log.e("MyAmplifyApp", "Tag not saved.", failure)
        );
    },
    failure -> Log.e("MyAmplifyApp", "Content not saved.", failure)
);
```
I get this crash:

```console
2020-11-30 11:18:15.800 15077-15118/com.example.relationshipjava I/MyAmplifyApp: Content saved.
2020-11-30 11:18:17.820 15077-15120/com.example.relationshipjava E/MyAmplifyApp: Tag not saved.
    DataStoreException{message=Unable to acquire orchestrator lock. Transition currently in progress., cause=null, recoverySuggestion=Retry your request}
        at com.amplifyframework.datastore.syncengine.Orchestrator.start(Orchestrator.java:195)
        at com.amplifyframework.datastore.AWSDataStorePlugin.lambda$start$9$AWSDataStorePlugin(AWSDataStorePlugin.java:259)
        at com.amplifyframework.datastore.-$$Lambda$AWSDataStorePlugin$FPjeqf-ENbST6bEgtzMenBBWB54.call(Unknown Source:6)
        at com.amplifyframework.datastore.-$$Lambda$TpN5VBwI7X4qLhZPz2Xh_RciFYY.run(Unknown Source:2)
[...]

PR #2551, opened 2023-08-09
fix(data): Resolve orchestrator transition threading issues.

tylerjroach:

*Issue #, if available:*
https://github.com/aws-amplify/amplify-flutter/issues/3017

*Description of changes:*

This PR intends to fix the crash caused by the UnicastSubject "buffer" in the SubscriptionProcessor being subscribed to more than once. I believe there were 2 root causes to this.

* The reachabilityMonitor subscriber was directly calling `startApiSync` and `stopApiSync` rather than use the transitionTo methods. The transitionTo methods read the current state, and ignore calls if the current state equals the requested state. I believe the crash was due to `startApiSync` being called multiple times, and through a race condition, there was a scenario that could allow the UnicastSubject to receive more than 1 subscription.
* When `stopApiSync` is called, the completable that is started in `startApiSync` does not stop by simply clearing the disposable. As a result, a call to start/stop/start could still result in a single buffer being subscribed to more than once. A check to stop additional processing was added in the startApiSync completable, if the disposable was cleared.

Lastly, I refactored the `currentState.set` calls to only be made inside the transitionTo api methods and added a lock for thread safety. 

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/.github/workflows/closed_issue_message.yml
+++ b/.github/workflows/closed_issue_message.yml
@@ -8,7 +8,7 @@ jobs:
     auto_comment:
         runs-on: ubuntu-latest
         steps:
-        - uses: aws-actions/closed-issue-message@v1
+        - uses: aws-actions/closed-issue-message@36b7048ea77bb834d16e7a7c5b5471ac767a4ca1 # v1
           with:
             # These inputs are both required
             repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/codecov_code_coverage.yml
+++ b/.github/workflows/codecov_code_coverage.yml
@@ -24,10 +24,10 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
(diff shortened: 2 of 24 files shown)


mattcreaser:
Issue #2458 is also referencing this same problem.

mattcreaser:
🤔 There are definitely some good improvements here - it's easy to see how the prior direct invocation of `startApiSync` could cause a crash when network status changed more quickly than the total sync time - but I am skeptical that the `!emitter.isDisposed()` check is sufficient to **completely** fix the issue. It seems that, although the door is _mostly_ closed, it is still theoretically possible for multiple calls to `startDrainingMutationBuffer` to interleave.

What about wrapping the contents of the completable in a lock, such that we can guarantee that no threads can interleave between `subscriptionProcessor.startSubscriptions()` and `subscriptionProcess.startDrainingMutationBuffer()`? 

Another option would be to use a single-threaded scheduler instead of `Schedulers.io()` to run the `startApiSync` completable. That has the downside of permanently reserving a single thread for this purpose however.

ankpshah:
Issue #1010 might also be referencing same problem

mattcreaser (approved):

ankpshah (approved):

PR #2580, opened 2023-08-31
fix(auth): Fix for checking current preferred type before setting the MFAType as enabled to ensure the current preference is not cleared out

gpanshu:
Fix for checking current preferred type before setting the MFAType as enabled to ensure the current preference is not cleared out

*Issue #, if available:*

*Description of changes:* For updating of mfa preferences we were just updating based on what customer had provided without taking into consideration what current preferences were which meant we could get into scenarios where 2 preferences were set to preferred. Hence in this PR I aim to fix that by prefetching the current preferences before updating it.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2224,57 @@ internal class RealAWSCognitoAuthPlugin(
...
-                                            enabled = it.mfaEnabled
-                                            it.mfaPreferred ?.let { preferred -> preferredMfa = preferred }
-                                        }
-                                    }
-                                    this.softwareTokenMfaSettings = totp?.let {
-                                        SoftwareTokenMfaSettingsType.invoke {
-                                            enabled = it.mfaEnabled
-                                            it.mfaPreferred ?.let { preferred -> preferredMfa = preferred }
+        if (sms == null && totp == null) {
+            onError.accept(InvalidParameterException())
```
tylerjroach, line 2228: A few issues here.

If we are going to callback in onError we need to return so we don't continue with another callback.

Also, the default messaging for InvalidParameterException is not very helpful. What does iOS say in this instance?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2224,57 @@ internal class RealAWSCognitoAuthPlugin(
...
                         }
                     }
-                }
 
-                else -> onError.accept(InvalidStateException())
+                    else -> onError.accept(InvalidStateException())
+                }
             }
-        }
+        }, {})
```
tylerjroach, line 2277: If fetch call fails, you aren't returning success or error back. There should be a test for this case.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2224,57 @@ internal class RealAWSCognitoAuthPlugin(
...
+                                val accessToken = getSession().userPoolTokensResult.value?.accessToken
+                                accessToken?.let { token ->
+                                    authEnvironment
+                                        .cognitoAuthService
+                                        .cognitoIdentityProviderClient
+                                        ?.setUserMfaPreference {
+                                            this.accessToken = token
+                                            this.smsMfaSettings = sms?.let { it ->
+                                                val preferredMFASetting = it.mfaPreferred
+                                                    ?: (userPreference.preferred == MFAType.SMS)
```
tylerjroach, line 2245: Faulty logic. If Disabled is set, we don't want it to remain preferred. Test should catch this.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2224,57 @@ internal class RealAWSCognitoAuthPlugin(
...
+                                                SmsMfaSettingsType.invoke {
+                                                    enabled = it.mfaEnabled
+                                                    preferredMfa = preferredMFASetting
+                                                }
+                                            }
+                                            this.softwareTokenMfaSettings = totp?.let { it ->
+                                                val preferredMFASetting = it.mfaPreferred
+                                                    ?: (userPreference.preferred == MFAType.TOTP)
```
tylerjroach, line 2253: Faulty logic. If Disabled is set, we don't want it to remain preferred. Test should catch this.

tylerjroach (changes requested):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: done

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: done.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: done

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
tylerjroach: This is a good change. Lets take the extra step to add in, but improve the default message/recovery suggestion to help the customer here.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2223,56 @@ internal class RealAWSCognitoAuthPlugin(
...
                     }
-                }
 
-                else -> onError.accept(InvalidStateException())
+                    else -> onError.accept(InvalidStateException())
+                }
             }
-        }
+        }, {
+            onError.accept(it)
```
tylerjroach, line 2274: We should be careful with this message being thrown. The message of failure here is if fetch fails, which will be confusing to the customer in an update call. Can we wrap the inner exception in an outer exception that has a message like "Failed to fetch current MFA preference, which is a prerequisite of updating MFA preferences.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2223,56 @@ internal class RealAWSCognitoAuthPlugin(
...
+                                val accessToken = getSession().userPoolTokensResult.value?.accessToken
+                                accessToken?.let { token ->
+                                    authEnvironment
+                                        .cognitoAuthService
+                                        .cognitoIdentityProviderClient
+                                        ?.setUserMfaPreference {
+                                            this.accessToken = token
+                                            this.smsMfaSettings = sms?.let { it ->
+                                                val preferredMFASetting = it.mfaPreferred
+                                                    ?: (userPreference.preferred == MFAType.SMS && it.mfaEnabled)
```
tylerjroach, line 2241: Change looks good. I am curious now though. Without the change to check for mfaEnabled. Will Cognito accept "enabled = false, preferred = true"?

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: done

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: done.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
gpanshu: this is not even allowed from an enum perspective but if sent we will get an invalidparameterexception

tylerjroach (approved):

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/RealAWSCognitoAuthPlugin.kt
```diff
@@ -2223,45 +2224,67 @@ internal class RealAWSCognitoAuthPlugin(
...
+                                            this.softwareTokenMfaSettings = totp?.let { it ->
+                                                val preferredMFASetting = it.mfaPreferred
+                                                    ?: (userPreference.preferred == MFAType.TOTP && it.mfaEnabled)
+                                                SoftwareTokenMfaSettingsType.invoke {
+                                                    enabled = it.mfaEnabled
+                                                    preferredMfa = preferredMFASetting
+                                                }
+                                            }
+                                        }?.also {
+                                            println("Ankit's test $it")
```
tylerjroach, line 2261: This line needs removed

gpanshu: done

tylerjroach (approved):

mattcreaser (approved):

Issue #2603, opened 2023-10-06
(Auth) forgetDevice(id) on remembered device locks that device from signing in


tjleing:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
implementation("com.amplifyframework:aws-auth-cognito:2.13.2")
```

</details>

### Environment information

<details>

```
Welcome to Gradle 8.0!

For more details see https://docs.gradle.org/8.0/release-notes.html


------------------------------------------------------------
Gradle 8.0
------------------------------------------------------------

Build time:   2023-02-13 13:15:21 UTC
Revision:     62ab9b7c7f884426cf79fbedcf07658b2dbe9e97

Kotlin:       1.8.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.6 (JetBrains s.r.o. 17.0.6+0-17.0.6b829.9-10027231)
OS:           Mac OS X 13.6 x86_64

```

</details>

### Describe the bug

Reported via JS team.  After signing in and calling `rememberDevice()`, calling `forgetDevice(id)` on another device or manually forgetting the device via Cognito leads to the device being unable to sign in.  The issue is that the original device still has the `deviceMetadata` and `deviceId` and provides it to Cognito, which no longer recognizes it as valid.  The error we get back from Cognito is `ResourceNotFoundException{message=Could not find the requested online resource., cause=ResourceNotFoundException(message=Device does not exist.)}`.  Swift/Flutter apparently have retries in place to catch this and resubmit without the device metadata.

### Reproduction steps (if applicable)

I have a sample app if necessary, but the steps are:

> Device A is remembered -> Device B calls `fetchDevices` -> Device B calls `forgetDevice` -> Device B forgets device A -> Device A signs in

or as I reproduced it:

> fresh install -> `rememberDevice()` -> `signOut()` -> go into Cognito user pool user and forget the most recent login -> `signIn()`

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
ResourceNotFoundException{message=Could not find the requested online resource., cause=ResourceNotFoundException(message=Device does not exist.)}
```

</details>

[...]

PR #2605, opened 2023-10-10
fix(auth): Fix for retry sign in when resourceNotFoundException is raised

gpanshu:

*Issue #, if available:* #2603

*Description of changes:* Reported via JS team. After signing in and calling rememberDevice(), calling forgetDevice(id) on another device or manually forgetting the device via Cognito leads to the device being unable to sign in. The issue is that the original device still has the deviceMetadata and deviceId and provides it to Cognito, which no longer recognizes it as valid. The error we get back from Cognito is ResourceNotFoundException{message=Could not find the requested online resource., cause=ResourceNotFoundException(message=Device does not exist.)}. Swift/Flutter apparently have retries in place to catch this and resubmit without the device metadata. 

Manually and integration tested

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [x] Added Integration Tests


tjleing (approved):
approving because the code looks the way I would expect code that fixes this issue to look and because the tests look like they properly test the feature.  not huge due diligence here, I'm not an auth expert

tylerjroach:
"After signing in and calling rememberDevice(), calling forgetDevice(id) on another device or manually forgetting the device via Cognito leads to the device being unable to sign in."

Can you clarify this? Is the user unable to sign in again after a sign out? Or does this impact their signed in user.

Does this behavior now match Swift exactly?

gpanshu:
> "After signing in and calling rememberDevice(), calling forgetDevice(id) on another device or manually forgetting the device via Cognito leads to the device being unable to sign in."
> 
> Can you clarify this? Is the user unable to sign in again after a sign out? Or does this impact their signed in user.
If the device ID is forgotten either through the console or through another session of the same user, after logout if they login on the same device whose deviceID was forgotten they will get a resourceNotFoundException which is now mitigated by this logic add. 
> 
> Does this behavior now match Swift exactly?
Yes it does exactly match Swift now.

tylerjroach (approved):

Issue #2609, opened 2023-10-17
DeviceKey is empty after logging in again after remembering device


DutchPixels:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
// Put output below this line
amplify = "2.14.0"
amplify = { module = "com.github.stkent:amplify", version.ref = "github-amplify" }
amplify-aws-api = { module = "com.amplifyframework:aws-api", version.ref = "amplify" }
amplify-aws-auth = { module = "com.amplifyframework:aws-auth-cognito", version.ref = "amplify" }
amplify-core = { module = "com.amplifyframework:core-kotlin", version.ref = "amplify" }
```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 8.1.1
------------------------------------------------------------

Build time:   2023-04-21 12:31:26 UTC
Revision:     1cf537a851c635c364a4214885f8b9798051175b

Kotlin:       1.8.10
Groovy:       3.0.15
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.8.1 (Amazon.com Inc. 17.0.8.1+8-LTS)
OS:           Mac OS X 14.0 aarch64


```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/lib/auth/device_features/q/platform/android/#terminology

### Describe the bug

Each time we login we're getting a new deviceKey even though the user should be remembered after entering the MFA code. In the console the device tracking is set to "always remember".
It seems to be caused by the key at which the deviceKey is being saved as, when using the debugger, the data is still available after logging in again.

When looking at the `AWSCognitoAuthCredentialStore` this can be validated, saving the deviceKey (available in the dashboard as well) is used with a key like this (in the `saveDeviceMetadata` method):
```
amplify.us-east-1_x4Z4E1i2Z.f9663a13-13db-40b1-bb7d-85d653915a67.deviceMetadata
```
How the key is being generated is fine, however the crucial part is the username in the middle: `f9663a13-13db-40b1-bb7d-85d653915a67`.
After logging out and logging in again the device metadata is being retrieved using this key:
```
[...]

PR #2614, opened 2023-10-23
fix(auth): Fix for when alias is used to sign in and deviceKey is stored with the entered username instead of retrieved username

gpanshu:

*Issue #, if available:* #2609

*Description of changes:*

When a user signs in with their alias instead of a username and the userpool is not configured to sign in with username, cognito assigns a username to that user (generally a generated UUID). However this breaks the current flow in a way that device ID gets generated each time as there is no stored device information for the new generated username for when we have to do respondToAuthChallenge we use the username returned from the service and not the username that was entered by the user. 
This PR will fix that flow for Custom Auth (with/without SRP and SRP).

Important to note: 

This fix does not fix device tracking for when AuthFlow is `USER_PASSWORD_AUTH ` as InitiateAuth in user_password_auth does not have access to deviceTracking. 

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [x] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/SRPCognitoActions.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/SRPCognitoActions.kt
@@ -65,7 +65,7 @@ internal object SRPCognitoActions : SRPActions {
                 secretHash?.let { authParams[KEY_SECRET_HASH] = it }
 
                 val encodedContextData = getUserContextData(event.username)
-                var deviceMetadata = getDeviceMetadata(event.username)
+                val deviceMetadata = getDeviceMetadata(event.username)
                 deviceMetadata?.let { authParams[KEY_DEVICE_KEY] = it.deviceKey }
                 val pinpointEndpointId = getPinpointEndpointId()
 
@@ -80,7 +80,7 @@ internal object SRPCognitoActions : SRPActions {
 
                 when (initiateAuthResponse?.challengeName) {
                     ChallengeNameType.PasswordVerifier -> {
-                        deviceMetadata = getDeviceMetadata(
+                        val updatedDeviceMetadata = getDeviceMetadata(
                             AuthHelper.getActiveUsername(
                                 username = event.username,
                                 alternateUsername = initiateAuthResponse.challengeParameters?.get(KEY_USERNAME),
@@ -91,7 +91,7 @@ internal object SRPCognitoActions : SRPActions {
                         )
 
                         initiateAuthResponse.challengeParameters?.let { params ->
-                            val challengeParams = deviceMetadata?.deviceKey?.let {
+                            val challengeParams = updatedDeviceMetadata?.deviceKey?.let {
                                 params.plus(KEY_DEVICE_KEY to it)
                             } ?: params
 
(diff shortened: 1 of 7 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/helpers/AuthHelper.kt
```diff
@@ -55,5 +55,16 @@ internal open class AuthHelper {
...
         }
+
+        /**
+         * This method is used to capture the activeUsername we should be using to store and retreive device Metadata.
+         * This becomes more important when an alias is used to signin instead of a username where cognito then
+         * generates the username on the customers behalf and returns it. We need to then use that username to store
+         * and retrieve device information as the username that the customer entered will no longer be available when
+         * respondToAuthChallenge/confirmSignIn calls are made.
+         * */
+        fun getActiveUsername(username: String, alternateUsername: String?, userIDForSRP: String?): String {
```
tylerjroach, line 66: May make sense to use the map of challenge parameters here as a map so that we aren't parsing out the usernames in multiple places

tylerjroach (approved):
1 comment suggestion but otherwise good.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/SRPCognitoActions.kt
```diff
@@ -64,7 +65,7 @@ internal object SRPCognitoActions : SRPActions {
                 secretHash?.let { authParams[KEY_SECRET_HASH] = it }
 
                 val encodedContextData = getUserContextData(event.username)
-                val deviceMetadata = getDeviceMetadata(event.username)
+                var deviceMetadata = getDeviceMetadata(event.username)
```
tylerjroach, line 68: Looks like deviceMetadata is only used in the password verifier block. Just create deviceMetadata in there and move the next line to assign KEY_DEVICE_KEY down in the when block.

tylerjroach (changes requested):
See comment on deviceMetadata

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/SRPCognitoActions.kt
gpanshu: No it is used in line 68(now 69) as well : See https://github.com/aws-amplify/amplify-android/pull/2614/files#diff-f40747d07791c3bcc88babaae31fcee22acf2c435db334d77fda44eb30293ee7R69

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/helpers/AuthHelper.kt
gpanshu: Instead of passing the map around I thought might be better to just pass in the strings. The parsing has to be done either way in all the places. Are you suggesting for better readability or are there performance concerns too ?

israx (approved):
LGTM!

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/helpers/AuthHelper.kt
gpanshu: Discussed offline and agreed upon to use primitives instead of map

tylerjroach (approved):

PR #2615, opened 2023-10-23
fix(liveness): Liveness web socket expiration retry

tjleing:

*Issue #, if available:*

*Description of changes:*
If device time is out of sync with true server time, skew the time sent to the Liveness websocket by that offset (calculated via server response vs. device time) and retry.

tested on gamma where liveness team has implemented the header.

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/liveness/InvalidSignatureException.kt
```diff
@@ -0,0 +1,28 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.predictions.aws.models.liveness
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Constructs a new AccessDeniedException with the specified error message.
```
gpanshu, line 21: Copy paste error? Should it be "InvalidSignatureException" here?

gradle/libs.versions.toml
```diff
@@ -15,9 +15,9 @@ androidx-test-junit = "1.1.2"
 androidx-test-orchestrator = "1.3.0"
 androidx-test-runner = "1.3.0"
 androidx-workmanager = "2.7.1"
-aws-kotlin = "0.29.1-beta" # ensure proper aws-smithy version also set
```
gpanshu, line 18: Kotlin SDK updated coroutines to 1.7.3. Should we do the same ? (We are currently at 1.6.3.
Also updated are Kotlin to 1.9.10 (current version in amplify is "1.7.10").
Please confirm the breaking change in v0.31.0-beta does not impact us: **prefix generated endpoint and auth scheme providers with client name and track upstream changes**
Please verify **BREAKING: Refactor HttpCall and HttpResponse types**  in v.32.0-beta does not impact us
Please verify https://github.com/awslabs/aws-sdk-kotlin/issues/946 **Refactor CredentialsProvider APIs** does not impact us

gpanshu (changes requested):
Comments added

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/liveness/InvalidSignatureException.kt
tjleing: maybe 😭

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -76,6 +78,10 @@ internal class LivenessWebSocket(
     private val signer = AWSV4Signer()
     private var credentials: Credentials? = null
 
+    private val sdf = SimpleDateFormat("yyyymmddhhmmss", Locale.US)
+    internal var offset = 0L
+    internal var closeExpected = false
```
tjleing, line 83: call it `reconnectWithOffset` maybe; set to false afterwards so that we only reconnect once (safeguard)

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -76,6 +78,10 @@ internal class LivenessWebSocket(
     private val signer = AWSV4Signer()
     private var credentials: Credentials? = null
 
+    private val sdf = SimpleDateFormat("yyyymmddhhmmss", Locale.US)
```
tjleing, line 81: unused?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -86,13 +92,31 @@ internal class LivenessWebSocket(
...
 
     @VisibleForTesting
     internal var webSocketListener = object : WebSocketListener() {
         override fun onOpen(webSocket: WebSocket, response: Response) {
             LOG.debug("WebSocket onOpen")
-            super.onOpen(webSocket, response)
-            this@LivenessWebSocket.webSocket = webSocket
+
+            // device time may be set incorrectly; read the header to skew time and retry
+            val sdf = SimpleDateFormat(datePattern, Locale.US)
```
tjleing, line 104: use device Locale

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -338,7 +369,7 @@ internal class LivenessWebSocket(
                         currentColor = currentColor,
                         previousColor = previousColor,
                         sequenceNumber = sequenceNumber,
-                        currentColorStartTimestamp = colorStartTime
+                        currentColorStartTimestamp = colorStartTime + offset
```
tjleing, line 372: `adjustedTimeWithOffset(time: long = Date().time)`

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
ankpshah, line 1: How do we ensure to keep the offset up to date?  It might be essential to consider the potential for time offset drift over the period of the application's usage. Time drift could happen due to various reasons, including daylight saving changes, manual adjustments on the client's device. 
What if the server uses Daylight Saving Time and the device doesn't, or vice versa? Twice a year, the offset could change by an hour. 

For example: Offset might be correctly set initially when web socket is created, lets say offset was 3 hours consider there is a time change (such as daylight saving adjustment) maybe during the same websocket session, actual time difference could be then 2 or 4 hours (depending on the direction of the change).
If offset is not updated to account for this, timestamps will be off by an hour until it is corrected by creating new web socket session.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -86,13 +106,31 @@ internal class LivenessWebSocket(
...
+
+            // device time may be set incorrectly; read the header to skew time and retry
+            val sdf = SimpleDateFormat(datePattern, Locale.getDefault())
+            val date = response.header("Date")?.let { sdf.parse(it) }
+            val tempOffset = if (date != null) {
+                date.time - adjustedDate()
+            } else 0
+
+            // if offset is > 5 minutes, server will reject the request
+            if (kotlin.math.abs(tempOffset) < FIVE_MINUTES) {
```
ankpshah, line 125: Can we also document the mechanism's reliance on session reinitialization to maintain time accuracy. Additionally, should we consider implementing a failsafe mechanism to close and reopen the connection if a session persists longer than expected or a warning that alerts when the session exceeds typical duration thresholds

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -317,6 +325,57 @@ internal class LivenessWebSocketTest {
         assertEquals(livenessWebSocket.getUserAgent(), "$baseline $additional")
     }
 
+    @Test
+    fun `web socket detects clock skew from server response`() {
+        val livenessWebSocket = createLivenessWebSocket()
+        mockkConstructor(WebSocket::class)
+        val socket: WebSocket = mockk()
+        livenessWebSocket.webSocket = socket
+        val sdf = SimpleDateFormat(livenessWebSocket.datePattern, Locale.getDefault())
```
ankpshah, line 334: The SimpleDateFormat is currently using Locale.getDefault(), which might introduce inconsistencies with date parsing if the default locale does not match the expected server's date header format, potentially leading to false negatives in tests.

Suggestion:

```kotlin
val sdf = SimpleDateFormat(livenessWebSocket.datePattern, Locale.US)
```
Using Locale.US is a common practice for parsing HTTP dates, as per RFC 1123.

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -317,6 +325,57 @@ internal class LivenessWebSocketTest {
...
+
+        assertTrue(livenessWebSocket.webSocket != null)
+        val originalRequest = livenessWebSocket.webSocket!!.request()
+
+        // make sure that followup request sends offset date
+        val sdfGMT = SimpleDateFormat("yyyyMMdd'T'HHmmss'Z'", Locale.getDefault())
+        sdfGMT.timeZone = TimeZone.getTimeZone("GMT")
+        val sentDate = originalRequest.url.queryParameter("X-Amz-Date") ?.let { sdfGMT.parse(it) }
+        val diff = abs(Date().time - sentDate?.time!!)
+        assert(oneHour - 10000 < diff && diff < oneHour + 10000)
```
ankpshah, line 366: Need some clarification as to why 10 seconds is an acceptable window here.

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -317,6 +325,57 @@ internal class LivenessWebSocketTest {
...
+
+        // also make sure that followup request is valid
+        assertEquals("AWS4-HMAC-SHA256", originalRequest.url.queryParameter("X-Amz-Algorithm"))
+        assertTrue(
+            originalRequest.url.queryParameter("X-Amz-Credential")!!.endsWith("//rekognition/aws4_request")
+        )
+        assertEquals("299", originalRequest.url.queryParameter("X-Amz-Expires"))
+        assertEquals("host", originalRequest.url.queryParameter("X-Amz-SignedHeaders"))
+        assertEquals("AWS4-HMAC-SHA256", originalRequest.url.queryParameter("X-Amz-Algorithm"))
```
ankpshah, line 375: Seems like a duplicate assert similar to L369

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -317,6 +325,57 @@ internal class LivenessWebSocketTest {
...
+        assertNotNull("x-amz-user-agent")
```
ankpshah, line 376: The assertion does not appear to test anything as it is passing a string literal instead of a variable or a function call result. Maybe you intended to test the query parameter in orignalRequest Object here.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
tylerjroach: The LivenessWebSocket will only live ~15 seconds. It will be recreated with each liveness run.

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
tylerjroach: We don't have an exact time because it was created within the request, so looks like its just a comparison to make sure the time is close (since unit test could vary in time taken. 10 seconds is probably way higher than needed but don't see a problem with it.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -86,13 +106,31 @@ internal class LivenessWebSocket(
...
 
     @VisibleForTesting
     internal var webSocketListener = object : WebSocketListener() {
         override fun onOpen(webSocket: WebSocket, response: Response) {
             LOG.debug("WebSocket onOpen")
-            super.onOpen(webSocket, response)
-            this@LivenessWebSocket.webSocket = webSocket
+
+            // device time may be set incorrectly; read the header to skew time and retry
+            val sdf = SimpleDateFormat(datePattern, Locale.getDefault())
```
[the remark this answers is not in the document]
tylerjroach: This is a good point. If we can verify this behavior that would be great, but good catch here.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -86,13 +106,31 @@ internal class LivenessWebSocket(
     private var webSocketError: PredictionsException? = null
     internal var clientStoppedSession = false
     val json = Json { ignoreUnknownKeys = true }
+    val FIVE_MINUTES = 1000 * 60 * 5
```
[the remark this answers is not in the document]
tylerjroach:
```
val FIVE_MINUTES = 1000 * 60 * 5
al datePattern = "EEE, d MMM yyyy HH:mm:ss z"
```
Are both best in companion object but do not want to delay moving this through on that.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
tylerjroach: Once a session has been successfully established, it can't be re-started because a sessionId for liveness can only be used once. First point on better documenting why we are doing what we are here, would be helpful for future.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -76,6 +78,24 @@ internal class LivenessWebSocket(
     private val signer = AWSV4Signer()
     private var credentials: Credentials? = null
 
+    internal var offset = 0L
+    internal enum class ReconnectState {
+        INITIAL,
+        RECONNECTING,
+        RECONNECTING_AGAIN;
```
tylerjroach, line 85: RECONNECTING_AGAIN is unused?

tylerjroach (commented):
Please see Ankit's date parsing concern. Outside of this, I am ok with PR.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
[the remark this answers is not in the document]
tjleing: `Locale.getDefault()` does throw an error, thanks @ankpshah for the callout -- fixed in most recent commit

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -86,13 +106,31 @@ internal class LivenessWebSocket(
...
+
+            // if offset is > 5 minutes, server will reject the request
+            if (kotlin.math.abs(tempOffset) < FIVE_MINUTES) {
+                super.onOpen(webSocket, response)
+                this@LivenessWebSocket.webSocket = webSocket
+            } else {
+                // server will close this websocket, don't report that failure back
+                reconnectState = ReconnectState.next(reconnectState)
+                offset = tempOffset
+                start()
```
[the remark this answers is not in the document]
tjleing: `start()` reinitializes the websocket with the new skew -- does that answer the question?

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
tjleing: `Locale.getDefault()` does throw an error, thanks @ankpshah for the callout -- fixed in most recent commit

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
tjleing: it was supposed to be caught by an if statement checking if `== ReconnectState.RECONNECTING`, but it looks like the logic wasn't working.  I'm just adding an if statement in `start()` to make sure

ankpshah (approved):

tylerjroach (approved):

Issue #2617, opened 2023-10-25
DataStore Crashes if `clear()` called while PersistentMutationOutbox semaphore is locked


tylerjroach:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

RxJava

### Amplify Categories

DataStore

### Gradle script dependencies

All known versions of DataStore

### Environment information

-

### Describe the bug

DatastoreCanaryTest.delete had to add a workaround to allow DataStore to hit a settled state before clear is called in teardown.

BasicCloudSyncInstrumentationTest.createThenDelete is ignored due to the same issue.

These tests both quickly initialize datastore, create an item, delete an item, then clear datastore. 

The failure is not easy to reproduce, but I believe the cause it due to delete calling `aquire` on a semaphore then `dataStore.clear` interrupting the thread.

The fix may be to simply try/catch semaphore.aquire, as an interruption may be safe that the assumption is do to clear being called.

### Reproduction steps (if applicable)

Re-enable BasicCloudSyncInstrumentationTest.createThenDelete and run the test multiple times.

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

PR #2618, opened 2023-10-25
fix(datastore): Temporary workaround to stabilize DataStore test.

tylerjroach:

*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/2617

*Description of changes:*
DataStore has a bug on `clear` that can cause crashes in some situations. While it is a race condition that is hard to replicate, the above explains the cause. 

The root of the issue needs fixed and can happen in production. While waiting for the correct fix, this PR should at least stabilize a `delete` test that has other valid purposes is its testing, and is important to not ignore, while waiting on a fix.

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreHubEventFilters.java
+++ b/aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreHubEventFilters.java
@@ -53,6 +53,31 @@ public final class DataStoreHubEventFilters {
     }
 
     /**
+     * Watches for publication (out of mutation queue) of a given model.
+     * Creates a filter that catches events from the mutation processor.
+     * Events will pass if they mention the provided model by its name and ID,
+     * and state that it has successfully been published off of the mutation queue.
+     * @param modelName Model name, e.g. "Post"
+     * @param modelId The ID of a model instance that might be published
+     * @return A filter that watches for publication of the provided model.
+     */
+    public static HubEventFilter enqueueOf(String modelName, String modelId) {
+        return event -> {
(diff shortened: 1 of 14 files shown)


aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreHubEventFilters.java
```diff
@@ -52,6 +52,31 @@ public static HubEventFilter publicationOf(String modelName, String modelId) {
...
+     * Events will pass if they mention the provided model by its name and ID,
+     * and state that it has successfully been published off of the mutation queue.
+     * @param modelName Model name, e.g. "Post"
+     * @param modelId The ID of a model instance that might be published
+     * @return A filter that watches for publication of the provided model.
+     */
+    public static HubEventFilter enqueueOf(String modelName, String modelId) {
+        return event -> {
+            if (!DataStoreChannelEventName.OUTBOX_MUTATION_ENQUEUED.toString().equals(event.getName())) {
+                return false;
```
gpanshu, line 67: both these conditions are hard to read. Would it be almost better to return true instead of doing `not` something and returning false as result?

gpanshu (approved):

gpanshu (approved):

aws-datastore/src/androidTest/java/com/amplifyframework/datastore/DataStoreHubEventFilters.java
tylerjroach: Can't return true as there are additional checks below it to run. This code is not new, and only exposed in test package. Made some changes to re-use this code instead of duplicating.

ankpshah (approved):

gpanshu (approved):

PR #2658, opened 2023-12-18
chore(liveness): add close codes to websocket apis

tjleing:

*Description of changes:*
Liveness can pass error code to websocket on close.  [UI PR](https://github.com/aws-amplify/amplify-ui-android/pull/106).

Local testing with breakpoint in `destroy()` to make sure close code is being sent; sending session ids to liveness team.

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
@@ -254,7 +254,7 @@ internal class LivenessWebSocket(
         val deviceManufacturer = Build.MANUFACTURER.replace(" ", "_")
         val deviceName = Build.MODEL.replace(" ", "_")
         var userAgent = "${UserAgent.string()} os/Android/${Build.VERSION.SDK_INT} md/device/$deviceName " +
-            "md/device-manufacturer/$deviceManufacturer api/rekognitionstreaming/$amplifyVersion"
+                "md/device-manufacturer/$deviceManufacturer api/rekognitionstreaming/$amplifyVersion"
 
         if (!livenessVersion.isNullOrBlank()) {
             userAgent += " api/liveness/$livenessVersion"
@@ -481,9 +481,9 @@ internal class LivenessWebSocket(
         }
     }
 
-    fun destroy() {
-        // Close gracefully
-        webSocket?.close(NORMAL_SOCKET_CLOSURE_STATUS_CODE, null)
+    fun destroy(reasonCode: Int = NORMAL_SOCKET_CLOSURE_STATUS_CODE) {
+        // Close with provided reason code
+        webSocket?.close(reasonCode, null)
     }
 
     fun adjustedDate(date: Long = Date().time): Long {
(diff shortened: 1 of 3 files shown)


aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -493,7 +493,7 @@ internal class LivenessWebSocket(
     private fun isTimeDiffSafe(diffInMillis: Long) = kotlin.math.abs(diffInMillis) < FOUR_MINUTES
 
     companion object {
-        private const val NORMAL_SOCKET_CLOSURE_STATUS_CODE = 1000
+        internal const val NORMAL_SOCKET_CLOSURE_STATUS_CODE = 1000
```
ankpshah, line 496: Suggestion: Should we keep access modifier unchanged? destroy() already has default value for parameter reasonCode= NORMAL_SOCKET_CLOSURE_STATUS_CODE, so we might not need to pass this value from caller site when reason code is null.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
```diff
@@ -179,8 +179,8 @@ internal class RunFaceLivenessSession(
         }
     }
 
-    private fun stopLivenessSession() {
+    private fun stopLivenessSession(reasonCode: Int?) {
         livenessWebSocket.clientStoppedSession = true
-        livenessWebSocket.destroy()
+        livenessWebSocket.destroy(reasonCode ?: LivenessWebSocket.NORMAL_SOCKET_CLOSURE_STATUS_CODE)
```
ankpshah, line 184: destroy() already has default value for parameter reasonCode= NORMAL_SOCKET_CLOSURE_STATUS_CODE.
Suggestion: This could be refactored like:
```kotlin
reasonCode?.let { livenessWebSocket.destroy(it) } ?: livenessWebSocket.destroy()
```

This will allow us to keep access modifier for NORMAL_SOCKET_CLOSURE_STATUS_CODE unchanged.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
```diff
@@ -16,14 +16,13 @@
...
 import com.amplifyframework.annotations.InternalAmplifyApi
-import com.amplifyframework.core.Action
 
 @InternalAmplifyApi
 class FaceLivenessSession(
     val challenges: List<FaceLivenessSessionChallenge>,
     private val onVideoEvent: (VideoEvent) -> Unit,
     private val onChallengeResponseEvent: (ChallengeResponseEvent) -> Unit,
-    private val stopLivenessSession: Action
+    private val stopLivenessSession: (Int?) -> Unit
```
tylerjroach, line 25: This check would break ui unit tests if we ran the tests against the latest version of Amplify

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
```diff
@@ -34,7 +33,8 @@ class FaceLivenessSession(
         onChallengeResponseEvent(challengeResponseEvent)
     }
 
-    fun stopSession() {
-        stopLivenessSession.call()
+    @JvmOverloads
```
tylerjroach, line 36: Lets investigate why this is necessary.

tylerjroach (approved):
I'd like to further investigate why jvmoverloads is needed when we are going Kotlin to Kotlin. May have something to do with how we've compiled against a different version of the lib.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
tjleing: unit tests have been fixed in the [UI PR](https://github.com/aws-amplify/amplify-ui-android/pull/106)

ankpshah (approved):

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
tjleing: Seems like the consensus is that the UI library at a lower version is compiled expecting no argument.  If UI was compiled against this updated version of Amplify it would have a virtual reference to the correct function.  The JVM overloads are useful because Kotlin compiles down to the JVM.

Issue #2667, opened 2024-01-03
DataStore createdAt and updatedAt fields aren't reconciled properly


tylerjroach:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

DataStore

### Gradle script dependencies

N/A

### Environment information

N/A

### Describe the bug

```
type Todo @model {
  id: ID!
  name: String!
}
```

Using an example schema for Todo as above, where createdAt and updatedAt are automatically added because of DataStore enablement.

Upon create a new Todo item, createdAt and updatedAt are initially expected to be null in the local db.

However, once the create request is sent to the server, the response should be reconciled with the local data.

In my test scenario, I'm noticing about 90% of the time, the models never reconcile createdAt and updatedAt values that were created by the server. These fields will remain null on the local device unless the model was updated from a different device (which would update version and force a sync reconciliation on the model).

From speaking with iOS team, it's possible reconciliation may be done on the subscription websocket and/or actual network request, whichever comes first. If this is Android's behavior, my speculation is that one of these processes is not reconciling the 2 fields, while the other one does.

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

tylerjroach:
The cause of the issue is:

1. Item is saved
2. Item is submitted via Outbox.
3. Subscription recognizes and begins processing the new item. At this time, the item is still in the outbox, so only the Metadata is saved.
4. The outbox response begins being processed. Because the item has sync metadata saved, and the version matches the response, the merge is ignored.

Need to figure out the proper sequence of events.

PR #2671, opened 2024-01-05
fix(Datastore): Ensure proper data merger reconciliation

tylerjroach:

*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/2667

*Description of changes:*

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/Merger.kt
@@ -86,10 +86,16 @@ internal class Merger(
             // fetch a Map of all model versions from Metadata table
             val localModelVersions = versionRepository.fetchModelVersions(modelsWithMetadata)
 
-            // fetch a Set of all pending mutation ids for type T
+            /*
+            Fetch a Set of all pending mutation ids for type T
+            We exclude in-flight mutations from the returned pending mutation. If a mutation is excluded, it is
+            likely that the subscription processor returned the item before the outbox response. We want to process
+            whichever comes first
+             */
             val pendingMutations = mutationOutbox.fetchPendingMutations(
-                modelsWithMetadata.map { it.model },
-                modelsWithMetadata.first().model.javaClass.name
+                models = modelsWithMetadata.map { it.model },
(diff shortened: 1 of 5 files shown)


tylerjroach:
See the bug description on the issue here:
https://github.com/aws-amplify/amplify-android/issues/2667

When `DataStore.save(model)` is called and remote reconciliation begins, there are 2 ways in which the remote model can re-sync with the local db (The websocket subscription and the POST network response from the model save). This merger reconciliation process is what will provide `createdAt` and `updatedAt` values.

The DataStore spec states that reconciliation should be completed from the 1st received response (subscription or POST response), and the later response should be discarded.

If the POST response arrived first, reconciliation would complete successfully.

If the websocket subscription message arrived first, we had a bug.


1. DataStore.save(model) called, saving the model to the Model table and PendingMutation table.
2. Outbox sees the save and POSTs to the API
3. Subscription message picks up successful model creation and starts merger.
4. When the merger runs, a local SQLite query is completed to determine if the model exists in the PendingMutation record. If a model exists in PendingMutation table, we only reconcile the sync ModelMetadata, not the model itself.
5. ModelMutation is saved to local db with version = 1, which will match the version of the saved model in DB.
6. The POST response is finally received, which kicks off the reconciliation merger.
7. This time, the merger detects that the model version from the server matches the version saved in the local db and it ignores the response. This will continue on any future syncs as records are only updated if the version increases.

The bug is at step 4. In a normal scenario, saving only the sync metadata would be ok, it causes issues here because we aren't recognizing that the model from the server is no longer a pending mutation. The pending mutation record will only get deleted in step 6 when the POST response is processed.

We have a Set of `inFlightMutations` where we are aware that the given model is in-flight. The change we make here is: If a model is "in-flight", we do not consider it a PendingMutation. This will allow the merger to update both the sync metadata AND the model itself. This matches Swift behavior.

Additional Notes:
`inFlightMutations` is a `Set`, but theoretically, should only ever have 1 model at a time, since we only POST model updates 1 at a time.

mattcreaser (approved):

ankpshah (approved):

Issue #2681, opened 2024-01-16
Crash when AndroidKeyStoreKeyGeneratorSpi.engineGenerateKey


geek5nan:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
// Put output below this line
com.amplifyframework:aws-core:2.14.8
com.amplifyframework:aws-auth-cognito:2.14.8
com.amplifyframework:aws-storage-s3:2.14.8

```

</details>

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

**Crashlytics**

This issue has occurred 178 times and has affected 9 users.

87% Xiaomi
86% Android 9

```
Fatal Exception: java.security.ProviderException: Keystore operation failed
       at android.security.keystore.AndroidKeyStoreKeyGeneratorSpi.engineGenerateKey(AndroidKeyStoreKeyGeneratorSpi.java:318)
       at javax.crypto.KeyGenerator.generateKey(KeyGenerator.java:612)
       at androidx.security.crypto.MasterKeys.generateKey(MasterKeys.java:130)
       at androidx.security.crypto.MasterKeys.getOrCreate(MasterKeys.java:88)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:37)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:34)
       at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository.getSharedPreferences$com_amplifyframework_core_release(EncryptedKeyValueRepository.kt:34)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository.get(EncryptedKeyValueRepository.kt:56)
       at com.amplifyframework.auth.cognito.data.AWSCognitoAuthCredentialStore.retrieveCredential(AWSCognitoAuthCredentialStore.kt:63)
[...]

Issue #2616, opened 2023-10-25
Crash: AndroidKeystoreAesGcm.encryptInternal


ygag-gopikrishnan:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Storage

### Gradle script dependencies

<details>

```groovy
// Put output below this line
api "com.amplifyframework:core-kotlin:2.8.2"
 api "com.amplifyframework:aws-storage-s3:2.8.2"
 api "com.amplifyframework:aws-auth-cognito:2.8.2"

```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 7.6-rc-1
------------------------------------------------------------

Build time:   2022-10-28 07:52:11 UTC
Revision:     62973455a8df25d003949f3f573593cd10d083a3

Kotlin:       1.7.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.18 (Homebrew 11.0.18+0)
OS:           Mac OS X 13.2.1 aarch64

```

</details>

### Describe the bug

```
Fatal Exception: java.security.KeyStoreException: the master key android-keystore://_androidx_security_master_key_ exists but is unusable
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.readOrGenerateNewMasterKey(AndroidKeysetManager.java:275)
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.build(AndroidKeysetManager.java:236)
       at androidx.security.crypto.EncryptedSharedPreferences.create(EncryptedSharedPreferences.java:123)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:35)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:34)
       at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository.getSharedPreferences$com_amplifyframework_core(EncryptedKeyValueRepository.kt:34)
[...]

Issue #2548, opened 2023-08-07
AndroidKeystoreAesGcm.encryptInternal error


r-elbouhi:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

Analytics

### Gradle script dependencies

<details>

```
implementation 'com.amplifyframework:aws-analytics-pinpoint:2.8.5'
implementation 'com.amplifyframework:aws-auth-cognito:2.8.5'
coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'

```

</details>

### Environment information

<details>

```
Kotlin: 1.6.21
Groovy: 3.0.10
JVM: 19.0.2


```

</details>

### Describe the bug

Our crash monitoring is reporting a lot of occurrences of a crash for one single user running Android 13 on Xiaomi Poco F3, with the app in the background ,RAM free: 3.02 GB and Disk free: 67.14 GB.

I couldn't reproduce the issue in any way

```
Fatal Exception: java.security.KeyStoreException: the master key android-keystore://_androidx_security_master_key_ exists but is unusable
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.readOrGenerateNewMasterKey(AndroidKeysetManager.java:275)
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.build(AndroidKeysetManager.java:236)
       at androidx.security.crypto.EncryptedSharedPreferences.create(EncryptedSharedPreferences.java:123)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:35)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository$sharedPreferences$2.invoke(EncryptedKeyValueRepository.kt:34)
       at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository.getSharedPreferences$com_amplifyframework_core(EncryptedKeyValueRepository.kt:34)
       at com.amplifyframework.core.store.EncryptedKeyValueRepository.get(EncryptedKeyValueRepository.kt:56)
[...]

Issue #2510, opened 2023-07-11
KeyStoreException when AWS Cognito tries to create EncryptedSharedPreferences


reisfernanda:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```groovy
implementation("com.squareup.okhttp3:okhttp:5.0.0-alpha.11")
implementation("com.amplifyframework:aws-auth-cognito:2.8.7")
implementation("com.amplifyframework:core-kotlin:2.8.7")
```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 7.6.2
------------------------------------------------------------

Build time:   2023-06-30 15:42:51 UTC
Revision:     dab132169006b16e7ada4ab2456e0c9d6415b52a

Kotlin:       1.7.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.12 (Oracle Corporation 11.0.12+8-LTS-237)
OS:           Windows 10 10.0 amd64
```

</details>

### Describe the bug

Our crash monitoring is reporting some occurrences of an `android.security.KeyStoreException` (`com.google.crypto.tink.integration.android.AndroidKeystoreAesGcm.encryptInternal`) under the creation of `EncryptedSharedPreferences` (`com.amplifyframework.core.store.EncryptedKeyValueRepository`).
I couldn't reproduce the issue in any way, but the rare users that find this problem can reproduce it frequently.
I'm not sure when this is happening but I believe it's when the app is starting since 100% of the occurrences happened with the app in background, and also because these occurrences happen before the application can send logs related to fragments creation.
This crash happens on the devices:
- Motorola Edge 30 Pro - Android 13
- Samsung - Android 12 (the crash monitoring tool does not inform the specific model of this Samsung device)

Another crash with the same exception also happens but for `com.google.crypto.tink.integration.android.AndroidKeystoreAesGcm.decryptInternal`

### Code Snippet

```kotlin
class AppApplication : Application() {

    override fun onCreate() {
        super.onCreate()
        try {
            Amplify.addPlugin(AWSCognitoAuthPlugin())
            Amplify.configure(context)
[...]

PR #2686, opened 2024-01-23
fix(all): Add fallback logic for corrupt keys to EncryptedKeyValueRepository

mattcreaser:

*Issue #, if available:*
#2681 
#2684 
#2548 
#2616 
#2510 

*Description of changes:*
This PR aims to implement workarounds for corrupted master keys causing crashes when attempting to open the encrypted shared preferences. 

Certain OEM devices have unreliable KeyStore implementations that can occasionally corrupt their key material. When this happens the keys become unusable, and the encrypted preferences are therefore unretrievable. The workaround is to catch this error, create a new key to use, and create a new set of encrypted preferences. Please see the [Google tracker bug](https://issuetracker.google.com/issues/176215143) and [tink issue](https://github.com/google/tink/issues/535) for more details.

**NB**: This is destructive, as a corrupt key means we can never recover any previously stored data.

At a high level, the logic in this change can be described as follows:
- For existing users that have repositories encrypted using the default master key there will be no change, such repositories will continue to use this key as normal.
- Any new repositories created will instead use an amplify-specific master key to derive their key material.
- If the default master key becomes corrupted then we delete the preferences instance and create a new one using the amplify master key.
- If the amplify master key becomes corrupted then we delete and recreate the key, then delete and recreate the preferences instance using the new key.

I simulated various keystore errors by manually throwing exceptions using a debugger. I tried various scenarios like an existing user upgrading to a new version without errors, an existing user encountering a corrupt default master key, and corrupt amplify master keys. I tried having an open repository and deleting the master key in a different instance.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
```diff
@@ -17,93 +17,193 @@ package com.amplifyframework.core.store
...
+            "amplify_master_key",
+            KeyProperties.PURPOSE_ENCRYPT or KeyProperties.PURPOSE_DECRYPT
+        ).setBlockModes(KeyProperties.BLOCK_MODE_GCM)
+            .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_NONE)
+            .setKeySize(256)
+            .build()
+
+        // This prefix is used to identify repositories encrypted with the amplifyMasterKey instead of the
+        // defaultMasterKey
+        @VisibleForTesting const val amplifyIdentifierPrefix = "__amplify__"
```
tylerjroach, line 207: make internal?

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
```diff
@@ -17,93 +17,193 @@ package com.amplifyframework.core.store
...
         try {
-            identifierFile.writeText(newIdentifier)
+            identifierFile.writeText(identifier)
         } catch (e: Exception) {
             // Failed to write identifier to file, session will be forced to be in memory
         }
-        return newIdentifier
+    }
+
+    companion object {
```
tylerjroach, line 189: Should this be internal?

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
```diff
@@ -17,93 +17,193 @@ package com.amplifyframework.core.store
...
 
-    override fun put(dataKey: String, value: String?) {
-        with(editor) {
-            putString(dataKey, value)
-            apply()
+    private fun getMasterKeyOrNull(spec: KeyGenParameterSpec): String? {
+        // Getting the Master Key should succeed, but keystore bugs in some OEM implementations mean that
+        // the master key occasionally becomes corrupted on some devices. We make multiple attempts to ensure that
+        // the error is not transient.
+        repeat(3) { attempt ->
```
tylerjroach, line 129: good idea with repeat. Not sure if the error is ever transient but good safety mechanism.

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
```diff
@@ -17,93 +17,193 @@ package com.amplifyframework.core.store
...
             AES256_GCM
         )
-    }
 
-    @VisibleForTesting
-    internal val editor: SharedPreferences.Editor by lazy {
-        sharedPreferences.edit()
+    private fun deleteSharedPreferences(fileName: String) = context.deleteSharedPreferences(fileName)
+
+    private fun deleteMasterKey(spec: KeyGenParameterSpec) = KeyStore.getInstance("AndroidKeyStore").run {
```
tylerjroach, line 120: should we make this method more focused? Ex: deleteAmplifyMasterKey? Especially since we would never want to accidentally introduce deleting app master key.

tylerjroach (commented):
A few general comments and before approval I'd like to do some manual test checks but this looks great overall.

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
```diff
@@ -17,93 +17,193 @@ package com.amplifyframework.core.store
...
+    )
+
+    private val sharedPreferences by lazy { getOrCreateSharedPreferences() }
+
+    override fun put(dataKey: String, value: String?) = edit { putString(dataKey, value) }
+    override fun get(dataKey: String): String? = sharedPreferences.getString(dataKey, null)
+    override fun remove(dataKey: String) = edit { remove(dataKey) }
+    override fun removeAll() = edit { clear() }
+
+    private inline fun edit(crossinline block: SharedPreferences.Editor.() -> Unit) = with(sharedPreferences.edit()) {
```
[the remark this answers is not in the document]
mattcreaser: Yes this is best practice when using shared preferences. Caching the editor leads to risk of cross-mingling changesets. It's just instantiating a new object so there's virtually zero overhead.

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
mattcreaser: Good callout, initially I was working with multiple master keys before settling on just one. I'll update.

tylerjroach (approved):

tylerjroach (approved):

core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt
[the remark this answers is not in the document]
ankpshah: Do we have a use case to apply set of changes atomically? While each `edit` operation is atomic, using multiple `edit` calls for related preferences might not guarantee atomic changes across all those changes.

ankpshah, line 55: Do we have a use case to apply set of changes atomically? While each edit operation is atomic, using multiple edit calls for related preferences might not guarantee atomic changes across all those changes.

mattcreaser: No, this class does not offer that functionality.

ankpshah (approved):

PR #2718, opened 2024-02-16
fix(datastore): Remove unnecessary synchronized causing subscription slowness

tylerjroach:
- [ ] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*

*Description of changes:*

When DataStore is started, Create, Update, Delete subscriptions are made for each model. If I have 10 models, this results in 30 subscriptions established. Subscriptions are established over a single websocket connection by sending a websocket event per subscription. We wait until all subscriptions have been established (ws response message for each subscription) before we proceed with syncing/merging items.

An errant `synchronized` is causing subscription events to being be processed 1 at a time, and not in parallel.

If I have 10 models (30 subscriptions) and each ws event takes the server 200ms to acknowledge, establishing subscriptions would take the current build ~ 6 seconds. Allowing ws events in parallel will complete this effort in ~ 200ms.

Buggy State: 
Send Model1 Create -> Model1 Create Response -> Send Model1 Delete -> Model1 Delete Response -> Send Model1 Update -> Model1 Update Response -> Send Model2 Create -> Model2 Create Response -> ...

Fixed State: 
Send Model1 Create, Send Model1 Delete, Send Model1 Update, Send Model2 Create .... -> Wait for all acks

* The method is synchronized (problem): [https://github.com/aws-amplify/amplify-android/blob/2145c0ac26ab9b406c36b299b3ab96[…]ain/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java](https://github.com/aws-amplify/amplify-android/blob/2145c0ac26ab9b406c36b299b3ab96c116ec2b60/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java#L120)
* Inside the method, there is another synchronized lock (necessary) to only create 1 WS: [https://github.com/aws-amplify/amplify-android/blob/2145c0ac26ab9b406c36b299b3ab96[…]ain/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java](https://github.com/aws-amplify/amplify-android/blob/2145c0ac26ab9b406c36b299b3ab96c116ec2b60/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java#L137)
[...]

PR #2719, opened 2024-02-16
datastore(fix): Fixes previous PR which was inadvertently creating multiple websockets.

tylerjroach:

*Issue #, if available:*

*Description of changes:*

Continuation of https://github.com/aws-amplify/amplify-android/pull/2718

The previous PR was inadvertently creating multiple websockets. This is because the websocket lock had a check 
`if (webSocketListener == null || webSocketListener.isDisconnectedState())` which would create a new ws. Because we were rapidly sending, every subscription request was in a disconnected state. 

This PR waits for the ws connection to open, before it adds the subscriptions

Additionally, I also discovered that disconnecting the websocket takes too long due to the same issue we observed in https://github.com/aws-amplify/amplify-android/pull/2718 with the creation of the subscriptions.

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -141,17 +141,17 @@ final class SubscriptionEndpoint {
             pendingSubscriptionIds.add(subscriptionId);
             socketListener = webSocketListener;
             socket = webSocket;
-        }
 
-        // Every request waits here for the connection to be ready.
-        Connection connection = socketListener.waitForConnectionReady();
-        if (connection.hasFailure()) {
-            // If the latch didn't count all the way down
-            if (pendingSubscriptionIds.remove(subscriptionId)) {
-                // The subscription was pending, so we need to emit an error.
-                onSubscriptionError.accept(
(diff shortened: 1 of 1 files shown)


mattcreaser (commented):
At first I was concerned because we would lose the ability to have multiple threads waiting for the connection concurrently, but that isn't a concerned since `waitForConnection` will return immediately after connected - so threads are simply waiting for `webSocket` lock instead, and there's no impact to throughput.

Removal of `synchronize` on `releaseSubscriptions` also looks good, and allows releasing multiple subscriptions in parallel to speed up the shutdown process.

I think my main question is whether the `isDisconnected()` check even makes sense while establishing subscriptions. If the socket is closed due to failure then error handling should have already been in progress, otherwise if it was a graceful close then `webSocketListener` will already be null. Recognize that isn't really part of this PR but something worth thinking about.

mattcreaser (approved):

mattcreaser:
👍 Looks good in testing. Subscription establishment is much faster than in `2.14.10` (and the improvement is more significant the more models you have) and `Datastore.stop()` is now also much faster. I also tested the "stop->start" scenario and that worked as expected as well.

ankpshah (approved):

Issue #2751, opened 2024-04-02
IndentityId get refreshed after one hour, provoking 401 code response


yktoledo:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication, REST API

### Gradle script dependencies

<details>

```groovy
// Put output below this line
    implementation 'com.amplifyframework:core:2.14.11'
    implementation 'com.amplifyframework:aws-api:2.14.11'
    implementation 'com.amplifyframework:aws-auth-cognito:2.14.11'

```

</details>

### Environment information

<details>

```
# Put output below this line

------------------------------------------------------------
Gradle 8.2
------------------------------------------------------------

Kotlin:       1.8.20
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          17.0.7 (JetBrains s.r.o. 17.0.7+7-b985.2)
OS:           Mac OS X 14.4.1 aarch64


```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/android/build-a-backend/auth/enable-guest-access/

### Describe the bug

I want to consume API Gateway resource using `authorizationType:AWS_IAM` and "Enable guest access" (Allow unauthenticated logins), based on the [documentation](https://docs.amplify.aws/android/build-a-backend/auth/enable-guest-access/#:~:text=you%20can%20obtain,method),  "you can obtain the unique Identity ID which is assigned to the device through the fetchAuthSession method", with the intention to use this "identity id" as the device identifier to communicate, after a little a bit more than one hour of fresh install,  the credentials get refreshed, which is perfect, but the identityId too, I will assume is not the expected behavior. 

Maybe I'm missing something, server configuration specific or android implementation detail, despite is so minimal the amount of code use, 4 line of code, add plugins call with api and auth, follow by the configure call, after that just the get call.

Following the doc https://docs.amplify.aws/android/build-a-backend/restapi/fetch-data/

```
// add this to log
Amplify.addPlugin(AndroidLoggingPlugin(LogLevel.VERBOSE))

Amplify.addPlugin(AWSApiPlugin())
Amplify.addPlugin(AWSCognitoAuthPlugin())
[...]

PR #2752, opened 2024-04-03
Fix issue where unauthenticated logins caused 401 because refresh caused new credentials to be created

marcoatblueowl:

*Issue #, if available:*
#2751

*Description of changes:*
These changes were based off how amplify-swift state machine behaves for same scenario.
For the iOS SDK:
When a RefreshSession event occurs for an IdentityPool only, then a RefeshUnAuthAwsCredentials event is created with the identity ID. [Refer to code here](https://github.com/aws-amplify/amplify-swift/blob/946600a7a2289fbded6ff2c3a1a8d19e98a7475f/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/InitializeRefreshSession.swift#L30)

Then when the event is processed in the state machine, it send an action to FetchAuthAWSCredentials with the passed identity ID and an empty login map. [Refer to code here](https://github.com/aws-amplify/amplify-swift/blob/946600a7a2289fbded6ff2c3a1a8d19e98a7475f/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/Resolvers/RefreshSession/RefreshSessionState%2BResolver.swift#L43)

However on Android SDK:
When a RefreshSession event occurs for an IdentityPool only, then a RefreshUnAuthSession event is created but the identity ID that the session needs to be refreshed for is not passed in. This is essentially where I believe we lose track of the anonymous user so when we create credentials, it creates it for a new user. [Refer to code here](https://github.com/aws-amplify/amplify-android/blob/0f61cc8e54adb65a3272ceeb9c06b7d2a0f09efc/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/AuthorizationCognitoActions.kt#L108)

Then when the event is processed in the state machine, it sends an action to RefreshAuthSessionAction instead of FetchAWSCredentialsAction with the identity ID and empty login map. [Refer to code here](https://github.com/aws-amplify/amplify-android/blob/0f61cc8e54adb65a3272ceeb9c06b7d2a0f09efc/aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/states/RefreshSessionState.kt#L90)

The updates to this are to reflect the same logic that is expressed on the iOS SDK as going thru the same scenario there the correct identity ID is being kept and session is being refreshed as well. Please let me know if I misunderstood the logic of both SDKs and if more information is needed. Thanks

- [X] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/AuthorizationCognitoActions.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/AuthorizationCognitoActions.kt
@@ -105,7 +105,10 @@ internal object AuthorizationCognitoActions : AuthorizationActions {
                     RefreshSessionEvent.EventType.RefreshUserPoolTokens(amplifyCredential.signedInData)
                 )
                 is AmplifyCredential.IdentityPool -> RefreshSessionEvent(
-                    RefreshSessionEvent.EventType.RefreshUnAuthSession(amplifyCredential.identityId, LoginsMapProvider.UnAuthLogins())
+                    RefreshSessionEvent.EventType.RefreshUnAuthSession(
+                        amplifyCredential.identityId,
+                        LoginsMapProvider.UnAuthLogins()
+                    )
                 )
                 is AmplifyCredential.IdentityPoolFederated -> {
                     AuthorizationEvent(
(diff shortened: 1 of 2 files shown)


mattcreaser:
Thanks for posting this PR @marcoatblueowl! A member of our team will have a look and help move this PR along.

joon-won (approved):

mattcreaser (approved):

Issue #2789, opened 2024-04-24
SDK v2.14.13 prevented events from being sent with the error “The security token included in the request is invalid” 


jin-jobs:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

Analytics

### Gradle script dependencies

<details>

```groovy
    api ("com.amplifyframework:aws-analytics-pinpoint:2.14.13") {
        exclude( group = "androidx.lifecycle", module = "lifecycle-viewmodel-ktx")
    }
    api ("com.amplifyframework:aws-auth-cognito:2.14.13") {
        exclude( group = "androidx.lifecycle", module = "lifecycle-viewmodel-ktx")
    }


```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 8.1.1
------------------------------------------------------------

Build time:   2023-04-21 12:31:26 UTC
Revision:     1cf537a851c635c364a4214885f8b9798051175b

Kotlin:       1.8.10
Groovy:       3.0.15
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          19.0.1 (Oracle Corporation 19.0.1+10-21)
OS:           Mac OS X 12.0.1 aarch64


```

</details>

### Describe the bug

In SDK v2.14.13, Events could not be sent with the error “The security token included in the request is invalid”.
Our app was previously using SDK v1.37.3, but we upgraded to SDK v2.14.13 in early April and applied it to our app.

During the internal testing process after applying the SDK, amplify event was intermittently not sent, and the following logs were generated in logcat at that time. 

```javascript
Failed to submit events aws.sdk.kotlin.services.pinpoint.model.PinpointException: The security token included in the request is expired
```
<img width="868" alt="스크린샷 2024-04-24 오전 9 19 03" src="https://github.com/aws-amplify/amplify-android/assets/17607205/3a9f0a0a-705f-4ec4-81cc-86142b1b9b2f">

Our team tried to reproduce the error, but it was difficult to get an easy reproduction path. 
We deployed the app with v2.14.13 to the market and checked the Amplify Event data about two weeks later, and found that events were being collected at a very low volume compared to before. 
We recognized this as a very big issue and are urgently investigating and analyzing the issue. 

Question:
[...]

PR #2799, opened 2024-04-30
fix(auth): Fix timestamp issue with v1 to v2 migration

vincetran:

*Issue #, if available:* https://github.com/aws-amplify/amplify-android/issues/2789

*Description of changes:* 
V1 was storing session expiration timestamp as epoch milliseconds and V2 was reading/storing it as epoch seconds. The migration code did not account for this and users that migrated would store a session expiration that would be many many years in the future.

This change adds some logic to the migration code to detect this and covert from milli to sec for new users that still need to migrate.

This change also adds a check in the isValidSession function for users who have already migrated and needs a way to fix themselves.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/AWSCognitoLegacyCredentialStore.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/AWSCognitoLegacyCredentialStore.kt
@@ -182,16 +182,16 @@ internal class AWSCognitoLegacyCredentialStore(
         val accessKey = idAndCredentialsKeyValue.get(namespace(AK_KEY))
         val secretKey = idAndCredentialsKeyValue.get(namespace(SK_KEY))
         val sessionToken = idAndCredentialsKeyValue.get(namespace(ST_KEY))
-        var expiration = idAndCredentialsKeyValue.get(namespace(EXP_KEY))?.toLongOrNull()
 
         // V2 AWSCredential expiration is in epoch seconds whereas legacy expiration may be in epoch milliseconds
         // Session expiration should be within 24 hours so if we see a date in the far future, we can assume the
         // timestamp is encoded in milliseconds.
-        val expirationInstant = expiration?.let { Instant.ofEpochSecond(it) }
-        if (expiration != null &&
-            expirationInstant?.isAfter(Instant.now().plus(365, ChronoUnit.DAYS)) == true
-        ) {
-            expiration /= 1000
+        val expiration = idAndCredentialsKeyValue.get(namespace(EXP_KEY))?.toLongOrNull()?.let {
+            if (Instant.ofEpochSecond(it).isAfter(Instant.now().plus(365, ChronoUnit.DAYS))) {
+                it / 1000
+            } else {
+                it
+            }
         }
 
         return if (accessKey == null && secretKey == null && sessionToken == null) {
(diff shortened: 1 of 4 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/AWSCognitoLegacyCredentialStore.kt
```diff
@@ -180,7 +182,17 @@ internal class AWSCognitoLegacyCredentialStore(
         val accessKey = idAndCredentialsKeyValue.get(namespace(AK_KEY))
         val secretKey = idAndCredentialsKeyValue.get(namespace(SK_KEY))
         val sessionToken = idAndCredentialsKeyValue.get(namespace(ST_KEY))
-        val expiration = idAndCredentialsKeyValue.get(namespace(EXP_KEY))?.toLongOrNull()
+        var expiration = idAndCredentialsKeyValue.get(namespace(EXP_KEY))?.toLongOrNull()
```
tylerjroach, line 185:
```
val expiration = idAndCredentialsKeyValue.get(namespace(EXP_KEY))?.toLongOrNull()?.let {
    if (Instant.ofEpochSecond(it).isAfter(Instant.now().plus(365, ChronoUnit.DAYS))) {
        it / 1000
    } else {
        it
    }
}
```

suggestion to prefer val over var and less null checking for same reuslt.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/helpers/SessionHelperTests.kt
```diff
@@ -34,6 +40,14 @@ class SessionHelperTests {
         expiration = 0
     )
 
+    // Encoded Date Time: April 30, 2024 02:00AM
+    private val currentInstant = Instant.ofEpochSecond(1714467600)
+
+    @After
+    fun teardown() {
+        unmockkStatic(Instant::class)
```
tylerjroach, line 48: Do you feel that it was better to mock Instant.now() static method versus using a real implementation? I think your test would work just the same without it since you aren't asserting on actual expiration values but just that the credential is valid or not.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/data/AWSCognitoLegacyCredentialStore.kt
vincetran: Good point, will change

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/helpers/SessionHelperTests.kt
vincetran: That's a good point! I think when I first started writing these, I had a more complicated test setup and when I made them simpler, I just left the mocking. I've removed the mocking and the tests work just fine still 👌

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/helpers/SessionHelperTests.kt
```diff
@@ -34,6 +36,9 @@ class SessionHelperTests {
         expiration = 0
     )
 
+    // Encoded Date Time: April 30, 2024 02:00AM
+    private val currentInstant = Instant.ofEpochSecond(1714467600)
```
tylerjroach, line 40: No longer needed.

vincetran: Can't believe ktlint didn't catch that 🤦

tylerjroach (approved):

yuhengshs (approved):

joon-won (approved):

PR #2808, opened 2024-05-07
DataStore Model Sync Parallelization

tylerjroach:

This PR attempts to speed up sync times when no associations are present on any models. This is to bring Android into parity with Swift.

Future optimizations could be:
1. To create model association groups to allow for parallelization when connected models are present.
2. Enable a high concurrency count by default. Right now, the proposal on Android is to default to 1.
    * I have some concerns about high memory usage on Android if the customer has a large number of models, with a large number of records per model. An Android app is only allowed an allotted size of memory. In the current Android implementation, network requests are accumulated while local merger reconciliation is processing. By increasing the parallelization limit, if the local merger reconciliation limit is the bottleneck, a large number of records could be allowed to accumulate in memory before the merger can discard network responses. Many customers have a large number of models, but small number of records per model. For these customers, the time savings by enabling concurrency will be significant. Swift apps are allowed to use a higher amount of memory so there is less of a concern on that platform.

*Issue #, if available:*

*Description of changes:*
Comparable Swift Optimization: https://github.com/aws-amplify/amplify-swift/pull/1267

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
```diff
@@ -126,16 +126,39 @@ Completable hydrate() {
             TopologicalOrdering.forRegisteredModels(schemaRegistry, modelProvider);
         Collections.sort(modelSchemas, ordering::compare);
         ArrayList<String> toBeSyncedModelArray = new ArrayList<>();
+        boolean syncInParallel = true;
```
lawmicha, line 129: [nit] I think there may be slight differences in the meaning of "concurrent" and "parallel", and so it may be better to keep the terminology the same, by naming it `syncConcurrently` or `canSyncConcurrently` since the configuration name is "syncMaxConcurrentModels". Later the code will read

"if canSyncConcurrently && syncMaxConcurrentModels > 1"

lawmicha (dismissed):

lawmicha (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -71,6 +74,8 @@ private DataStoreConfiguration(Builder builder) {
         this.syncIntervalInMinutes = builder.syncIntervalInMinutes;
         this.syncExpressions = builder.syncExpressions;
         this.doSyncRetry = builder.doSyncRetry;
+        this.syncMaxConcurrentModels = builder.syncMaxConcurrentModels != null ?
```
mattcreaser, line 77: This has to be nullable in the builder because of the weird user provided thing, right?

mattcreaser (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
tylerjroach: Yep

PR #2815, opened 2024-05-10
chore: Move to using the Maven Publish DSL

mattcreaser:

*Issue #, if available:*

*Description of changes:*
Migrated publishing to use the [publishing DSL](https://developer.android.com/build/publish-library/configure-pub-variants#single-pub-var). This is a pre-requisite to upgrading to Gradle 8+.

The publishing DSL adds `scope` information and publishes Gradle metadata by default. Although those are beneficial, they could potentially impact customers and should not be added in a minor version, so I've disabled/removed both of those. They can be enabled in a future major version release.

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -180,6 +180,12 @@ fun Project.configureAndroid() {
                 )
             )
         }
+
+        publishing {
+            singleVariant("release") {
+                withSourcesJar()
+            }
+        }
     }
 
     dependencies {
--- a/configuration/publishing.gradle
+++ b/configuration/publishing.gradle
@@ -42,12 +42,6 @@ def getRepositoryPassword() {
 }
 
 afterEvaluate { project ->
-    task androidSourcesJar(type: Jar) {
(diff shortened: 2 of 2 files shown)


mattcreaser:
Note: the one difference I've noted with this PR is that it adds the Kotlin standard library as a dependency to all poms, specifically 
`<artifactId>kotlin-stdlib-jdk8</artifactId>`.

I am not actually sure why it wasn't being added before since the Kotlin plugin adds it to all projects so it always has been a dependency. Looking into whether there is any problem here.

vincetran (approved):

tylerjroach (approved):

mattcreaser:
After some additional testing I don't believe there is any downside to this additional dependency being added. If app is running on JDK < 26 it does **not** add a requirement for desugaring unless we invoke one of the JDK8 APIs, which already would have been requiring desugaring previously.

Issue #2770, opened 2024-04-05
Combining API KEY authorization "read" and "listen" access doesn't work


gpavlov2016:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-js/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-js/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-js/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### JavaScript Framework

Next.js

### Amplify APIs

Authentication, GraphQL API

### Amplify Version

v6

### Amplify Categories

auth

### Backend

Amplify Gen 2 (Preview)

### Environment information

<details>

```
# Put output below this line
System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i9-13900H
    Memory: 3.07 GB / 31.68 GB
  Binaries:
    Node: 18.19.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.15.4 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Chromium (122.0.2365.92)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    %name%:  0.1.0
    @ampproject/toolbox-optimizer:  undefined ()
    @aws-amplify/backend: ^0.13.0-beta.13 => 0.13.0-beta.13
    @aws-amplify/backend-cli: ^0.12.0-beta.15 => 0.12.0-beta.15
    @aws-amplify/ui-react: ^6.1.6 => 6.1.6
    @aws-amplify/ui-react-internal:  undefined ()
    @aws-sdk/client-s3: ^3.540.0 => 3.540.0
    @aws-sdk/s3-presigned-post: ^3.540.0 => 3.540.0
    @aws-sdk/signature-v4-crt: ^3.535.0 => 3.535.0
    @babel/core:  undefined ()
    @babel/runtime:  7.22.5
    @edge-runtime/cookies:  4.1.0
    @edge-runtime/ponyfill:  2.4.2
[...]

PR #2828, opened 2024-05-22
fix(api): Don't prevent subscribing with API_KEY when there is also an owner-based rule

mattcreaser:

*Issue #, if available:*
#2770 

*Description of changes:*
This fix applies to models that have multiple types of authentication, e.g:

```typescript
const schema = a.schema({
  Todo: a
    .model({
      content: a.string(),
    })
    .authorization(allow => [
        allow.publicApiKey().to(['read']),
        allow.owner()
    ]),
});
```

This schema translates into a model that has the following auth rules:

```java
@ModelConfig(pluralName = "Todos", type = Model.Type.USER, version = 1, authRules = {
  @AuthRule(allow = AuthStrategy.PUBLIC, provider = "apiKey", operations = { ModelOperation.READ }),
  @AuthRule(allow = AuthStrategy.OWNER, ownerField = "owner", identityClaim = "cognito:username", provider = "userPools", operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ })
}, hasLazySupport = true)
public final class Todo implements Model
```

We would previously disallow subscribing to this model using `AuthorizationType.API_KEY`, because of  the presence of the `AuthStrategy.OWNER` rule. You should be able to subscribe using either type. If using `AuthorizationType.API_KEY` then you receive events for all models, while if using `AuthorizationType.COGNITO_USER_POOLS` you receive events only for your own models.

I also fixed a bug I noticed that could (rarely) prevent proper shutting down of the subscription when using MultiAuth subscriptions.

- Verified that user can subscribe to a model with both public and owner read using API key, whether they are signed in or not, and receive all events for that model. The same user can also subscribe using cognito user pools (only when signed in) to receive only the events for models they create.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


aws-api/src/main/java/com/amplifyframework/api/aws/auth/AuthRuleRequestDecorator.java
```diff
@@ -135,6 +139,17 @@ && userNotInReadRestrictingGroups(readAuthorizedGroupsMap, authType)) {
...
+    private boolean doesRuleAllowNonOwnerSubscribe(AuthRule authRule, AuthorizationType authMode) {
+        AuthorizationType typeForRule = AuthorizationType.from(authRule.getAuthProvider());
+        AuthStrategy strategy = authRule.getAuthStrategy();
+        List<ModelOperation> operations = authRule.getOperationsOrDefault();
+        return strategy != AuthStrategy.OWNER && strategy != AuthStrategy.GROUPS
+            && typeForRule != AuthorizationType.AMAZON_COGNITO_USER_POOLS
+            && typeForRule != AuthorizationType.OPENID_CONNECT
+            && typeForRule == authMode
+            && (operations.contains(ModelOperation.LISTEN) || operations.contains(ModelOperation.READ));
+    }
```
lawmicha, line 151: Can this be simplified, or will equality checks make it easier to understand in the future?

```
strategy == AuthStrategy.PUBLIC
            && typeForRule == AuthorizationType.API_KEY
            && authMode == AuthorizationType.API_KEY
            && (operations.contains(ModelOperation.LISTEN) || operations.contains(ModelOperation.READ);
```

aws-api/src/main/java/com/amplifyframework/api/aws/auth/AuthRuleRequestDecorator.java
```diff
@@ -83,13 +83,16 @@ public <R> GraphQLRequest<R> decorate(
...
+        boolean subscribeAllowedForNonOwner = false;
 
         // Note that we are intentionally supporting only one owner rule with a READ operation at this time.
         // If there is more than one, the operation will fail because AppSync generates a parameter for each
         // one. The question then is which one do we pass. JavaScript currently doesn't support this use case
         // and it's not clear what a good solution would be until AppSync supports real time filters.
         for (AuthRule authRule : appSyncRequest.getModelSchema().getAuthRules()) {
-            if (isReadRestrictingOwner(authRule)) {
+            if (doesRuleAllowNonOwnerSubscribe(authRule, authType)) {
+                subscribeAllowedForNonOwner = true;
```
lawmicha, line 94: we might be able to `return request` here if the remaining auth rules does not matter, since we're doing `!subscribeAllowedForNonOwner` below

mattcreaser: We could return but doing a return buried in a loop is not encouraged for clarity purposes. What I can do here is `break` though as there is no reason to continue with the loop.

aws-api/src/main/java/com/amplifyframework/api/aws/auth/AuthRuleRequestDecorator.java
mattcreaser: 🤔 this rewrite only covers a single case - public subscribe with an API key. Although that's the issue in the bug and likely the most common scenario, it's likely not the only one. I'll think about it a little more.

mattcreaser: I've updated to only address the case from the linked issue, when there is a public read rule and call is made using API_KEY. We can address other cases if other issues are opened in the future.

vincetran (approved):

lawmicha (approved):

PR #2841, opened 2024-06-04
feat(predictions): Add support for a no light freshness challenge

vincetran:

*Issue #, if available:* N/A

*Description of changes:*
This adds support for a no-light challenge to the Prediction plugin's Face LIveness feature and adds telemetry tracking as requested by the Rekognition team. There's no API change to the core plugin API and **additive** changes to the `FaceLivenessSessionInformation` data class that's used as part of the API.

Notable callouts:
- `FaceLivenessSessionInformation`
   - Changed the `String` `challenge` parameter to be nullable. This is required to allow backward compatibility with the existing Liveness UI component library.
   - Added a new parameter `challengeVersions` to replace the previous `challenge` which takes in a `List<Challenge>`. These `Challenge`s represent both the challenge name and the version for that challenge. 
   - Added a new parameter `preCheckViewEnabled` which signifies if the default "Start View" was enabled. Used for telemetry.
   - Added a new parameter `attemptCount` which signifies how many times the user attempted to perform the liveness check. Used for telemetry.
- Added a new supported challenge called `FaceMovementChallenge` that will live alongside the existing supported `FaceMovementAndLightChallenge` 
- Updated the websocket endpoint generation to use a `Uri.Builder()` for code cleanliness
- Added support for a new web socket event called `ChallengeType` which is used for determining what type of challenge is requested from the backend to render to the user. 

- Updated and added unit tests for the new ChallengeType event
- Tested with a corresponding Amplify UI change that accepts the new challenge type and renders it
- Tested with the existing version of Amplify UI to ensure backwards compatibility

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
@@ -97,7 +97,7 @@ internal class LivenessWebSocket(
     @VisibleForTesting
     internal var webSocket: WebSocket? = null
     internal val challengeId = UUID.randomUUID().toString()
-    lateinit var challengeType: FaceLivenessChallengeType
+    var challengeType: FaceLivenessChallengeType? = null
     private var initialDetectedFace: BoundingBox? = null
     private var faceDetectedStart = 0L
     private var videoStartTimestamp = 0L
@@ -139,16 +139,6 @@ internal class LivenessWebSocket(
             }
         }
 
-        init {
-            // If the client session is requesting a 1.0.0 FaceMovementAndLight challenge, the backend won't return a
-            // ChallengeEvent so we have to set the challengeType manually
-            clientSessionInformation.challengeVersions.forEach {
-                if (it.compareType(Challenge.FaceMovementAndLightChallenge("1.0.0"))) {
-                    challengeType = FaceLivenessChallengeType.FaceMovementAndLightChallenge
-                }
-            }
-        }
-
         override fun onMessage(webSocket: WebSocket, text: String) {
             LOG.debug("WebSocket onMessage text")
             super.onMessage(webSocket, text)
@@ -162,9 +152,20 @@ internal class LivenessWebSocket(
                         if (response.challengeEvent != null) {
                             challengeType = response.challengeEvent.challengeType
                         } else if (response.serverSessionInformationEvent != null) {
+
+                            val clientRequestedOldLightChallenge = clientSessionInformation.challengeVersions
+                                .any { it == Challenge.FaceMovementAndLightChallenge("1.0.0") }
+
+                            if (challengeType == null && clientRequestedOldLightChallenge) {
+                                // For the  1.0.0 version of FaceMovementAndLight challenge, backend doesn't send a
+                                // ChallengeEvent so we need to manually check and set it if that specific challenge
+                                // was requested.
+                                challengeType = FaceLivenessChallengeType.FaceMovementAndLightChallenge
+                            }
+
                             // If challengeType hasn't been initialized by this point it's because server sent an
                             // unsupported challenge type so return an error to the client.
-                            if (!this@LivenessWebSocket::challengeType.isInitialized) {
+                            if (challengeType == null) {
                                 onErrorReceived.accept(
                                     PredictionsException(
                                         "Received an unsupported ChallengeType from the backend",
@@ -392,44 +393,63 @@ internal class LivenessWebSocket(
             width = initialFaceRect.width() / clientSessionInformation.videoWidth
         )
         faceDetectedStart = adjustedDate(videoStartTime)
-        val clientInfoEvent =
-            ClientSessionInformationEvent(
+
+        if (challengeType == null) {
+            onErrorReceived.accept(
+                PredictionsException(
+                    "Failed to send an initial face detected event",
+                    AmplifyException.TODO_RECOVERY_SUGGESTION
+                )
+            )
+        } else {
+            val clientInfoEvent =
+                ClientSessionInformationEvent(
+                    challenge = buildClientChallenge(
+                        challengeType = challengeType!!,
+                        challengeId = challengeId,
+                        initialFace = InitialFace(
+                            boundingBox = initialDetectedFace!!,
+                            initialFaceDetectedTimestamp = faceDetectedStart
+                        ),
+                        videoStartTimestamp = videoStartTimestamp
+                    )
+                )
+            sendClientInfoEvent(clientInfoEvent)
+        }
+    }
+
+    fun sendFinalEvent(targetFaceRect: RectF, faceMatchedStart: Long, faceMatchedEnd: Long) {
+        if (challengeType == null) {
+            onErrorReceived.accept(
+                PredictionsException(
+                    "Failed to send an initial face detected event",
+                    AmplifyException.TODO_RECOVERY_SUGGESTION
+                )
+            )
+        } else {
+            val finalClientInfoEvent = ClientSessionInformationEvent(
                 challenge = buildClientChallenge(
-                    challengeType = challengeType,
+                    challengeType = challengeType!!,
                     challengeId = challengeId,
+                    videoEndTimestamp = videoEndTimestamp,
                     initialFace = InitialFace(
                         boundingBox = initialDetectedFace!!,
                         initialFaceDetectedTimestamp = faceDetectedStart
                     ),
-                    videoStartTimestamp = videoStartTimestamp
-                )
-            )
-        sendClientInfoEvent(clientInfoEvent)
-    }
-
-    fun sendFinalEvent(targetFaceRect: RectF, faceMatchedStart: Long, faceMatchedEnd: Long) {
-        val finalClientInfoEvent = ClientSessionInformationEvent(
-            challenge = buildClientChallenge(
-                challengeType = challengeType,
-                challengeId = challengeId,
-                videoEndTimestamp = videoEndTimestamp,
-                initialFace = InitialFace(
-                    boundingBox = initialDetectedFace!!,
-                    initialFaceDetectedTimestamp = faceDetectedStart
-                ),
-                targetFace = TargetFace(
-                    faceDetectedInTargetPositionStartTimestamp = adjustedDate(faceMatchedStart),
-                    faceDetectedInTargetPositionEndTimestamp = adjustedDate(faceMatchedEnd),
-                    boundingBox = BoundingBox(
-                        left = targetFaceRect.left / clientSessionInformation.videoWidth,
-                        top = targetFaceRect.top / clientSessionInformation.videoHeight,
-                        height = targetFaceRect.height() / clientSessionInformation.videoHeight,
-                        width = targetFaceRect.width() / clientSessionInformation.videoWidth
+                    targetFace = TargetFace(
+                        faceDetectedInTargetPositionStartTimestamp = adjustedDate(faceMatchedStart),
+                        faceDetectedInTargetPositionEndTimestamp = adjustedDate(faceMatchedEnd),
+                        boundingBox = BoundingBox(
+                            left = targetFaceRect.left / clientSessionInformation.videoWidth,
+                            top = targetFaceRect.top / clientSessionInformation.videoHeight,
+                            height = targetFaceRect.height() / clientSessionInformation.videoHeight,
+                            width = targetFaceRect.width() / clientSessionInformation.videoWidth
+                        )
                     )
                 )
             )
-        )
-        sendClientInfoEvent(finalClientInfoEvent)
+            sendClientInfoEvent(finalClientInfoEvent)
+        }
     }
 
     fun sendColorDisplayedEvent(
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
@@ -169,13 +169,12 @@ internal class LivenessWebSocket(
                             }
 
                             // If challengeType hasn't been initialized by this point it's because server sent an
-                            // unsupported challenge type so return an error to the client.
+                            // unsupported challenge type so return an error to the client and close the web socket.
                             val resolvedChallengeType = challengeType
                             if (resolvedChallengeType == null) {
                                 webSocketError = FaceLivenessUnsupportedChallengeTypeException()
                                 destroy(UNSUPPORTED_CHALLENGE_CLOSURE_STATUS_CODE)
                             } else {
-                                // send non-null challenge Type
                                 onSessionResponseReceived.accept(
                                     SessionResponse(
                                         response.serverSessionInformationEvent.sessionInformation,
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
@@ -30,6 +30,7 @@ import com.amplifyframework.predictions.PredictionsException
 import com.amplifyframework.predictions.aws.BuildConfig
 import com.amplifyframework.predictions.aws.exceptions.AccessDeniedException
 import com.amplifyframework.predictions.aws.exceptions.FaceLivenessSessionNotFoundException
+import com.amplifyframework.predictions.aws.exceptions.FaceLivenessUnsupportedChallengeTypeException
 import com.amplifyframework.predictions.aws.models.liveness.BoundingBox
 import com.amplifyframework.predictions.aws.models.liveness.ClientChallenge
 import com.amplifyframework.predictions.aws.models.liveness.ClientSessionInformationEvent
@@ -78,10 +79,14 @@ internal class LivenessWebSocket(
     val region: String,
     val clientSessionInformation: FaceLivenessSessionInformation,
     val livenessVersion: String?,
-    val onSessionInformationReceived: Consumer<SessionInformation>,
+    val onSessionResponseReceived: Consumer<SessionResponse>,
     val onErrorReceived: Consumer<PredictionsException>,
     val onComplete: Action
 ) {
+    internal data class SessionResponse(
+        val faceLivenessSession: SessionInformation,
+        val livenessChallengeType: FaceLivenessChallengeType
+    )
 
     private val signer = AWSV4Signer()
     private var credentials: Credentials? = null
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSessionInformation.kt
```diff
@@ -20,6 +20,23 @@ import com.amplifyframework.annotations.InternalAmplifyApi
 data class FaceLivenessSessionInformation(
     val videoWidth: Float,
     val videoHeight: Float,
-    val challengeVersions: String,
-    val region: String
+    val challenge: String? = null,
+    val region: String,
+    val challengeVersions: List<Challenge>? = null,
+    val preCheckViewEnabled: Boolean? = null,
+    val attemptCount: Int? = null
```
phantumcode, line 27: curious why `attemptCount` and `preCheckViewEnabled` are nullable here? is this for backward compatibility? does the UI component not pinned to an Amplify Android version?

tylerjroach: Its to make sure the constructor used by older versions of Amplify Liveness aren't broken if Amplify dependency is updated. While we resolve to a specific version, if a customer updates their Amplify dependency to a higher version, it will resolve to that one. 

We have a minimum Amplify version on the Liveness component.
Amplify Android must support current + all previous versions of Liveness

tylerjroach: Suggest something like this to make constructors cleaner at liveness call site.

```
@InternalAmplifyApi
class FaceLivenessSessionInformation {

    val videoWidth: Float
    val videoHeight: Float
    val challengeVersions: List<Challenge>
    val region: String
    val preCheckViewEnabled: Boolean?
    val attemptCount: Int?
    
    @Deprecated("Keeping compatibility for <= Amplify Liveness 1.2.6")
    constructor(
        videoWidth: Float,
        videoHeight: Float,
        challenge: String,
        region: String
    ) {
        this.videoWidth = videoWidth,
        this.videoHeight = videoHeight,
        this.challengeVersions = listOf(Challenge(ChallengeType.FaceMovementAndLightChallenge, challenge)),
        this.region = region
        this.preCheckViewEnabled = null
        this.attemptCount = null
    }

    constructor(
        videoWidth: Float,
        videoHeight: Float,
        region: String,
        challengeVersions: List<Challenge>,
        preCheckViewEnabled: Boolean,
        attemptCount: Int
    ) {
        this.videoWidth = videoWidth
        this.videoHeight = videoHeight
        this.region = region
        this.challengeVersions = challengeVersions
        this.preCheckViewEnabled = preCheckViewEnabled
        this.attemptCount = attemptCount
    }
}
```

If we can convert `challenge` the String to a list with a single item of `challengeVersions` it also makes downstream implementation easier.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSessionInformation.kt
```diff
@@ -20,6 +20,23 @@ import com.amplifyframework.annotations.InternalAmplifyApi
...
+@InternalAmplifyApi
+data class Challenge(
+    val type: ChallengeType,
+    val version: String
+) {
+    fun toQueryParamString(): String = "${type.name}_$version"
+}
+
+@InternalAmplifyApi
+enum class ChallengeType {
```
tylerjroach, line 39: Thoughts?

```
@InternalAmplifyApi
sealed class Challenge private constructor(val name: String, val version: String) {
    
    @InternalAmplifyApi
    class FaceMovementChallenge(version: String) : Challenge("FaceMovementChallenge", version)
    
    @InternalAmplifyApi
    class FaceMovementAndLightChallenge(version: String) : Challenge("FaceMovementAndLightChallenge", version)

    fun toQueryParamString(): String = "${name}_$version"
}
```

This may be safer if a single challenge type needs more information than another challenge type.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
```diff
@@ -19,6 +19,8 @@ import com.amplifyframework.annotations.InternalAmplifyApi
 
 @InternalAmplifyApi
 class FaceLivenessSession(
+    val challengeId: String,
+    val challengeType: FaceLivenessChallengeType,
     val challenges: List<FaceLivenessSessionChallenge>,
```
tylerjroach, line 24: Not sure there is much we can do here without breaking API to liveness lib, but its a bit confusing that a "challenge" has "challenges" itself. It requires some context thought to the overall outer challenge and the inner challenges that make up the outer one.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
```diff
@@ -29,41 +30,45 @@ import com.amplifyframework.predictions.aws.models.FaceTargetChallengeResponse
...
-        sessionInformation = sessionInformation,
+        endpoint = buildWebSocketEndpoint(),
+        region = clientSessionInformation.region,
+        clientSessionInformation = clientSessionInformation,
         livenessVersion = livenessVersion,
-        onSessionInformationReceived = { sessionInformation ->
-            val challenges = processSessionInformation(sessionInformation)
+        onSessionInformationReceived = { serverSessionInformation ->
+            val challenges = processSessionInformation(serverSessionInformation)
+            val challengeType = processChallengeType()
```
tylerjroach, line 64: Shouldn't we just provide `onError` if we don't understand the challenge type

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
```diff
@@ -78,12 +83,75 @@ internal class RunFaceLivenessSession(
...
+                // A ChallengeType event is not returned when the v1 version of FaceMovementAndLightChallenge
+                // is requested so we'll check here for backwards compat
+                if (clientSessionInformation.challengeVersions != null &&
+                    clientSessionInformation.challengeVersions!!.contains(
+                            Challenge(ChallengeType.FaceMovementAndLightChallenge, "1.0.0")
+                        )
+                ) {
+                    FaceLivenessChallengeType.FaceMovementAndLightChallenge
+                } else {
+                    FaceLivenessChallengeType.Unknown
```
tylerjroach, line 146: I don't think we need unknown. We should catch exception and provide exception in onError

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessChallengeType.kt
```diff
@@ -0,0 +1,24 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.predictions.models
+
+import com.amplifyframework.annotations.InternalAmplifyApi
+
+@InternalAmplifyApi
+enum class FaceLivenessChallengeType {
```
tylerjroach, line 20: Do we need to duplicate ChallengeType and FaceLivenessChallengeType. They are essentially the same.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -145,7 +148,9 @@ internal class LivenessWebSocket(
             try {
                 when (val response = LivenessEventStream.decode(bytes, json)) {
                     is LivenessResponseStream.Event -> {
-                        if (response.serverSessionInformationEvent != null) {
+                        if (response.challengeEvent != null) {
+                            challengeType = response.challengeEvent.challengeType
+                        } else if (response.serverSessionInformationEvent != null) {
```
tylerjroach, line 153: we should just set challengeType here since we know what it is.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -523,6 +526,57 @@ internal class LivenessWebSocket(
...
+                faceMovementChallenge = FaceMovementClientChallenge(
+                    challengeId = challengeId,
+                    videoStartTimestamp = videoStartTimestamp,
+                    videoEndTimestamp = videoEndTimestamp,
+                    initialFace = initialFace,
+                    targetFace = targetFace
+                )
+            )
+        }
+        null -> {
```
tylerjroach, line 563: remove this case

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/liveness/ChallengeEvent.kt
```diff
@@ -0,0 +1,25 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.predictions.aws.models.liveness
+
+import com.amplifyframework.predictions.models.ChallengeType
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class ChallengeEvent(
```
tylerjroach, line 22: Have you tested what happens when Liveness sends an event we can't parse into a ChallengeType we understand. Also still confused when ChallengeType and FaceLivenessChallengeType is used.

tylerjroach (commented):
Can you add some tests on the new UrlBuilder being used on the websocket?

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSessionInformation.kt
vincetran: Ah this is a really good way to do this! Will update 

> If we can convert challenge the String to a list with a single item of challengeVersions it also makes downstream implementation easier.

I can make this change but wanted to better understand how this will make the downstream implementation easier?

tylerjroach: @vincetran Doing so would remove `challenge` entirely, so you no longer have to check if its null and attempt to use if present. If we do the conversion up front to `challengeVersions` then we have 1 field downstream to use.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
vincetran: I agree with this. I modified the flow and have it so that we'll call `onError` in the web socket class.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
vincetran: With the changes I'm making, `challengeType` should never be null by the time we get to this point. An exception will be provided earlier in the flow.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessChallengeType.kt
vincetran: For some reason I had this assumption that we needed to keep server/client models separate but yes they are exactly the same. I can keep this core FaceLivenessChallengeType and reference this from the server model and remove `ChallengeType`

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSession.kt
vincetran: For sure, I had a bit of trouble ramping up to the code trying to figure out the difference. When we're planning to make breaking changes, I'd want to rename the inner "challenges" to like "challengeSteps" or something that doesn't include the word "challenge" to explicitly differentiate them.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/models/liveness/ChallengeEvent.kt
vincetran: Before, a new event would fail to get decoded and fail silently and the client app would get confused as to what to do (oops!). Added an exception that gets thrown in the websocket class when a challengetype fails to get parsed.

And from a previous comment reply, I've removed ChallengeType in favor of FaceLivenessChallengeType

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -96,7 +97,7 @@ internal class LivenessWebSocket(
     @VisibleForTesting
     internal var webSocket: WebSocket? = null
     internal val challengeId = UUID.randomUUID().toString()
-    var challengeType: ChallengeType? = null
+    lateinit var challengeType: FaceLivenessChallengeType
```
tylerjroach, line 100: I'd rather leave this nullable rather than relying on a server response to guarantee lateinit behavior.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -151,9 +162,21 @@ internal class LivenessWebSocket(
                         if (response.challengeEvent != null) {
                             challengeType = response.challengeEvent.challengeType
                         } else if (response.serverSessionInformationEvent != null) {
-                            onSessionInformationReceived.accept(
-                                response.serverSessionInformationEvent.sessionInformation
-                            )
+                            // If challengeType hasn't been initialized by this point it's because server sent an
+                            // unsupported challenge type so return an error to the client.
+                            if (!this@LivenessWebSocket::challengeType.isInitialized) {
```
tylerjroach, line 167: Lets just use nullable here. No benefit to using lateinit as it makes it harder to understand it may not have been given a value.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSessionInformation.kt
```diff
@@ -17,26 +17,58 @@ package com.amplifyframework.predictions.models
...
-    FaceMovementAndLightChallenge,
+sealed class Challenge private constructor(val name: String, val version: String) {
+
+    @InternalAmplifyApi
+    class FaceMovementChallenge(version: String) : Challenge("FaceMovementChallenge", version)
+
+    @InternalAmplifyApi
+    class FaceMovementAndLightChallenge(version: String) : Challenge("FaceMovementAndLightChallenge", version)
+
+    fun compareType(challenge: Challenge): Boolean {
```
tylerjroach, line 69: Can't you just make these data classes and then you can just do an `==` check

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -135,6 +139,16 @@ internal class LivenessWebSocket(
             }
         }
 
+        init {
+            // If the client session is requesting a 1.0.0 FaceMovementAndLight challenge, the backend won't return a
+            // ChallengeEvent so we have to set the challengeType manually
+            clientSessionInformation.challengeVersions.forEach {
+                if (it.compareType(Challenge.FaceMovementAndLightChallenge("1.0.0"))) {
+                    challengeType = FaceLivenessChallengeType.FaceMovementAndLightChallenge
```
tylerjroach, line 147: What if the client asks for 1.00 FaceMovementAndLight and then a NoLightChallenge in the array? Is that valid? If so, this is doing a loop and setting it if it sees a single item asking for it.

core/src/main/java/com/amplifyframework/predictions/models/FaceLivenessSessionInformation.kt
vincetran: You know I never thought to use data classes in sealed classes lol yes I've updated this

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -162,9 +152,20 @@ internal class LivenessWebSocket(
...
+                                // ChallengeEvent so we need to manually check and set it if that specific challenge
+                                // was requested.
+                                challengeType = FaceLivenessChallengeType.FaceMovementAndLightChallenge
+                            }
+
                             // If challengeType hasn't been initialized by this point it's because server sent an
                             // unsupported challenge type so return an error to the client.
-                            if (!this@LivenessWebSocket::challengeType.isInitialized) {
+                            if (challengeType == null) {
                                 onErrorReceived.accept(
```
tylerjroach, line 169: Sorry I'm just now noticing this but don't we need to kill the websocket at this point?

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/RunFaceLivenessSession.kt
```diff
@@ -128,7 +128,7 @@ internal class RunFaceLivenessSession(
 
     private fun getChallengeId(): String = livenessWebSocket.challengeId
 
-    private fun getChallengeType(): FaceLivenessChallengeType = livenessWebSocket.challengeType
+    private fun getChallengeType(): FaceLivenessChallengeType = livenessWebSocket.challengeType!!
```
tylerjroach, line 131: Lets remove this method and instead pass challenge type through to:

```
onSessionInformationReceived.accept(
    response.serverSessionInformationEvent.sessionInformation
)
```
We can change type of the consumer to contain a class wrapping SessionInformation and ChallengeType `val onSessionInformationReceived: Consumer<SessionInformation>`

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -392,44 +393,63 @@ internal class LivenessWebSocket(
...
+                            initialFaceDetectedTimestamp = faceDetectedStart
+                        ),
+                        videoStartTimestamp = videoStartTimestamp
+                    )
+                )
+            sendClientInfoEvent(clientInfoEvent)
+        }
+    }
+
+    fun sendFinalEvent(targetFaceRect: RectF, faceMatchedStart: Long, faceMatchedEnd: Long) {
```
tylerjroach, line 421: create a local challengeType variable here to remove !! later on in this method.

tylerjroach (approved):

phantumcode (approved):

Issue #2404, opened 2023-04-19
Storage.getUrl("ExampleKey") should be able to throw a meaningful exception if the file does not exist in the storage


kvramyasri7:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Not applicable

### Amplify Categories

Storage

### Gradle script dependencies

<details>

```groovy
// Put output below this line


```

</details>

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

When we call getUrl from storage category we are receiving a presigned url even though the file does not exist.The expected behavior is to return a meaningful exception if the file does not exist in the storage.

### Reproduction steps (if applicable)

amplify add storage
Call Amplify.Storage.getUrl("NonexistedFile");. call getUrl with a non-existed File.

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

kvramyasri7:
Hello All, Wondering when this is going to be resolved. 😉. Just an ETA will be fine or you can circle back to me when this is fixed.

phantumcode:
Fix released in version[ 2.20.0](https://github.com/aws-amplify/amplify-android/releases/tag/release_v2.20.0)

PR #2848, opened 2024-06-12
feat(storage): add object existence validation option to get presigned url

phantumcode:

*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/2404

*Description of changes:*
Update Storage getURL options to have a boolean flag to check for object existence.  This provides feature parity with Amplify Swift

Doc PR: https://github.com/aws-amplify/docs/pull/7736

* Manually tested getURL with new option enabled/disabled with and without object
* Added unit tests

- [ ] No
- [x] Yes (Please include a PR link for the documentation update)

*General Checklist*
- [x] Added Unit Tests
- [x] Added Integration Tests


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/request/AWSS3StorageGetPresignedUrlRequest.java
```diff
@@ -26,13 +27,15 @@
  * @deprecated  this class is only constructed internally through deprecated transfer methods.
  */
 @Deprecated
+@InternalAmplifyApi
```
tylerjroach, line 30: We can't add this as it would cause a breaking change in forcing customer to add annotation to allow usage.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -309,6 +309,7 @@ public StorageGetUrlOperation<?> getUrl(
         return getUrl(path, StorageGetUrlOptions.defaultInstance(), onSuccess, onError);
     }
 
+    @OptIn(markerClass = InternalAmplifyApi.class)
```
tylerjroach, line 312: This opt in can be removed once removing annotation from the Request class as mentioned.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StorageGetPresignedUrlOperation.java
```diff
@@ -85,10 +88,20 @@ public void start() {
                         prefix -> {
                             try {
                                 String serviceKey = prefix.concat(getRequest().getKey());
+
+                                if (getRequest().validateObjectExistence()) {
+                                    try {
+                                        storageService.validateObjectExists(serviceKey);
```
tylerjroach, line 94: See comments on StoragePath operation class

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
```diff
@@ -85,6 +88,30 @@ internal class AWSS3StorageService(
         return URL(presignUrlRequest.url.toString())
     }
 
+    /**
+     * Validate if S3 object exists for the given key.
+     * Throws StorageException if NoSuchKey S3 client exception is caught.
+     * @param serviceKey S3 service key
+     */
+    override fun validateObjectExists(serviceKey: String) {
```
tylerjroach, line 96: lets not try catch here. Instead, lets catch when making the service call. This pattern is used throughout the rest of Amplifiy.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StoragePathGetPresignedUrlOperation.kt
```diff
@@ -48,6 +48,24 @@ internal class AWSS3StoragePathGetPresignedUrlOperation(
                 return@submit
             }
 
+            if (request.validateObjectExistence) {
+                try {
+                    storageService.validateObjectExists(serviceKey)
+                } catch (se: StorageException) {
```
tylerjroach, line 54: I like the way the StoragePath operation is written more than the key one.

Instead of checking for StorageException, as suggested in another comment, lets remove the try/catch on the service call and check for "Not Found" exception here. I think we can also improve the backup exception case which is likely no network. The message can be "Encountered an issue while validating the existence of object" vs "Encountered an issue while generating pre-signed URL"

tylerjroach (commented):
Nice work, a few suggested changes.

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/operation/AWSS3StorageGetPresignedUrlOperationTest.kt
```diff
@@ -138,4 +139,81 @@ public class AWSS3StorageGetPresignedUrlOperationTest {
         awsS3StorageGetPresignedUrlOperation.start()
         Mockito.verify(storageService).getPresignedUrl(expectedKey, 1, false)
     }
+
+    @Test
+    fun `getPresignedUrl fails with non existent S3 path when validateObjectExistence is enabled`() {
```
lawmicha, line 144: this would be a good E2E test to have instead of using mock storage service, we would just need to create a random path so it's non existent and pass in the option, right?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
```diff
@@ -85,6 +88,30 @@ internal class AWSS3StorageService(
...
+        try {
+            runBlocking {
+                s3Client.headObject(
+                    HeadObjectRequest {
+                        bucket = s3BucketName
+                        key = serviceKey
+                    }
+                )
+            }
+        } catch (ex: NotFound) {
```
lawmicha, line 106: if we're catching a specific exception, should we also have a catch base exception below this? If S3 ever increases its exception types, then the second catch on the base type should catch it

lawmicha, line 106: I'm guessing this is the same implementaton as Swift? to check head object and catch NotFound. Odd the documentation shows "NoSuchKey" error and nothing about "NotFound" https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors

tylerjroach: See my comments that we should not catch here but instead in the operation itself. We can specifically catch on this message to provide better error messaging and let the others fall through to more standard messaging.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/request/AWSS3StorageGetPresignedUrlRequest.java
phantumcode: annotation removed.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
phantumcode: annotation removed

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: try/catch removed in favor of catching at the operation level

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StoragePathGetPresignedUrlOperation.kt
phantumcode: good suggestions, code updated

aws-storage-s3/src/test/java/com/amplifyframework/storage/s3/operation/AWSS3StorageGetPresignedUrlOperationTest.kt
phantumcode: yes, E2E were added for bot positive/negative cases, see `AWSS3StoragePathGetUrlTest.kt`

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: exception handling moved to operation per @tylerjroach suggestions.

phantumcode: yes, this exception type is the same on [Amplify Swift](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Service/Storage/AWSS3StorageService%2BGetPreSignedURLBehavior.swift#L35C26-L35C34) and verified with E2E testing.

tylerjroach (approved):

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StoragePathGetPresignedUrlOperation.kt
```diff
@@ -48,6 +49,30 @@ internal class AWSS3StoragePathGetPresignedUrlOperation(
                 return@submit
             }
 
+            if (request.validateObjectExistence) {
+                try {
+                    storageService.validateObjectExists(serviceKey)
+                } catch (nfe: NotFound) {
```
lawmicha, line 55: This seems odd, `storageService` is of type **StorageService**, yet we're catching a specific Kotlin SDK exception. If we need to swap out the implementation or mock the returned error, we'll be dependent on the Kotlin SDK.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
lawmicha: The reason we're using try-catch here is because the implementation requires it: we're utilizing S3's `headObject` API and catching `NotFound`, which is a Kotlin SDK type. This leverages the SDK to validate if the object exists.

If we don't catch the exception, we'll expose the abstraction of **StorageService**,  (see the import of `aws.sdk.kotlin.services.s3.model.NotFound` in AWSS3StoragePathGetPresignedUrlOperation.kt)

One alternative I can think of is try-catching here and changing `validateObjectExists` to return a bool indicating whether the object exists or not for the caller to handle

tylerjroach: Not a bad idea, but we would still need try/catch for the other failure scenarios. I see what you are saying in that NotFound is really as success case of validating that the object didn't exist, and not a true failure.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StoragePathGetUrlTest.kt
```diff
@@ -79,4 +82,25 @@ class AWSS3StoragePathGetUrlTest {
...
+    }
+
+    @Test
+    fun testGetUrlWithStorageExceptionObjectNotFoundThrown() {
+        assertThrows(StorageException::class.java) {
+            synchronousStorage.getUrl(
+                StoragePath.fromString("public/SOME_UNKNOWN_FILE"),
+                AWSS3StorageGetPresignedUrlOptions.builder().setValidateObjectExistence(true).expires(30).build()
+            )
+        }
```
lawmicha, line 104: can we assert to make sure this exception is the excepted notFound exception? In current commit, that should be checking inner exception i believe

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: I've refactored the exception handling to isolate `aws.sdk.kotlin.services.s3.model.NotFound` to `StorageService`.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/operation/AWSS3StoragePathGetPresignedUrlOperation.kt
phantumcode: I've refactored the exception handling to isolate `aws.sdk.kotlin.services.s3.model.NotFound` to `StorageService`.

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StoragePathGetUrlTest.kt
phantumcode: integration test updated to specifically check for `NotFound` exception

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
lawmicha: Awesome, thanks for taking the time to address my consideration

aws-storage-s3/src/androidTest/java/com/amplifyframework/storage/s3/AWSS3StoragePathGetUrlTest.kt
```diff
@@ -79,4 +83,27 @@ class AWSS3StoragePathGetUrlTest {
...
+    fun testGetUrlWithStorageExceptionObjectNotFoundThrown() {
+        val exception = assertThrows(StorageException::class.java) {
+            synchronousStorage.getUrl(
+                StoragePath.fromString("public/SOME_UNKNOWN_FILE"),
+                AWSS3StorageGetPresignedUrlOptions.builder().setValidateObjectExistence(true).expires(30).build()
+            )
+        }
+
+        assertTrue(exception.cause is NotFound)
+    }
```
lawmicha, line 108: Just thought of the case where `ValidateObjectExistence` is set to false (or not set at all) and the path is SOME_UNKNOWN_FILE, the expectation is that it doesn't throw an exception but when trying to resolve the retrieved URL, the response is not found. I don't think this case is written here yet

lawmicha (approved):

tylerjroach (approved):

lawmicha (approved):

Issue #2151, opened 2022-12-02
DataStoreException prevents syncing when saving same object quickly


PS-MS:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

GraphQL API, DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line


```

</details>

### Environment information

<details>

```
# Put output below this line


```

</details>

### Describe the bug

Our website was failing to receive updates after saving some data, after some investigation it turned out to be developer error saving the same object multiple times. e.g.

Click save -> save data -> save event callback -> save data

`DataStoreException{message=Outbox was asked to remove a mutation with ID = 18f158a3-7230-11ed-9b69-65a8d87b846b. However, there was no mutation with that ID in the outbox, to begin with`.

Although it was a mistake to save twice, I'd still expect DataStore to manage the mutation id's correctly and not fall over

### Reproduction steps (if applicable)

Note: the bug cause was with saving parent and child data, saving around 30 entries twice however I've been able to reproduce it with the below. If this doesn't cause the exception, increase the repeat count as it appears to be about timing.

1. Set up a model
2. Set up loop that saves in quick succession:

```
repeat(5) {
    Amplify.DataStore.save(modelObj)
}
```

3. Trigger code and watch debug

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

mikepschneider:
I was able to reproduce this in a flutter app while getting a slightly different error.  Once the error happens, the device db is in an inconsistent state and syncing up to the cloud is broken until the app data is reset.  

The error I got looks like this:
```
W/amplify:aws-datastore(29489): DataStoreException{message=Wanted 1 metadata for item with id = 54452c51-6522-4d2d-8786-0861dda220c5, but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}
W/amplify:aws-datastore(29489): 	at com.amplifyframework.datastore.syncengine.VersionRepository.extractVersion(VersionRepository.java:91)
[...]

PR #2849, opened 2024-06-12
fix(DataStore): should not crash on missing version metadata

lawmicha:

*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/2151#issuecomment-2135346064

*Description of changes:*

**The app should not crash when DataStore cannot retrieve the version for update and delete mutations. DataStore should attempt to send the mutation without a version and handle the response.**

This PR is looking to address the following crash that customers are experiencing

```
DataStoreException{message=Wanted 1 metadata for item with id = e7eb4237-c0d4-4ce6-8800-3b9d5fb603b8, but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}
     at com.amplifyframework.datastore.syncengine.VersionRepository.extractVersion(VersionRepository.java:91)
    at com.amplifyframework.datastore.syncengine.VersionRepository.lambda$findModelVersion$0$com-amplifyframework-datastore-syncengine-VersionRepository(VersionRepository.java:64)
     at com.amplifyframework.datastore.syncengine.VersionRepository$$ExternalSyntheticLambda0.accept(Unknown Source:8)
     at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter.lambda$query$4$com-amplifyframework-datastore-storage-sqlite-SQLiteStorageAdapter(SQLiteStorageAdapter.java:407)
     at com.amplifyframework.datastore.storage.sqlite.SQLiteStorageAdapter$$ExternalSyntheticLambda6.run(Unknown Source:10)
     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:463)
     at java.util.concurrent.FutureTask.run(FutureTask.java:264)
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
     at java.lang.Thread.run(Thread.java:1012)
```

The message 
```
Wanted 1 metadata for item with id = e7eb4237-c0d4-4ce6-8800-3b9d5fb603b8, but had 0.
```
comes from https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt#L148

Code paths
```
MutationProcessor.publishToNetwork → 
    MutationProcessor.update → 
        VersionRepository.findModelVersion → 
            VersionRepository.extractVersion

MutationProcessor.publishToNetwork → 
   MutationProcessor.delete → 
        VersionRepository.findModelVersion → 
            VersionRepository.extractVersion
```

The **unexpected behavior** observed by the customer is that DataStore is deciding to crash when publishing mutations requires a version for updates and deletes, but it could not find it.

According to this customer https://github.com/aws-amplify/amplify-android/issues/2151#issuecomment-2135346064 This happens in bad network scenarios during saving and updating a model. 

**The cause of this issue**:

1. `DataStore.save(model)` is called for a new model. This inserts into the local database and queues up a pending create mutation.
2. `DataStore.save(model)` is called for the existing model, performing some updates to the fields. This updates the existing record in the local database, and queues up a pending update mutation. We assume that the MutationProcessor performs (1) process the create mutation, such that it is marked in flight, and the update mutation is not merged into the create mutation when it is inserted into the mutation database, and has to be sent on its own by the MutationProcessor.

MutationProcessor (running in parallel):

1. Process create mutation by publishing it to the network. The network is down. The request times out and fails. MutationProcessor can handle this in two ways: retry indefinitely or drop the mutation event.
    1. Retrying indefinitely will only occur when it accurately classifies the response as a transient network failure.
    2. If the response is classified as a non-retryable error, the error is returned to the customer on the `errorHandler` and the mutation will be dropped. Subsequently there will not be any reconciliation, and no local version will be persisted.
2. Process update mutation. Query the local metadata table for the model’s version, cannot find it, throw exception which crashes the app.

The cause of this issue is the create mutation was dropped, which could happen for a variety of reasons decided by the MutationProcessor’s logic in handling the response of the request, and the subsequent update mutation did not have a version to send with the mutation.

**MutationProcessor error classification**

https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L353-L355

```
nonRetryableExceptions.add(DataStoreException.GraphQLResponseException.class);
nonRetryableExceptions.add(ApiException.NonRetryableException.class);
```

**`DataStoreException.GraphQLResponseException`**

https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L387

This exception is thrown when the response is successful with a GraphQL response payload containing errors. If any of the GraphQLError’s do not contain the AppSync “conflictUnhandled” type then throw this.

This means the conditions for classification of non-retryable is as follows:

1. A successful GraphQL response containing errors
2. The error is not ConflictUnhandled, such as 
    * ConditionalCheckFailedException
    * OperationDisabled
    * Unauthorized

Since we are discussing the scenario with bad network conditions, it’s unlikely that it was classified as non-retyable due to this condition since (1) A successful GraphQL response would not have occurred.

**`ApiException.NonRetryableException`**
This exception is thrown when the response is 400 to 499. 
```
private static final int START_OF_CLIENT_ERROR_CODE = 400;
private static final int END_OF_CLIENT_ERROR_CODE = 499;

/// ...
if (response.code() >= START_OF_CLIENT_ERROR_CODE && response.code() <= END_OF_CLIENT_ERROR_CODE) {
    onFailure.accept(new ApiException
            .NonRetryableException("OkHttp client request failed.", "Irrecoverable error")
    );
    return;
}

```

This includes 408 Request Timeout. To be verified by manual testing, but this could be the response being handled as non-retryable even though it was a transient timeout error (bad network conditions).

In summary, there is an existing scenario to be verified: During bad network conditions, DataStore’s MutationProcessor should handle failures as retryable. 

Consider the scenario in which Unauthorized is treated as non-retryable (as seen above in DataStoreException.GraphQLResponseException). Unauthorized is very common when the user is signed out, or their token or session is expired. The mutation is then dropped, and update mutation will attempt to get the version and crash the app. 

This PR looks to address this scenario while another PR independent from this will look into verifying MutationProcessor's classification of transient network errors as retryable. The app should not crash when DataStore cannot retrieve the version for update and delete mutations. DataStore should attempt to send the mutation without a version and handle the response. The response will be handled by DataStore by sending it back to the customer through the `errorHandler` and then dropping it, moving on to the next mutation in the queue.

**What kind of response are we expecting?**

The conflict resolution enabled backend does not enforce version to be a required field. If we cannot get a version locally, we can send the update/delete mutation without a version, AppSync will respond successfully with a GraphQL response. This is inline with Swift DataStore's implementation, as it will also send update/delete mutations without a version if it cannot retrieve it at the time it's processing the mutation event.

Example 1. Update Mutation on existing model
```
mutation UpdateWithNoVersionHelloUpdated {
  updateBlog(input: {id: "f92dfb85-c580-4c3e-9eb4-93e3b76c33b5", name: "hello updated"}) {
    _deleted
    _lastChangedAt
    _version
    createdAt
    id
    name
    owner
    updatedAt
  }
}

// response
{
  "data": {
    "updateBlog": {
      "_deleted": null,
      "_lastChangedAt": 1718203981917,
      "_version": 6,
      "createdAt": "2024-06-12T14:49:26.329Z",
      "id": "f92dfb85-c580-4c3e-9eb4-93e3b76c33b5",
      "name": "hello",
      "owner": "michael",
      "updatedAt": "2024-06-12T14:49:26.329Z"
    }
  }
}
```

In the above example, version is updated, and the name is not. The conflict resolution strategy is enabled is auto-merge. DataStore will then reconcile this response data into the local database as the latest version of this model.

Example 2: Update mutation on model not yet created in AppSync
```
mutation UpdateInvalidIdWithNoVersion {
  updateBlog(input: {id: "notYetCreated", name: "hello updated"}) {
    _deleted
    _lastChangedAt
    _version
    createdAt
    id
    name
    owner
    updatedAt
  }
}
// response
{
  "data": {
    "updateBlog": null
  },
  "errors": [
    {
      "path": [
        "updateBlog"
      ],
      "data": null,
      "errorType": "Unauthorized",
      "errorInfo": null,
      "locations": [
        {
          "line": 27,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "Not Authorized to access updateBlog on type Mutation"
    }
  ]
}
```
The example above closely resembles the scenario where we continue to send the update mutation without a version. When the create mutation fails, the blog id does not exist in the AppSync. When the id is used for update mutations w/ Cognito User Pool auth, and without version, the response will be unauthorized.

According to the MutationProcessor logic, this will be categorized as a non retryable error, and the error will be returned to the customer through the `errorHandler`, before the update mutation is dropped. Based on their use cases, they can decide to reconcile this themselves or not, however, new devices or cleared devices will not sync this data back down to the app since it does not exist in the remote store.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
@@ -112,7 +112,7 @@ public interface AppSync {
     <T extends Model> Cancelable update(
             @NonNull T model,
             @NonNull ModelSchema modelSchema,
-            @NonNull Integer version,
+            @Nullable Integer version,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure
     );
@@ -132,7 +132,7 @@ public interface AppSync {
     <T extends Model> Cancelable update(
             @NonNull T model,
             @NonNull ModelSchema modelSchema,
-            @NonNull Integer version,
+            @Nullable Integer version,
             @NonNull QueryPredicate predicate,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure
@@ -152,7 +152,7 @@ public interface AppSync {
     <T extends Model> Cancelable delete(
             @NonNull T model,
             @NonNull ModelSchema modelSchema,
-            @NonNull Integer version,
+            @Nullable Integer version,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure
     );
@@ -172,7 +172,7 @@ public interface AppSync {
     <T extends Model> Cancelable delete(
             @NonNull T model,
             @NonNull ModelSchema modelSchema,
-            @NonNull Integer version,
+            @Nullable Integer version,
             @NonNull QueryPredicate predicate,
             @NonNull Consumer<GraphQLResponse<ModelWithMetadata<T>>> onResponse,
             @NonNull Consumer<DataStoreException> onFailure
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
@@ -138,16 +138,14 @@ internal class VersionRepository(private val localStorageAdapter: LocalStorageAd
         metadataIterator: Iterator<ModelMetadata>
     ): Optional<Int> {
         val results: MutableList<ModelMetadata> =
-                ArrayList()
+            ArrayList()
         while (metadataIterator.hasNext()) {
             results.add(metadataIterator.next())
         }
         // There should be only one metadata for the model....
         if (results.size != 1) {
             LOG.warn(
-                    "Wanted exactly 1 metadata for item with id = " + model.primaryKeyString + ", but had " + results.size +
-                            "." +
-                    "This is likely a bug. please report to AWS."
+                "Wanted 1 metadata for item with id = " + model.primaryKeyString + ", but had " + results.size + "."
             )
             return Optional.empty()
         } else {
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
@@ -29,6 +29,7 @@ import com.amplifyframework.datastore.extensions.getMetadataSQLitePrimaryKey
 import com.amplifyframework.datastore.storage.LocalStorageAdapter
 import io.reactivex.rxjava3.core.Single
 import io.reactivex.rxjava3.core.SingleEmitter
+import java.util.Optional
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
@@ -48,8 +49,8 @@ internal class VersionRepository(private val localStorageAdapter: LocalStorageAd
      * @param <T> Type of model
      * @return Current version known locally
      </T> */
-    fun <T : Model> findModelVersion(model: T): Single<Int> {
-        return Single.create { emitter: SingleEmitter<Int> ->
+    fun <T : Model> findModelVersion(model: T): Single<Optional<Int>> {
+        return Single.create { emitter: SingleEmitter<Optional<Int>> ->
             // The ModelMetadata for the model uses the same ID as an identifier.
             localStorageAdapter.query(
                 ModelMetadata::class.java,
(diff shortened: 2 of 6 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
```diff
@@ -129,32 +130,29 @@ internal class VersionRepository(private val localStorageAdapter: LocalStorageAd
...
             results.add(metadataIterator.next())
         }
-
         // There should be only one metadata for the model....
         if (results.size != 1) {
-            throw DataStoreException(
-                "Wanted 1 metadata for item with id = " + model.primaryKeyString + ", but had " + results.size +
-                    ".",
-                "This is likely a bug. please report to AWS."
+            LOG.warn(
```
mattcreaser, line 147: It sounds like from the description this is not necessarily a bug, right?

aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSync.java
```diff
@@ -112,7 +112,7 @@ <T extends Model> Cancelable create(
     <T extends Model> Cancelable update(
             @NonNull T model,
             @NonNull ModelSchema modelSchema,
-            @NonNull Integer version,
+            @Nullable Integer version,
```
tylerjroach, line 115: Just noting that this changes a public contract. However, these are annotations that may throw warnings, but the change shouldn't explicitly break anything. This filed is really likely only public due to Java interface rules

tylerjroach (commented):
Can you describe the behavior of what happens if:

AppSync has a model that is on version 5.
If we attempt to update that model and don't pass a version, would it actually update?

Or is a null version passed to appsync side only going to insert as a v1 version if the model doesn't already exist.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/VersionRepository.kt
lawmicha: right, removed the "likely a bug" line. i think keeping the log still makes sense so we have more info in the logs.

tylerjroach (approved):

vincetran (approved):

Issue #2840, opened 2024-06-04
TransferWorkerObserver$attachObserverForPendingTransfer$1.invokeSuspend (API 25)


onseok:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin

### Amplify Categories

Storage

### Gradle script dependencies

<details>

```kotlin
// libs.versions.toml
[versions]
aws = "2.16.0"

[libraries]
aws-storage-s3 = { group = "com.amplifyframework", name = "aws-storage-s3", version.ref = "aws" }
aws-auth-cognito = { group = "com.amplifyframework", name = "aws-auth-cognito", version.ref = "aws" }
aws-core-kotlin = { group = "com.amplifyframework", name = "core-kotlin", version.ref = "aws" }

// build.gradle.kts (Module: app)
implementation(libs.aws.storage.s3)
implementation(libs.aws.auth.cognito)

// build.gradle.kts (Module: sync)
implementation(libs.aws.storage.s3)
implementation(libs.aws.core.kotlin)
```

</details>

### Environment information

<details>

```
# Put output below this line
------------------------------------------------------------
Gradle 8.7
------------------------------------------------------------

Build time:   2024-03-22 15:52:46 UTC
Revision:     650af14d7653aa949fce5e886e685efc9cf97c10

Kotlin:       1.9.22
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          17.0.10 (Amazon.com Inc. 17.0.10+7-LTS)
OS:           Mac OS X 14.3.1 aarch64

```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/gen1/android/build-a-backend/storage/upload/#upload-files

Upload files with StoragePath
### Describe the bug

The amplify version that i am using is 2.16.0, and I can't re-produce the problem because I counter it on Firebase Crashlytics.
But it occurs when I use workmanager for upload log files periodically using PeriodicWorkRequest.

### Code Snippet

```kotlin
// Put your code below this line.

// This is a simplified version of the code.
import android.content.Context
import androidx.hilt.work.HiltWorker
import androidx.work.CoroutineWorker
import androidx.work.ForegroundInfo
[...]

PR #2851, opened 2024-06-14
fix(storage): prevent transfer work exception

tylerjroach:

We did not have a try/catch around a method that can throw `https://developer.android.com/reference/android/database/Cursor#getColumnIndexOrThrow(java.lang.String)`.

In rare cases (likely db close) a column may not be available. If this happens, we break out of reading the cursor.

See customer issue for more details.

*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/2840

*Description of changes:*
See reported issue for details

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
@@ -192,7 +192,8 @@ internal class TransferWorkerObserver private constructor(
                         val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
                         attachObserver(id.toString())
                     } catch (exception: IllegalStateException) {
-                        // do nothing. Cursor is likely closed
+                        // Stop iterating over cursor. Cursor is likely closed
+                        break
                     }
                 }
             }
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
@@ -186,8 +186,14 @@ internal class TransferWorkerObserver private constructor(
                 )
             )?.use {
                 while (it.moveToNext()) {
-                    val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
-                    attachObserver(id.toString())
+                    // getInt has been reported to  throw in rare cases where it appears another thread
+                    // may have closed cursor
+                    try {
+                        val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
+                        attachObserver(id.toString())
+                    } catch (exception: IllegalStateException) {
+                        // do nothing. Cursor is likely closed
+                    }
                 }
             }
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## [Release 2.19.1](https://github.com/aws-amplify/amplify-android/releases/tag/release_v2.19.1)
+
+### Bug Fixes
+- **auth:** Fix isSignedIn states ([#2830](https://github.com/aws-amplify/amplify-android/issues/2830))
+- **DataStore:** should not crash on missing version metadata ([#2849](https://github.com/aws-amplify/amplify-android/issues/2849))
+
+[See all changes between 2.19.0 and 2.19.1](https://github.com/aws-amplify/amplify-android/compare/release_v2.19.0...release_v2.19.1)
+
+## [Release 2.19.0](https://github.com/aws-amplify/amplify-android/releases/tag/release_v2.19.0)
+
+### Miscellaneous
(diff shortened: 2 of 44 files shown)


vincetran (dismissed):

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
```diff
@@ -186,8 +186,14 @@ internal class TransferWorkerObserver private constructor(
                 )
             )?.use {
                 while (it.moveToNext()) {
```
lawmicha, line 188: if we catch and continue, what happens on the next `it.moveToNext()` ? does it return null and break out? would it be safer to catch and break out?

tylerjroach: It would be false the next time and break out itself

tylerjroach: It is not reported to throw: https://developer.android.com/reference/android/database/Cursor#moveToNext()

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
```diff
@@ -186,8 +186,14 @@ internal class TransferWorkerObserver private constructor(
...
-                    val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
-                    attachObserver(id.toString())
+                    // getInt has been reported to  throw in rare cases where it appears another thread
+                    // may have closed cursor
+                    try {
+                        val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
+                        attachObserver(id.toString())
+                    } catch (exception: IllegalStateException) {
+                        // do nothing. Cursor is likely closed
+                    }
```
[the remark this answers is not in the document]
tylerjroach: https://developer.android.com/reference/android/database/Cursor#getColumnIndexOrThrow(java.lang.String)

IllegalStateException is the only exception type to be handled here. Since these Android APIs are written in Java, we can trust that the exceptions thrown should only be the ones listed in documentation.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
tylerjroach: I went ahead and added a break as looking through cursor code, I couldn't find where cursor count would be reset to 0 when closed, and didn't see moveToNext checking if cursor were open.

lawmicha (approved):

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
```diff
@@ -186,8 +186,15 @@ internal class TransferWorkerObserver private constructor(
...
-                    val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
-                    attachObserver(id.toString())
+                    // getInt has been reported to  throw in rare cases where it appears another thread
+                    // may have closed cursor
+                    try {
+                        val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
+                        attachObserver(id.toString())
+                    } catch (exception: IllegalStateException) {
+                        // Stop iterating over cursor. Cursor is likely closed
+                        break
```
mattcreaser, line 196: Is this situation worthy of an error long? What is the impact of failing to attach an observer?

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferWorkerObserver.kt
```diff
@@ -186,8 +186,15 @@ internal class TransferWorkerObserver private constructor(
                 )
             )?.use {
                 while (it.moveToNext()) {
-                    val id = it.getInt(it.getColumnIndexOrThrow(TransferTable.COLUMN_ID))
-                    attachObserver(id.toString())
+                    // getInt has been reported to  throw in rare cases where it appears another thread
+                    // may have closed cursor
```
mattcreaser, line 190: Any ideas how this could happen?

tylerjroach:
Closing this PR as the fix was a bit misguided.

1. The cursor is never closed in the app, so this isn't a closed cursor.
2. getInt is in fact throwing (even though it doesn't say it throws. It's crashing at a native jni layer.
3. Couldn't read at row 912, column 0 means that a row in the db was not able to be read. Column 0 should just be the "_id" column of TransferTable. In research, this type of issue typically appears if we attempt to read column data containing 2+MB of data. This shouldn't be the case here being only the _id. My only other speculation at this time (outside of device bugs outside of our control), is that multithreaded operations hitting the table could be clearing the cursor where row 912 doesn't actually exist at the time we try to read data from it.

PR #2895, opened 2024-08-14
feat(storage): update AWSS3StoragePlugin to support multiple buckets

phantumcode:
**Multi-bucket support part 1**

*Issue #, if available:*

*Description of changes:*
This the first part of a series of changes/pull request to support multi-bucket in the S3 Storage Plugin.  The series of changes will be reviewed and merged into a feature branch and combined into a single pull request to be reviewed and merged into `main` branch.

Changes in this pull request includes:
* Update `AmplifyOutputsData` to support multi-bucket configuration
* Add new data structures and update `StorageOption` to specify S3 bucket
* Update `AWSS3StoragePlugin` to support interfacing with unique bucket/region by creating and maintaining multiple `AWSS3StorageService` instances.  Creating multiple `AWSS3StorageService` at the plugin level is the least invasive change to support multiple buckets/regions; the alternative is to create and maintain multiple `S3Client` at the operation layer and result in updating every operations in the Storage plugin.
* Update `getUrl` API to support multiple buckets.  The rest of the Storage plugin API will be updated in subsequent pull requests.
* Added unit tests to support new behaviors

Integration tests will be added in a separate pull request.

* verified via unit tests and manual testing

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
@@ -533,13 +533,13 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
         );
 
         AWSS3StoragePathDownloadFileOperation operation = new AWSS3StoragePathDownloadFileOperation(
-            request,
-            defaultStorageService,
-            executorService,
-            authCredentialsProvider,
-            onProgress,
-            onSuccess,
-            onError
+                request,
+                defaultStorageService,
+                executorService,
+                authCredentialsProvider,
+                onProgress,
+                onSuccess,
+                onError
         );
         operation.start();
 
@@ -663,13 +663,13 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
         );
 
         AWSS3StoragePathUploadFileOperation operation = new AWSS3StoragePathUploadFileOperation(
-            request,
-            defaultStorageService,
-            executorService,
-            authCredentialsProvider,
-            onProgress,
-            onSuccess,
-            onError
+                request,
+                defaultStorageService,
+                executorService,
+                authCredentialsProvider,
+                onProgress,
+                onSuccess,
+                onError
         );
         operation.start();
 
@@ -870,12 +870,12 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
 
         AWSS3StoragePathRemoveOperation operation =
                 new AWSS3StoragePathRemoveOperation(
-                    defaultStorageService,
-                    executorService,
-                    authCredentialsProvider,
-                    request,
-                    onSuccess,
-                    onError);
+                        defaultStorageService,
+                        executorService,
+                        authCredentialsProvider,
+                        request,
+                        onSuccess,
+                        onError);
 
         operation.start();
 
@@ -931,14 +931,14 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
                         case DOWNLOAD:
                             AWSS3StorageDownloadFileOperation
                                 downloadFileOperation = new AWSS3StorageDownloadFileOperation(
-                                    transferId,
-                                    new File(transferRecord.getFile()),
-                                    defaultStorageService,
-                                    executorService,
-                                    authCredentialsProvider,
-                                    awsS3StoragePluginConfiguration,
-                                    null,
-                                    transferObserver);
+                                transferId,
+                                new File(transferRecord.getFile()),
+                                defaultStorageService,
+                                executorService,
+                                authCredentialsProvider,
+                                awsS3StoragePluginConfiguration,
+                                null,
+                                transferObserver);
                             onReceived.accept(downloadFileOperation);
                             break;
                         default:
@@ -1030,12 +1030,12 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
 
         AWSS3StoragePathListOperation operation =
                 new AWSS3StoragePathListOperation(
-                    defaultStorageService,
-                    executorService,
-                    authCredentialsProvider,
-                    request,
-                    onSuccess,
-                    onError);
+                        defaultStorageService,
+                        executorService,
+                        authCredentialsProvider,
+                        request,
+                        onSuccess,
+                        onError);
 
         operation.start();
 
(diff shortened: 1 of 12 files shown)


core/src/test/java/com/amplifyframework/core/configuration/AmplifyOutputsDataTest.kt
```diff
@@ -253,6 +253,44 @@ class AmplifyOutputsDataTest {
...
+        )
+
+        val outputs = AmplifyOutputsData.deserialize(json)
+
+        outputs.storage.shouldNotBeNull()
+        outputs.storage?.run {
+            awsRegion shouldBe "us-east-1"
+            bucketName shouldBe "myBucket"
+            buckets.size shouldBe 2
+            buckets[0].name shouldBe "name1"
```
tylerjroach, line 288: suggest 

```
buckets[0].apply {
  name should be "myBucket"
  ...
}
```

Another option wouild be to just constructed an expectedList and do a shouldBe comparison on that.

core/src/main/java/com/amplifyframework/storage/StorageBucket.kt
```diff
@@ -0,0 +1,32 @@
...
+        @JvmStatic
+        fun fromOutputs(name: String): StorageBucket = OutputsStorageBucket(name)
+        @JvmStatic
+        fun fromBucketInfo(bucketInfo: BucketInfo): StorageBucket = ResolvedStorageBucket(bucketInfo)
+    }
+}
+
+// While class may be technically public, the constructor is internal,
+// Customer is never expected to see or attempt to use this extended class
+data class OutputsStorageBucket internal constructor(val name: String) : StorageBucket()
```
tylerjroach, line 28: Can add `@InternalAmplifyApi` annotation for each of these instead of the code comments

core/src/main/java/com/amplifyframework/storage/InvalidStorageBucketException.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.amplifyframework.storage
+
+import com.amplifyframework.AmplifyException
+import com.amplifyframework.annotations.InternalAmplifyApi
+
+/**
+ * Exception thrown when an invalid StorageBucket is specified.
+ */
+class InvalidStorageBucketException @InternalAmplifyApi constructor(
```
tylerjroach, line 23: the constructoer here should be marked `internal`, not that annotation.

I think this is a legitimate public error, but if not, the annotation would go on the class (not constructor) and that is only if you couldn't mark the class itself as internal, for various reasons.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -252,6 +265,7 @@ public void configure(@NonNull AmplifyOutputsData configuration, @NonNull Contex
             );
         }
 
+        this.configuredBuckets = storage.getBuckets();
```
tylerjroach, line 268: I would consider refactoring the private configure method that all other configure's go through just in case we introduce another path that supports multiple buckets.

```
private void configure(
    @NonNull Context context,
    @NonNull ResolvedStorageBucket defaultBucket,
    @NonNull Map<String, ResolvedStorageBucket> buckets
)
```

Can convert default region, bucket to a ResolvedStorageBucket.
Can convert buckets from config to a map of the friendly name to a ResolveStorageBucket.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -121,11 +129,16 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
     private final ExecutorService executorService;
     private final AuthCredentialsProvider authCredentialsProvider;
     private final AWSS3StoragePluginConfiguration awsS3StoragePluginConfiguration;
-    private AWSS3StorageService storageService;
+    private AWSS3StorageService defaultStorageService;
     @SuppressWarnings("deprecation")
     private StorageAccessLevel defaultAccessLevel;
     private int defaultUrlExpiration;
 
+    private Map<String, AWSS3StorageService> awsS3StorageServicesByBucketName;
```
tylerjroach, line 137: What if 2 buckets have same name but different region? I think the key here should be ResolvedStorageBucket.

tylerjroach: Can we also add the default storage service in this map so that we don't end up with a duplicate if customer explicitly requests the default bucket.

tylerjroach (changes requested):
There are a few options types in `package com.amplifyframework.storage.options` that still need updated.

Main concern at the moment is the storage service lookup. I may have additional comments on the AWSS3StoragePlugin once the changes are considered.

If at all possible, can you try to revert the unnecessary formatting changes so its easier to see what has changed due to feature implementation.

core/src/test/java/com/amplifyframework/core/configuration/AmplifyOutputsDataTest.kt
phantumcode: good suggestion, unit test updated.

core/src/main/java/com/amplifyframework/storage/StorageBucket.kt
phantumcode: `@InternalAmplifyApi` annotation added

core/src/main/java/com/amplifyframework/storage/InvalidStorageBucketException.kt
phantumcode: updated per suggestion.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
phantumcode: great input, I didn't think of it.  I've updated per suggestion.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
phantumcode: Yes, the default storage service is always added by default at configure time [see line 290](https://github.com/aws-amplify/amplify-android/pull/2895/files#diff-68c08b683554135bdca3ec8741ef1160af2a1add61a87d302af6719c214b46cdR290)

Bucket names are unique globally across all regions in S3; so I think it's appropriate to use it as the key here.  It's not possible to have two buckets with the same name but different region.

phantumcode:
I've resolved the review comments.  I've intentionally excluded the other option types in package `com.amplifyframework.storage.options` to reduce the scope of this review and focus on the service lookup behavior.  Once this pull request is reviewed and approved, the other option types will be updated in a separate pull request.

tylerjroach (approved):
Thanks for addressing. I was not aware about bucket names being unique regardless of region

vincetran (approved):

PR #2896, opened 2024-08-15
feat(storage): update storage API options to include bucket

phantumcode:
**Multi-bucket support part 2**

This the second part of a series of changes/pull request to support multi-bucket in the S3 Storage Plugin. The series of changes will be reviewed and merged into a feature branch and combined into a single pull request to be reviewed and merged into main branch.

part 1: https://github.com/aws-amplify/amplify-android/pull/2895

- [x] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*

*Description of changes:*
* Updated Storage plugin api options in `com.amplifyframework.storage.options` to support specifying `StorageBucket`.

* Integration test will be added in a separate followup PR (https://github.com/aws-amplify/amplify-android/pull/2900)

*How did you test these changes?*
(Please add a line here how the changes were tested)

*Documentation update required?*
- [x] No
- [ ] Yes (Please include a PR link for the documentation update)

*General Checklist*
- [x] Added Unit Tests
- [ ] ~Added Integration Tests~
- [x] Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2899, opened 2024-08-20
feat(storage): update multipart operations to support multiple buckets/regions

phantumcode:
**Multi-bucket support part 3**

This the third part of a series of changes/pull request to support multi-bucket in the S3 Storage Plugin. The series of changes will be reviewed and merged into a feature branch and combined into a single pull request to be reviewed and merged into main branch.

part 1: https://github.com/aws-amplify/amplify-android/pull/2895
part 2: https://github.com/aws-amplify/amplify-android/pull/2896

*Issue #, if available:*

*Description of changes:*
* Updated transfer table to upgrade to new version with region column added.
* Refactored transfer workers to get new s3 client from a client provider that gets the S3 client based on the transfer record's region and bucket name
* Updated transfer operations to get `AWSS3Service` based on bucket name and region

Integration tests will be added in a separate followup PR (https://github.com/aws-amplify/amplify-android/pull/2900)

* Manually tested storage apis and verified uploads/downloads works as expected.
* Verified migrating from old transfer db table to new transfer db table works as expecting while in the middle of an large file upload.  Large file upload was able to continue and complete successfully after table migration/upgrade and region column was default to null.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] ~Added Integration Tests~

### DIFF
--- a/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
+++ b/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
@@ -90,6 +90,7 @@ import com.amplifyframework.storage.s3.request.AWSS3StoragePathUploadRequest;
 import com.amplifyframework.storage.s3.request.AWSS3StorageRemoveRequest;
 import com.amplifyframework.storage.s3.request.AWSS3StorageUploadRequest;
 import com.amplifyframework.storage.s3.service.AWSS3StorageService;
+import com.amplifyframework.storage.s3.service.AWSS3StorageServiceContainer;
 import com.amplifyframework.storage.s3.transfer.S3StorageTransferClientProvider;
 import com.amplifyframework.storage.s3.transfer.StorageTransferClientProvider;
 import com.amplifyframework.storage.s3.transfer.TransferObserver;
@@ -102,9 +103,7 @@ import org.json.JSONObject;
 
 import java.io.File;
 import java.io.InputStream;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -136,8 +135,7 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
     private StorageAccessLevel defaultAccessLevel;
     private int defaultUrlExpiration;
 
-    private final Map<String, AWSS3StorageService> awsS3StorageServicesByBucketName = new HashMap<>();
-    private Context context;
+    private AWSS3StorageServiceContainer awss3StorageServiceContainer;
     @SuppressLint("UnsafeOptInUsageError")
     private List<AmplifyOutputsData.StorageBucket> configuredBuckets;
 
@@ -146,13 +144,11 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
             = new S3StorageTransferClientProvider((region, bucketName) -> {
                 if (region != null && bucketName != null) {
                     StorageBucket bucket = StorageBucket.fromBucketInfo(new BucketInfo(bucketName, region));
-                    return getAWSS3StorageService((ResolvedStorageBucket) bucket).getClient();
+                    return awss3StorageServiceContainer.get((ResolvedStorageBucket) bucket).getClient();
                 }
 
                 if (region != null) {
-                    // unable to create a new S3Client from java code,
-                    // redirecting to AWSS3Service to create S3 Client from kotlin
-                    return AWSS3StorageService.getS3Client(region, authCredentialsProvider);
+                    return S3StorageTransferClientProvider.getS3Client(region, authCredentialsProvider);
                 }
                 return defaultStorageService.getClient();
             });
@@ -195,6 +191,7 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
     @VisibleForTesting
     AWSS3StoragePlugin(AuthCredentialsProvider authCredentialsProvider,
                        AWSS3StoragePluginConfiguration awss3StoragePluginConfiguration) {
+
         this((context, region, bucket, clientProvider) ->
                 new AWSS3StorageService(
                     context,
@@ -300,14 +297,15 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
             @NonNull ResolvedStorageBucket bucket
     ) throws StorageException {
         try {
-            this.context = context;
             this.defaultStorageService = storageServiceFactory.create(
                     context,
                     region,
                     bucket.getBucketInfo().getName(),
                     clientProvider);
-            this.awsS3StorageServicesByBucketName.clear();
-            this.awsS3StorageServicesByBucketName.put(bucket.getBucketInfo().getName(), this.defaultStorageService);
+            this.awss3StorageServiceContainer = new AWSS3StorageServiceContainer(
+                    context, storageServiceFactory,
+                    (S3StorageTransferClientProvider) clientProvider);
+            this.awss3StorageServiceContainer.put(bucket.getBucketInfo().getName(), this.defaultStorageService);
         } catch (RuntimeException exception) {
             throw new StorageException(
                 "Failed to create storage service.",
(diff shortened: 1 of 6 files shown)


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
```diff
@@ -126,20 +127,36 @@ public final class AWSS3StoragePlugin extends StoragePlugin<S3Client> {
...
+    private StorageTransferClientProvider clientProvider
+            = new S3StorageTransferClientProvider((region, bucketName) -> {
+                if (region != null && bucketName != null) {
+                    StorageBucket bucket = StorageBucket.fromBucketInfo(new BucketInfo(bucketName, region));
+                    return getAWSS3StorageService((ResolvedStorageBucket) bucket).getClient();
+                }
+
+                if (region != null) {
+                    // unable to create a new S3Client from java code,
+                    // redirecting to AWSS3Service to create S3 Client from kotlin
```
lawmicha, line 154: i don't quite understand this comment, from the checks it looks like we'll dynamically create the S3 client based on 
1. bucket and region
2. region only- bucket inferred?
3. bucket and region inferred?

what does it mean to "create S3 Client from kotlin"?

lawmicha (approved):

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/AWSS3StoragePlugin.java
phantumcode: ideally, we already have the region here in `AWSStoragePlugin.java` and can create an S3Client directly using java syntax i.e. `S3Client.builder().setRegion(region).setCredentialsProvider(authCredentialsProvider).build()`.  But this syntax is not possible from java code since the AWS Kotlin SDK doesn't expose the builder to set the region and credentials.  The only one to create the S3Client is by using kotlin syntax, so the S3Client creation is defined in `AWSS3Service.kt`

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferTable.kt
```diff
@@ -219,6 +222,9 @@ internal class TransferTable {
             if (TABLE_VERSION_9 in (oldVersion + 1)..newVersion) {
                 addVersion9Columns(database)
             }
+            if (TABLE_VERSION_10 in (oldVersion + 1)..newVersion) {
+                addVersion10Columns(database)
```
tylerjroach, line 226: This is good for the update path, but the create code is missing the new region column

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/StorageTransferClientProvider.kt
```diff
@@ -0,0 +1,23 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.storage.s3.transfer
+
+import aws.sdk.kotlin.services.s3.S3Client
+import com.amplifyframework.annotations.InternalApiWarning
+
+@InternalApiWarning
```
tylerjroach, line 20: We don't need to add annotations to internal classes, only public classes that are not part of public contract.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/S3StorageTransferClientProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+package com.amplifyframework.storage.s3.transfer
+
+import aws.sdk.kotlin.services.s3.S3Client
+import com.amplifyframework.auth.AuthCredentialsProvider
+import com.amplifyframework.storage.StorageException
+
+internal class S3StorageTransferClientProvider(
+    private val createS3Client: (region: String?, bucketName: String?) -> S3Client
+) : StorageTransferClientProvider {
+    override fun getStorageTransferClient(region: String?, bucketName: String?): S3Client {
```
tylerjroach, line 24: Do we need to also inject the credentials provider as that is a global s3 config option I believe.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
```diff
@@ -55,16 +55,27 @@ internal class AWSS3StorageService(
...
     private val s3BucketName: String,
     private val authCredentialsProvider: AuthCredentialsProvider,
-    private val awsS3StoragePluginKey: String
+    private val awsS3StoragePluginKey: String,
+    private val clientProvider: StorageTransferClientProvider
 ) : StorageService {
 
+    companion object {
+        @JvmStatic
+        fun getS3Client(region: String, authCredentialsProvider: AuthCredentialsProvider): S3Client {
```
tylerjroach, line 64: Should we just use the ClientProvider so that all s3 providers are created in same location. Also see comment about the credentialsProvider. I think we need to use this value for all created S3Clients.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/S3StorageTransferClientProvider.kt
tylerjroach: Another thought. At this point, could we make region and bucketName required? We should be able to provide the default values for both of these at any point that this method is getting called. It would allow us to remove all of the fallback logic elsewhere if region and bucketName were no longer null here.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferTable.kt
phantumcode: if the the new column in the create code, we'll get a column already exists error.  The database gets created at version 1 and then upgrades to 10. 
https://github.com/aws-amplify/amplify-android/pull/2899/files#diff-26d789d324d49670cba8878a5cdd9b2eca0f1ea1ad29f684c46d5a2c8cd5e14dR186-R187

tylerjroach: Yikes, I see this now. This is NOT how it is supposed to work. Lets huddle on this.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/S3StorageTransferClientProvider.kt
phantumcode: The clientProvider is mainly/directly used to create S3 clients from the transfer records which doesn't have context or access to to credentials. 
The transfer record has region and bucket name fields as null by default, so bucket name and region can't be required here as it is possible for either of those fields to be null when reading from the database.
Here's an example of usage:
https://github.com/aws-amplify/amplify-android/pull/2899/files#diff-9fdcc6ff34286fc160d66a1b782c065eb10867f6c21e70e2b37fdf9287ba3f17R44

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: This is essentially part of the client provider delegated work/implementation.  It's the work-around to create an S3Client when region is the only field available and I can't create an S3Client in java code.

https://github.com/aws-amplify/amplify-android/pull/2899/files#diff-68c08b683554135bdca3ec8741ef1160af2a1add61a87d302af6719c214b46cdR152-R156

Let me know if there's a simpler way to do this.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/StorageTransferClientProvider.kt
phantumcode: Good catch, annotation removed.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferTable.kt
tylerjroach: TransferDBHelper extends SQLiteOpenHelper which is a SQLite helper class that Android provides.

onCreate, onUpdate, etc are supposed to be called by Android, not manually called.

Looks like we have `TransferTable` to then route these calls to. Right now, TransferDbHelper sets the version of the database, not `TransferTable`.

So our `TransferTable's create statement is always creating v1, and then running through all upgrades to arrive at how VERSION 10 should actually look.

While efficiency would take a small hit (probably negligible in this case), there's no reason we should be calling onUpgrade from within onCreate. This is not how DB creation/migrations are supposed to work and could result in us making errors in the future.

The correct think to do would be to maintain a single create statement and then proper onUpgrade/onDowngrade sqlite statements. Looks like this is nothing introduced by your changes, and is how the db is written to behave in other areas of our code as well.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: Alternatively, I can remove if `if(region != null)` block completely and always return the S3Client from the `defaultStorageService` to simplify the clientProvider delegate implementation.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/S3StorageTransferClientProvider.kt
tylerjroach: I don't think we currently expose it publicly, its mainly for tests. But the AWSS3StoragePlugin takes a global credentialsProvider. We inject that provider into the s3 client. The S3Client from within the Kotlin SDK will then resolve the credentials. All of these created S3Clients are missing a way to inject the proper credentials provider provided in the plugin constructor.

```
private var s3Client: S3Client = S3Client {
        region = awsRegion
        credentialsProvider = authCredentialsProvider
}
```

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/TransferTable.kt
phantumcode: I've added code comments to clarify the current design and prevent future updates to the create sql statement.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/S3StorageTransferClientProvider.kt
phantumcode: Discussed offline and verified that all `S3Client` instances have the appropriate region and credentials injected.

aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.kt
phantumcode: Refactored and move the static method to `S3StorageTransferClientProvider`

tylerjroach (approved):

PR #2937, opened 2024-10-10
fix(datastore): Store and use syncExpression field in LastSyncMetadata table to retrieve adjusted lastSyncTime

edisooon:

*Issue #, if available:*

*Description of changes:*
This is **the second out of two PRs** in an effort to address a current issue customers had reported in our Datastore category.

The issue is that the "changing sync expression in Runtime" doesn't work as expected as described in [our doc](https://docs.amplify.aws/gen1/react/build-a-backend/more-features/datastore/sync-to-cloud/#reevaluate-expressions-at-runtime).

To see the full description of the bug & change, please refer to [the description in this PR](https://github.com/aws-amplify/amplify-android/pull/2936).

**[TODO]**

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
@@ -17,6 +17,7 @@ package com.amplifyframework.datastore.syncengine;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.core.model.query.Where;
 import com.amplifyframework.core.model.query.predicate.QueryField;
@@ -40,18 +41,19 @@ final class SyncTimeRegistry {
         this.localStorageAdapter = localStorageAdapter;
     }
 
-    //TODO: add the second argument: current sync expression here
-    Single<SyncTime> lookupLastSyncTime(@NonNull String modelClassName) {
+    Single<SyncTime> lookupLastSyncTime(@NonNull String modelClassName, @NonNull String syncExpression) {
         return Single.create(emitter -> {
             QueryPredicate hasMatchingModelClassName = QueryField.field("modelClassName").eq(modelClassName);
 
             localStorageAdapter.query(LastSyncMetadata.class, Where.matches(hasMatchingModelClassName), results -> {
                 try {
                     LastSyncMetadata syncMetadata = extractSingleResult(modelClassName, results);
-                    //TODO: syncMetadata should contain the previous serialized sync expression
-                    //TODO: compare the previous sync expression with the current one
-                    //TODO: emmit SyncTime.never() if different
-                    emitter.onSuccess(SyncTime.from(syncMetadata.getLastSyncTime()));
+                    SyncTime lastSyncTime = SyncTime.from(syncMetadata.getLastSyncTime());
+                    String lastSyncExpression = syncMetadata.getSyncExpression();
+                    if(!ObjectsCompat.equals(lastSyncExpression, syncExpression)){
+                        lastSyncTime = SyncTime.never();
+                    }
+                    emitter.onSuccess(lastSyncTime);
                 } catch (DataStoreException queryResultFailure) {
                     emitter.onError(queryResultFailure);
                 }
@@ -59,8 +61,7 @@ final class SyncTimeRegistry {
         });
     }
 
-    //TODO: change the name to saveLastDeltaSync, and add the second argument for sync expression
-    Completable saveLastDeltaSyncTime(@NonNull String modelClassName, @Nullable SyncTime syncTime, @Nullable String syncExpression) {
+    Completable saveLastDeltaSync(@NonNull String modelClassName, @Nullable SyncTime syncTime, @Nullable String syncExpression) {
         LastSyncMetadata metadata = syncTime != null && syncTime.exists() ?
             LastSyncMetadata.deltaSyncedAt(modelClassName, syncTime.toLong(), syncExpression) :
             LastSyncMetadata.neverSynced(modelClassName);
@@ -76,8 +77,7 @@ final class SyncTimeRegistry {
         );
     }
 
-    //TODO: change the name to saveLastDeltaSync, and add the second argument for sync expression
-    Completable saveLastBaseSyncTime(@NonNull String modelClassName, @Nullable SyncTime syncTime, @Nullable String syncExpression) {
+    Completable saveLastBaseSync(@NonNull String modelClassName, @Nullable SyncTime syncTime, @Nullable String syncExpression) {
         LastSyncMetadata metadata = syncTime != null && syncTime.exists() ?
             LastSyncMetadata.baseSyncedAt(modelClassName, syncTime.toLong(), syncExpression) :
             LastSyncMetadata.neverSynced(modelClassName);
--- a/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
+++ b/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncProcessor.java
@@ -46,7 +46,9 @@ import com.amplifyframework.datastore.utils.ErrorInspector;
 import com.amplifyframework.hub.HubChannel;
 import com.amplifyframework.hub.HubEvent;
 import com.amplifyframework.logging.Logger;
+import com.amplifyframework.util.GsonFactory;
 import com.amplifyframework.util.Time;
+import com.google.gson.Gson;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -80,6 +82,7 @@ final class SyncProcessor {
     private final DataStoreConfigurationProvider dataStoreConfigurationProvider;
     private final QueryPredicateProvider queryPredicateProvider;
     private final RetryHandler requestRetry;
+    private final Gson gson;    //TODO: for pre-deserialization of QueryPredicate, will remove shortly
 
     /**
      * The `isSyncRetryEnabled` value is being passed down all the way from the `AWSDataStorePlugin` or the
@@ -97,7 +100,7 @@ final class SyncProcessor {
         this.queryPredicateProvider = builder.queryPredicateProvider;
         this.requestRetry = builder.requestRetry;
         this.isSyncRetryEnabled = builder.isSyncRetryEnabled;
-
+        this.gson = GsonFactory.instance();     //TODO: for pre-deserialization of QueryPredicate, will remove shortly
         if (!this.isSyncRetryEnabled) {
             LOG.warn("Disabling sync retries will be deprecated in a future version.");
         }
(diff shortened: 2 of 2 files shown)


aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+
+import com.amplifyframework.core.Amplify;
+import com.amplifyframework.core.category.CategoryType;
+import com.amplifyframework.logging.Logger;
+import com.amplifyframework.util.Wrap;
+
+/**
+ * Add SyncExpression (TEXT) column to LastSyncMetadata table.
+ */
+public final class AddSyncExpressionToLastSyncMetadata implements ModelMigration {
```
mattcreaser, line 29: Should not be public

mattcreaser, line 29: Let's write any added files in Kotlin please

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
```diff
@@ -512,6 +512,20 @@ public Builder syncExpression(@NonNull String modelName,
...
+        /**
+         * Sets sync expressions for models to filter which data is synced locally.
+         * The expression is evaluated each time DataStore is started.
+         * The QueryPredicate is applied on both sync and subscriptions.
+         * [NOTE: this will override the previous syncExpression configuration if there's any]
+         * @param syncExpressions the Map of {@link DataStoreSyncExpression}s to filter data
+         * @return Current builder
+         */
+        @NonNull
+        public Builder syncExpressions(@NonNull Map<String, DataStoreSyncExpression> syncExpressions) {
```
mattcreaser, line 524: This API change hasn't been reviewed, please remove

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+    private boolean needsMigration() {
+        final String checkColumnSql = "SELECT COUNT(*) FROM pragma_table_info('LastSyncMetadata') " +
+            "WHERE name=" + Wrap.inSingleQuotes(newSyncExpColumnName);
+        try (Cursor queryResults = database.rawQuery(checkColumnSql, new String[]{})) {
+            if (queryResults.moveToNext()) {
+                int recordNum = queryResults.getInt(0);
+                return recordNum == 0; // needs to be upgraded if there's no column named ${newSyncExpColumnName}
+            }
+        }
+        return false;
```
mattcreaser, line 80: Should this default be true?

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
edisooon: Make sense! I had addressed this in my latest commit

aws-datastore/src/main/java/com/amplifyframework/datastore/DataStoreConfiguration.java
edisooon: Gotcha! I just removed this in my latest comment

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
edisooon: This should be false / shouldn't matter IMO since the only scenario when it reaches this default return is that we change the table name "LastSyncMetadata" in our schema, causing the Cursor to be empty.
We would need a new migration class being executed before this class if we actually want to change this table name.

blakedunson (approved):

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -39,40 +41,49 @@ public final class LastSyncMetadata implements Model {
...
      * Creates an instance of an {@link LastSyncMetadata}, indicating that the provided
      * model has been base sync'd, and that the last sync occurred at the given time.
      * @param modelClassName Name of model
      * @param lastSyncTime Last time it was synced
+     * @param syncExpression the corresponding sync expression being used during last sync
      * @param <T> t type of Model.
      * @return {@link LastSyncMetadata} for the model class
      */
+    @InternalApiWarning
     public static <T extends Model> LastSyncMetadata baseSyncedAt(@NonNull String modelClassName,
```
tylerjroach, line 66: This is a breaking change. Need to leave alone and created method overload. Internally, we should check for usages and make sure we are't using this method. Can mark this one as deprecated.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
```diff
@@ -84,22 +95,25 @@ static <T extends Model> LastSyncMetadata deltaSyncedAt(@NonNull String modelCla
      */
     public static <T extends Model> LastSyncMetadata neverSynced(@NonNull String modelClassName) {
```
tylerjroach, line 96: This is a breaking change. Need to leave alone and created method overload. Internally, we should check for usages and make sure we are't using this method. Can mark this one as deprecated.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/ModelMigration.java
```diff
@@ -15,9 +15,12 @@
 
 package com.amplifyframework.datastore.storage.sqlite.migrations;
 
+import com.amplifyframework.annotations.InternalApiWarning;
+
 /**
  * Interface that defines the contract of an in-place model migration.
  */
+@InternalApiWarning
```
tylerjroach, line 23: I'd prefer that we don't go adding this annotation to existing public classes. It can give the impression later on that this annotation has always been there and that we are free to break the contract, which isn't the case.

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Alter LastSyncMetadata table with new column.
+     * Existing rows in LasySyncMetadata will have 'null' for ${newSyncExpColumnName} value,
+     * until the next sync/hydrate operation.
+     */
+    private void addNewSyncExpColumnName() {
+        try {
+            database.beginTransaction();
+            final String addColumnSql = "ALTER TABLE LastSyncMetadata ADD COLUMN " +
```
tylerjroach, line 59: If we are injecting a value, we need to use a parameterized query

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddSyncExpressionToLastSyncMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+            LOG.debug("Successfully upgraded LastSyncMetadata table with new field: " + newSyncExpColumnName);
+        } finally {
+            if (database.inTransaction()) {
+                database.endTransaction();
+            }
+        }
+    }
+
+    private boolean needsMigration() {
+        final String checkColumnSql = "SELECT COUNT(*) FROM pragma_table_info('LastSyncMetadata') " +
```
tylerjroach, line 72: If we are injecting a value, we need to use a parameterized query

aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/migrations/AddModelNameToModelMetadataKey.java
```diff
@@ -30,6 +31,7 @@
 /**
  * Add model name to ModelMetadata key.
  */
+@InternalApiWarning
```
tylerjroach, line 34: I'd prefer that we don't go adding this annotation to existing public classes. It can give the impression later on that this annotation has always been there and that we are free to break the contract, which isn't the case.

aws-api-appsync/src/main/java/com/amplifyframework/core/model/query/predicate/GsonPredicateAdapters.java
```diff
@@ -33,13 +34,15 @@ public final class GsonPredicateAdapters {
     private GsonPredicateAdapters() {}
```
tylerjroach, line 34: Can you explain why the change is needed in this class? Its not obvious from the review why we are having to change serialization code around QueryOperator and QueryPredicate as neither of these files changed.

tylerjroach (commented):
Lets say a customer runs the app for the first time after updating with this fix.

You migrate the database and add `null` to the syncExpression. How do you differentiate between a real `null` sync expression, and one from the migration? We don't really know if thats accurate or not right? Would we be better or not removing last sync for models? I understand this would force a full sync for all initial app opens on this upgrade.

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
```diff
@@ -40,24 +41,31 @@ final class SyncTimeRegistry {
...
         return Single.create(emitter -> {
             QueryPredicate hasMatchingModelClassName = QueryField.field("modelClassName").eq(modelClassName);
 
             localStorageAdapter.query(LastSyncMetadata.class, Where.matches(hasMatchingModelClassName), results -> {
                 try {
                     LastSyncMetadata syncMetadata = extractSingleResult(modelClassName, results);
-                    emitter.onSuccess(SyncTime.from(syncMetadata.getLastSyncTime()));
+                    SyncTime lastSyncTime = SyncTime.from(syncMetadata.getLastSyncTime());
+                    QueryPredicate lastSyncExpression = syncMetadata.getSyncExpression();
+                    if (!ObjectsCompat.equals(lastSyncExpression, syncExpression)) {
```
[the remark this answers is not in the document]
edisooon: @tylerjroach why we could use Objects.equals here specifically? I learned that ObjectsCompat would provide compatibility across different Android API levels

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/LastSyncMetadata.java
edisooon: Make sense!

aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/SyncTimeRegistry.java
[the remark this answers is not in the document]
tylerjroach: It really doesn't matter, just leave it. Looks like it was used elsewhere throughout the project. We don't use a low enough API level on Android for it to matter.

tylerjroach (dismissed):

tylerjroach (approved):

PR #2992, opened 2025-02-21
chore(auth): Migrate TOTP setup to use case classes

mattcreaser:

*Issue #, if available:*

*Description of changes:*
- Moves the TOTP Setup functions into use case classes.
- Fixes some bugs in `setupTotp` where Amplify would hang (never invoke callbacks) in certain scenarios.

- Unit test

- [ ] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.kt
@@ -414,7 +414,7 @@ class AWSCognitoAuthPlugin : AuthPlugin<AWSCognitoAuthService>() {
     }
 
     override fun setUpTOTP(onSuccess: Consumer<TOTPSetupDetails>, onError: Consumer<AuthException>) =
-        enqueue(onSuccess, onError) { queueFacade.setUpTOTP() }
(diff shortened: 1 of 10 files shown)


tylerjroach:
I've noticed this error will never surface anymore:

```
CognitoAuthExceptionConverter.lookup(
  error,
  "Cannot find a multi-factor authentication (MFA) method."
)
```

This looks like it was just a generic catch all but want to make sure the change is expected.

mattcreaser:
The string there is the message set on unmapped exception types (i.e. if we return an `UnknownException`). It was indeed intentionally omitted since we map exceptions at the top level now, which is much simpler, although it does cause a loss of context. I also thought the message didn't make any sense given what the user is trying to do.

If we wanted to retain the contextual message we can do something like:

```kotlin
internal inline fun <R> mapExceptions(fallbackMessage: String, block: () -> R): R {
    try {
        return block()
    } catch (e: Exception) {
        throw CognitoAuthExceptionConverter.lookup(e, fallbackMessage)
    }
}
```

and then `execute` becomes:

```kotlin
suspend fun execute(): TOTPSetupDetails = mapExceptions("Cannot setup TOTP") {
  // ....
}
```

tylerjroach:
Thanks for clarification. I am ok with original change.

tylerjroach (approved):

Issue #2938, opened 2024-10-11
StorageException/S3Exception: Something went wrong with your AWS S3 Storage download file operation/The requested range is not satisfiable


Orbyt:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin, Kotlin - Coroutines

### Amplify Categories

Authentication, Storage

### Gradle script dependencies

<details>

```groovy
implementation("com.amplifyframework.ui:authenticator:1.2.0")
// Issue still present on latest (2.23.0).
implementation("com.amplifyframework:aws-storage-s3:2.19.1")
implementation("com.amplifyframework:aws-storage-s3:2.23.0")

```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 8.7
------------------------------------------------------------
Kotlin:       1.9.22
Groovy:       3.0.17

```

</details>

### Please include any relevant guides or documentation you're referencing

Amplify documentation

### Describe the bug

Amplify throws an `StorageException` when attempting to _download_ a file with the same name to the same location where the original file was _uploaded_.

```kotlin
val downloadsDirectory = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS)
// This is simply the file name from the path used to download the file.
// For example, if path = "/customers/12345/some_file.pdf", this will equal "some_file.pdf".
val fileNameFromPath = path.split("/").last()
val file = File(downloadsDirectory,"$fileNameFromPath")
Amplify.Storage.downloadFile(StoragePath.fromString(path), file, { 
     // Success.
},{
     // Failure.
})
```

The SDK has no problem overwriting a file with the same name in the same location that was already downloaded via the SDK. This issue is only present when an original file is uploaded from some location on the files (e.g. "downloads") and when, immediately afterwards, the SDK attempts to download that file from S3 to the same location with the same file name.

Stack trace
```
DownloadWorker failed with exception: aws.sdk.kotlin.services.s3.model.S3Exception: The requested range is not satisfiable, Request ID: ...., Extended request ID: .....
    at aws.sdk.kotlin.services.s3.serde.GetObjectOperationDeserializerKt.throwGetObjectError(GetObjectOperationDeserializer.kt:115)
[...]

PR #3056, opened 2025-05-12
fix(storage): Improve exception handling when attempting to overwrite a file during download

vincetran:

*Issue #, if available:* https://github.com/aws-amplify/amplify-android/issues/2938 

*Description of changes:* 
There is an edge case where if the app tries to overwrite a file that the app doesn't have permission to (as per the recent [scoped storage changes](https://developer.android.com/training/data-storage#scoped-storage)). When we recognize that a file already exists before downloading, we first delete the file before the download. However, if the app doesn't have the permission to delete the file, it silently fails, and then it has a cascading effect of being unable to download the file. 

This change throws a custom StorageException that calls out that the app did not have permissions to overwrite the file. 

~TODO: Update storage documentation to call out this exception handling and that the developer needs to request write permission to that file before calling download.~

- Tested happy case, overwriting a file was successful
- Tested overwrite fail, the exception was properly thrown
- Tested case where customer app requests permission to modify a file and end user accepts, overwriting a file was successful
- Tested case where customer app requests permission to modify a file and end user declines, overwriting a file files and exception was propery thrown

- [ ] No
- [x] Yes https://github.com/aws-amplify/docs/pull/8361

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/extensions/StorageExceptionExtensions.kt
```diff
@@ -31,3 +32,12 @@ internal fun StoragePathValidationException.Companion.unsupportedStoragePathExce
     ),
     "Provided StoragePath not supported by AWS S3 Storage Plugin"
 )
+
+internal fun StorageFilePermissionException.Companion.unableToOverwriteFileException() = StorageException(
```
tylerjroach, line 36: Feel like companion object on StorageException would be more appropriate. Could also get rid of the companion object on StorageFilePermissionException.

tylerjroach (approved):

Issue #3064, opened 2025-05-16
Autosignin does not trigger hub AuthChannelEventName.SIGNED_IN


niqo01:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```kotlin
amplify = "2.27.4"

amplify-auth-cognito = { module = "com.amplifyframework:aws-auth-cognito", version.ref = "amplify" }
amplify-core-kotlin = { module = "com.amplifyframework:core-kotlin", version.ref = "amplify" }
```

</details>

### Environment information

<details>

```
Gradle 8.13
------------------------------------------------------------

Build time:    2025-02-25 09:22:14 UTC
Revision:      073314332697ba45c16c0a0ce1891fa6794179ff

Kotlin:        2.0.21
Groovy:        3.0.22
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  21.0.7 (Azul Systems, Inc. 21.0.7+6-LTS)
Daemon JVM:    /Library/Java/JavaVirtualMachines/zulu-21.jdk/Contents/Home (no JDK specified, using current Java home)
OS:            Mac OS X 15.4.1 aarch64



```

</details>

### Please include any relevant guides or documentation you're referencing

https://docs.amplify.aws/android/build-a-backend/auth/connect-your-frontend/listen-to-auth-events/

### Describe the bug

when `auth.autoSignIn()` succeed `Amplify.Hub` does not trigger `AuthChannelEventName.SIGNED_IN.name` event name.

### Reproduction steps (if applicable)

1. subscribe to hub event
```
        Amplify.Hub.subscribe(HubChannel.AUTH).collect {
          logger.d{ "Event: $it" }
        }
```
2. auto sign in successfully `auth.autoSignIn()`
3. notice that no event are logged.

mattcreaser:
This was fixed in version **2.28.0**. Please let us know if there are any issues!

PR #3068, opened 2025-05-20
fix(auth): Emit SIGNED_IN event after autoSignIn

mattcreaser:

*Issue #, if available:* #3064

*Description of changes:*
Fixes two issues found in AutoSignInUseCase:
- The hub emitter was being called after emitting the sign in result. This doesn't work because of the use of the `first()` terminal operator, which cancels flow collection as soon as the sign in result is received. This would cancel the suspended coroutine and thus it never resumes to proceed to emitting the hub event.
- The use case was reusing the same hub emitter for multiple instances. This is the wrong pattern and would result in subsequent events potentially not being emitted if there were multiple sign ups within a single app session.

I also changed the transforming function from `transformWhile` to `mapNotNull` to make it more obvious that the collector is cancelled when the result is emitted.

- Verified event is emitted using autoSignIn
- Added unit test to ensure event is emitted

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


mattcreaser:
Codecov says coverage dropped but that appears to be an issue with Kover missing lines. That would maybe be fixed with a Kover version update as we are quite behind.

PR #3123, opened 2025-09-04
chore(auth): Add typesafe classes for handling auth tokens

mattcreaser:

*Issue #, if available:*

*Description of changes:*
Updates internal handling of auth tokens to wrap them in typesafe classes instead of bare strings. This makes the codebase easier to understand, drastically cuts down on the number of times the JWT strings are parsed, and makes it less likely that token content is accidentally leaked (because the masking is applied at the wrapper class level).

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/scripts/Gemfile
+++ b/scripts/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gem "fastlane"
 gem "addressable", ">= 2.8.0"
-gem "rexml", ">= 3.3.9"
+gem "rexml", ">= 3.4.2"
 
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
 eval_gemfile(plugins_path) if File.exist?(plugins_path)
--- a/scripts/Gemfile.lock
+++ b/scripts/Gemfile.lock
@@ -183,7 +183,7 @@ GEM
       trailblazer-option (>= 0.1.1, < 0.2.0)
       uber (< 0.2.0)
     retriable (3.1.2)
-    rexml (3.3.9)
+    rexml (3.4.4)
     rouge (2.0.7)
     ruby2_keywords (0.0.5)
     rubyzip (2.3.2)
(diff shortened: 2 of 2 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/AmplifyCredential.kt
```diff
@@ -134,9 +141,9 @@ internal data class AWSCredentials(
     }
 
     override fun toString(): String = "AWSCredentials(" +
-        "accessKeyId = ${accessKeyId?.substring(0..4)}***, " +
-        "secretAccessKey = ${secretAccessKey?.substring(0..4)}***, " +
-        "sessionToken = ${sessionToken?.substring(0..4)}***, " +
+        "accessKeyId = ${accessKeyId.mask()}, " +
```
tylerjroach, line 109: This really probably doesn't matter since its a masked value but this method includes 1 less character.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/AmplifyCredential.kt
```diff
@@ -95,16 +96,22 @@ internal data class FederatedToken(val token: String, val providerName: String)
  */
 @Serializable
 internal data class CognitoUserPoolTokens(
```
tylerjroach: What does the toString look like now? I think the tokens are still tracked properly after traversing through the code, but notice expiration is now included (probably for the best).

mattcreaser: I had not noticed that expiration was excluded before, but that seems like an error. The updated toString looks like this:

```
CognitoUserPoolTokens(idToken=eyJh***, accessToken=eyJh***, refreshToken=eyJh***, expiration=1758044817)
```

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/AmplifyCredential.kt
mattcreaser: Noted, but agree that doesn't matter. We can make it 4 or 5 but it should be consistent.

PR #3156, opened 2025-11-07
fix(storage): Download Accuracy Fix

tylerjroach:

*Issue #, if available:*

*Description of changes:*
Our integration tests occasionally failed with the downloaded file not always matching the expected original. Through logging and file inspections, I was able to determine that in a rapid pause/resume scenario, the DownloadWorker that is cancelled on pause, may still write buffered data to the downloaded file. The new DownloadWorker may have already begun, setting an offset for download that was the file.length at the time. This value may end up being wrong if the previous worker flushes its last bits of data before cancelling. This would result in a section of the file that ends up duplicating contents.

This PR ensures that no buffered data is written to the file once the pause has been initiated.

- [ ] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/testutils/src/main/java/com/amplifyframework/testutils/random/RandomTempFile.java
+++ b/testutils/src/main/java/com/amplifyframework/testutils/random/RandomTempFile.java
@@ -78,10 +78,7 @@ public final class RandomTempFile extends File {
     private void createFile(long byteSize) throws IOException {
         deleteOnExit();
         try (RandomAccessFile raf = new RandomAccessFile(this, "rw")) {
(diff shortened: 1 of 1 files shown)


mattcreaser:
Should we also change the `ExistingWorkPolicy` to `REPLACE`?

tylerjroach:
> Should we also change the `ExistingWorkPolicy` to `REPLACE`?

I didn't see any behavioral difference when testing. I'm guessing this is because we manually cancel the task. Otherwise, it still had the same issue.

PR #3337, opened 2026-07-06
feat(analytics): add Event Enrichment client module

ekjotmultani:
## Summary

Adds a new `aws-event-enrichment` Kotlin module: a standalone, backend-agnostic client that attaches device, app, session, and identity context onto analytics events and serializes them to a structured analytics JSON envelope (`event_version` 3.1).

The module has zero AWS SDK dependencies. It does not talk to any AWS service directly; transport is delegated to an injectable `EventSink`, so applications can forward enriched events to Kinesis, Firehose, or any custom destination. This is the Android counterpart of the event enrichment client already available on Flutter.

This PR contains the package/library only. Unit tests are in a stacked PR so they can be reviewed independently.

## What it does

- `EventEnrichmentClient` auto-resolves the persistent client id from a `SharedPreferences` key shared across Amplify clients (`com.amplifyframework.device_id`, read-or-create) and resolves device metadata from `android.os.Build` and the default `Locale`. Callers provide only the app id and SDK metadata. When a full `AppMetadata` is supplied, its app id must agree with the app id argument.
- `SessionManager` tracks Active, Paused, and Stopped states with a configurable background timeout (default 5 seconds). It starts a fresh session on the next `record()` after a stop, so a stopped session is never reused.
- `EnrichedEvent.toJson` emits the envelope with the documented presence rules: optional fields are omitted when null, empty attribute and metric maps are omitted, and the device object is always present. Attributes are String-valued and metrics are Double-valued, with no length or count caps.
- `AndroidLifecycleObserver` drives session pause and resume from app foreground/background transitions using `ActivityLifecycleCallbacks`, avoiding an extra lifecycle dependency.

## Conventions

The module mirrors the `aws-kinesis` module conventions: same Gradle convention plugins, `POM_ARTIFACT_ID` in `gradle.properties`, a checked-in `api/aws-event-enrichment.api` for the binary compatibility validator, and registration in `settings.gradle.kts`. It builds clean under `ktlintCheck`, `apiCheck`, and `lintRelease`.

## Testing

Covered by the stacked tests PR (47 unit tests). See the follow-up PR based on this branch.


aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/GlobalFieldsManager.kt
```diff
@@ -0,0 +1,55 @@
...
+package com.amplifyframework.eventenrichment
+
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Manages global attributes and metrics that are stamped on every event.
+ *
+ * Values are in-memory only and not persisted between sessions.
+ */
+class GlobalFieldsManager {
```
mattcreaser: can be `internal`

aws-event-enrichment/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+apply(from = rootProject.file("configuration/checkstyle.gradle"))
+
+android {
+    namespace = "com.amplifyframework.eventenrichment"
+}
+
+dependencies {
+    implementation(project(":foundation"))
+
+    implementation(libs.androidx.appcompat)
```
mattcreaser: Can be removed

aws-event-enrichment/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    implementation(libs.kotlin.coroutines)
+    implementation(libs.kotlin.serializationJson)
+
+    testImplementation(libs.test.junit)
```
mattcreaser, line 36: If #3356 gets merged first please migrate to test bundle

aws-event-enrichment/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    testImplementation(libs.test.mockk)
+    testImplementation(libs.test.robolectric)
+    testImplementation(libs.test.androidx.junit)
+    testImplementation(libs.test.androidx.core)
+    testImplementation(libs.test.kotlin.coroutines)
+    testImplementation(libs.test.kotest.assertions)
+    testImplementation(libs.test.kotest.assertions.json)
+    testImplementation(project(":aws-event-enrichment"))
```
mattcreaser: Can be removed

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/metadata/DeviceMetadataProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Default [DeviceMetadataProvider] backed by [android.os.Build] and
+ * [java.util.Locale].
+ *
+ * Resolves the platform name, OS version, manufacturer, model, and locale so
+ * that events carry real device context. This is wired automatically by
+ * [com.amplifyframework.eventenrichment.EventEnrichmentClient]; there is no
+ * need to construct it directly outside of testing.
+ */
+class AndroidDeviceMetadataProvider : DeviceMetadataProvider {
```
mattcreaser: Can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/SessionManager.kt
```diff
@@ -0,0 +1,155 @@
...
+ * The clock ([now]), id generator ([generateId]), and [scheduler] are injectable
+ * so tests can drive session behavior deterministically.
+ *
+ * @param appId Application id used to build session ids.
+ * @param sessionTimeout Time the app may stay backgrounded before the session stops.
+ * @param now Clock supplying the current instant.
+ * @param generateId Supplier of unique ids for session id generation.
+ * @param scheduler Schedules the background pause timeout.
+ */
+class SessionManager(
```
mattcreaser: Can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/TimeoutScheduler.kt
```diff
@@ -0,0 +1,59 @@
...
+import kotlin.time.Duration
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+
+/** Handle for a scheduled timeout that can be cancelled before it fires. */
+fun interface TimeoutHandle {
```
mattcreaser: Everything in this file can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/lifecycle/LifecycleObserver.kt
```diff
@@ -0,0 +1,33 @@
...
+package com.amplifyframework.eventenrichment.lifecycle
+
+/**
+ * Abstract interface for observing app lifecycle transitions.
+ *
+ * Implement this to bridge platform-specific lifecycle events to the session
+ * manager. The default [AndroidLifecycleObserver] uses
+ * [android.app.Application.ActivityLifecycleCallbacks].
+ */
+interface LifecycleObserver {
```
mattcreaser: Can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/lifecycle/AndroidLifecycleObserver.kt
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Uses [Application.ActivityLifecycleCallbacks] and ref-counts started
+ * activities to detect when the whole app moves to the background (the last
+ * activity stops) or returns to the foreground (the first activity starts).
+ * This avoids a dependency on `androidx.lifecycle-process`.
+ *
+ * @param application Application to register lifecycle callbacks on.
+ * @param sessionManager Session manager driven by the lifecycle transitions.
+ */
+class AndroidLifecycleObserver(
```
mattcreaser: Can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/clientid/SharedPreferencesClientIdProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+ * generated, persisted, and returned.
+ *
+ * The storage key [CLIENT_ID_STORAGE_KEY] is a cross-package integration
+ * contract shared with the Connect client so that a device gets ONE id
+ * everywhere. Whichever client initializes first generates the UUID; the other
+ * reads it.
+ *
+ * @param context Android context used to access SharedPreferences.
+ */
+class SharedPreferencesClientIdProvider(context: Context) : ClientIdProvider {
```
mattcreaser: Can be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/SessionManager.kt
```diff
@@ -0,0 +1,155 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.eventenrichment.session
+
+import java.time.Duration as JavaDuration
+import java.time.Instant
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
```
mattcreaser: Want to avoid using any `java.time.*` because it requires desugaring, which we want to eventually get rid of.

Use `kotlin.time` instead for the Duration and Instant. For the formatting consider some alternatives, but the easiest is probably adding `kotlinx-datetime` as a dependency.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/clientid/SharedPreferencesClientIdProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+        val id = UUID.randomUUID().toString()
+        preferences.edit().putString(CLIENT_ID_STORAGE_KEY, id).apply()
+        return id
+    }
+
+    companion object {
+        /**
+         * SharedPreferences key for the persistent client/device ID.
+         *
+         * Shared with the Connect client so a device resolves to a single id
```
mattcreaser, line 64: This is a bit scary. We should consider migrating this class to a shared utility location.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/TimeoutScheduler.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    fun schedule(delay: Duration, action: () -> Unit): TimeoutHandle
+}
+
+/**
+ * Default [TimeoutScheduler] backed by a coroutine [delay] on the provided
+ * [scope]. The scope defaults to a supervised scope on [Dispatchers.Default].
+ */
+class CoroutineTimeoutScheduler(
+    private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
mattcreaser: This scope needs to be closed when the client closes. The scope should probably be owned by EventEnrichmentClient and injected as needed. Then when the client closes it cancels the scope.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+ * @param context Android context used to resolve device metadata, the client
+ *   id, and app lifecycle callbacks.
+ * @param appId Application identifier stamped on every event.
+ * @param sdkMetadata SDK metadata stamped on every event.
+ * @param options Configuration options with sensible defaults.
+ * @param sink Optional transport for enriched events.
+ * @param appMetadata Optional full application metadata. When provided, its
+ *   [AppMetadata.appId] must equal [appId].
+ */
+@OptIn(InternalAmplifyApi::class)
```
mattcreaser: This is applied globally across the project and can be removed

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+    )
+
+    private val logger: Logger = AmplifyLogging.logger<EventEnrichmentClient>()
+
+    /**
+     * The session manager for this client. Exposed for advanced integrations
+     * and platform lifecycle observers.
+     */
+    @get:VisibleForTesting
+    val sessionManager: SessionManager = sessionManager
```
mattcreaser: Should be `internal`

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+import com.amplifyframework.eventenrichment.metadata.AppMetadata
+import com.amplifyframework.eventenrichment.metadata.DeviceMetadata
+import com.amplifyframework.eventenrichment.metadata.DeviceMetadataProvider
+import com.amplifyframework.eventenrichment.metadata.SdkMetadata
+import com.amplifyframework.eventenrichment.session.SessionManager
+import com.amplifyframework.eventenrichment.session.SessionState
+import com.amplifyframework.foundation.logging.AmplifyLogging
+import com.amplifyframework.foundation.logging.Logger
+import com.amplifyframework.foundation.result.Result
+import java.time.Instant
```
mattcreaser: Use `kotlin.time` instead

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+            attributes = mergedAttributes,
+            metrics = mergedMetrics,
+            device = deviceMetadata,
+            app = appMetadata,
+            sdk = sdkMetadata,
+            clientId = clientId,
+            userId = userId
+        )
+
+        sink?.send(event)
```
mattcreaser: This is synchronous and blocking and is called from the caller's thread. This could introduce unintended performance hits into a customer's application.

Recommendation: launch a coroutine on the IO dispatcher (inject the dispatcher for tests) in the scope that we're going to add to this class (see other comments) and call it from there instead.

Also we need to catch any exception and log a warning if this callback fails.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+        attributes: Map<String, String> = emptyMap(),
+        metrics: Map<String, Double> = emptyMap()
+    ): Result<EnrichedEvent, EventEnrichmentException> {
+        if (closed) return Result.Failure(EventEnrichmentClosedException())
+
+        // A stopped session is still exposed by the manager for inspection, so
+        // start a fresh one instead of stamping the stopped session (which
+        // carries a stop_timestamp) onto a new event.
+        if (sessionManager.session == null || sessionManager.state == SessionState.STOPPED) {
+            sessionManager.startSession()
```
mattcreaser: `record` may be called often, and from different threads. There is a race condition here where the session may be started multiple times or may be ended by the time line 181 is hit.

For this I recommend adding a `ReentrantLock` to this class and injecting it into the SessionManager. Then add a function like this to the SessionManager and call it here:

```kotlin
  fun ensureActiveSession(): Session = lock.withLock {
      if (session == null || state == SessionState.STOPPED) startSession()  // startSession also takes lock (reentrant)
      session!!  // safe: we hold the lock, startSession just set it
  }
```

Ensure any mutations to the session within sessionManager acquire the lock.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
```diff
@@ -0,0 +1,256 @@
...
+     * @param attributes Per-event attributes merged over the global attributes.
+     * @param metrics Per-event metrics merged over the global metrics.
+     */
+    @JvmOverloads
+    fun record(
+        eventType: String,
+        attributes: Map<String, String> = emptyMap(),
+        metrics: Map<String, Double> = emptyMap()
+    ): Result<EnrichedEvent, EventEnrichmentException> {
+        if (closed) return Result.Failure(EventEnrichmentClosedException())
```
mattcreaser: There is also a race condition here where close could be called concurrently with record.

Recommend wrapping `close` using the same lock as in the other comment, and then wrapping the first part of this function. Do **not** call `sink.send` from within the lock.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/GlobalFieldsManager.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/SessionManager.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/TimeoutScheduler.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/lifecycle/LifecycleObserver.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/lifecycle/AndroidLifecycleObserver.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/clientid/SharedPreferencesClientIdProvider.kt
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Done.

aws-event-enrichment/build.gradle.kts
ekjotmultani: Done.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Removed.

aws-event-enrichment/build.gradle.kts
ekjotmultani: Removed. It was supplying `androidx.annotation` transitively so I added that one directly instead, same as aws-connect does.

aws-event-enrichment/build.gradle.kts
ekjotmultani: #3356 is still open so I left these as is. Happy to migrate once it lands.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/metadata/DeviceMetadataProvider.kt
ekjotmultani: Made `AndroidDeviceMetadataProvider` internal. The fun interface itself stays public since it's a public constructor parameter type.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/SessionManager.kt
ekjotmultani: Done, the module is java.time free now. Went with `kotlin.time` Duration/Instant and `kotlinx-datetime` 0.7.1 for formatting, which only pulls kotlin-stdlib transitively. Timestamp output is unchanged. One note: `kotlin.time` Instant/Clock are still `@ExperimentalTime` on Kotlin 2.2.0, so those files carry a file-level OptIn.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Done, covered with the SessionManager migration.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/clientid/SharedPreferencesClientIdProvider.kt
ekjotmultani: Agreed it's fragile. aws-connect ships its own internal copy of this same store, so I'd rather move both modules onto one shared synchronized utility in a dedicated PR than do it here, especially since a shared class can't stay module-internal and we'd need to pick where it lives. For this PR I've made it internal. Sound ok?

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/session/TimeoutScheduler.kt
ekjotmultani: Done. The client owns the scope and injects it into the scheduler, and `close()` cancels it.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Done. The callback now runs in a coroutine on an injected dispatcher (IO by default) inside the client scope, wrapped in a catch that logs a warning.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Added the `ReentrantLock` and `ensureActiveSession` exactly as you sketched. All session mutations in SessionManager take the lock now.

aws-event-enrichment/src/main/java/com/amplifyframework/eventenrichment/EventEnrichmentClient.kt
ekjotmultani: Done. `close` takes the same lock and `sink.send` stays outside it.

mattcreaser (approved):

PR #3338, opened 2026-07-06
feat(notifications): add backend-agnostic push notifications common module

ekjotmultani:
## Summary

Adds `aws-push-notifications-common`, a backend-agnostic library of Android push notification primitives. It is a Pinpoint-free successor to `aws-push-notifications-pinpoint-common`, providing the same building blocks without any coupling to Amazon Pinpoint or the AWS SDK.

The module provides:

- **Generic FCM payload parsing** (`PushNotificationPayload`): builds a displayable payload from a `NotificationPayload`, a raw FCM data map, a `RemoteMessage`, or an `Intent`, reading standard payload keys (title, body, imageUrl, url, deeplink, openApp, channelId, silentPush). It parses plain FCM pushes that carry no backend-specific keys and never drops them, so a valid non-Pinpoint push is still displayable.
- **Notification display** (`PushNotificationsUtils`): `showNotification` backed by `NotificationCompat`, plus `isAppInForeground` and `areNotificationsEnabled`.
- **Permissions** (`PushNotificationPermission`, `PermissionRequestResult`): `POST_NOTIFICATIONS` handling for API 33+.
- **Constants and channel helpers** (`PushNotificationsConstants`, `PushNotificationChannels`): the generic payload keys and a notification channel DSL.

The public `NotificationPayload` and `NotificationContentProvider` types continue to come from `common-core` and are re-exported, so consumers get them transitively.

## Why

This is a foundation library that other consumers depend on:

- The Amplify JavaScript React Native Android bridge currently depends on `aws-push-notifications-pinpoint-common`. This module gives it a Pinpoint-free replacement so it can migrate off Pinpoint with close to a dependency swap.
- The Flutter Android push layer, which today routes through the Pinpoint-specific parser and drops non-Pinpoint FCM pushes, can consume this instead to fix that gap.
- Native app developers who want the low level pieces directly.

The public symbol names and shapes are kept aligned with what these consumers already import so the migration stays small.

## Dependencies

Depends only on `common-core`, `annotations`, androidx, and `firebase-messaging`. No AWS SDK and no Pinpoint dependency.

## Testing

16 unit tests, all passing, including a regression guard that a payload with no Pinpoint keys is parsed and not dropped. Build, ktlintCheck, apiCheck, and lintRelease all pass against current `main` after rebase.

## Open questions for reviewers

- **Public vs internal API surface.** The successor API is exposed as a clean public API, whereas the original Pinpoint payload and constants were marked `@InternalAmplifyApi` (ERROR level opt-in). Making it public means consumers can drop their `@OptIn` annotations, but it also commits us to binary compatibility on these symbols. Happy to mark them internal instead if we would rather not commit yet.
- **`firebase-messaging` as an `api` dependency.** `PushNotificationPayload.fromRemoteMessage(RemoteMessage)` exposes a Firebase type, so `firebase-messaging` is currently an `api` dependency and leaks transitively. If we would rather keep Firebase off the public surface, I can drop that overload and keep only `fromData(Map<String, String>)`, which is what the RN bridge already uses.


soberm (commented):
Major finding:

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationsUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+     * @param notificationId a stable id for the notification
+     * @param payload the parsed notification content
+     * @param targetClass the Activity to launch when the notification is tapped; falls back to the
+     *   payload's own target class when provided
+     */
+    @Suppress("DEPRECATION")
+    @SuppressLint("NewApi")
+    fun showNotification(notificationId: Int, payload: PushNotificationPayload, targetClass: Class<*>?) {
+        CoroutineScope(Dispatchers.IO).launch {
+            val largeImageIcon = payload.imageUrl?.let { downloadImage(it) }
```
soberm, line 159: [major] `CoroutineScope(Dispatchers.IO)` with no `Job` or lifecycle attachment creates a scope that's never cancelled. If the context is destroyed while the image download is running, the coroutine keeps going with a captured `context` reference and then calls `notify()` on it. Pass a `CoroutineScope` into the constructor so callers can tie the lifetime to a `LifecycleOwner.lifecycleScope` or cancel it themselves:

```kotlin
class PushNotificationsUtils(
    private val context: Context,
    private val scope: CoroutineScope,
    private val channelId: String = ...
)
```

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationsUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+                NotificationCompat.Builder(context, payload.channelId ?: notificationChannel.id)
+            } else {
+                NotificationCompat.Builder(context)
+            }
+
+            builder.apply {
+                setContentTitle(payload.title)
+                setContentText(payload.body)
+                setSmallIcon(R.drawable.ic_launcher_foreground)
+                setContentIntent(pendingIntent)
```
soberm, line 180: [major] `R.drawable.ic_launcher_foreground` references this library's own resource, not the consuming app's. On API 26+ notification small icons must be monochrome/silhouette — adaptive launcher icons render as a solid white square in the status bar. There's also no way for library consumers to provide their own icon.

The small icon should be a constructor parameter:

```kotlin
class PushNotificationsUtils(
    private val context: Context,
    private val smallIconRes: Int,
    ...
)
```

soberm (commented):
Major finding:

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationsUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+                setPriority(NotificationCompat.PRIORITY_DEFAULT)
+                setLargeIcon(largeImageIcon)
+                setAutoCancel(true)
+                setStyle(NotificationCompat.BigTextStyle().bigText(payload.body))
+            }
+
+            with(NotificationManagerCompat.from(context)) {
+                notify(notificationId, builder.build())
+            }
```
soberm, line 189: [major] On API 33+, calling `notify()` without `POST_NOTIFICATIONS` permission silently drops the notification — no exception, no log. There's already an `areNotificationsEnabled()` helper on this class; worth calling it here and returning early so callers aren't left debugging phantom missing notifications. (Pre-existing behaviour carried over from `aws-push-notifications-pinpoint-common`, but worth fixing before this surface goes public.)

soberm (commented):
Major finding:

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationsUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+ * @param context an Android [Context]
+ * @param channelId the default notification channel id to use when a payload does not specify one
+ */
+class PushNotificationsUtils(
+    private val context: Context,
+    private val channelId: String = PushNotificationsConstants.DEFAULT_NOTIFICATION_CHANNEL_ID
+) {
+    init {
+        retrieveNotificationChannel()
+    }
```
soberm, line 65: [nit] `retrieveNotificationChannel()` returns `NotificationChannel?` and the result is silently dropped here. Since the channel is re-fetched on every `showNotification()` call anyway, this eager call is mostly a side-effect trigger — a debug log on null return would make channel-creation failures visible during integration testing.

soberm (commented):
Another finding:

soberm (commented):
Additional findings:

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/permissions/PermissionRequestChannel.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Communication channel for relaying permission request results from the [PermissionsRequestActivity] to
+ * the [PushNotificationPermission] utility class.
+ */
+internal object PermissionRequestChannel {
+    private class IdAndResult(val requestId: String, val result: PermissionRequestResult)
+
+    private val flow = MutableSharedFlow<IdAndResult>(
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST
+    )
```
soberm, line 37: [minor] With `extraBufferCapacity = 1` and `DROP_OLDEST`, if two `requestPermission()` calls are in-flight simultaneously and both activities post results before either collector runs, the first result is dropped and the corresponding `listen().first()` suspends forever (no timeout, no cancellation signal). Bumping the buffer capacity or adding a `withTimeout` on the `first()` call side would close this window.

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationPayload.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        private fun parseAction(data: Map<String, String>): Map<String, String> {
+            val action = mutableMapOf<String, String>()
+            data[PushNotificationsConstants.OPEN_APP]?.let { action[PushNotificationsConstants.OPEN_APP] = it }
+            data[PushNotificationsConstants.URL]?.let {
+                // force HTTPS URL scheme
+                action[PushNotificationsConstants.URL] = it.replaceFirst("http://", "https://")
+            }
+            data[PushNotificationsConstants.DEEPLINK]?.let { action[PushNotificationsConstants.DEEPLINK] = it }
+            return action
```
soberm, line 134: [minor] `replaceFirst("http://", "https://")` mutates the sender-provided URL silently. If the target is an intranet address that only speaks HTTP, the tap action will fail quietly. Also `replaceFirst` matches on substring position rather than the URI scheme component — `URI.scheme` would be more defensive. Worth a note in the `PushNotificationsConstants.URL` KDoc so senders know the library enforces HTTPS.

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/PushNotificationChannels.kt
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+    var showBadge: Boolean = true
+        set(value) {
+            field = value
+            builder.setShowBadge(value)
+        }
+
+    var sound: Uri? = Settings.System.DEFAULT_NOTIFICATION_URI
+        set(value) {
```
soberm, line 133: [nit] `sound` and `audioAttributes` are mutually dependent: each setter calls `builder.setSound(sound, audioAttributes)` using the *current* value of the other. The interaction is subtle: a caller who sets `audioAttributes` first then `sound = null` gets the right result; the reverse order leaves the builder with the default sound URI even though silence was intended. A brief comment spelling out the dependency would save the next reader a trip through both setters.

aws-push-notifications-common/src/main/java/com/amplifyframework/notifications/pushnotifications/permissions/PushNotificationPermission.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Opens the application's settings page, where the user can manually grant/revoke application permissions
+     */
+    fun openSettings() {
+        // Build a uri like "package:com.example.myapplication". See docs for ACTION_APPLICATION_DETAILS_SETTINGS.
+        val uri = Uri.fromParts("package", context.packageName, null)
+        val intent = Intent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS, uri)
+        context.startActivity(intent)
+    }
```
soberm, line 80: [major] `startActivity()` from a non-Activity `Context` (service, application context, background receiver) requires `Intent.FLAG_ACTIVITY_NEW_TASK` — without it this crashes at runtime with `AndroidRuntimeException`. Since `PushNotificationPermission` accepts a bare `Context`, this is reachable in practice. `requestPermission()` correctly sets the flag on its intent; `openSettings()` needs the same:

```kotlin
intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
context.startActivity(intent)
```

(Pre-existing in `aws-push-notifications-pinpoint-common`, but a public API is a good opportunity to fix it.)

soberm (commented):
## Parity & integration notes

This is a well-scoped foundation layer and the Pinpoint-free design is the right direction. A few behavioral changes and one structural concern are worth flagging before this gets consumed by downstream packages.

**`body` / `message` key preference flip (silent behavioral change)**
The old parser preferred `data["message"]` and fell back to `data["pinpoint.notification.body"]`. The new parser prefers `data["body"]` and falls back to `data["message"]`. For any sender that includes *both* keys in the same payload the displayed body will silently change after migration. Low probability in the wild, but worth a note in the migration guide.

**Default channel ID changed: `"PINPOINT.NOTIFICATION"` → `"amplify.notifications"`**
On devices that have already configured per-channel notification preferences (muted, custom sound, etc.) for the old channel, those settings will be lost silently on migration — Android channel prefs are keyed by ID with no migration path. Worth calling out in the changelog/migration guide so app developers can decide whether to keep the old channel alive as an alias.

**`notificationId` deduplication will degrade for Pinpoint-originated pushes**
The RN bridge currently computes a stable notification ID by hashing `pinpoint.campaign.campaign_id + campaign_activity_id` (or journey equivalents), which prevents duplicate notifications from the same campaign activity stacking up. Those keys are not present in this module. Once the bridge migrates, campaign/journey pushes will get random IDs on every delivery, so a second delivery of the same campaign activity will show two notifications instead of replacing the first. If deduplication matters for downstream consumers, the new module (or the bridge migration PR) should define a `notificationId` strategy — either a standard `"notificationId"` FCM data key, or a hash of title+body.

**Four `[major]` items in `PushNotificationsUtils`** — see inline comments. The `openSettings()` `FLAG_ACTIVITY_NEW_TASK` gap and the missing `areNotificationsEnabled()` pre-check are pre-existing bugs carried over from `aws-push-notifications-pinpoint-common` (not regressions introduced here), but since this is now a public API it's a good opportunity to fix them before they reach more consumers.

PR #3339, opened 2026-07-06
test: add Event Enrichment client unit tests

ekjotmultani:
## Summary

Unit tests for the `aws-event-enrichment` module. These are split into their own branch stacked on the package branch so the implementation and its tests can be reviewed independently

Based this on `feat/event-enrichment-client`

## Coverage

47 tests across the module:

- Envelope parity: `EnrichedEvent.toJson` emits the full `event_version` 3.1 envelope and honors the presence rules (optional fields omitted, empty attribute and metric maps omitted, device object always present, platform and locale sub-objects omitted when their contents are null, `arrival_timestamp` mirrors `event_timestamp`).
- Fresh session after stop: a stopped session is never reused. The next `record()` starts a new session with no stop metadata. Verified at both the `SessionManager` and the client level.
- Client id: `SharedPreferencesClientIdProvider` read-or-create semantics under the shared `com.amplifyframework.device_id` key, including reading an id written by another client.
- App id assertion: a mismatched `AppMetadata` app id is rejected at construction.
- No caps: large attribute and metric counts and long values pass through unchanged.
- Session lifecycle: session id format, short app id padding, pause and resume within the timeout, timeout expiry, and start-over after timeout, driven by an injected clock and scheduler.
- Device metadata: `AndroidDeviceMetadataProvider` resolves real values rather than an empty object.
- Options: defaults (auto session tracking on, 5 second timeout) and builder overrides.

All tests pass. The module is clean under `ktlintCheck`.

### DIFF
--- a/aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/EventEnrichmentClientTest.kt
+++ b/aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/EventEnrichmentClientTest.kt
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.eventenrichment
+
+import com.amplifyframework.eventenrichment.exception.EventEnrichmentClosedException
+import com.amplifyframework.eventenrichment.metadata.AppMetadata
+import com.amplifyframework.eventenrichment.metadata.DeviceMetadata
+import com.amplifyframework.eventenrichment.metadata.SdkMetadata
+import com.amplifyframework.eventenrichment.session.SessionManager
+import com.amplifyframework.eventenrichment.session.SessionState
+import com.amplifyframework.foundation.result.Result
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.nulls.shouldNotBeNull
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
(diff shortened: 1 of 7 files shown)


aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/session/SessionManagerTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        var cancelled = false
+
+        override fun schedule(delay: Duration, action: () -> Unit): TimeoutHandle {
+            scheduledAction = action
+            cancelled = false
+            return TimeoutHandle { cancelled = true }
+        }
+
+        fun fireTimeout() {
+            scheduledAction?.invoke()
```
mattcreaser: Perhaps assert `cancelled != null` here before firing the action so that you don't accidentally fire it in an invalid situation.

aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/EventEnrichmentClientTest.kt
```diff
@@ -0,0 +1,193 @@
...
+    fun `record starts a fresh session after a stop instead of reusing the stopped one`() {
+        val client = client(autoSessionTracking = true)
+
+        val first = (client.record("first") as Result.Success).data
+        val firstSessionId = first.session.id
+
+        client.stopSession()
+        clock = clock.plusMillis(1_000)
+
+        val second = (client.record("second") as Result.Success).data
```
mattcreaser: In this file instead of `(client.record(...) as Result.Success).data` use `client.record(...).get()`

aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/session/SessionManagerTest.kt
ekjotmultani: Done, with one tweak: `cancelled` is a non-nullable Boolean in the fake, so the thing that actually guards an invalid fire is `scheduledAction`. The assert is `scheduledAction.shouldNotBeNull()` before invoking.

aws-event-enrichment/src/test/java/com/amplifyframework/eventenrichment/EventEnrichmentClientTest.kt
ekjotmultani: Done. Swept all 7 in this file plus 3 more of the same pattern in AutoResolveTest.

mattcreaser (dismissed):

mattcreaser (approved):

PR #3341, opened 2026-07-13
feat(notifications): add Amplify Connect client for Customer Profiles

ekjotmultani:
## Summary

Adds the Amplify Connect client for Android (`aws-connect` module) for the Pinpoint deprecation effort, at parity with the Flutter client. It identifies users against the backend-notifications construct's Lambda endpoints and registers devices for push delivery through Customer Profiles. The public surface is a single `identifyUser` call; device registration folds into the identify options, and the client auto-fills a stable device id (plain SharedPreferences under `com.amplifyframework.device_id`, shared with the event enrichment client) plus the platform.

## What's included

- Authenticated identify: OkHttp `POST /identify-user` with the Cognito access token (Bearer), profile keyed on the Cognito sub
- Guest identify: `POST /identify-user-guest` with inline SigV4 signing for unauthenticated Identity Pool users, keyed on identity id, with merge on sign-in
- Device registration through identify options, with platform and device id auto-population
- Reads endpoint and region from the `analytics.amazon_connect_customer_profiles` configuration
- Unit tests included in this PR (the module is new, so its tests ship alongside it)

## Testing

32 unit tests pass; all Gradle checks are green.

## Backend configuration

This client reads its endpoint and region from the `notifications.amazon_connect_customer_profiles` section of `amplify_outputs.json` (fields `endpoint` and `aws_region`), which is what the backend-notifications construct writes and what amplify-js parses. Guest and authenticated profiles are treated as separate by the backend, so the earlier guest identity linking option has been removed from the identify options.

## Why HTTP instead of the AWS SDK

This client intentionally uses OkHttp rather than an AWS SDK service client. All calls go to the REST API that the backend-notifications construct deploys in the customer's account (an API Gateway endpoint vended through `amplify_outputs.json`), not to an AWS service, so no SDK client models it. The authenticated path is a plain Bearer token header, and the guest path uses a small SigV4 implementation for `execute-api` signing (the smithy-kotlin signer is marked internal, so the module does not depend on it). Routing through the backend also keeps Customer Profiles IAM permissions in the backend Lambda instead of on customer devices, which was the reason an earlier direct-to-service design using the AWS SDK Kotlin CustomerProfiles client was dropped. The JS provider uses the same pattern (plain fetch to the construct endpoint).


aws-connect/src/main/java/com/amplifyframework/connect/ConnectClientConfiguration.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    companion object {
+        /**
+         * Parses configuration from the `custom.CustomerProfiles` section of a
+         * decoded amplify_outputs map.
+         *
+         * Expects:
+         * ```json
+         * {
+         *   "custom": {
```
thisisabhash: needs update to the newer amplify_outputs format

aws-connect/src/main/java/com/amplifyframework/connect/IdentifyUserOptions.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+data class IdentifyUserOptions(
+    val userAttributes: Map<String, List<String>>? = null,
+    val address: String? = null,
+    val deviceId: String? = null,
+    val channelType: ChannelType? = null,
+    val platform: String? = null,
+    val appVersion: String? = null,
+    val optOut: OptOut? = null,
+    val previousGuestIdentityId: String? = null
```
thisisabhash: renamed to `guestIdentityId`

thisisabhash (dismissed):

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
```diff
@@ -0,0 +1,159 @@
...
+ * @param credentialsProvider Resolves AWS credentials for SigV4 signing
+ * @param deviceIdStore Persistent device id store (shared key with enrichment)
+ * @param platform Client OS platform (e.g. "Android")
+ * @param appVersion Application version string
+ * @param channelType Push channel type for this device
+ */
+class AmplifyConnectClient(
+    configuration: ConnectClientConfiguration,
+    private val credentialsProvider: ConnectCredentialsProvider,
+    context: android.content.Context,
```
mattcreaser: Add import for Context, and Context should always be the first parameter in Android

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * POST /remove-device with the given [body], SigV4-signed with [credentials].
+     */
+    suspend fun removeDevice(credentials: AwsCredentials, body: String) {
+        sendSigV4("$endpoint$REMOVE_DEVICE_PATH", credentials, body)
+    }
+
+    private suspend fun sendSigV4(url: String, credentials: AwsCredentials, body: String) =
```
mattcreaser: There are now many different implementations of sigv4 logic in this codebase. We don't have to address that in this PR but I'm flagging as a followup task to consolidate them into one.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
```diff
@@ -0,0 +1,159 @@
...
+ * implemented here.
+ *
+ * @param configuration Endpoint and region from amplify_outputs
+ * @param credentialsProvider Resolves AWS credentials for SigV4 signing
+ * @param deviceIdStore Persistent device id store (shared key with enrichment)
+ * @param platform Client OS platform (e.g. "Android")
+ * @param appVersion Application version string
+ * @param channelType Push channel type for this device
+ */
+class AmplifyConnectClient(
```
mattcreaser, line 62: Should we release this initially as an `@AmplifyExperimentalApi`?

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
```diff
@@ -0,0 +1,159 @@
...
+     * @throws AmplifyConnectException for endpoint errors
+     */
+    suspend fun removeDevice() {
+        val credentials = credentialsProvider.resolve()
+        val deviceId = deviceIdStore.getOrCreate()
+        val body = buildJsonObject {
+            put("deviceId", deviceId)
+        }
+        service.removeDevice(credentials, body.toString())
+        logger.info { "removeDevice sent" }
```
mattcreaser: These logs can be modified, since they will only be emitted if the request was successful (the service throws if not)

aws-connect/src/main/java/com/amplifyframework/connect/ConnectClientConfiguration.kt
```diff
@@ -0,0 +1,96 @@
...
+         *   }
+         * }
+         * ```
+         *
+         * @throws ConnectConfigurationException if the section or either required
+         *   field is missing
+         */
+        @JvmStatic
+        fun fromAmplifyOutputs(amplifyOutputs: Map<String, Any?>): ConnectClientConfiguration {
+            val notifications = amplifyOutputs["notifications"]
```
mattcreaser: This should be using `AmplifyOutputsData` and not directly parsing JSON.

mattcreaser: To update `AmplifyOutputs` to match the 1.5 spec we can use the following shape:

```kotlin
  @Serializable
  data class Notifications(
      val awsRegion: String? = null,
      val amazonPinpointAppId: String? = null,
      val channels: List<AmazonPinpointChannels> = emptyList(),
      val amazonConnect: AmazonConnect? = null
  ) {
      @Serializable
      data class AmazonConnect(
          val awsRegion: String,
          val endpoint: String
      )
  }
```

aws-connect/src/main/java/com/amplifyframework/connect/UserProfile.kt
```diff
@@ -0,0 +1,47 @@
...
+ * @param name Display name
+ * @param phone Phone number
+ * @param customAttributes Additional key-value string attributes
+ * @param location Geographic location attributes
+ */
+data class UserProfile(
+    val email: String? = null,
+    val name: String? = null,
+    val phone: String? = null,
+    val customAttributes: Map<String, String>? = null,
```
mattcreaser: Use empty collection instead of a null collection
```suggestion
    val customAttributes: Map<String, String> = emptyMap(),
```

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * The [deviceId] is resolved from the shared device-id store.
+     *
+     * @throws ConnectNotSignedInException if credentials cannot be resolved
+     * @throws ConnectNetworkException on transport failure
+     * @throws AmplifyConnectException for endpoint errors
+     */
+    suspend fun removeDevice() {
+        val credentials = credentialsProvider.resolve()
+        val deviceId = deviceIdStore.getOrCreate()
```
mattcreaser: This should just be `get()` so we don't create a new UUID when trying to remove the device.

aws-connect/src/main/java/com/amplifyframework/connect/internal/DeviceIdStore.kt
```diff
@@ -0,0 +1,81 @@
...
+ */
+internal class DeviceIdStore(context: Context) {
+
+    private val prefs: SharedPreferences = context.applicationContext
+        .getSharedPreferences(PREFERENCES_NAME, Context.MODE_PRIVATE)
+
+    /**
+     * Returns the persisted device ID, or generates and persists a new one.
+     */
+    fun getOrCreate(): String {
```
mattcreaser: All of the prefs access should be main safe:
```suggestion
    suspend fun getOrCreate(): String = withContext(dispatchers.IO) {
```

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
mattcreaser: Also drop these logs to `debug`

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Sends user profile attributes to the Customer Profiles endpoint.
+     *
+     * POST /identify-user with body `{ userProfile }`.
+     *
+     * @param userProfile User profile attributes to send
+     * @throws ConnectNotSignedInException if credentials cannot be resolved
+     * @throws ConnectNetworkException on transport failure
+     * @throws AmplifyConnectException for endpoint errors
+     */
+    suspend fun identifyUser(userProfile: UserProfile) {
```
mattcreaser: Each of these APIs should return a result with an empty data type:

```kotlin
class IdentifyUserData internal constructor()
typealias IdentifyUserResult = Result<IdentifyUserData, AmplifyConnectException>

// ...
suspend fun identifyUser(userProfile: UserProfile): IdentifyUserResult
```

This is to allow the APIs to evolve over time without breaking changes. See KinesisClient for examples.

mattcreaser: Add a helper like this:

```kotlin
  private suspend fun <T> connectResult(
      block: suspend () -> T
  ): Result<T, AmplifyConnectException> =
      try {
          Result.Success(block())
      } catch (e: CancellationException) {
          throw e                                    // MUST propagate — never becomes a Failure
      } catch (e: AmplifyConnectException) {
          Result.Failure(e)                          // already-typed, expected failures
      } catch (e: Exception) {
          Result.Failure(                            // last-resort: unforeseen bug, don't leak raw
              ConnectServiceException(
                  detail = "Unexpected error calling AWS Connect",
                  cause = e
              )
          )
      }
```

and use it for each function:

```kotlin
suspend fun identifyUser(
      userProfile: UserProfile
  ): IdentifyUserResult = connectResult {
      // ... existing body
      IdentifyUserData()  
  }
```

aws-connect/src/main/java/com/amplifyframework/connect/internal/InputValidation.kt
```diff
@@ -0,0 +1,57 @@
...
+ */
+package com.amplifyframework.connect.internal
+
+import com.amplifyframework.connect.ConnectValidationException
+import com.amplifyframework.connect.UserProfile
+import com.amplifyframework.connect.UserProfileLocation
+
+/**
+ * Client-side input length validation mirroring backend limits (construct head
+ * aeffa206). MAX_ATTRIBUTE_LENGTH = 255 on all string fields.
```
mattcreaser: This comment can be cleaned up to remove the commit sha

aws-connect/src/main/java/com/amplifyframework/connect/internal/DeviceIdStore.kt
mattcreaser: And inject the dispatcher in the constructor for testing, see e.g. `AutoFlushScheduler` for an example.

aws-connect/src/main/java/com/amplifyframework/connect/ChannelType.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+enum class ChannelType(val value: String) {
+    /** Firebase Cloud Messaging (Android). */
+    GCM("GCM"),
+
+    /** Apple Push Notification Service. */
+    APNS("APNS"),
+
+    /** Apple Push Notification Service (Sandbox). */
+    APNS_SANDBOX("APNS_SANDBOX")
```
mattcreaser, line 31:
```suggestion
    /** Firebase Cloud Messaging (Android). */
    Gcm("GCM"),

    /** Apple Push Notification Service. */
    Apns("APNS"),

    /** Apple Push Notification Service (Sandbox). */
    ApnsSandbox("APNS_SANDBOX")
```

aws-connect/src/main/java/com/amplifyframework/connect/ChannelType.kt
```diff
@@ -0,0 +1,29 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.connect
+
+/**
+ * Push notification channel type for device registration.
+ */
+enum class ChannelType(val value: String) {
```
mattcreaser:
```suggestion
enum class ChannelType(internal val value: String) {
```

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
```diff
@@ -0,0 +1,204 @@
...
+                .addHeader("Authorization", authHeader)
+                .post(body.toRequestBody(JSON_MEDIA_TYPE))
+
+            if (credentials is AwsCredentials.Temporary) {
+                requestBuilder.addHeader("X-Amz-Security-Token", credentials.sessionToken)
+            }
+
+            val response = try {
+                httpClient.newCall(requestBuilder.build()).execute()
+            } catch (e: Exception) {
```
mattcreaser: This broad `catch (e: Exception)` will swallow `CancellationException`: if the coroutine is cancelled mid-request, OkHttp's `execute()` throws an `IOException` from the interrupt, and we rethrow it wrapped as `ConnectNetworkException` — so a cancelled call surfaces as a bogus network failure instead of propagating cancellation.

Suggest narrowing the catch and re-checking cancellation:
```kotlin
} catch (e: IOException) {
    currentCoroutineContext().ensureActive() // rethrows CancellationException if we were cancelled
    throw ConnectNetworkException(cause = e)
}
```
`ensureActive()` is a no-op when active, so genuine network failures still map to `ConnectNetworkException`. `execute()` only throws `IOException`, so narrowing from `Exception` is safe here too.

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * All routes are SigV4-signed for service `execute-api`. The backend Lambda
+ * derives the principal identity from the signer (Cognito sub for authed
+ * users, identityId for guests).
+ */
+internal class ConnectService(
+    private val endpoint: String,
+    private val region: String,
+    private val httpClient: OkHttpClient = OkHttpClient()
+) {
```
mattcreaser, line 61:
```suggestion
internal class ConnectService(
    private val endpoint: String,
    private val region: String,
    private val httpClient: OkHttpClient = OkHttpClient(),
    private val dispatcher = Dispatchers.IO
) {
```

Inject the dispatcher to use when calling `withContext` so that tests stay on the test dispatcher.

aws-connect/src/test/java/com/amplifyframework/connect/internal/ConnectServiceTest.kt
```diff
@@ -0,0 +1,164 @@
...
+    @Test
+    fun `identifyUser sends POST to identify-user with SigV4`() = runTest {
+        server.enqueue(MockResponse().setResponseCode(200).setBody("{}"))
+
+        service.identifyUser(testCredentials, """{"userProfile":{"name":"Alice"}}""")
+
+        val request = server.takeRequest()
+        request.method shouldBe "POST"
+        request.path shouldBe "/identify-user"
+        request.getHeader("Authorization")!! shouldContain "AWS4-HMAC-SHA256"
```
mattcreaser, line 74: These tests only assert the `Authorization` header *contains* `AWS4-HMAC-SHA256` — they never verify the computed signature. MockWebServer accepts any signature, so a bug in the hand-rolled signing (canonical request construction, string-to-sign, signing-key derivation, header ordering) would pass all of these.

Since the SigV4 implementation here is hand-rolled, consider adding a known-answer test: fixed credentials + a pinned timestamp + a fixed body, asserting the exact expected `Signature=` value. Pinning the clock will require making the timestamp injectable (it currently calls `ZonedDateTime.now(...)` internally). A single KAT vector would lock down the signing math and catch regressions the current `shouldContain` assertions can't.

mattcreaser (commented):
Looking good overall, a variety of minor feedback below.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
ekjotmultani: Done, imported `Context` and moved it to the first parameter.

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
ekjotmultani: Agreed, good followup. I'll track it separately so we can consolidate the sigv4 implementations across the repo.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
ekjotmultani: Good idea, done. Annotated the client and its public types `@ExperimentalAmplifyApi` for the initial release so we can evolve the surface without a breaking change.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
ekjotmultani: Done, reworded them and dropped to `debug`.

aws-connect/src/main/java/com/amplifyframework/connect/ConnectClientConfiguration.kt
ekjotmultani: Done. Config parses through `AmplifyOutputsData` now instead of raw JSON, with your `amazonConnect` shape added to `Notifications`. I also made the Pinpoint fields nullable so a Connect-only outputs file deserializes, since that's what the Connect backend actually emits. Added a test for the Connect-only case.

aws-connect/src/main/java/com/amplifyframework/connect/UserProfile.kt
ekjotmultani: Done, thanks.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
ekjotmultani: Good catch, done. `removeDevice` reads the persisted id and returns early when there isn't one, so it won't mint a UUID just to remove it.

aws-connect/src/main/java/com/amplifyframework/connect/internal/DeviceIdStore.kt
ekjotmultani: Done. Moved the `withContext(dispatcher)` into `getOrCreate`/`get`/`clear` and injected the dispatcher through the constructor, following `AutoFlushScheduler`.

aws-connect/src/main/java/com/amplifyframework/connect/internal/InputValidation.kt
ekjotmultani: Done.

aws-connect/src/main/java/com/amplifyframework/connect/ChannelType.kt
ekjotmultani: Done, thanks.

aws-connect/src/main/java/com/amplifyframework/connect/AmplifyConnectClient.kt
ekjotmultani: Done, went with this. Each method returns a `Result` with an empty data type now (`IdentifyUserData` and friends), and I added the `connectResult { }` helper for the mapping, following KinesisClient. It rethrows `CancellationException`, maps `AmplifyConnectException` to `Failure`, and wraps anything else in `ConnectServiceException`.

aws-connect/src/main/java/com/amplifyframework/connect/ChannelType.kt
ekjotmultani: Done, `value` is `internal` now, and I regenerated the api golden to drop the public `getValue()` so apiCheck stays green.

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
ekjotmultani: Good catch. Narrowed the catch to `IOException` and added `currentCoroutineContext().ensureActive()` before wrapping, so a cancelled request propagates `CancellationException` instead of surfacing as a bogus network failure.

aws-connect/src/main/java/com/amplifyframework/connect/internal/ConnectService.kt
ekjotmultani: Done, `ConnectService` takes an injected `dispatcher` (defaults to `Dispatchers.IO`) and uses it in `withContext`.

aws-connect/src/test/java/com/amplifyframework/connect/internal/ConnectServiceTest.kt
ekjotmultani: Added one. Pulled the signing into an internal pure `buildAuthorizationHeader(...)` with an injectable timestamp, and added a known-answer test with fixed credentials, a pinned timestamp, and a fixed body asserting the exact `Signature=`. The expected value was computed independently with a Python SigV4 reference rather than captured from our own output, so it locks down the canonical-request and signing-key math instead of just the header shape.

mattcreaser (dismissed):

mattcreaser (approved):

PR #3376, opened 2026-08-13
chore: add v3 cloudwatch client public API definitions

thisisabhash:

*Issue #, if available:*

N/A

*Description of changes:*

Scaffolds a new standalone `:aws-cloudwatch` module (package `com.amplifyframework.cloudwatch`) with the **public API shells** for the v3 CloudWatch client, following the same standalone-client pattern as `:aws-kinesis` (Kotlin, coroutines, `Result<T, E>`, builder-style options, `AmplifyFoundation`/`foundation-bridge` only — no `:core`/`:aws-core` dependency).

This is a shells-only PR — public types and signatures are defined to match the approved API Review; method bodies are `TODO()`. Internal implementation (rotating file store, batching/consumer, filtering, flush monitor, session management) will follow in later PRs, largely in the shared module described in the design's dev strategy.

Added:
- `AmplifyCloudWatchClient` — implements `LogSink`; `events: SharedFlow<LoggingEvent>`, `isEnabledFor`, `emit`, `enable`, `disable`, `setUserIdentifier`, `setLoggingConstraints`, `suspend flushLogs(): Result<FlushData, AmplifyCloudWatchException>`, `getCloudWatchLogsClient()`.
- `AmplifyCloudWatchClientOptions` — data class + no-arg `builder()` / `invoke { }` DSL (`logGroupName`, `localStoreMaxSizeInMB` = 5, `flushStrategy`, `loggingConstraints`, `configureClient`), plus the `CloudWatchLogsClientConfigurationProvider` typealias.
- `FlushStrategy` — sealed class with `Interval(Duration = 60.seconds)` (no `None`; `disable()` covers that case).
- `LoggingConstraints` / `UserLogLevel` — `defaultLogLevel`, `namespaceLogLevel`, `userLogLevel` (v2's category-level filtering maps to namespace-level).
- `LoggingEvent` — sealed class `WriteLogFailure` / `FlushLogFailure` (`context`, `cause`).
- `FlushData` — `success: Boolean = true`.
- `AmplifyCloudWatchException` — sealed hierarchy (`Storage` / `Service` / `Configuration` / `Unknown`) over `AmplifyException`.
- Module scaffold: `build.gradle.kts`, `gradle.properties`, `.gitignore`, and registration in `settings.gradle.kts`.

`./gradlew :aws-cloudwatch:compileDebugKotlin` — compiles successfully. No behavior to unit-test yet (all method bodies are `TODO()`).

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests


aws-cloudwatch/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+plugins {
+    alias(libs.plugins.amplify.android.library)
+    alias(libs.plugins.kotlin.serialization)
```
mattcreaser: This plugin is unused — there's no `@Serializable` type in the module. Same for `libs.kotlin.serializationJson` on line 35. Worth dropping both unless the `@Serializable` configuration type is landing in a follow-up.

aws-cloudwatch/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+dependencies {
+    implementation(project(":foundation"))
+    implementation(project(":foundation-bridge"))
+
+    implementation(libs.androidx.appcompat)
+    implementation(libs.aws.cloudwatchlogs)
+    implementation(libs.aws.http)
+    implementation(libs.kotlin.serializationJson)
```
mattcreaser: These need to be `api`, not `implementation` — public signatures expose types from all three of these dependencies:

- `:foundation` — `LogSink`, `LogLevel`, `LogMessage`, `AwsCredentialsProvider<AwsCredentials>`, `Result`, `AmplifyException`
- `:foundation-bridge` — `SdkClientConfigurationProvider`, the target of the `CloudWatchLogsClientConfigurationProvider` typealias
- `libs.aws.cloudwatchlogs` — `CloudWatchLogsClient` returned by `getCloudWatchLogsClient()`, and `Config.Builder` in the typealias

As written, a consumer of the published AAR can't name the return type of `flushLogs()` or `getCloudWatchLogsClient()`. Compare `aws-kinesis/build.gradle.kts`, which uses `api(...)` for all of these.

Also missing `api(platform(libs.aws.bom))`. `aws-cloudwatchlogs` is declared version-less in the catalog and takes its version from the BOM — right now it resolves only because `foundation-bridge` exposes the platform transitively. Every other AWS-consuming module declares it explicitly.

Minor, while you're here: `libs.androidx.appcompat` (line 32) is unused.

aws-cloudwatch/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+    androidTestImplementation(project(":testutils"))
+    androidTestImplementation(libs.test.androidx.core)
+    androidTestImplementation(project(":core"))
+    androidTestImplementation(project(":aws-core"))
+    androidTestImplementation(project(":aws-auth-cognito"))
+    androidTestImplementation(libs.test.androidx.runner)
+    androidTestImplementation(libs.test.kotlin.coroutines)
+    androidTestImplementation(libs.test.androidx.junit)
+    androidTestImplementation(libs.test.kotest.assertions)
+    androidTestImplementation(project(":aws-cloudwatch"))
```
mattcreaser: A few cleanups here:

- `testImplementation(project(":aws-cloudwatch"))` (line 45) and `androidTestImplementation(project(":aws-cloudwatch"))` (line 56) are self-dependencies — test source sets already see main. Remove both.
- Prefer `libs.bundles.test.unit` / `libs.bundles.test.unit.android` over enumerating the individual test deps, matching `aws-kinesis` and `aws-event-enrichment`.
- The `androidTestImplementation` block pulls `:core`, `:aws-core`, and `:aws-auth-cognito` — the exact dependencies this module exists to avoid — with no `androidTest` sources present. Drop it until there are instrumentation tests (which also need registering in `configuration/instrumentation-tests.gradle`).

aws-cloudwatch/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=aws-cloudwatch
```
mattcreaser, line 1: `aws-cloudwatch` is too generic for a logs-only client — it claims the whole CloudWatch namespace (metrics, alarms, RUM, Logs Insights), sitting next to the existing `aws-logging-cloudwatch` v2 artifact. The approved design uses `AmplifyCloudWatchLoggingClient` / `:aws-logging-cloudwatch-client`.

Since a published artifact id is effectively permanent, let's settle this before merge — it also affects the class names (`AmplifyCloudWatchClient`, `AmplifyCloudWatch*Exception`) and the package (`com.amplifyframework.cloudwatch`).

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/FlushStrategy.kt
```diff
@@ -0,0 +1,35 @@
...
+ * [AmplifyCloudWatchClient.disable] instead.
+ */
+@ExperimentalAmplifyApi
+sealed class FlushStrategy {
+    /**
+     * Automatically flush at a regular interval.
+     *
+     * @param interval Time between automatic flush operations. Defaults to 60 seconds.
+     */
+    data class Interval(val interval: Duration = 60.seconds) : FlushStrategy()
```
mattcreaser, line 31: Please add a `None` case, matching `com.amplifyframework.recordcache.FlushStrategy`. The KDoc rationale — "automatic flushing is turned off via `disable()` instead" — isn't equivalent: `disable()` stops log capture *and* flushing, so there's no way to keep buffering locally and flush only on demand. That paragraph should come out along with the change.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/FlushData.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import com.amplifyframework.annotations.ExperimentalAmplifyApi
+
+/**
+ * The successful result of a [AmplifyCloudWatchClient.flushLogs] call.
+ *
+ * @param success Whether the flush completed successfully
+ */
+@ExperimentalAmplifyApi
+data class FlushData(val success: Boolean = true)
```
mattcreaser: Reserving a domain type here is right — going from `Unit` to a data type later would be breaking. But drop the `success` flag: a `Result.Success` payload that can carry `success = false` is contradictory, and the flag conveys nothing `Result` doesn't already. (I think this came from kinesis's `RecordData(val success: Boolean = true)`; kinesis's own `FlushData(recordsFlushed, flushInProgress)` is the better model for what this eventually carries.)

`aws-connect/ConnectResults.kt` already has the idiom for an intentionally-empty result, including the rationale in the KDoc:

```kotlin
/**
 * Data returned by a successful [AmplifyCloudWatchClient.flushLogs] call. It
 * currently carries no fields and exists so the operation can grow its result
 * shape without a breaking change.
 */
@ExperimentalAmplifyApi
class FlushData internal constructor()

/** Result of [AmplifyCloudWatchClient.flushLogs]. */
@ExperimentalAmplifyApi
typealias FlushResult = Result<FlushData, AmplifyCloudWatchException>
```

Note it has to be `class`, not `data class` — Kotlin requires at least one primary constructor parameter for a data class. The result typealias is also worth adding for consistency; both `aws-connect` and `aws-kinesis` pair each data type with one.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,118 @@
...
+ */
+@ExperimentalAmplifyApi
+class AmplifyCloudWatchClient(
+    context: Context,
+    region: String,
+    credentialsProvider: AwsCredentialsProvider<AwsCredentials>,
+    options: AmplifyCloudWatchClientOptions
+) : LogSink {
+
+    private val eventsFlow = MutableSharedFlow<LoggingEvent>()
```
mattcreaser: Defaults here are `replay = 0, extraBufferCapacity = 0`, which means `emit` suspends until every subscriber has collected — so a slow collector back-pressures whatever coroutine is reporting a write or flush failure, and events emitted before anyone subscribes are dropped. For a failure-notification channel, `extraBufferCapacity` with `BufferOverflow.DROP_OLDEST` is usually what you want. Easier to settle now than after the internals land on top of it.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,118 @@
...
+ * ```kotlin
+ * val cloudWatch = AmplifyCloudWatchClient(
+ *     context = applicationContext,
+ *     region = "us-east-1",
+ *     credentialsProvider = credentialsProvider,
+ *     options = AmplifyCloudWatchClientOptions(logGroupName = "/app/my-android-app")
+ * )
+ *
+ * // Register as a sink to capture all AmplifyLogging messages
+ * AmplifyLogging.addSink(cloudWatch)
```
mattcreaser, line 62: Flagging for later, not this PR: `AmplifyLogging.logger(name)` is `@InternalAmplifyApi`, so once a customer registers this sink there's no public way for them to obtain a logger and emit through it — the only public path is hand-constructing a `LogMessage` and calling `emit`. Fine while the primary use case is capturing Amplify's own logs, but we should open an issue to expose a public logger in `foundation`; CloudWatch makes it considerably more valuable.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
mattcreaser: Probably a relatively low `extraBufferCapacity` to minimize memory implications.

aws-cloudwatch/gradle.properties
thisisabhash: Trying to follow same convention as existing clients Amplify-servicename-Client (AmplifyKinesisClient). Same as in Swift.

mattcreaser: Okay. I suppose we could always add e.g. `aws-cloudwatch-metrics` if necessary in the future, or simply colocate a new client in the same artifact (the cost to customers is low)

mattcreaser (approved):

PR #3308, opened 2026-04-28
feat(api): add aws-appsync module with public interface

jvh-aws:
- [ ] PR title and description conform to [Pull Request](https://github.com/aws-amplify/amplify-android/blob/main/CONTRIBUTING.md#pull-request-guidelines) guidelines.

*Issue #, if available:*

*Description of changes:*

Add module and public interface of the AppSync client

*How did you test these changes?*
(Please add a line here how the changes were tested)

*Documentation update required?*
- [ ] No
- [ ] Yes (Please include a PR link for the documentation update)

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests
- [ ] Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)
- [ ] Ensure commit message has the appropriate scope (e.g `fix(storage): message`, `feat(auth): message`, `chore(all): message`)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3377, opened 2026-08-13
chore: relocate aws-appsync to top level and decouple from aws-api

mattcreaser:
## What

Moves the standalone `aws-appsync` client module from `appsync/aws-appsync` to the repo top level, and removes its dependency on the `aws-api` plugin. Also fixes two pre-existing latent build failures in the module (see below).

## Why

**Relocation.** `aws-appsync` was nested under `appsync/`, which hosts the *separately-versioned* AppSync **events** libraries (`aws-sdk-appsync-*`). Living there forced the client to carry a bespoke `version.properties` override. The AppSync client is not separately versioned — it ships as part of the main Amplify release. Moving it to the top level (alongside other standalone clients like `aws-kinesis`) lets it inherit the root `VERSION_NAME` and stops it from being conflated with the events SDK family.

**Decoupling from `aws-api`.** The client is a standalone module that will eventually replace the GraphQL functionality of `AWSApiPlugin`. The tempting path — move the plugin's internals into the client and share them — creates a dependency cycle once the plugin is meant to use (or be deleted in favor of) the client, and only defers the "reverse the module dependency" problem.

Because in Phase 1 the plugin does **not** consume the client, *moving* the plugin's private internals into the client now would just force the client to re-expose them as `@InternalAmplifyApi` so the plugin can still compile — the same cross-boundary complexity, pointing the other way, plus premature coupling. Instead, the client will **reimplement** the internals it needs privately, per-feature, in later PRs. That keeps the two modules fully independent during Phase 1 (no dependency in either direction, no new `@InternalAmplifyApi` surface), and the separate-module boundary *mechanically enforces* the client/plugin split so it can't silently re-tangle over the deprecation window.

This PR is the structural groundwork for that: the client now depends only on `core`, `aws-api-appsync` (shared base), `foundation`, and `foundation-bridge` — never on `aws-api`.

## Changes

- **Relocate** `appsync/aws-appsync` → `aws-appsync/`; update `settings.gradle.kts` (top-level `include`, no `projectDir` remap); drop the `version.properties` override so the module inherits the root version.
- **Decouple**: remove `implementation(project(":aws-api"))` from the module.
- **Fix pre-existing #3308 debt** (the module never built green): a latent ktlint function-signature violation (`ktlintFormat`, pure formatting) and a missing `apiCheck` baseline (`apiDump` — records the already-existing public interface).
- **Reverts** The visibility changes to classes in `aws-api`
- **Updates** the build file to use the test bundles present on main

## Out of scope

- Reimplementing plugin-private internals in the client (later per-feature PRs, each with ported Kotlin tests).
- Refactoring `AWSApiPlugin` to delegate to the client

## Verification

- `./gradlew :aws-appsync:build` → BUILD SUCCESSFUL (compiles + tests + ktlint + `apiCheck`).
- `aws-api/` is untouched by this branch; no dangling references to the old path.


aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
cadivus, line 65: It it intended that these stay TODO?

aws-appsync/build.gradle.kts
cadivus, line 31: Is this import still needed?

aws-appsync/build.gradle.kts
```diff
@@ -42,10 +32,5 @@ dependencies {
     api(project(":foundation"))
     implementation(project(":foundation-bridge"))
```
cadivus, line 33: Is this import still needed?

aws-appsync/build.gradle.kts
```diff
@@ -13,27 +13,17 @@
...
-project.setProperty("VERSION_NAME", readVersion())
-
 android {
     namespace = "com.amazonaws.appsync"
 }
 
 dependencies {
     api(project(":core"))
-    implementation(project(":aws-api"))
     implementation(project(":aws-api-appsync"))
```
cadivus: Is this import still needed?

aws-appsync/build.gradle.kts
```diff
@@ -13,27 +13,17 @@
...
 
     implementation(libs.okhttp)
```
cadivus: I am not sure here. This is marked implementation but in the public API in `AmplifyAppSyncClient.kt`. Is this right?

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
mattcreaser: Yes

aws-appsync/build.gradle.kts
mattcreaser: Yes it will be necessary

aws-appsync/build.gradle.kts
mattcreaser: Yes for followup PR

aws-appsync/build.gradle.kts
mattcreaser: Yes good catch this should be `api`

cadivus (approved):

cadivus (approved):

PR #3378, opened 2026-08-17
chore: add cloudwatch client implementation

thisisabhash:

*Issue #, if available:*

N/A

*Description of changes:*

Implements the v3 standalone CloudWatch logging client: a lightweight `AmplifyCloudWatchClient` (`:aws-cloudwatch`, package `com.amplifyframework.cloudwatch`) that depends only on `AmplifyFoundation` / `foundation-bridge` and the `aws-cloudwatchlogs` SDK (no `:core` / `:aws-core`), following the `:aws-kinesis` standalone-client pattern (Kotlin, coroutines, `Result<T, E>`, builder + DSL options). The public API is marked `@ExperimentalAmplifyApi`.

**`:aws-cloudwatch` (new v3 client)**
- `AmplifyCloudWatchClient` — implements `LogSink`. `isEnabledFor(level)` reports the client's enabled state; `emit(message)` filters against `LoggingConstraints` then persists; `enable()`/`disable()` toggle capture + automatic flushing (disable preserves already-buffered entries); `setUserIdentifier(id)`; `setLoggingConstraints(constraints)`; `suspend flushLogs(): FlushResult`; `getCloudWatchLogsClient()`; `events: SharedFlow<LoggingEvent>` (write/flush failures, `DROP_OLDEST` buffered).
- `CloudWatchLogManager` — the batching/upload engine (per-request limits: 10,000 events / 1,048,576 bytes / +26 bytes per event / 24h span; events over the per-event limit are dropped so they can't block the buffer), log-stream creation, delete-on-success, too-new-event retry, and coroutine + WorkManager auto-flush (`CloudWatchRouterWorker` / `CloudWatchLogsSyncWorker` / `CloudWatchWorkerFactory`).
- Persistence: the SQLCipher-backed layer (`CloudWatchDatabase`, `CloudWatchDatabaseHelper`, `LogEventTable`, `LogEvent` in `com.amplifyframework.cloudwatch.db`; `CloudWatchLogEvent` in `.models`) plus `CloudWatchPreferences`, with the database passphrase persisted via foundation's `AmplifyKeyValueRepository`.
- `CloudWatchLoggingFilter` — namespace-based filtering with user > namespace > default precedence (namespace keys case-insensitive, user keys case-sensitive; `None` blocks all).
- `AmplifyCloudWatchClientOptions` (builder + `invoke { }` DSL): `logGroupName`, `localStoreMaxSizeInMB` = 5, `flushStrategy`, `loggingConstraints`, `configureClient`, plus the `CloudWatchLogsClientConfigurationProvider` typealias.
- `FlushStrategy` — `Interval(Duration = 60.seconds)` and `None`; `FlushData` (empty result) + `FlushResult` typealias.
- `LoggingConstraints` / `UserLogLevel` — `defaultLogLevel` (default `Error`), `namespaceLogLevel`, `userLogLevel`.
- `LoggingEvent` — sealed `WriteLogFailure` / `FlushLogFailure` (`context`, `cause`).
- `AmplifyCloudWatchException` — sealed hierarchy (`Storage` / `Service` / `Configuration` / `Unknown`) over foundation `AmplifyException`, with an internal `from(...)` mapper.

**Multi-instance isolation** — the worker factory, the WorkManager unique-work name, and the on-device store (database + passphrase preferences) are all keyed by log group, so two clients targeting different log groups can't hijack each other's delegate factory, clobber each other's scheduled flush, or share a buffer.

**`:foundation` — `KeyValueRepository` moved out of `:core`**
- Moved the `AmplifyKeyValueRepository` cluster (`KeyValueRepository`, `AmplifyKeyValueRepository`, `EncryptedKeyValueRepository`, `InMemoryKeyValueRepository`(`Provider`)) from `:core` into `foundation/androidMain`, so standalone clients can persist a database passphrase without depending on `:core`. `KeyValueRepository` and `EncryptedKeyValueRepository` stay public (their public API moves from `core.api` to `foundation.api`); `AmplifyKeyValueRepository` remains `@InternalAmplifyApi` and `InMemoryKeyValueRepository` `@InternalApiWarning`.
- Enabled Robolectric-based host tests in the KMP module (`jvmToolchain(17)` + host-test Android resources) so the relocated keystore / `EncryptedSharedPreferences` tests run in `:foundation`.
- `LogLevel.allows` is promoted to `@InternalAmplifyApi` and now encapsulates all `None` handling (a `None` message is never emitted; a `None` threshold blocks everything); the CloudWatch filter reuses it instead of duplicating the comparison.

**`:core` and consumers**
- Removed the `KeyValueRepository` cluster from `:core`'s public API and repointed `:aws-auth-cognito`, `:aws-pinpoint-core`, `:aws-analytics-pinpoint`, `:aws-push-notifications-pinpoint`, and `:aws-auth-plugins-core` to the `:foundation` types (behavior unchanged). `apiDump` updated for `:core` and `:foundation`.
- Note on the shared keystore key: all Amplify components (Cognito, Pinpoint, this client, …) share the single `"amplify_master_key"` alias managed by the one `EncryptedKeyValueRepository`; on keystore corruption the recovery path resets those Amplify-encrypted stores together. This is documented in the class as an accepted trade-off — consolidating removes the duplicate keystore-management implementation but does not isolate the components' key material.

**`:aws-logging-cloudwatch` (v2 plugin)**
- Now depends on `:aws-cloudwatch` and reuses its persistence layer (`db/*`, `models/CloudWatchLogEvent`, `CloudWatchPreferences`) instead of its own copies. Public API and behavior are unchanged (apiCheck passes); it keeps its own engine, Hub events, remote config, and `CategoryType` constraints.

- 49 unit tests in `:aws-cloudwatch` (Robolectric + MockK): filter precedence + `None` handling, exception mapping, options builder/DSL, client behavior (emit filtering, disabled-drop, flush `Result` mapping, enable/disable, `setUserIdentifier`/`setLoggingConstraints`), and the manager engine — batching boundary cases (crossing the 10,000-event / 1 MB / 24h limits plus oversized-event drop), stream creation/reuse, too-new retry, failure callbacks, WorkManager scheduling, and per-log-group instance isolation.
- The relocated `KeyValueRepository` / `EncryptedKeyValueRepository` unit tests now run as `:foundation` Robolectric host tests; the storage instrumentation test moved into `:aws-cloudwatch`.
- Full `build` (compile, ktlint, checkstyle, Android lint, `apiCheck`, unit tests, coverage) passes for `:aws-cloudwatch`, `:foundation`, `:core`, `:aws-logging-cloudwatch`, and the five repointed consumer modules; v2's existing unit tests still pass with its public API unchanged.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * Update the current user identifier. Pending events are flushed to the previous user's stream
+     * before the identifier changes.
+     */
+    fun setUserIdentifier(identifier: String?) {
+        coroutineScope.launch {
+            syncLogEventsWithCloudwatch()
+            userIdentityId = identifier
+        }
+    }
```
mattcreaser: 🔴 **Critical — this can crash the app and permanently kill the manager's scope.**

`coroutineScope` (line 84) is `CoroutineScope(coroutineDispatcher)` — a plain `Job`, no `SupervisorJob`, no `CoroutineExceptionHandler`. Unlike the v2 plugin, v3's `syncLogEventsWithCloudwatch` **rethrows** (line 178), so any flush failure here becomes an uncaught exception: it reaches the default handler (app crash on Android) *and* cancels the scope, after which `clearCache()` and every later `setUserIdentifier` silently no-op.

It also leaves `userIdentityId` on the old value while `AmplifyCloudWatchClient.setUserIdentifier` (line 213) already advanced its own copy, so filtering and stream naming disagree from then on.

```suggestion
    fun setUserIdentifier(identifier: String?) {
        coroutineScope.launch {
            try {
                syncLogEventsWithCloudwatch()
            } catch (e: Exception) {
                logger.warn(e) { "Failed to flush pending events before switching user" }
            } finally {
                userIdentityId = identifier
            }
        }
    }
```

Worth adding `SupervisorJob()` to the scope on line 84 regardless, so one failed child can't take down the rest.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,225 @@
...
+        logManager.stopSync()
+    }
+
+    /** Flush all pending log entries to CloudWatch. */
+    suspend fun flushLogs(): FlushResult = try {
+        logManager.syncLogEventsWithCloudwatch()
+        Result.Success(FlushData())
+    } catch (error: Exception) {
+        Result.Failure(AmplifyCloudWatchException.from(error))
+    }
```
mattcreaser: 🟡 **`CancellationException` is converted into a failure result.**

`CancellationException` is an `Exception`, so cancelling the caller's scope yields `Result.Failure(AmplifyCloudWatchUnknownException)` instead of propagating — that breaks structured concurrency and reports a cancellation to the customer as a CloudWatch error.

```suggestion
    /** Flush all pending log entries to CloudWatch. */
    suspend fun flushLogs(): FlushResult = try {
        logManager.syncLogEventsWithCloudwatch()
        Result.Success(FlushData())
    } catch (error: CancellationException) {
        throw error
    } catch (error: Exception) {
        Result.Failure(AmplifyCloudWatchException.from(error))
    }
```

Same issue in `CloudWatchLogManager.saveLogEvent` (lines 101-104), where a cancellation is surfaced through `onWriteLogFailure`.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+    private val coroutineScope = CoroutineScope(coroutineDispatcher)
+    private var isSyncInProgress = AtomicBoolean(false)
+    private val logger = AmplifyLogging.logger(CloudWatchLogManager::class.java.simpleName)
+
+    init {
+        // Register the delegate factory so the WorkManager-instantiated CloudWatchRouterWorker can
+        // create a CloudWatchLogsSyncWorker wired to this manager.
+        CloudWatchRouterWorker.workerFactories[CloudWatchRouterWorker.WORKER_FACTORY_KEY] =
+            CloudWatchWorkerFactory(this)
+    }
```
mattcreaser: 🟡 **A second client instance silently hijacks the first.**

Three process-global singletons are keyed by constants rather than by client identity: this factory slot, `CloudWatchLogsSyncWorker.WORKER_NAME_TAG`, and `CLIENT_DATABASE_NAME` (`AmplifyCloudWatchClient.kt:52`). Two clients targeting different log groups therefore share one database and one scheduled flush, and the second construction overwrites the first's factory here — so all buffered events ship to whichever log group registered last.

`AmplifyCloudWatchClient`'s KDoc states the one-instance-per-log-group constraint, but nothing enforces or even detects a violation. `:aws-kinesis` already solves this by keying storage on identity (`SQLiteRecordStorage(dbPrefix = "kinesis_records", identifier = region)`) — worth doing the same with the log group name here, or failing fast on a second registration.

aws-cloudwatch/src/test/java/com/amplifyframework/cloudwatch/CloudWatchLogManagerTest.kt
```diff
@@ -0,0 +1,292 @@
...
+        localStoreMaxSizeInMB = 5,
+        flushIntervalInSeconds = flushIntervalInSeconds,
+        awsCloudWatchLogsClient = cloudWatchLogsClient,
+        databaseName = "test.cloudwatch.db",
+        passphrasePreferencesName = "test.prefs",
+        onWriteLogFailure = { _, cause -> writeFailures.add(cause) },
+        onFlushLogFailure = { _, cause -> flushFailures.add(cause) },
+        cloudWatchLoggingDatabase = database,
+        coroutineDispatcher = testDispatcher
+    )
```
mattcreaser: 🟡 **Batching boundaries are untested.**

The PR description lists "batching" as covered, but every test here feeds one or two events, so `getNextBatch` never actually splits a batch: the 10,000-event cap, the 1 MB byte cap, and the 24-hour span check have no coverage. That is precisely why the millisecond/second unit bug on `CloudWatchLogManager.kt:223` survived the port from v2.

Worth adding three cases that each cross exactly one boundary and assert a single batch split:
- more than 10,000 queued events
- cumulative message size crossing 1,048,576 bytes
- events spanning more than 24 hours

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    /** Returns the underlying AWS CloudWatch Logs SDK client. */
+    fun getCloudWatchLogsClient(): CloudWatchLogsClient = cloudWatchLogsClient
```
mattcreaser: 🔵 Expose the SDK client as a property, for consistency with the other standalone clients.

`AmplifyKinesisClient` exposes `val kinesisClient: KinesisClient`. Same JVM signature (`getCloudWatchLogsClient()`), more idiomatic Kotlin.

```suggestion
    /** The underlying AWS CloudWatch Logs SDK client, for direct access. */
    val cloudWatchLogsClient: CloudWatchLogsClient
```

(promote the constructor parameter on line 87 and drop this accessor)

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+        passphrasePreferencesName = passphrasePreferencesName
+    ),
+    private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+
+    private var stopSync = false
+
+    @Volatile
+    private var userIdentityId: String? = null
+    private val todayDate: String = SimpleDateFormat("MM-dd-yyyy", Locale.US).format(Date())
```
mattcreaser: 🔵 `todayDate` is frozen at construction.

A process that stays alive across midnight keeps writing to the previous day's stream name (line 147). Computing the date where the stream name is built would keep streams aligned with the day the events were actually emitted.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLoggingFilter.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.amplifyframework.annotations.ExperimentalAmplifyApi
+import com.amplifyframework.foundation.logging.LogLevel
+
+/**
+ * Decides whether a log message passes the current [LoggingConstraints]. Precedence, highest first:
+ * per-user override (by exact user id), then per-namespace override (namespace keys matched
+ * case-insensitively), then the default level. A threshold of [LogLevel.None] blocks everything.
+ */
+@ExperimentalAmplifyApi
+internal class CloudWatchLoggingFilter(initialConstraints: LoggingConstraints) {
```
mattcreaser: 🔵 `@ExperimentalAmplifyApi` has no effect on an `internal` class.

Opt-in annotations only matter on the public surface, and this class is excluded from the API dump — safe to drop.

```suggestion
internal class CloudWatchLoggingFilter(initialConstraints: LoggingConstraints) {
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientOptions.kt
```diff
@@ -0,0 +1,118 @@
...
+            @JvmSynthetic set
+
+        var flushStrategy: FlushStrategy = FlushStrategy.Interval()
+            @JvmSynthetic set
+
+        var loggingConstraints: LoggingConstraints = LoggingConstraints()
+            @JvmSynthetic set
+
+        var configureClient: CloudWatchLogsClientConfigurationProvider? = null
+            @JvmSynthetic private set
```
mattcreaser: 🔵 This setter is asymmetric with every sibling property.

`logGroupName`, `localStoreMaxSizeInMB`, `flushStrategy`, and `loggingConstraints` all have a public `@JvmSynthetic set`, so `AmplifyCloudWatchClientOptions { logGroupName = "…" }` works — but `configureClient = provider` does not compile, because the setter is `private`. The DSL quietly loses one knob relative to the Java builder.

Either match the siblings, or make it consistently non-assignable in both languages so the asymmetry is deliberate and visible.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/FlushStrategy.kt
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param interval Time between automatic flush operations. Defaults to 60 seconds.
+     */
+    data class Interval(val interval: Duration = 60.seconds) : FlushStrategy()
+
+    /**
+     * Disable automatic flushing. Log events continue to be buffered locally and
+     * must be flushed manually by calling [AmplifyCloudWatchClient.flushLogs].
+     */
+    data object None : FlushStrategy()
```
mattcreaser: 🔵 Duplicates `recordcache`'s equivalent, and `Interval` is unreachable from Java.

Two things worth an explicit decision in the API-review docs rather than a code change here:

1. `com.amplifyframework.recordcache.FlushStrategy` (in `:aws-kinesis`) is the same sealed `Interval(Duration) / None` shape with a 30 s rather than 60 s default. An app using both clients ends up with two same-named types that collide on import. Reasonable as-is since `:aws-cloudwatch` can't depend on `:aws-kinesis` — but it argues for extracting the record-cache primitives into a shared module.

2. Because `Duration` is a value class, both of `Interval`'s constructors compile to `synthetic` and the getter mangles to `getInterval-UwyO8pc`, so **Java consumers cannot construct `FlushStrategy.Interval` at all** — they're limited to `None`. `:aws-kinesis` has the identical gap, so this is an ecosystem-wide question, not a defect introduced here. Given `AGENTS.md` §5 commits to Java interop, it deserves a deliberate answer (a `Long`-seconds factory alongside the `Duration` one, perhaps).

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: 🔴 **Critical — this can crash the app and permanently kill the manager's scope.**

`coroutineScope` (line 84) is `CoroutineScope(coroutineDispatcher)` — a plain `Job`, no `SupervisorJob`, no `CoroutineExceptionHandler`. Unlike the v2 plugin, v3's `syncLogEventsWithCloudwatch` **rethrows** (line 178), so any flush failure here becomes an uncaught exception: it reaches the default handler (app crash on Android) *and* cancels the scope, after which `clearCache()` and every later `setUserIdentifier` silently no-op.

It also leaves `userIdentityId` on the old value while `AmplifyCloudWatchClient.setUserIdentifier` (line 213) already advanced its own copy, so filtering and stream naming disagree from then on.

```suggestion
    fun setUserIdentifier(identifier: String?) {
        coroutineScope.launch {
            try {
                syncLogEventsWithCloudwatch()
            } catch (e: Exception) {
                logger.warn(e) { "Failed to flush pending events before switching user" }
            } finally {
                userIdentityId = identifier
            }
        }
    }
```

Worth adding `SupervisorJob()` to the scope on line 84 regardless, so one failed child can't take down the rest.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/FlushData.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.amplifyframework.annotations.ExperimentalAmplifyApi
+import com.amplifyframework.foundation.result.Result
+
+/**
+ * Data returned by a successful [AmplifyCloudWatchClient.flushLogs] call. It
+ * currently carries no fields and exists so the operation can grow its result
+ * shape without a breaking change.
+ */
+@ExperimentalAmplifyApi
+class FlushData internal constructor()
```
mattcreaser: 🔵 Approved shape was `data class FlushData(val success: Boolean = true)`.

The API review doc specifies a `success` field. The forward-compatibility rationale in the KDoc is sound and arguably better than a field that's always `true` — just flagging it as an unrecorded deviation, since `FlushData` is the one type the two platforms don't otherwise share a shape for.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/LoggingConstraints.kt
```diff
@@ -0,0 +1,45 @@
...
+ * @param defaultLogLevel The default minimum [LogLevel] to capture
+ * @param namespaceLogLevel Per-namespace overrides of the minimum [LogLevel]
+ * @param userLogLevel Per-user overrides, keyed by user identifier
+ */
+@ExperimentalAmplifyApi
+data class LoggingConstraints(
+    val defaultLogLevel: LogLevel = LogLevel.Error,
+    val namespaceLogLevel: Map<String, LogLevel> = emptyMap(),
+    val userLogLevel: Map<String, UserLogLevel> = emptyMap()
+)
```
mattcreaser: 🔵 Approved spec marks both of these `@Serializable`.

`LoggingConstraints` and `UserLogLevel` (line 42) are both `@Serializable` in the API review doc, including in the post-revision version that removed `CloudWatchLoggingClientConfiguration` — so the annotation was meant to outlive the configuration object, presumably so customers can persist or remote-fetch constraints themselves now that `RemoteLoggingConstraintProvider` is gone.

Field names and defaults otherwise match the approved spec exactly, including the `categoryLogLevel` → `namespaceLogLevel` rename from the first review round. 👍

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: 🟡 **A second client instance silently hijacks the first.**

Three process-global singletons are keyed by constants rather than by client identity: this factory slot, `CloudWatchLogsSyncWorker.WORKER_NAME_TAG`, and `CLIENT_DATABASE_NAME` (`AmplifyCloudWatchClient.kt:52`). Two clients targeting different log groups therefore share one database and one scheduled flush, and the second construction overwrites the first's factory here — so all buffered events ship to whichever log group registered last.

`AmplifyCloudWatchClient`'s KDoc states the one-instance-per-log-group constraint, but nothing enforces or even detects a violation. `:aws-kinesis` already solves this by keying storage on identity (`SQLiteRecordStorage(dbPrefix = "kinesis_records", identifier = region)`) — worth doing the same with the log group name here, or failing fast on a second registration.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+    private val coroutineScope = CoroutineScope(coroutineDispatcher)
+    private var isSyncInProgress = AtomicBoolean(false)
+    private val logger = AmplifyLogging.logger(CloudWatchLogManager::class.java.simpleName)
```
mattcreaser: 🔵 Use the reified logger factory and the lazy message overloads.

`AGENTS.md` §3b calls for lazy suppliers, and `AmplifyCloudWatchClient` in this same PR already uses `AmplifyLogging.logger<AmplifyCloudWatchClient>()` with `logger.info { … }`. This class uses the string factory plus the eager overloads on lines 102 (`logger.error("failed to save event", e)`) and 126 (`logger.debug("stopping sync")`).

```suggestion
    private val logger = AmplifyLogging.logger<CloudWatchLogManager>()
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLoggingFilter.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        return constraints.defaultLogLevel.allows(level)
+    }
+
+    private fun Map<String, LogLevel>.matching(namespace: String): LogLevel? =
+        entries.firstOrNull { it.key.equals(namespace, ignoreCase = true) }?.value
+
+    // A threshold allows a message when the message level is at or above the threshold. LogLevel is
+    // ordered Verbose < Debug < Info < Warn < Error < None, so None as a threshold allows nothing.
+    private fun LogLevel.allows(level: LogLevel): Boolean = this != LogLevel.None && this <= level
```
mattcreaser: 🔵 This shadows `foundation`'s `allows` with different semantics.

`foundation/src/commonMain/kotlin/com/amplifyframework/foundation/logging/LogLevel.kt` already declares `internal infix fun LogLevel.allows(level: LogLevel) = this <= level`. Same name, different behaviour, one module apart — easy to confuse when reading either site.

Either rename this one (`permits`?) or lift the `None` handling into the foundation helper so there's a single definition.

The precedence logic itself matches the approved spec (user → namespace → default), and the case-insensitive namespace matching is a nice touch. 👍

foundation/src/androidMain/kotlin/com/amplifyframework/foundation/store/EncryptedKeyValueRepository.kt
```diff
@@ -0,0 +1,213 @@
...
+    internal companion object {
+        private val logger = AmplifyLogging.logger(EncryptedKeyValueRepository::class.simpleName!!)
+
+        private fun getDefaultMasterKeySpec() = MasterKeys.AES256_GCM_SPEC
+
+        // We create our own KeyGenParameterSpec that is exactly like MasterKeys.AES256_GCM_SPEC except with a different
+        // alias. This allows us to safely delete this key should it become corrupted without potentially impacting any
+        // other part of the customer's application.
+        private fun getAmplifyMasterKeySpec() = KeyGenParameterSpec.Builder(
+            "amplify_master_key",
```
mattcreaser, line 208: 🔴 **Critical — this copy shares its keystore alias with `:core`'s copy, and the design doc called for a move rather than a copy.**

`core/src/main/java/com/amplifyframework/core/store/EncryptedKeyValueRepository.kt:200` uses the identical `"amplify_master_key"` alias. Both copies own a `deleteAmplifyMasterKey()` that unconditionally deletes that keystore entry when they judge it corrupt, and neither knows the other exists — so a CloudWatch passphrase-store hiccup can delete the key protecting `:core`'s Cognito credential store, signing the user out. The comment directly above stopped being accurate the moment this class was duplicated.

The design doc's Android development-strategy section specified a move, not a copy:

> `AmplifyKeyValueRepository` and its supporting types (`KeyValueRepository`, `EncryptedKeyValueRepository`, `InMemoryKeyValueRepository`) will be **moved** to the `AmplifyFoundation` library so they can be shared across all Amplify clients without depending on `:core`.

This one has no Swift counterpart to compare against — the keystore/`EncryptedSharedPreferences` machinery is Android-only — so there's no cross-platform precedent to lean on here.

If keeping both copies is a deliberate staging decision, the foundation copy needs its own alias so the two can't fight:

```suggestion
        // We create our own KeyGenParameterSpec that is exactly like MasterKeys.AES256_GCM_SPEC except with a different
        // alias. This allows us to safely delete this key should it become corrupted without potentially impacting any
        // other part of the customer's application. The alias is deliberately distinct from :core's copy of this class
        // so that neither module's corruption-recovery path can delete the key the other depends on.
        private fun getAmplifyMasterKeySpec() = KeyGenParameterSpec.Builder(
            "amplify_foundation_master_key",
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser, line 139: 🔴 **Critical — this can crash the app and permanently kill the manager's scope.**

`coroutineScope` (line 84) is `CoroutineScope(coroutineDispatcher)` — a plain `Job`, no `SupervisorJob`, no `CoroutineExceptionHandler`. Unlike the v2 plugin, v3's `syncLogEventsWithCloudwatch` **rethrows** (line 178), so any flush failure here becomes an uncaught exception: it reaches the default handler (app crash on Android) *and* cancels the scope, after which `clearCache()` and every later `setUserIdentifier` silently no-op.

It also leaves `userIdentityId` on the old value while `AmplifyCloudWatchClient.setUserIdentifier` (line 213) already advanced its own copy, so filtering and stream naming disagree from then on.

```suggestion
    fun setUserIdentifier(identifier: String?) {
        coroutineScope.launch {
            try {
                syncLogEventsWithCloudwatch()
            } catch (e: Exception) {
                logger.warn(e) { "Failed to flush pending events before switching user" }
            } finally {
                userIdentityId = identifier
            }
        }
    }
```

Worth adding `SupervisorJob()` to the scope on line 84 regardless, so one failed child can't take down the rest.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,225 @@
...
+        isEnabled = true
+        scope.launch { logManager.startSync() }
+    }
+
+    /** Disable logging and automatic flushing. Messages emitted while disabled are dropped. */
+    fun disable() {
+        logger.info { "Disabling CloudWatch logging and automatic flushing" }
+        isEnabled = false
+        logManager.stopSync()
+    }
```
mattcreaser: 🟡 **`disable()` destroys buffered logs here, but preserves them on Swift.**

`stopSync()` calls `clearCache()` (`CloudWatchLogManager.kt:126`, 239-243), which wipes the entire local table — every event captured but not yet uploaded is gone.

Swift's client, for the same public API, does not: [`AmplifyCloudWatchClient.disable()`](https://github.com/aws-amplify/amplify-swift/blob/feat/v3-cloudwatch-client/AmplifyClients/AmplifyCloudWatchClient/Sources/AmplifyCloudWatchClient.swift) sets `enabled = false` and tears down each session controller, and `CloudWatchLoggingSessionController.disable()` only clears `batchSubscription` / `session` / `consumer`. It has a `resetCurrentLogs()` but deliberately doesn't call it from `disable()`, so the on-disk batches survive and a later `flushLogs()` still ships them.

Of everything I compared between the two clients, this is the only place they behave differently on an identically-named public API — so it looks unintended rather than a platform-appropriate choice. It's also the sharp edge that would distinguish `disable()` from `FlushStrategy.None`, which preserves the buffer.

Either match Swift and drop the `clearCache()`, or if the wipe is intended, say so:

```suggestion
    /**
     * Disable logging and automatic flushing. Messages emitted while disabled are dropped, and any
     * log entries still buffered locally are discarded — call [flushLogs] first if you need them.
     */
    fun disable() {
        logger.info { "Disabling CloudWatch logging and automatic flushing" }
        isEnabled = false
        logManager.stopSync()
    }
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Engine for buffering and delivering CloudWatch log events for [AmplifyCloudWatchClient]. Failures
+ * are reported through the [onWriteLogFailure] / [onFlushLogFailure] callbacks, user identity is set
+ * explicitly via [setUserIdentifier], and log persistence is delegated to the shared
+ * `:aws-cloudwatch-common` store.
+ *
+ * @param flushIntervalInSeconds interval between automatic flushes; `null` disables automatic
+ *   flushing (used for `FlushStrategy.None`).
+ */
+internal class CloudWatchLogManager(
```
mattcreaser, line 70: 🟡 **Four latent defects were cloned along with the v2 engine.**

The design doc listed `CloudWatchLogManager` among the logic to extract into the shared module, and `:aws-cloudwatch-common` ended up holding only the database layer plus `CloudWatchPreferences`. I'd flag that as a divergence except Swift made exactly the same call — its `InternalCloudWatchLogging` shares only `LogActor` / `LogFile` / `LogRotation`, the `LogBatch` protocols, constants, and the network monitor, while `CloudWatchLoggingSession`, `CloudWatchLoggingSessionController`, `CloudWatchLoggingConsumer`, and `RotatingLogger` are all duplicated between plugin and client. Share the persistence layer, duplicate the engine, on both platforms. So the doc is what's stale here, not the code.

The cost is still real, though: four defects now exist in both copies and have to be fixed twice.

- the 24-hour batch-span check comparing ms against seconds (line 223)
- the check-then-set flush re-entrancy race (lines 131-137)
- the oversized-event permanent stall (lines 152-154)
- `Result.retry()` fighting the `finally` re-enqueue (`worker/CloudWatchLogsSyncWorker.kt`)

The batching/`getNextBatch` logic in particular is byte-identical to v2's and has no v3-specific behaviour, so it's the one piece that could move into this module cheaply and fix all four in one place.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    suspend fun syncLogEventsWithCloudwatch() {
+        if (isSyncInProgress.get()) {
+            return
+        }
+        withContext(coroutineDispatcher) {
+            var inputLogEventsIdToBeDeleted: List<Long> = emptyList()
+            try {
+                isSyncInProgress.set(true)
```
mattcreaser: 🟡 **The re-entrancy guard is a check-then-set race.**

A manual `flushLogs()`, a cache-full flush from `saveLogEvent` (line 99), and the WorkManager flush can all clear the `get()` before any of them reaches `set(true)`, so several uploads run concurrently and delete each other's rows. Inherited from v2, but v3 adds a third concurrent trigger.

```suggestion
    suspend fun syncLogEventsWithCloudwatch() {
        if (!isSyncInProgress.compareAndSet(false, true)) {
            return
        }
        withContext(coroutineDispatcher) {
            var inputLogEventsIdToBeDeleted: List<Long> = emptyList()
            try {
```

(the `isSyncInProgress.set(true)` on line 137 then becomes redundant)

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+            val cloudWatchEvent = iterator.next()
+            totalBatchSize = totalBatchSize.plus(cloudWatchEvent.message.length).plus(26)
+            if (
+                // The maximum number of log events in a batch is 10,000.
+                inputLogEvents.size >= 10000 ||
+                // The maximum batch size is 1,048,576 bytes.
+                totalBatchSize >= 1048576 ||
+                // A batch of log events in a single request cannot span more than 24 hours.
+                // Otherwise, the operation fails.
+                cloudWatchEvent.timestamp - firstEvent.timestamp >= 24 * 60 * 60L
```
mattcreaser: 🟡 **Off by 1000× — millisecond timestamps compared against a seconds constant.**

Timestamps come from `System.currentTimeMillis()` (`AmplifyCloudWatchClient.kt:171`), so this threshold is 86,400 **milliseconds** = 86.4 seconds, not 24 hours. Batches split roughly every 90 seconds of log span, producing far more `PutLogEvents` calls than intended.

Ported verbatim from v2 — but this is a new module, so a good moment to not carry it forward.

```suggestion
                // A batch of log events in a single request cannot span more than 24 hours.
                // Otherwise, the operation fails.
                cloudWatchEvent.timestamp - firstEvent.timestamp >= 24 * 60 * 60 * 1000L
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+                                    logStreamName = streamName
+                                }
+                            ).also { response ->
+                                response.rejectedLogEventsInfo?.tooNewLogEventStartIndex?.let {
+                                    inputLogEventsIdToBeDeleted = inputLogEventsIdToBeDeleted.slice(
+                                        IntRange(0, it - 1)
+                                    ).toMutableList()
+                                }
+                                cloudWatchLoggingDatabase.bulkDelete(inputLogEventsIdToBeDeleted)
+                            }
```
mattcreaser: 🟡 **`tooNewLogEventStartIndex == 0` produces an unbounded upload loop.**

When index `0` comes back (every event in the batch rejected as too new), `slice(IntRange(0, -1))` is empty and `bulkDelete` is a no-op. The inner loop drains the in-memory list, then the outer `while (true)` on line 139 re-queries the same undeleted rows and repeats — calling `PutLogEvents` in a tight loop indefinitely.

```suggestion
                            ).also { response ->
                                response.rejectedLogEventsInfo?.tooNewLogEventStartIndex?.let {
                                    inputLogEventsIdToBeDeleted = inputLogEventsIdToBeDeleted.slice(
                                        IntRange(0, it - 1)
                                    ).toMutableList()
                                }
                                // Nothing was accepted, so re-querying would return the same rows forever.
                                if (inputLogEventsIdToBeDeleted.isEmpty()) return@withContext
                                cloudWatchLoggingDatabase.bulkDelete(inputLogEventsIdToBeDeleted)
                            }
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+
+                            // Default format: MM-dd-yyyy.deviceId.userId
+                            val streamName = "$todayDate.$deviceId.${userIdentityId ?: "guest"}"
+
+                            val nextBatch = getNextBatch(queriedEvents)
+                            val inputLogEvents = nextBatch.first
+                            inputLogEventsIdToBeDeleted = nextBatch.second
+                            if (inputLogEvents.isEmpty()) {
+                                return@withContext
+                            }
```
mattcreaser: 🟡 **A single oversized event permanently stalls the whole buffer.**

If the oldest buffered event is on its own ≥ 1,048,550 chars, the `totalBatchSize >= 1048576` check on line 220 trips on the first iteration of `getNextBatch`, so it returns an empty batch and this `return@withContext` aborts. The event is never uploaded and never deleted, and it blocks every event queued behind it for the lifetime of the store.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
mattcreaser, line 211: 🟡 **`CancellationException` is converted into a failure result.**

`CancellationException` is an `Exception`, so cancelling the caller's scope yields `Result.Failure(AmplifyCloudWatchUnknownException)` instead of propagating — that breaks structured concurrency and reports a cancellation to the customer as a CloudWatch error.

```suggestion
    /** Flush all pending log entries to CloudWatch. */
    suspend fun flushLogs(): FlushResult = try {
        logManager.syncLogEventsWithCloudwatch()
        Result.Success(FlushData())
    } catch (error: CancellationException) {
        throw error
    } catch (error: Exception) {
        Result.Failure(AmplifyCloudWatchException.from(error))
    }
```

Same issue in `CloudWatchLogManager.saveLogEvent` (lines 101-104), where a cancellation is surfaced through `onWriteLogFailure`.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser, line 107: 🟡 **A second client instance silently hijacks the first.**

Three process-global singletons are keyed by constants rather than by client identity: this factory slot, `CloudWatchLogsSyncWorker.WORKER_NAME_TAG`, and `CLIENT_DATABASE_NAME` (`AmplifyCloudWatchClient.kt:52`). Two clients targeting different log groups therefore share one database and one scheduled flush, and the second construction overwrites the first's factory here — so all buffered events ship to whichever log group registered last.

`AmplifyCloudWatchClient`'s KDoc states the one-instance-per-log-group constraint, but nothing enforces or even detects a violation. Note Swift has no equivalent exposure — its client keys session controllers by `LoggerKey(namespace:logLevel:)` on the instance, with no process-global registry. `:aws-kinesis` also already solves this by keying storage on identity (`SQLiteRecordStorage(dbPrefix = "kinesis_records", identifier = region)`) — worth doing the same with the log group name here, or failing fast on a second registration.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+    private val onFlushLogFailure: (context: String?, cause: Throwable?) -> Unit,
+    private val cloudWatchLoggingDatabase: CloudWatchDatabase = CloudWatchDatabase(
+        context,
+        databaseName = databaseName,
+        passphrasePreferencesName = passphrasePreferencesName
+    ),
+    private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+
+    private var stopSync = false
```
mattcreaser, line 89: 🟡 **Shared mutable state without `@Volatile`.**

`stopSync` is written by `stopSync()`/`startSync()` and read by `enqueueSync()` (line 260) from WorkManager threads, the client's IO scope, and this manager's own scope. The neighbouring `userIdentityId` got `@Volatile` for exactly this reason.

```suggestion
    @Volatile
    private var stopSync = false
```

`isSyncInProgress` on line 85 should also be `val` rather than `var` — the reference never changes.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/worker/CloudWatchLogsSyncWorker.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    override suspend fun doWork(): Result = try {
+        cloudWatchLogManager.syncLogEventsWithCloudwatch()
+        Result.success()
+    } catch (exception: Exception) {
+        Result.retry()
+    } finally {
+        cloudWatchLogManager.enqueueSync()
+    }
```
mattcreaser, line 54: 🟡 **`Result.retry()` and the `finally` re-enqueue conflict.**

`enqueueSync()` uses `ExistingWorkPolicy.REPLACE` on the same unique name (`CloudWatchLogManager.kt:255-263`), so the re-enqueue supersedes the `Result.retry()` this method just returned. WorkManager's exponential backoff never applies, and a failed flush waits a full interval instead of retrying quickly.

Inherited from v2, but worth resolving now — either return `Result.success()` and rely entirely on the re-enqueue, or drop the `finally` and let `retry()` do its job.

Separately, `catch (exception: Exception)` swallows the throwable without logging it, so the reason for a retry is invisible.

aws-cloudwatch/src/test/java/com/amplifyframework/cloudwatch/CloudWatchLogManagerTest.kt
mattcreaser: 🟡 **Batching boundaries are untested.**

The PR description lists "batching" as covered, but every test here feeds one or two events, so `getNextBatch` never actually splits a batch: the 10,000-event cap, the 1 MB byte cap, and the 24-hour span check have no coverage. That is precisely why the millisecond/second unit bug on `CloudWatchLogManager.kt:223` survived the port from v2.

Worth adding three cases that each cross exactly one boundary and assert a single batch split:
- more than 10,000 queued events
- cumulative message size crossing 1,048,576 bytes
- events spanning more than 24 hours

The client-level tests are otherwise thorough — filter precedence, disabled-drop, `Result` mapping, and the enable/disable transitions all look good. 👍

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchException.kt
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @param message Error message describing what went wrong
+ * @param recoverySuggestion Suggested action to resolve the error
+ * @param cause Underlying cause of the exception
+ */
+sealed class AmplifyCloudWatchException(
+    message: String,
+    recoverySuggestion: String,
+    cause: Throwable? = null
+) : AmplifyException(message, recoverySuggestion, cause) {
```
mattcreaser, line 40: 🔵 These four are the only public types in the module without `@ExperimentalAmplifyApi`.

Both review docs state that all public APIs are annotated `@RequiresOptIn(ExperimentalAmplifyApi)`, and everything else here follows that — `AmplifyCloudWatchClient`, its options, `FlushStrategy`, `FlushData`, `LoggingConstraints`, `LoggingEvent`. The exception hierarchy is reachable from `FlushResult`, so it's part of the same experimental surface.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: 🔵 Use the reified logger factory and the lazy message overloads.

`AGENTS.md` §3b calls for lazy suppliers, and `AmplifyCloudWatchClient` in this same PR already uses `AmplifyLogging.logger<AmplifyCloudWatchClient>()` with `logger.info { … }`. This class uses the string factory plus the eager overloads on lines 102 (`logger.error("failed to save event", e)`) and 126 (`logger.debug("stopping sync")`).

```suggestion
    private val logger = AmplifyLogging.logger<CloudWatchLogManager>()
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,281 @@
...
+
+    private fun getNextBatch(queriedEvents: MutableList<LogEvent>): Pair<List<InputLogEvent>, List<Long>> {
+        var totalBatchSize = 0L
+        val inputLogEvents = mutableListOf<InputLogEvent>()
+        val inputLogEventsIdToBeDeleted = mutableListOf<Long>()
+        val firstEvent = queriedEvents[0]
+        val iterator = queriedEvents.iterator()
+        while (iterator.hasNext()) {
+            val cloudWatchEvent = iterator.next()
+            totalBatchSize = totalBatchSize.plus(cloudWatchEvent.message.length).plus(26)
```
mattcreaser: 🔵 Batch size counts characters, not UTF-8 bytes.

CloudWatch measures the 1,048,576-byte batch limit in UTF-8. Any non-ASCII content (localised messages, emoji, symbols) undercounts here, so a batch can be built past the real limit and fail the whole `PutLogEvents` request.

```suggestion
            totalBatchSize = totalBatchSize
                .plus(cloudWatchEvent.message.toByteArray(Charsets.UTF_8).size)
                .plus(26)
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: 🔵 `todayDate` is frozen at construction.

A process that stays alive across midnight keeps writing to the previous day's stream name (line 147). Computing the date where the stream name is built would keep streams aligned with the day the events were actually emitted.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLoggingFilter.kt
mattcreaser: 🔵 `@ExperimentalAmplifyApi` has no effect on an `internal` class.

Opt-in annotations only matter on the public surface, and this class is excluded from the API dump — safe to drop.

```suggestion
internal class CloudWatchLoggingFilter(initialConstraints: LoggingConstraints) {
```

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLoggingFilter.kt
mattcreaser: 🔵 This shadows `foundation`'s `allows` with different semantics.

`foundation/src/commonMain/kotlin/com/amplifyframework/foundation/logging/LogLevel.kt` already declares `internal infix fun LogLevel.allows(level: LogLevel) = this <= level`. Same name, different behaviour, one module apart — easy to confuse when reading either site.

Recommend lifting the `None` handling into the foundation helper so there's a single definition, marking it with `@InternalAmplifyApi`.

The precedence logic itself matches the approved spec and Swift's `CloudWatchLoggingFilter` (user → namespace → default), and the case-insensitive namespace matching is a nice touch. 👍

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: Also prefer the `+` operator to `.plus()` here

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/worker/CloudWatchLogsSyncWorker.kt
mattcreaser: Add a log warning if keeping the try/catch

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
mattcreaser: Recommend using kotlinx.time Duration to do this math unambigiously.

```kotlin
private val MAX_BATCH_SPAN = 24.hours
// ...
cloudWatchEvent.timestamp - firstEvent.timestamp >= MAX_BATCH_SPAN.inWholeMilliseconds
```

foundation/src/androidMain/kotlin/com/amplifyframework/foundation/store/EncryptedKeyValueRepository.kt
mattcreaser: I think we just want to move it.

aws-cloudwatch-common/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
+/*
```
mattcreaser: Given that we want to implement the plugin via the client, this new module doesn't seem to offer many benefits. We could add the classes to the `aws-cloudwatch` module instead, and have `aws-logging-cloudwatch` add an `implementation` dependency on `aws-cloudwatch`, as it will need to do in the next phase anyways.

mattcreaser (approved):

PR #3383, opened 2026-08-19
chore: Fix Notify Pull Request workflow for PRs from forks

mattcreaser:

*Issue #, if available:* N/A

*Description of changes:*

The **Notify Pull Request** workflow always failed for pull requests opened from forks. It was triggered by the `pull_request` event, which runs fork-originated jobs with a read-only token and **no access to repository secrets**, so `secrets.SLACK_PR_WEBHOOK_URL` was empty and the `curl` step failed.

Changes:
- Switch the trigger from `pull_request` to `pull_request_target`. This event runs in the context of the base repository, so the secret is available even for fork PRs. It is safe here because the workflow only reads PR metadata and posts a Slack notification — it never checks out or executes PR code, and the workflow definition that runs is always the trusted version from the base branch (not the fork's copy).
- Harden the notify step while here:
  - Build the JSON payload with `jq` so special characters in the (untrusted) PR title can no longer break the payload.
  - Use `-X POST` explicitly (`POST` was previously parsed by curl as a URL).
  - Add `-f` so a failing webhook surfaces as a failed step instead of passing silently.

Untrusted values (PR title, author, URL) are still passed via `env` and never interpolated into the shell, so there is no command-injection surface.

This is a CI workflow change. Validated the YAML and confirmed the `pull_request_target` semantics: the workflow file that executes comes from the base branch, and secrets are available to fork-originated runs. Full end-to-end verification will occur when the workflow runs on this PR / subsequent fork PRs.

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/.github/workflows/notify_pull_request.yml
+++ b/.github/workflows/notify_pull_request.yml
@@ -1,7 +1,12 @@
 name: Notify Pull Request
 
+# Use pull_request_target so the workflow runs in the context of the base
+# repository and has access to secrets (e.g. SLACK_PR_WEBHOOK_URL) even for
+# pull requests opened from forks. This is safe because the workflow does not
+# check out or execute any PR code — it only reads PR metadata and posts a
+# Slack notification.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, ready_for_review, reopened]
 
 permissions:
@@ -19,4 +24,12 @@ jobs:
           TITLE: ${{ github.event.pull_request.title }}
           USER: ${{ github.event.pull_request.user.login }}
         shell: bash
-        run: curl -s POST "$WEBHOOK_URL" -H "Content-Type:application/json" --data "{\"url\":\"$URL\", \"title\":\"$TITLE\", \"user\":\"$USER\"}"
+        run: |
+          payload="$(jq -n \
+            --arg url "$URL" \
+            --arg title "$TITLE" \
+            --arg user "$USER" \
+            '{url: $url, title: $title, user: $user}')"
+          curl -sf -X POST "$WEBHOOK_URL" \
+            -H "Content-Type: application/json" \
+            --data "$payload"
--- a/.github/workflows/notify_pull_request.yml
+++ b/.github/workflows/notify_pull_request.yml
@@ -5,12 +5,20 @@ name: Notify Pull Request
 # pull requests opened from forks. This is safe because the workflow does not
 # check out or execute any PR code — it only reads PR metadata and posts a
 # Slack notification.
+#
+# Two consequences of pull_request_target worth noting:
+#   - The workflow definition that runs comes from the PR's base branch, so
+#     changes to this file only take effect once merged to that branch.
+#   - Unlike pull_request, runs are not gated behind maintainer approval for
+#     first-time contributors, so notifications fire immediately on fork PRs.
 on:
   pull_request_target:
     types: [opened, ready_for_review, reopened]
 
-permissions:
-  contents: read
+# This job only posts to an external webhook — it makes no GitHub API calls and
+# checks out no code, so the GITHUB_TOKEN needs no scopes. This matters under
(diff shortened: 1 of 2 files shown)


.github/workflows/notify_pull_request.yml
```diff
@@ -19,4 +24,12 @@ jobs:
...
-        run: curl -s POST "$WEBHOOK_URL" -H "Content-Type:application/json" --data "{\"url\":\"$URL\", \"title\":\"$TITLE\", \"user\":\"$USER\"}"
+        run: |
+          payload="$(jq -n \
+            --arg url "$URL" \
+            --arg title "$TITLE" \
+            --arg user "$USER" \
+            '{url: $url, title: $title, user: $user}')"
+          curl -sf -X POST "$WEBHOOK_URL" \
+            -H "Content-Type: application/json" \
+            --data "$payload"
```
mattcreaser: 🟡 **Warning — `-sf` makes the failure undiagnosable**

The PR description's goal is "a failing webhook surfaces as a failed step instead of passing silently." `-f` achieves the failure, but `-s` also suppresses curl's error message, so the step fails with an empty log.

Reproduced locally:
```
$ curl -sf  -X POST https://httpbin.org/status/500 --data '{}'   # exit=56, no output at all
$ curl -sSf -X POST https://httpbin.org/status/500 --data '{}'   # exit=56, "curl: (56) The requested URL returned error: 500"
```

`-S` re-enables error reporting while keeping the progress meter suppressed. Same applies if the secret is ever unset — `curl -sf -X POST ""` exits 2 silently.

```suggestion
          curl -sSf -X POST "$WEBHOOK_URL" \
            -H "Content-Type: application/json" \
            --data "$payload"
```

If you also want the Slack response body on failure, `--fail-with-body` (curl ≥ 7.76, available on `ubuntu-latest`) is the stronger option.

.github/workflows/notify_pull_request.yml
```diff
@@ -19,4 +24,12 @@ jobs:
...
           USER: ${{ github.event.pull_request.user.login }}
         shell: bash
-        run: curl -s POST "$WEBHOOK_URL" -H "Content-Type:application/json" --data "{\"url\":\"$URL\", \"title\":\"$TITLE\", \"user\":\"$USER\"}"
+        run: |
+          payload="$(jq -n \
+            --arg url "$URL" \
+            --arg title "$TITLE" \
+            --arg user "$USER" \
+            '{url: $url, title: $title, user: $user}')"
+          curl -sf -X POST "$WEBHOOK_URL" \
```
mattcreaser: 🔵 **Suggestion — no request timeout**

There is no `--max-time`, so a hung or blackholed webhook endpoint holds the runner until the 6-hour job limit. Consider `--max-time 30` on this call, or `timeout-minutes: 5` on the `notify` job. Not folded into the suggestion above to keep the two changes separable.

.github/workflows/notify_pull_request.yml
```diff
@@ -1,7 +1,12 @@
...
+# repository and has access to secrets (e.g. SLACK_PR_WEBHOOK_URL) even for
+# pull requests opened from forks. This is safe because the workflow does not
+# check out or execute any PR code — it only reads PR metadata and posts a
+# Slack notification.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, ready_for_review, reopened]
 
 permissions:
```
mattcreaser: 🔵 **Suggestion — tighten permissions now that the token is privileged**

Under `pull_request`, fork-originated runs got a read-only `GITHUB_TOKEN` for free. Under `pull_request_target` the default is read/write, so this block is now the only thing scoping it. It works as-is, but this job makes no API calls and does no checkout, so it needs no token scopes at all:

```suggestion
permissions: {}
```

Optional hardening rather than a defect — `contents: read` is already a long way from the default.

ekjotmultani (approved):

PR #3384, opened 2026-08-19
refactor: decouple lazy loading from request factory internals

mattcreaser:

*Issue #, if available:*

No issue. PR 1 of 3 in Phase 1 of the V3 AppSync client work.

*Description of changes:*

### The goal behind it

The new V3 client needs customers to be able to write `client.query(ModelQuery.get(Todo::class.java, "id-123"))`. Today `ModelQuery` and its siblings live in `aws-api` — the old plugin module — and the new client deliberately does not depend on `aws-api`. So those helpers have to move into `aws-api-appsync`, a module both the old plugin and the new client can use.

That move is the follow-up PR. **This PR is what makes it possible.**

### The obstacle

The helpers cannot move on their own, because all three just delegate to `AppSyncGraphQLRequestFactory` — so the factory has to move too. And the factory was stuck.

Two of its methods were marked `internal`, which in Kotlin means "visible only inside this module". Both were called by the lazy-loading code, which has to *stay* in `aws-api` because it is plugin machinery. The moment the factory crosses the module boundary, those calls stop compiling.

One was worse than that. `buildModelPageQuery` referenced `ApiModelPage`, which lives in `aws-api` and cannot move either — it is declared alongside the lazy-loading runtime classes that use `Amplify.API` directly. So the factory needed a type from `aws-api`, and `aws-api` needed a method from the factory. Neither could leave first.

### How this PR solves it

It cuts those ties while leaving every file exactly where it is.

**Moved some duplicated logic into the factory.** The lazy-loading code had its own copy of the "work out the query variables from the model's schema" loop, which the factory already knew how to do. That is now `buildQueryFromKeyMap`, and the lazy code calls it. The useful side effect: the lazy code no longer mentions `GraphQLRequestVariable`, so that type is free to travel with the factory.

**A method became private.** Once the lazy code stopped calling `buildQueryInternal`, nothing outside the factory was using it — so it went from `internal` to `private`. Worth noting because it means this PR *reduces* cross-module surface rather than adding to it.

**Flipped who knows about the page type.** `buildModelPageQuery` no longer mentions `ApiModelPage`. The caller passes the page class and the factory parameterizes it with the model class. The factory now knows nothing whatsoever about lazy loading, and the circle is gone.

Both new entry points are `@InternalAmplifyApi`, which the binary compatibility validator excludes, so `apiCheck` passes with no `apiDump`.

### Why it is a separate PR

So the follow-up is boring. After this lands, moving those six files is a pure relocation — a reviewer can read that diff as "these files changed directory" instead of trying to spot a real behaviour change buried inside a thousand-line move. All the actual thinking is isolated here, in about 30 lines.

### What it does not change

Nothing customer-facing. No behaviour change and no public API change.

One subtlety is deliberate and worth not "fixing". `ModelSchema.fromModelClass` throws a plain `AmplifyException`, and the lazy-loading code has two different `catch` blocks that treat that differently from an `ApiException`. Making it match the factory's sibling overload, which wraps into an unchecked `IllegalStateException`, would silently change which handler fires — so `buildQueryFromKeyMap` intentionally lets it escape unwrapped.

All tasks run with `--rerun-tasks`:

- `:aws-api:testDebugUnitTest` — **210 tests, 0 failures**, including the lazy-loading and factory suites that cover the touched paths: `ApiLazyModelListTest` (8), `ApiLazyModelReferenceTest` (6), `AppSyncGraphQLRequestFactoryTest` (16), `AppSyncGraphQLRequestAndResponseCPKTest` (13), `ModelQueryTest` (12), `ModelMutationTest` (10), `ModelSubscriptionTest` (8).
- New `AppSyncGraphQLRequestFactoryKeyMapTest` (3) covers `buildQueryFromKeyMap`, which is now the only request-building path lazy loading takes: a single-field primary key, a composite key asserting each schema-resolved GraphQL type, and equivalence with the `objectId` overload.
- `:aws-api:apiCheck` — passes unmodified, confirming no public API change.
- `:aws-api:ktlintCheck`, `:aws-api:checkstyle`, `:aws-datastore:compileDebugKotlin`, and `:aws-appsync:` compile / test / `apiCheck` / ktlint.

Note for anyone reproducing locally: `./gradlew :aws-api:build` fails at `:core:lintAnalyzeDebugAndroidTest` on a dangling gitignored symlink (`core/src/androidTest/res/raw/amplifyconfiguration.json` → `~/.aws-amplify/...`). It reproduces on the unmodified base branch with these changes stashed, so it is environmental and unrelated.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests — not applicable; internal refactor with no behaviour change

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/ApiModelListTypes.kt
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/ApiModelListTypes.kt
@@ -28,7 +28,6 @@ import com.amplifyframework.core.model.PaginationToken
 import com.amplifyframework.core.model.query.predicate.QueryField
 import com.amplifyframework.core.model.query.predicate.QueryPredicate
 import com.amplifyframework.core.model.query.predicate.QueryPredicates
-import com.amplifyframework.util.TypeMaker
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -91,7 +90,7 @@ internal class ApiLazyModelList<out M : Model> constructor(
         AppSyncGraphQLRequestFactory.buildModelPageQuery(
             clazz,
             queryPredicate,
-            TypeMaker.getParameterizedType(ApiModelPage::class.java, clazz),
+            ApiModelPage::class.java,
             (paginationToken as? ApiPaginationToken)?.nextToken
         )
 
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/ApiModelListTypes.kt
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/ApiModelListTypes.kt
@@ -28,6 +28,7 @@ import com.amplifyframework.core.model.PaginationToken
 import com.amplifyframework.core.model.query.predicate.QueryField
 import com.amplifyframework.core.model.query.predicate.QueryPredicate
 import com.amplifyframework.core.model.query.predicate.QueryPredicates
+import com.amplifyframework.util.TypeMaker
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -90,6 +91,7 @@ internal class ApiLazyModelList<out M : Model> constructor(
         AppSyncGraphQLRequestFactory.buildModelPageQuery(
             clazz,
             queryPredicate,
+            TypeMaker.getParameterizedType(ApiModelPage::class.java, clazz),
             (paginationToken as? ApiPaginationToken)?.nextToken
         )
 
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
@@ -15,6 +15,7 @@
 package com.amplifyframework.api.aws
 
 import com.amplifyframework.AmplifyException
+import com.amplifyframework.annotations.InternalAmplifyApi
 import com.amplifyframework.api.graphql.GraphQLRequest
 import com.amplifyframework.api.graphql.MutationType
 import com.amplifyframework.api.graphql.Operation
@@ -189,7 +190,37 @@ object AppSyncGraphQLRequestFactory {
         }
     }
 
-    internal fun <R, T : Model, P : ModelPath<T>> buildQueryInternal(
+    /**
+     * Creates a [GraphQLRequest] that represents a query for a single value, identified by a map of
+     * primary/sort key field names to values.
+     *
+     * This is the entry point used by lazy loading, where the identifying fields are already known as a
+     * map rather than as a [ModelIdentifier]. The variable types are resolved from the model schema.
+     *
+     * @param modelClass the model class.
+     * @param keyMap the primary index field names mapped to their values.
+     * @param <R> the response type.
+     * @param <T> the concrete model type.
+     * @return a valid [GraphQLRequest] instance.
+     * @throws AmplifyException when a schema cannot be generated for [modelClass].
+     * @throws IllegalStateException when the model schema does not contain the expected information.
+     </T></R> */
+    @JvmStatic
+    @InternalAmplifyApi
+    fun <R, T : Model> buildQueryFromKeyMap(modelClass: Class<T>, keyMap: Map<String, Any>): GraphQLRequest<R> {
+        val modelSchema = ModelSchema.fromModelClass(modelClass)
+        val variables = modelSchema.primaryIndexFields.map { key ->
+            // Find target field to pull type info
+            val targetField = requireNotNull(modelSchema.fields[key])
+            val requiredSuffix = if (targetField.isRequired) "!" else ""
+            val targetTypeString = "${targetField.targetType}$requiredSuffix"
+            val value = requireNotNull(keyMap[key])
+            GraphQLRequestVariable(key, value, targetTypeString)
+        }
+        return buildQueryInternal(modelClass, null, *variables.toTypedArray())
+    }
+
+    private fun <R, T : Model, P : ModelPath<T>> buildQueryInternal(
         modelClass: Class<T>,
         includes: ((P) -> List<PropertyContainerPath>)?,
         vararg variables: GraphQLRequestVariable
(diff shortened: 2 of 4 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
```diff
@@ -189,7 +190,37 @@ object AppSyncGraphQLRequestFactory {
...
+     * @param keyMap the primary index field names mapped to their values.
+     * @param <R> the response type.
+     * @param <T> the concrete model type.
+     * @return a valid [GraphQLRequest] instance.
+     * @throws AmplifyException when a schema cannot be generated for [modelClass].
+     * @throws IllegalStateException when the model schema does not contain the expected information.
+     </T></R> */
+    @JvmStatic
+    @InternalAmplifyApi
+    fun <R, T : Model> buildQueryFromKeyMap(modelClass: Class<T>, keyMap: Map<String, Any>): GraphQLRequest<R> {
```
mattcreaser: **WARNING - `buildQueryFromKeyMap` needs `@Throws(AmplifyException::class)` for Java callers**

`AmplifyException extends Exception` (`core/src/main/java/com/amplifyframework/AmplifyException.java:29`) - it is *checked*. This is the one entry point in this file that deliberately lets it escape unwrapped, and without a `@Throws` the generated method has no `throws` clause, so a Java caller cannot write `catch (AmplifyException e)` - javac rejects it as "exception is never thrown in body of corresponding try statement".

That matters specifically because the point of `@InternalAmplifyApi` here is cross-module consumption (the same marker gates the Amplify Flutter bridge), and `aws-api` already uses `@Throws` for exactly this reason elsewhere (`ApiQuery.kt:29`, `AWSGraphQLOperation.kt:39`). Every sibling `buildQuery` overload sidesteps the issue by wrapping into unchecked `IllegalStateException`; this one intentionally does not, so it is the one that needs the annotation.

```suggestion
    @JvmStatic
    @InternalAmplifyApi
    @Throws(AmplifyException::class)
    fun <R, T : Model> buildQueryFromKeyMap(modelClass: Class<T>, keyMap: Map<String, Any>): GraphQLRequest<R> {
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
```diff
@@ -231,14 +262,29 @@ object AppSyncGraphQLRequestFactory {
...
+    fun <R, T : Model> buildModelPageQuery(
         modelClass: Class<T>,
         predicate: QueryPredicate,
+        pageType: Type,
         pageToken: String?
-    ): GraphQLRequest<R> {
-        val dataType = TypeMaker.getParameterizedType(ApiModelPage::class.java, modelClass)
-        return buildListQueryInternal(modelClass, predicate, DEFAULT_QUERY_LIMIT, dataType, null, pageToken)
-    }
+    ): GraphQLRequest<R> = buildListQueryInternal(modelClass, predicate, DEFAULT_QUERY_LIMIT, pageType, null, pageToken)
```
mattcreaser: **WARNING - `pageType: Type` widens the contract more than the decoupling requires**

Breaking the `ApiModelPage` dependency is the right call, but exporting a bare `java.lang.reflect.Type` means nothing constrains the caller to pass a `ModelPage` parameterization at all - and because the page type only matters at *deserialization* time (`registerTypeHierarchyAdapter(ModelPage::class.java, ...)`), a wrong type fails at runtime rather than compile time.

Note the factory still imports and uses `TypeMaker` for `PaginatedResult` (`buildQuery`, `buildPaginatedResultQuery`), so keeping the `TypeMaker` call inside the factory costs nothing in coupling terms - only `ApiModelPage` had to leave. Taking the page *class* and parameterizing internally keeps the cycle broken, narrows what a caller can get wrong, and stops the `TypeMaker.getParameterizedType(..., clazz)` knowledge from being duplicated at each call site:

```suggestion
    fun <R, T : Model> buildModelPageQuery(
        modelClass: Class<T>,
        predicate: QueryPredicate,
        pageClass: Class<*>,
        pageToken: String?
    ): GraphQLRequest<R> = buildListQueryInternal(
        modelClass,
        predicate,
        DEFAULT_QUERY_LIMIT,
        TypeMaker.getParameterizedType(pageClass, modelClass),
        null,
        pageToken
    )
```

The call site then becomes `buildModelPageQuery(clazz, queryPredicate, ApiModelPage::class.java, token)` and `ApiModelListTypes.kt` no longer needs the new `TypeMaker` import. If you would rather tighten it further, `Class<out ModelPage<*>>` makes the intent explicit - `ModelPage` lives in `core`, which the factory already depends on. The KDoc `@param pageType` line needs the matching rename either way.

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
```diff
@@ -189,7 +190,37 @@ object AppSyncGraphQLRequestFactory {
...
+     * This is the entry point used by lazy loading, where the identifying fields are already known as a
+     * map rather than as a [ModelIdentifier]. The variable types are resolved from the model schema.
+     *
+     * @param modelClass the model class.
+     * @param keyMap the primary index field names mapped to their values.
+     * @param <R> the response type.
+     * @param <T> the concrete model type.
+     * @return a valid [GraphQLRequest] instance.
+     * @throws AmplifyException when a schema cannot be generated for [modelClass].
+     * @throws IllegalStateException when the model schema does not contain the expected information.
```
mattcreaser: **SUGGESTION - KDoc omits `IllegalArgumentException`**

Both `requireNotNull` calls below throw `IllegalArgumentException`, not `IllegalStateException`, and neither is caught here - the `IllegalStateException` in this `@throws` only comes from `buildQueryInternal`'s wrapping. The `keyMap[key]` one is genuinely caller-controllable (a `keyMap` missing a primary index field), so it is worth stating on a newly public entry point.

```suggestion
     * @throws IllegalStateException when the model schema does not contain the expected information.
     * @throws IllegalArgumentException when [keyMap] has no value for a primary index field.
```

aws-api/src/main/java/com/amplifyframework/api/aws/AppSyncGraphQLRequestFactory.kt
```diff
@@ -189,7 +190,37 @@ object AppSyncGraphQLRequestFactory {
...
+     </T></R> */
+    @JvmStatic
+    @InternalAmplifyApi
+    fun <R, T : Model> buildQueryFromKeyMap(modelClass: Class<T>, keyMap: Map<String, Any>): GraphQLRequest<R> {
+        val modelSchema = ModelSchema.fromModelClass(modelClass)
+        val variables = modelSchema.primaryIndexFields.map { key ->
```
mattcreaser: **SUGGESTION - this schema-to-variables loop now exists three times in this file**

Identical bodies at lines 124-135 and 171-182 (the `ModelIdentifier` overloads) differ only in how `value` is derived. Since this PR is already consolidating one copy out of `ApiLazyModelReference`, a private helper would fold in the other two:

```kotlin
private fun variablesFor(
    modelSchema: ModelSchema,
    valueFor: (index: Int, key: String) -> Any
) = modelSchema.primaryIndexFields.mapIndexed { i, key ->
    val targetField = requireNotNull(modelSchema.fields[key])
    val requiredSuffix = if (targetField.isRequired) "!" else ""
    GraphQLRequestVariable(key, valueFor(i, key), "${targetField.targetType}$requiredSuffix")
}
```

Entirely optional and arguably out of scope for a decoupling PR - but it is cheap while the file is already open, and it shrinks what has to move in the follow-up.

aws-api/src/main/java/com/amplifyframework/api/aws/ApiModelListTypes.kt
```diff
@@ -90,6 +91,7 @@ internal class ApiLazyModelList<out M : Model> constructor(
         AppSyncGraphQLRequestFactory.buildModelPageQuery(
             clazz,
             queryPredicate,
+            TypeMaker.getParameterizedType(ApiModelPage::class.java, clazz),
```
mattcreaser: **SUGGESTION - no test pins the page type now that the caller owns it**

Before this change the parameterization was the factory's invariant and unit-testable there; now it is this call site's responsibility, and `ApiLazyModelListTest` mocks the `query(...)` call, so nothing asserts the request's `responseType`. A caller that passed the wrong type would still pass all 8 existing tests and only fail at deserialization.

A short assertion over `createRequest(...)`'s `responseType` (expecting `TypeMaker.getParameterizedType(ApiModelPage::class.java, clazz)`) would guard the newly caller-owned invariant, and would be exactly the regression test that protects the follow-up file move. The same argument applies to `buildQueryFromKeyMap` - `AppSyncGraphQLRequestFactoryTest` covers the sibling overloads but not this new one, even though it is now the only path lazy loading takes.

thisisabhash (approved):

PR #3387, opened 2026-08-20
fix(api): percent-encode semicolons in REST query parameter values

VarshithaPamisetty:
## Description

`RestRequestFactory` builds the URL then `Uri.decode`s it, and OkHttp leaves `;` unencoded. API Gateway treats a raw `;` as a query delimiter, so a value like `"X;Y"` was truncated to `"X"`. This encodes `;` to `%3B` in the query so the value arrives intact.

Android counterpart of the Swift fix aws-amplify/amplify-swift#4274; raised after @thisisabhash flagged the same gap on Android during that review.

## Testing

- Regression tests (raw and pre-encoded `;`) on the built URL and on the wire via MockWebServer; all fail without the change.
- aws-api unit suite 210/210; `ktlintCheck`/`checkstyle`/`apiCheck` clean.
- Integration tests not run locally (need a provisioned AWS backend).

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
@@ -74,8 +74,19 @@ public final class RestRequestFactory {
         }
 
         try {
-            String encodedUrl = builder.build().url().toString();
-            return new URL(Uri.decode(encodedUrl));
+            String decodedUrl = Uri.decode(builder.build().url().toString());
+            // API Gateway treats an unencoded ';' in the query as a parameter delimiter
+            // and drops everything after it, so encode it explicitly (query only).
+            // https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html
+            int queryStart = decodedUrl.indexOf('?');
+            if (queryStart >= 0) {
+                int fragmentStart = decodedUrl.indexOf('#', queryStart);
+                int queryEnd = fragmentStart < 0 ? decodedUrl.length() : fragmentStart;
+                decodedUrl = decodedUrl.substring(0, queryStart + 1)
+                    + decodedUrl.substring(queryStart + 1, queryEnd).replace(";", "%3B")
+                    + decodedUrl.substring(queryEnd);
+            }
+            return new URL(decodedUrl);
         } catch (MalformedURLException error) {
             throw new MalformedURLException(error.getMessage());
         }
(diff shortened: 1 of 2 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/utils/RestRequestFactory.java
```diff
@@ -74,8 +74,19 @@ public static URL createURL(
...
         try {
-            String encodedUrl = builder.build().url().toString();
-            return new URL(Uri.decode(encodedUrl));
+            String decodedUrl = Uri.decode(builder.build().url().toString());
+            // API Gateway treats an unencoded ';' in the query as a parameter delimiter
+            // and drops everything after it, so encode it explicitly (query only).
+            // https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html
+            int queryStart = decodedUrl.indexOf('?');
+            if (queryStart >= 0) {
+                int fragmentStart = decodedUrl.indexOf('#', queryStart);
```
mattcreaser: The fragment carve-out defeats the fix when a query *value* contains a `#` before a `;`. This factory intentionally decodes `#` back to a raw `#` inside the query (the existing `RestRequestFactoryTest.createValidURLWithSpecialCharacters` asserts exactly this), so `indexOf('#', queryStart)` will latch onto a query-embedded `#` and treat everything after it as a "fragment" that is left untouched.

Concretely, `createURL(endpoint, path, mapOf("filter" to "a#b;c"))` produces `...?filter=a#b;c`; `fragmentStart` points at the `#`, `queryEnd` stops there, and the `;` in `b;c` is never encoded — so the exact truncation this PR fixes still occurs.

Since REST URLs from this factory never carry a real fragment, the simplest fix is to drop the fragment special-casing and re-encode `;` across the entire substring after the first `?`:

```java
int queryStart = decodedUrl.indexOf('?');
if (queryStart >= 0) {
    decodedUrl = decodedUrl.substring(0, queryStart + 1)
        + decodedUrl.substring(queryStart + 1).replace(";", "%3B");
}
```

This is both simpler and handles the `#`-before-`;` case correctly.

aws-api/src/test/java/com/amplifyframework/api/aws/utils/RestQuerySemicolonEncodingTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        val url = RestRequestFactory.createURL(
+            "http://amplify-android.com",
+            "/path",
+            mapOf("filter" to "X%3BY")
+        )
+        url.toString() shouldBe "http://amplify-android.com/path?filter=X%3BY"
+    }
+
+    @Test
+    fun `semicolon in a query value is percent-encoded on the wire`() {
```
mattcreaser: Consider adding a regression case where a query value contains a `#` before the `;` (e.g. `mapOf("filter" to "a#b;c")`). With the current fragment carve-out the `;` is left raw; that test would fail today and would lock in the corrected behavior once the inline suggestion above is applied.

thisisabhash (approved):

PR #3388, opened 2026-08-20
fix(auth): recover device metadata keyed by a pre-2.30.3 username

cadivus:

*Issue #, if available:* n/a

*Description of changes:*

Device metadata was keyed in the credential store by the value typed at sign-in until 2.30.2, and by the Cognito username from 2.30.3 onward (#3149). On a device-tracking pool where users sign in with an alias (e.g. an email), upgrading an app across that boundary orphans the stored entry: the refresh path looks it up under the Cognito username, misses, omits `DeviceKey`, and Cognito rejects the call with `NotAuthorizedException: Invalid Refresh Token.` That surfaces as `SessionExpiredException`, so every affected user is signed out once and has to re-authenticate.

On a lookup miss, the refresh path now retries the alias values from the user's own id token (`email`, `phone_number`, `preferred_username`), migrates the entry to the canonical key and clears the stale one. Affected installs self-heal on their next refresh. Alias values come from the user's own token, so a migrated entry always belongs to that user. `signInAliases` never throws, so a token that cannot be parsed simply yields no candidates and refresh behaves as before.

- 11 new unit tests (`AuthEnvironmentDeviceMetadataTest`, `IdTokenSignInAliasesTest`); full `:aws-auth-cognito:testDebugUnitTest` suite green (551 tests), plus `ktlintCheck` and `apiCheck`.
- On an emulator against a device-tracking user pool, with an identical pre-upgrade state (sign in on 2.30.0 + one refresh): an in-place upgrade to unpatched `main` fails the next refresh with `Invalid Refresh Token.`, while the patched build succeeds and logs the migration once. Two further refreshes also succeed, confirming the migration persisted.
- Regression checks: clean install is unaffected (no migration, refreshes succeed), and a genuinely forgotten device still fails with `Invalid Refresh Token.` — the fallback does not mask real device-binding failures.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
@@ -106,12 +106,40 @@ internal class AuthEnvironment internal constructor(
         return userContextDataProvider?.getEncodedContextData(username, deviceId)
     }
 
-    suspend fun getDeviceMetadata(username: String): DeviceMetadata.Metadata? {
-        var deviceCredentials =
+    /**
+     * Loads device metadata for [username].
+     *
+     * SDK versions <= 2.30.2 keyed device metadata by the value the user typed at sign-in (an alias
+     * such as an email) rather than by the Cognito username. On those installs the metadata is
+     * unreachable under the current key, so the DeviceKey is omitted from refresh and Cognito
+     * rejects it with "Invalid Refresh Token.". [legacyUsernames] are alias values taken from the
+     * user's own tokens; the first match is migrated to [username] so this happens at most once.
+     */
+    suspend fun getDeviceMetadata(
+        username: String,
+        legacyUsernames: List<String> = emptyList()
+    ): DeviceMetadata.Metadata? {
+        loadDeviceMetadata(username)?.let { return it }
+
+        for (legacyUsername in legacyUsernames.filter { it != username }) {
+            val legacyMetadata = loadDeviceMetadata(legacyUsername) ?: continue
+            logger.info("Migrating device metadata stored by an earlier SDK version.")
+            credentialStoreClient.storeCredentials(
+                CredentialType.Device(username),
+                AmplifyCredential.DeviceData(legacyMetadata)
+            )
+            credentialStoreClient.clearCredentials(CredentialType.Device(legacyUsername))
+            return legacyMetadata
+        }
+        return null
+    }
+
+    private suspend fun loadDeviceMetadata(username: String): DeviceMetadata.Metadata? {
+        val deviceCredentials =
             credentialStoreClient.loadCredentials(CredentialType.Device(username)) as? AmplifyCredential.DeviceData
         if (deviceCredentials == null) {
             logger.warn("loadCredentials returned unexpected AmplifyCredential Type.")
-            deviceCredentials = AmplifyCredential.DeviceData(DeviceMetadata.Empty)
+            return null
         }
         return deviceCredentials.deviceMetadata as? DeviceMetadata.Metadata
     }
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
@@ -44,7 +44,9 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
             val evt = try {
                 val username = signedInData.username
                 val tokens = signedInData.cognitoUserPoolTokens
-                val deviceMetadata: DeviceMetadata.Metadata? = getDeviceMetadata(username)
+                // Alias candidates let us recover device metadata keyed by an older SDK version.
+                val deviceMetadata: DeviceMetadata.Metadata? =
+                    getDeviceMetadata(username, tokens.idToken?.signInAliases ?: emptyList())
 
                 val response = cognitoAuthService.cognitoIdentityProviderClient?.getTokensFromRefreshToken {
                     refreshToken = tokens.refreshToken?.tokenValue
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
@@ -76,7 +76,10 @@ internal abstract class Jwt {
         UserSub("sub"),
         Username("username"), // username claim in the access token
         TokenRevocationId("origin_jti"),
-        CognitoUsername("cognito:username") // username claim in the identity token
+        CognitoUsername("cognito:username"), // username claim in the identity token
+        Email("email"),
+        PhoneNumber("phone_number"),
+        PreferredUsername("preferred_username")
     }
 }
 
@@ -87,6 +90,19 @@ internal class IdToken(override val tokenValue: String) : Jwt() {
         get() = getClaim(Claim.UserSub)
     val username: String?
         get() = getClaim(Claim.CognitoUsername)
+
+    // Sign-in aliases. Used to locate device metadata written by older SDK versions, which keyed it
+    // by the value the user typed at sign-in rather than by the Cognito username. Never throws: a
+    // token we cannot parse simply yields no candidates.
+    val signInAliases: List<String>
+        get() = runCatching {
+            listOfNotNull(
+                getClaim(Claim.Email),
+                getClaim(Claim.PhoneNumber),
+                getClaim(Claim.PreferredUsername)
+            )
+        }.getOrDefault(emptyList())
+
     val expiration: Instant? by lazy {
         getClaim(Claim.Expiration)?.let { Instant.ofEpochSecond(it.toLong()) }
     }
--- a/aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
+++ b/aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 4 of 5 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,12 +106,40 @@ internal class AuthEnvironment internal constructor(
...
+            val legacyMetadata = loadDeviceMetadata(legacyUsername) ?: continue
+            logger.info("Migrating device metadata stored by an earlier SDK version.")
+            credentialStoreClient.storeCredentials(
+                CredentialType.Device(username),
+                AmplifyCredential.DeviceData(legacyMetadata)
+            )
+            credentialStoreClient.clearCredentials(CredentialType.Device(legacyUsername))
+            return legacyMetadata
+        }
+        return null
```
mattcreaser: :yellow_circle: **New failure mode for a user who has no device metadata of their own**

The alias-uniqueness invariant holds at a point in time, but not across attribute changes. Consider:

1. User A signs in on 2.30.0 with `a@example.com`; metadata is stored under `a@example.com`.
2. User A changes their email to `new@example.com`, freeing the old value.
3. User B registers/claims `a@example.com` and signs in on this device.

B has no entry under their own canonical key, so the loop runs, finds A's entry under `a@example.com`, and B's refresh is sent with A's `DeviceKey`. Cognito rejects a `DeviceKey` that is not bound to B, so B gets `NotAuthorizedException` -> `SessionExpiredException`.

Before this change B's refresh omitted `DeviceKey` entirely and **succeeded** (B never remembered a device). So this path can sign out a user who was previously fine, which is the same symptom the PR is fixing.

Worth considering: retry the refresh once without `DeviceKey` when a migrated (as opposed to canonically-keyed) `DeviceKey` is rejected. That keeps the "genuinely forgotten device still fails" property you verified for canonical entries, while making a bad alias match self-correcting rather than terminal.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
```diff
@@ -87,6 +90,19 @@ internal class IdToken(override val tokenValue: String) : Jwt() {
...
+    // by the value the user typed at sign-in rather than by the Cognito username. Never throws: a
+    // token we cannot parse simply yields no candidates.
+    val signInAliases: List<String>
+        get() = runCatching {
+            listOfNotNull(
+                getClaim(Claim.Email),
+                getClaim(Claim.PhoneNumber),
+                getClaim(Claim.PreferredUsername)
+            )
+        }.getOrDefault(emptyList())
```
mattcreaser: :large_blue_circle: **`runCatching` swallows more than intended**

`runCatching` catches `Throwable`, so this also absorbs `OutOfMemoryError`, `StackOverflowError`, and — if this property is ever read from inside a coroutine — `CancellationException`. `getClaim` only ever throws `UnknownException`, so catching that is both precise and self-documenting about what "cannot be parsed" means.

```suggestion
        get() = try {
            listOfNotNull(
                getClaim(Claim.Email),
                getClaim(Claim.PhoneNumber),
                getClaim(Claim.PreferredUsername)
            )
        } catch (e: UnknownException) {
            emptyList()
        }
```

`UnknownException` is already imported in this file.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
```diff
@@ -76,7 +76,10 @@ internal abstract class Jwt {
         UserSub("sub"),
         Username("username"), // username claim in the access token
         TokenRevocationId("origin_jti"),
-        CognitoUsername("cognito:username") // username claim in the identity token
+        CognitoUsername("cognito:username"), // username claim in the identity token
+        Email("email"),
+        PhoneNumber("phone_number"),
+        PreferredUsername("preferred_username")
```
mattcreaser, line 82: :large_blue_circle: **Annotate which token these claims belong to**

`Claim` is shared by `IdToken` and `AccessToken`, and the neighbouring entries carry a comment for exactly this reason (`Username` — access token, `CognitoUsername` — identity token). All three new claims are id-token-only.

```suggestion
        Email("email"), // id token only
        PhoneNumber("phone_number"), // id token only
        PreferredUsername("preferred_username") // id token only
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,12 +106,40 @@ internal class AuthEnvironment internal constructor(
...
+     * rejects it with "Invalid Refresh Token.". [legacyUsernames] are alias values taken from the
+     * user's own tokens; the first match is migrated to [username] so this happens at most once.
+     */
+    suspend fun getDeviceMetadata(
+        username: String,
+        legacyUsernames: List<String> = emptyList()
+    ): DeviceMetadata.Metadata? {
+        loadDeviceMetadata(username)?.let { return it }
+
+        for (legacyUsername in legacyUsernames.filter { it != username }) {
```
mattcreaser, line 124: :large_blue_circle: **Deduplicate the candidates**

`preferred_username` is commonly set to the same value as `email`, in which case this reads the same credential-store key twice. Each `loadCredentials` is a round trip through the credential-store state machine plus a keystore decrypt, so it is not free.

```suggestion
        for (legacyUsername in legacyUsernames.distinct().filter { it != username }) {
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,12 +106,40 @@ internal class AuthEnvironment internal constructor(
...
+    private suspend fun loadDeviceMetadata(username: String): DeviceMetadata.Metadata? {
+        val deviceCredentials =
             credentialStoreClient.loadCredentials(CredentialType.Device(username)) as? AmplifyCredential.DeviceData
         if (deviceCredentials == null) {
             logger.warn("loadCredentials returned unexpected AmplifyCredential Type.")
-            deviceCredentials = AmplifyCredential.DeviceData(DeviceMetadata.Empty)
+            return null
         }
         return deviceCredentials.deviceMetadata as? DeviceMetadata.Metadata
     }
```
mattcreaser, line 145: :large_blue_circle: **This null branch is unreachable for `CredentialType.Device`, and the message is misleading**

`CredentialStoreCognitoActions` unconditionally wraps device retrieval as `AmplifyCredential.DeviceData(credentialStore.retrieveDeviceMetadata(...))`, so the `as?` cast cannot fail here — a missing entry arrives as `DeviceData(DeviceMetadata.Empty)` and falls out of the final `as? DeviceMetadata.Metadata` as `null`. That is why your mock in `storeContains` is faithful, and also why this `return null` is never executed by any test.

The behaviour is equivalent to the previous `DeviceData(DeviceMetadata.Empty)` substitution, so this is not a regression — but the retained `logger.warn("loadCredentials returned unexpected AmplifyCredential Type.")` now sits on dead code while reading as though it reports a missing entry. Either drop it or reword it to reflect that it only fires on a genuine type mismatch.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
```diff
@@ -44,7 +44,9 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
             val evt = try {
                 val username = signedInData.username
                 val tokens = signedInData.cognitoUserPoolTokens
-                val deviceMetadata: DeviceMetadata.Metadata? = getDeviceMetadata(username)
+                // Alias candidates let us recover device metadata keyed by an older SDK version.
+                val deviceMetadata: DeviceMetadata.Metadata? =
+                    getDeviceMetadata(username, tokens.idToken?.signInAliases ?: emptyList())
```
mattcreaser: :large_blue_circle: **Sibling callers are left on the single-argument overload — worth a note in the PR description**

The refresh path is the right place for the primary fix, and I confirmed the sign-in paths do not need it: `SRPCognitoActions`, `DeviceSRPCognitoSignInActions`, `UserAuthSignInCognitoActions`, and `SignInCustomCognitoActions` all pass `event.username`, which on an affected install *is* the legacy key, so they hit it directly.

The two that do miss are `RememberDeviceUseCase` and `ForgetDeviceUseCase` — both read `stateMachine.requireSignedInState().signedInData.username`, i.e. the canonical username, and so get a `null` `deviceKey` on an affected install that has not yet refreshed. The window is narrow because refresh runs early and migrates, but `Amplify.Auth.forgetDevice()` called before the first refresh will silently forget nothing.

Not asking you to widen the change — just calling out that the fix is refresh-triggered, so this is a real if short-lived gap.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    private val environment = AuthEnvironment(
+        mockk<Context>(relaxed = true),
+        mockk(relaxed = true),
+        mockk(relaxed = true),
+        credentialStoreClient,
+        null,
+        null,
+        logger
+    )
```
mattcreaser, line 47: :large_blue_circle: **Use named arguments for the `AuthEnvironment` constructor**

Seven positional arguments, five of which are bare `mockk(relaxed = true)`, means a future reordering of `AuthEnvironment`'s `internal constructor` compiles fine and silently binds the wrong mocks. Naming at least `credentialStoreClient` and `logger` makes the intent explicit and the breakage loud.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        storeContains()
+
+        environment.getDeviceMetadata(username, listOf(email)) shouldBe null
+    }
+
+    @Test
+    fun `falls back to metadata stored under a legacy alias key`() = runTest {
+        storeContains(email to metadata)
+
+        environment.getDeviceMetadata(username, listOf(email)) shouldBe metadata
```
mattcreaser, line 75: :large_blue_circle: **Two gaps in an otherwise thorough suite**

1. **Candidate ordering / first-match-wins.** Every test passes a single-element `listOf(email)`, so the loop's "try each in turn, stop at the first hit" behaviour is never exercised. A case like `storeContains(preferredUsername to metadata)` with `listOf(email, phone, preferredUsername)` would cover the `?: continue` path.

2. **The wiring is untested.** Nothing asserts that `FetchAuthSessionCognitoActions` actually forwards `tokens.idToken?.signInAliases`. Both halves are well covered in isolation, but deleting that argument at the call site would leave all 11 new tests green while restoring the bug. Given this is a regression fix, a test pinning the call site is the one that stops it coming back.

PR #3389, opened 2026-08-20
chore: add cloudwatch client integration tests

thisisabhash:

*Issue #, if available:*

N/A

*Description of changes:*

Adds connected integration tests for the standalone v3 CloudWatch client (`AmplifyCloudWatchClient`), structured after the `:aws-kinesis` client's instrumentation tests. (Stacked on the client implementation — this PR is only the tests.)

- `AmplifyCloudWatchClientInstrumentationTest` (`:aws-cloudwatch/src/androidTest`, extends `DeviceFarmTestBase`) exercises the real client against a provisioned CloudWatch Logs backend:
  - `testGetEscapeHatch` — `getCloudWatchLogsClient()` returns the underlying AWS SDK client.
  - `testFlushLogWithMessages` — emitted messages are flushed and land in the log group.
  - `testFlushLogWithVerboseMessageAfterEnabling` — a verbose message emitted while enabled is delivered.
  - `testFlushLogWithVerboseMessageAfterDisabling` — a verbose message emitted while disabled is dropped (not delivered).

  Delivery is confirmed by reading the log group back via `filterLogEvents`, polling with a re-flush between attempts since CloudWatch ingestion is eventually consistent.
- Credentials come from the provisioned backend's Cognito identity pool (guest access — no sign-in). Backend config is read from two gitignored, CI-injected resources under `src/androidTest/res/raw/`: `amplify_outputs.json` (Auth) and `amplifyconfiguration_logging.json` (region, log group, etc.). The CI test-config step must supply both, and the target log group must be provisioned.
- Adds `:core`, `:aws-core`, and `:aws-auth-cognito` as `androidTestImplementation` dependencies for the Amplify Auth / Cognito credential setup.

- Ran `./gradlew :aws-cloudwatch:connectedDebugAndroidTest` on an emulator against the provisioned backend — 10/10 connected tests pass (these 4 client tests plus the module's 6 existing DB instrumentation tests).

- [x] No

*General Checklist*
- [ ] Added Unit Tests
- [x] Added Integration Tests

### DIFF
--- a/aws-cloudwatch/build.gradle.kts
+++ b/aws-cloudwatch/build.gradle.kts
@@ -22,13 +22,6 @@ apply(from = rootProject.file("configuration/checkstyle.gradle"))
 
 android {
     namespace = "com.amplifyframework.cloudwatch"
-
-    // Persist app data across connected test methods so the client's integration tests keep a stable device
-    // id and write to a single CloudWatch stream (one per test otherwise). Overrides the shared convention's
-    // clearPackageData=true; the module's own DB instrumentation test self-cleans in @After.
-    defaultConfig {
-        testInstrumentationRunnerArguments["clearPackageData"] = "false"
-    }
 }
 
 dependencies {
--- a/aws-cloudwatch/build.gradle.kts
+++ b/aws-cloudwatch/build.gradle.kts
@@ -22,6 +22,13 @@ apply(from = rootProject.file("configuration/checkstyle.gradle"))
 
 android {
     namespace = "com.amplifyframework.cloudwatch"
+
+    // Persist app data across connected test methods so the client's integration tests keep a stable device
+    // id and write to a single CloudWatch stream (one per test otherwise). Overrides the shared convention's
+    // clearPackageData=true; the module's own DB instrumentation test self-cleans in @After.
+    defaultConfig {
+        testInstrumentationRunnerArguments["clearPackageData"] = "false"
+    }
 }
 
 dependencies {
@@ -44,4 +51,8 @@ dependencies {
 
     androidTestImplementation(libs.bundles.test.android)
     androidTestImplementation(project(":testutils"))
+    // Integration tests obtain guest AWS credentials via Cognito (identity pool) and Amplify.configure.
+    androidTestImplementation(project(":core"))
+    androidTestImplementation(project(":aws-core"))
+    androidTestImplementation(project(":aws-auth-cognito"))
 }
--- a/aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
+++ b/aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+@file:OptIn(ExperimentalAmplifyApi::class)
+
+package com.amplifyframework.cloudwatch
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import aws.sdk.kotlin.services.cloudwatchlogs.model.FilterLogEventsRequest
+import com.amplifyframework.annotations.ExperimentalAmplifyApi
+import com.amplifyframework.auth.CognitoCredentialsProvider
+import com.amplifyframework.auth.cognito.AWSCognitoAuthPlugin
+import com.amplifyframework.cloudwatch.test.R
+import com.amplifyframework.core.Amplify
+import com.amplifyframework.core.configuration.AmplifyOutputs
+import com.amplifyframework.foundation.credentials.AwsCredentials
+import com.amplifyframework.foundation.credentials.AwsCredentialsProvider
+import com.amplifyframework.foundation.credentials.toAwsCredentialsProvider
+import com.amplifyframework.foundation.logging.LogLevel
+import com.amplifyframework.foundation.logging.LogMessage
+import com.amplifyframework.testutils.DeviceFarmTestBase
+import io.kotest.matchers.collections.shouldBeEmpty
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.nulls.shouldNotBeNull
+import io.kotest.matchers.string.shouldContain
+import java.util.UUID
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import org.json.JSONObject
+import org.junit.Before
+import org.junit.BeforeClass
+import org.junit.Test
+import org.junit.runner.RunWith
+
+// Give the async on-device write a moment to land before the first flush.
+private const val SAVE_WAIT_MS = 2_000L
+
+// CloudWatch ingestion is eventually consistent; poll with a re-flush between attempts.
+private const val INGEST_WAIT_MS = 20_000L
+private const val MAX_FLUSH_ATTEMPTS = 6
+
+/**
+ * Integration tests for [AmplifyCloudWatchClient] against a real CloudWatch Logs backend.
+ *
+ * Ported from the Amplify Swift `CloudWatchLoggingClientIntegrationTests`, and structured like the
+ * `:aws-kinesis` client instrumentation tests. Credentials come from the provisioned backend's
+ * Cognito identity pool (guest access — no sign-in required).
+ *
+ * Backend config is supplied at test time (not committed) under `src/androidTest/res/raw/`:
+ * - `amplify_outputs.json` — Auth (Cognito user pool + identity pool)
+ * - `amplifyconfiguration_logging.json` — `{ "cloudWatchClient": { region, logGroupName, ... } }`
+ *
+ * The target log group must be provisioned in the test account.
+ */
+@RunWith(AndroidJUnit4::class)
+class AmplifyCloudWatchClientInstrumentationTest : DeviceFarmTestBase() {
+
+    companion object {
+        private lateinit var credentialsProvider: AwsCredentialsProvider<AwsCredentials>
+        private lateinit var testRegion: String
+        private lateinit var testLogGroupName: String
+        private var testLocalStoreMaxSizeInMB: Int = 1
+        private var testFlushIntervalInSeconds: Long = 60
+        private lateinit var testDefaultLogLevel: LogLevel
+        private var configured = false
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpClass() {
+            val context = ApplicationProvider.getApplicationContext<Context>()
+            if (!configured) {
+                Amplify.Auth.addPlugin(AWSCognitoAuthPlugin())
+                Amplify.configure(AmplifyOutputs(R.raw.amplify_outputs), context)
+                configured = true
+            }
+            credentialsProvider = CognitoCredentialsProvider().toAwsCredentialsProvider()
+
+            val json = context.resources.openRawResource(R.raw.amplifyconfiguration_logging)
+                .bufferedReader().use { it.readText() }
+            val config = JSONObject(json).getJSONObject("cloudWatchClient")
+            testRegion = config.getString("region")
+            testLogGroupName = config.getString("logGroupName")
+            testLocalStoreMaxSizeInMB = config.getInt("localStoreMaxSizeInMB")
+            testFlushIntervalInSeconds = config.getLong("flushIntervalInSeconds")
+            testDefaultLogLevel = logLevelOf(config.getJSONObject("loggingConstraints").getString("defaultLogLevel"))
+        }
+
+        // Config stores levels as e.g. "VERBOSE"; map to the LogLevel enum ("Verbose", "Error", ...).
+        private fun logLevelOf(value: String): LogLevel =
+            LogLevel.valueOf(value.lowercase().replaceFirstChar { it.uppercase() })
+    }
+
+    private lateinit var client: AmplifyCloudWatchClient
+
+    @Before
+    fun setUp() {
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        client = AmplifyCloudWatchClient(
+            context = context,
+            region = testRegion,
+            credentialsProvider = credentialsProvider,
+            options = AmplifyCloudWatchClientOptions {
+                logGroupName = testLogGroupName
+                localStoreMaxSizeInMB = testLocalStoreMaxSizeInMB
+                flushStrategy = FlushStrategy.Interval(testFlushIntervalInSeconds.seconds)
+                loggingConstraints = LoggingConstraints(defaultLogLevel = testDefaultLogLevel)
+            }
+        )
+    }
+
+    /** The escape hatch exposes the underlying AWS CloudWatch Logs client. */
+    @Test
+    fun testGetEscapeHatch() {
+        client.getCloudWatchLogsClient().shouldNotBeNull()
+    }
+
+    /** Emitted messages are flushed to CloudWatch and appear in the log group. */
+    @Test
+    fun testFlushLogWithMessages(): Unit = runBlocking {
+        val namespace = UUID.randomUUID().toString()
+        val message = "this is an error message in the integration test ${System.currentTimeMillis()}"
+
+        client.emit(LogMessage(LogLevel.Error, namespace, message, null))
+        client.emit(LogMessage(LogLevel.Debug, namespace, message, null))
+        client.emit(LogMessage(LogLevel.Warn, namespace, message, null))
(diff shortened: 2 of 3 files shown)


aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+        fun setUpClass() {
+            val context = ApplicationProvider.getApplicationContext<Context>()
+            if (!configured) {
+                Amplify.Auth.addPlugin(AWSCognitoAuthPlugin())
+                Amplify.configure(AmplifyOutputs(R.raw.amplify_outputs), context)
+                configured = true
+            }
+            credentialsProvider = CognitoCredentialsProvider().toAwsCredentialsProvider()
+
+            val json = context.resources.openRawResource(R.raw.amplifyconfiguration_logging)
```
mattcreaser: 🔴 **Critical — androidTest source set does not compile; CI config script not updated.**

```
e: AmplifyCloudWatchClientInstrumentationTest.kt:87:52 Unresolved reference 'raw'.
e: AmplifyCloudWatchClientInstrumentationTest.kt:92:60 Unresolved reference 'raw'.
```

That is `./gradlew :aws-cloudwatch:compileDebugAndroidTestKotlin` on this branch. `R.raw.*` fields are generated from files present at compile time, so with no `aws-cloudwatch/src/androidTest/res/raw/` the whole source set fails to build — which also takes down the module's 6 existing DB instrumentation tests.

The PR description says "the CI test-config step must supply both", but `scripts/pull_backend_config_from_s3` is the mechanism that does that and it is unchanged. Compare the `:aws-kinesis` precedent, which is registered there (lines 68-70). Please add:

```bash
    # CloudWatch
    "aws-cloudwatch/src/androidTest/res/raw/amplify_outputs.json"
    "aws-cloudwatch/src/androidTest/res/raw/amplifyconfiguration_logging.json"
```

and upload both objects to the config bucket. Note `pr_verification.yml` only runs `koverXmlReport` / `ktlintCheck checkstyle apiCheck lint`, so GitHub Actions will stay green and this will only surface in the Device Farm run — i.e. the one pipeline this PR exists to feed.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+    private suspend fun awaitEvents(message: String, expectedCount: Int): List<String> {
+        var events = emptyList<String>()
+        repeat(MAX_FLUSH_ATTEMPTS) { attempt ->
+            client.flushLogs()
+            delay(INGEST_WAIT_MS)
+            events = filterEvents(message, windowMinutes = attempt + 2)
+            if (events.size >= expectedCount) return events
+        }
+        return events
+    }
```
mattcreaser, line 214: 🔴 **Critical — the discarded `flushLogs()` result makes the negative test vacuous.**

`flushLogs()` returns `FlushResult = Result<FlushData, AmplifyCloudWatchException>`; it catches everything and reports failure in the return value rather than throwing. Discarding it means a flush that fails on credentials, a missing log group, or throttling is indistinguishable from a successful one.

That is merely a bad error message for the positive tests, but it is fatal for `testFlushLogWithVerboseMessageAfterDisabling`: with `expectedCount = 0` the `shouldBeEmpty()` assertion holds when every single flush failed. The test would pass against a completely broken client, so it certifies nothing.

```suggestion
    private suspend fun awaitEvents(message: String, expectedCount: Int): List<String> {
        var events = emptyList<String>()
        repeat(MAX_FLUSH_ATTEMPTS) { attempt ->
            client.flushLogs().shouldBeSuccess()
            delay(INGEST_WAIT_MS)
            events = filterEvents(message, windowMinutes = attempt + 2)
            if (events.size >= expectedCount) return events
        }
        return events
    }
```

`shouldBeSuccess` is the existing `:testutils` assertion for `Result`. Even with that in place the disabled-path test has no positive control — consider emitting one `Error`-level message alongside the dropped `Verbose` one and asserting the error arrives while the verbose one does not, so delivery failure can no longer masquerade as correct filtering.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+        val message = "this is a verbose message after enabling ${System.currentTimeMillis()}"
+
+        client.enable()
+        client.emit(LogMessage(LogLevel.Verbose, namespace, message, null))
+        delay(SAVE_WAIT_MS)
+
+        val events = awaitEvents(message, expectedCount = 1)
+
+        events shouldHaveSize 1
+        events.first().lowercase() shouldContain "verbose"
```
mattcreaser: 🟡 **Warning — this assertion cannot fail.**

The intent is to check the client's `"${level.name.lowercase()}/${name}: ..."` prefix, but `message` is `"this is a verbose message after enabling …"` (line 156) — it already contains the word "verbose". The substring check is satisfied by the body regardless of the level actually recorded, so an `Error`-level prefix would pass too.

Anchor it to the prefix instead:

```suggestion
        events shouldHaveSize 1
        events.first() shouldStartWith "verbose/$namespace: "
```

(`io.kotest.matchers.string.shouldStartWith`; this subsumes the namespace assertion on line 167.) Renaming the message payload so it does not contain the level name would also help keep the two independent.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        client = AmplifyCloudWatchClient(
+            context = context,
+            region = testRegion,
+            credentialsProvider = credentialsProvider,
+            options = AmplifyCloudWatchClientOptions {
+                logGroupName = testLogGroupName
+                localStoreMaxSizeInMB = testLocalStoreMaxSizeInMB
+                flushStrategy = FlushStrategy.Interval(testFlushIntervalInSeconds.seconds)
+                loggingConstraints = LoggingConstraints(defaultLogLevel = testDefaultLogLevel)
```
mattcreaser: 🟡 **Warning — the pass/fail threshold is read from external config while the expected counts are hard-coded.**

`defaultLogLevel` comes from `amplifyconfiguration_logging.json`, but `testFlushLogWithMessages` asserts exactly 4 events for `Error`/`Debug`/`Warn`/`Info` and `testFlushLogWithVerboseMessageAfterEnabling` asserts 1 for `Verbose`. `CloudWatchLoggingFilter` drops anything below the threshold, so an edit to the S3-hosted config silently breaks the suite: `Error` yields 1 of 4, and anything above `Verbose` makes the verbose test fail outright. The tests only pass today because the config happens to say `VERBOSE`.

Pin what the assertions depend on:

```suggestion
            options = AmplifyCloudWatchClientOptions {
                logGroupName = testLogGroupName
                localStoreMaxSizeInMB = testLocalStoreMaxSizeInMB
                flushStrategy = FlushStrategy.None
                loggingConstraints = LoggingConstraints(defaultLogLevel = LogLevel.Verbose)
            }
```

This also drops `testDefaultLogLevel` and its `logLevelOf` parsing helper (lines 78, 99, 102-104) — the config's `loggingConstraints` block no longer needs reading. If you would rather keep exercising the config value, derive the expectations from it instead of hard-coding 4 and 1.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+    fun setUp() {
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        client = AmplifyCloudWatchClient(
+            context = context,
+            region = testRegion,
+            credentialsProvider = credentialsProvider,
+            options = AmplifyCloudWatchClientOptions {
+                logGroupName = testLogGroupName
+                localStoreMaxSizeInMB = testLocalStoreMaxSizeInMB
+                flushStrategy = FlushStrategy.Interval(testFlushIntervalInSeconds.seconds)
```
mattcreaser: 🟡 **Warning — interval auto-flush races the explicit flushes.**

`FlushStrategy.Interval` starts a WorkManager sync from `init`, so a background flush can fire between an `emit()` and the `filterEvents()` poll. The `:aws-kinesis` tests deliberately avoid this — every builder there passes `FlushStrategy.None` with the comment *"Don't auto-flush as it may impact other tests"* — except in the two tests specifically covering auto-flush.

Since `awaitEvents` drives flushing explicitly, the interval buys nothing here and adds nondeterminism (it is also why `clearPackageData=false` matters more than it should). Use `FlushStrategy.None`, which makes `testFlushIntervalInSeconds` and its config read (lines 77, 98) dead too. If interval flushing is worth covering, it deserves its own test that asserts delivery *without* calling `flushLogs()`.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+                loggingConstraints = LoggingConstraints(defaultLogLevel = testDefaultLogLevel)
+            }
+        )
+    }
```
mattcreaser, line 106: 🟡 **Warning — no `@After`, so every test leaks a live client.**

Each test method builds a new `AmplifyCloudWatchClient` in `@Before` and nothing ever shuts it down. The client owns a `CoroutineScope(SupervisorJob() + Dispatchers.IO)`, a `CloudWatchLogsClient`, a SQLCipher store, and (with `FlushStrategy.Interval`) scheduled WorkManager sync — so by the last test four clients are concurrently flushing the same log group and the same shared on-device DB, since `databaseName` is keyed on `logGroupName` alone.

`AmplifyCloudWatchClient` exposes no `close()`, so `disable()` (which calls `logManager.stopSync()`) is the best available teardown. `:aws-kinesis`'s base class does exactly this:

```suggestion
    @Before
    fun setUp() {
        val context = ApplicationProvider.getApplicationContext<Context>()
        client = AmplifyCloudWatchClient(
            context = context,
            region = testRegion,
            credentialsProvider = credentialsProvider,
            options = AmplifyCloudWatchClientOptions {
                logGroupName = testLogGroupName
                localStoreMaxSizeInMB = testLocalStoreMaxSizeInMB
                flushStrategy = FlushStrategy.Interval(testFlushIntervalInSeconds.seconds)
                loggingConstraints = LoggingConstraints(defaultLogLevel = testDefaultLogLevel)
            }
        )
    }

    @After
    fun tearDown() {
        client.disable()
    }
```

That the client has no way to release its scope may itself be worth raising against the client PR.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+
+        client.emit(LogMessage(LogLevel.Error, namespace, message, null))
+        client.emit(LogMessage(LogLevel.Debug, namespace, message, null))
+        client.emit(LogMessage(LogLevel.Warn, namespace, message, null))
+        client.emit(LogMessage(LogLevel.Info, namespace, message, null))
+        delay(SAVE_WAIT_MS)
+
+        val events = awaitEvents(message, expectedCount = 4)
+
+        events shouldHaveSize 4
```
mattcreaser: 🔵 **Suggestion — `>=` in the poll loop versus an exact-size assertion.**

`awaitEvents` returns as soon as `events.size >= expectedCount`, then this asserts exactly 4. All four messages share one `message` string and `filterPattern` matches on it, so any at-least-once redelivery (a flush retry, or leftovers from an earlier run now that `clearPackageData=false` preserves the DB) returns 5 and fails a test that arguably should tolerate it.

Either assert `events.size shouldBeGreaterThanOrEqual 4`, or give each of the four emissions a distinct message and assert one event per level — the latter is stronger, since the current form cannot tell four delivered messages from the same message delivered four times.

aws-cloudwatch/build.gradle.kts
```diff
@@ -22,6 +22,13 @@ apply(from = rootProject.file("configuration/checkstyle.gradle"))
 
 android {
     namespace = "com.amplifyframework.cloudwatch"
+
+    // Persist app data across connected test methods so the client's integration tests keep a stable device
+    // id and write to a single CloudWatch stream (one per test otherwise). Overrides the shared convention's
+    // clearPackageData=true; the module's own DB instrumentation test self-cleans in @After.
+    defaultConfig {
+        testInstrumentationRunnerArguments["clearPackageData"] = "false"
+    }
```
mattcreaser: 🔵 **Suggestion — is the module-wide `clearPackageData` override actually needed?**

The stated reason is keeping a stable device id so tests share one log stream, but the assertions never look at a stream: `filterEvents` calls `filterLogEvents` scoped to `logGroupName`, which spans every stream in the group. So stream fragmentation would not fail these tests, and turning this off weakens isolation for the whole module — including `CloudWatchDatabaseInstrumentationTest` — and lets a failed run's unflushed events persist into the next one.

If something does depend on it, please spell that out in the comment; if not, dropping the block keeps the shared convention's `clearPackageData=true`. Worth confirming on-device either way, since I could not run the connected suite.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpClass() {
+            val context = ApplicationProvider.getApplicationContext<Context>()
+            if (!configured) {
+                Amplify.Auth.addPlugin(AWSCognitoAuthPlugin())
+                Amplify.configure(AmplifyOutputs(R.raw.amplify_outputs), context)
+                configured = true
+            }
```
mattcreaser: 🔵 **Suggestion — the `configured` guard is narrower than the problem it addresses.**

`@BeforeClass` already runs once per class, so within this class the flag is redundant. Across classes it does not help: it is private to this companion, so if another test class in the module configures Amplify first, this still throws `AlreadyConfiguredException` — and `clearPackageData=false` keeps everything in one process, making that more likely.

Catching the actual condition is more robust:

```suggestion
            if (!configured) {
                Amplify.Auth.addPlugin(AWSCognitoAuthPlugin())
                try {
                    Amplify.configure(AmplifyOutputs(R.raw.amplify_outputs), context)
                } catch (e: Amplify.AlreadyConfiguredException) {
                    // Another test class in this process already configured Amplify.
                }
                configured = true
            }
```

Separately, `credentialsProvider` on line 90 is reassigned on every entry and `testLocalStoreMaxSizeInMB` / `testFlushIntervalInSeconds` (lines 76-77) carry initialisers that are always overwritten — `lateinit`/`val` would express the intent better.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
mattcreaser: Use `Resources.getRawResourceId` to reference the file instead of directly using the `R` class, as this won't compile for anyone who doesn't have a backend available.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
mattcreaser: We don't need to guard against reconfiguration here since this is run in `@BeforeClass`, so the test orchestrator will only invoke this once per process.

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
```diff
@@ -0,0 +1,217 @@
...
+ */
+@RunWith(AndroidJUnit4::class)
+class AmplifyCloudWatchClientInstrumentationTest : DeviceFarmTestBase() {
+
+    companion object {
+        private lateinit var credentialsProvider: AwsCredentialsProvider<AwsCredentials>
+        private lateinit var testRegion: String
+        private lateinit var testLogGroupName: String
+        private var testLocalStoreMaxSizeInMB: Int = 1
+        private var testFlushIntervalInSeconds: Long = 60
```
mattcreaser: These can be `lateinit` too

aws-cloudwatch/src/androidTest/kotlin/com/amplifyframework/cloudwatch/AmplifyCloudWatchClientInstrumentationTest.kt
mattcreaser: The client should have a `close` function that cancels the scope

mattcreaser (approved):

PR #3390, opened 2026-08-20
fix(auth): report why Cognito rejected a refresh token

cadivus:

*Issue #, if available:* n/a

*Description of changes:*

Cognito rejects a refresh token with `NotAuthorizedException` for several unrelated reasons and distinguishes them only by the message text. `FetchAuthSessionCognitoActions` caught all of them and reported a plain `SessionExpiredException`, so app code could not tell a recoverable device-binding failure from a genuine expiry, a revocation, or a disabled/deleted user without string matching the cause. In practice apps treat them all as an expired session and sign the user out, which is wrong for the recoverable cases and gives no signal for diagnosing the rest.

Refresh failures now raise `AWSCognitoSessionExpiredException`, which extends `SessionExpiredException` so existing `catch` blocks and behaviour are unchanged, and adds:
- `reason: RefreshFailureReason` — `Expired`, `Revoked`, `DeviceBindingFailed`, `Invalid`, `UserDisabled`, `UserDeleted`, `Unknown`
- `serviceMessage: String?` — the service message, unmodified

Classification is by message text because that is all the service provides; it is lenient, documented as best-effort, and anything unrecognised maps to `Unknown` with the raw message still available. API change is additive only.

Note this exposes the reason, it does not prevent the rejection: a refresh that Cognito refuses still fails.

- 9 new unit tests covering each reason, plus unknown/empty/null input and that the exception is still catchable as `SessionExpiredException`. The expected strings are the messages Cognito actually returned when each scenario was exercised against a device-tracking user pool. Full `:aws-auth-cognito:testDebugUnitTest` suite green (549 tests), `ktlintCheck` and `apiCheck` pass.
- On an emulator against a device-tracking pool: forgetting the device then refreshing yields `reason=DeviceBindingFailed` with service message `Invalid Refresh Token.`, while `AdminUserGlobalSignOut` then refreshing yields `reason=Revoked` with `Refresh Token has been revoked` — so the reasons are distinguished rather than collapsed.
- A sample app branching on `reason` re-authenticated silently for `DeviceBindingFailed` and recovered a working session with no user-visible logout, took the non-recoverable path for `Revoked`, and reported a healthy session unchanged (no false positives).
- On the unpatched build the same app sees only `com.amplifyframework.auth.exceptions.SessionExpiredException` with no reason available.

- [ ] No
- [x] Yes (Please include a PR link for the documentation update)

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
@@ -20,8 +20,8 @@ import aws.sdk.kotlin.services.cognitoidentity.model.GetIdRequest
 import aws.sdk.kotlin.services.cognitoidentityprovider.getTokensFromRefreshToken
 import aws.smithy.kotlin.runtime.time.Instant
 import com.amplifyframework.auth.cognito.AuthEnvironment
+import com.amplifyframework.auth.cognito.exceptions.session.AWSCognitoSessionExpiredException
 import com.amplifyframework.auth.exceptions.NotAuthorizedException
-import com.amplifyframework.auth.exceptions.SessionExpiredException
 import com.amplifyframework.auth.exceptions.SignedOutException
 import com.amplifyframework.statemachine.Action
 import com.amplifyframework.statemachine.codegen.actions.FetchAuthSessionActions
@@ -79,7 +79,13 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
                 }
             } catch (notAuthorized: aws.sdk.kotlin.services.cognitoidentityprovider.model.NotAuthorizedException) {
                 logger.error("$id Failed to refresh user pool tokens: NotAuthorizedException", notAuthorized)
-                val error = SessionExpiredException(cause = notAuthorized)
+                // Cognito distinguishes refresh rejections only by message, so classify it and pass
+                // it through instead of reporting every rejection as a plain expired session.
+                val error = AWSCognitoSessionExpiredException(
+                    reason = AWSCognitoSessionExpiredException.classify(notAuthorized.message),
+                    serviceMessage = notAuthorized.message,
+                    cause = notAuthorized
+                )
                 AuthorizationEvent(AuthorizationEvent.EventType.ThrowError(error))
             } catch (e: Exception) {
                 logger.error("$id Failed to refresh user pool tokens", e)
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.cognito.exceptions.session
+
+import com.amplifyframework.auth.exceptions.SessionExpiredException
+
+/**
+ * Thrown when Cognito rejects a refresh token. Subclasses [SessionExpiredException] so existing
+ * handling keeps working, and adds [reason] plus [serviceMessage] so an app can tell a recoverable
+ * device-binding failure apart from a genuine expiry or a revocation instead of treating every
+ * rejection as an expired session.
+ */
+class AWSCognitoSessionExpiredException internal constructor(
+    /** Best-effort classification of the rejection. */
+    val reason: RefreshFailureReason,
+    /** The message returned by Cognito, unmodified. Null if the service supplied none. */
+    val serviceMessage: String?,
+    cause: Throwable? = null
+) : SessionExpiredException(cause = cause) {
+
+    internal companion object {
+        /**
+         * Cognito signals the cause of a refresh rejection only through the NotAuthorizedException
+         * message, so classification is by text. Matching is lenient because these strings are
+         * service-controlled and have varied (for example "Invalid Refresh Token." is returned with
+         * a trailing period for a device-binding failure and without one when the token itself
+         * cannot be validated).
+         */
+        fun classify(serviceMessage: String?): RefreshFailureReason {
+            val message = serviceMessage?.trim() ?: return RefreshFailureReason.Unknown
+            return when {
+                message.contains("has expired", ignoreCase = true) -> RefreshFailureReason.Expired
+                message.contains("has been revoked", ignoreCase = true) -> RefreshFailureReason.Revoked
+                message.contains("been deleted", ignoreCase = true) -> RefreshFailureReason.UserDeleted
+                message.contains("is disabled", ignoreCase = true) -> RefreshFailureReason.UserDisabled
+                // The trailing period distinguishes a device-binding failure from a token that
+                // could not be validated at all.
+                message.equals("Invalid Refresh Token.", ignoreCase = true) ->
+                    RefreshFailureReason.DeviceBindingFailed
+                message.contains("Invalid Refresh Token", ignoreCase = true) -> RefreshFailureReason.Invalid
+                else -> RefreshFailureReason.Unknown
+            }
+        }
+    }
+}
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/RefreshFailureReason.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/RefreshFailureReason.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.auth.cognito.exceptions.session
+
+/**
+ * Why Cognito rejected a refresh token. Cognito returns every rejection as a NotAuthorizedException
+ * and distinguishes the cause only by the message text, so these values are a best-effort
+ * classification of that text. [Unknown] is used for anything unrecognised, and the raw service
+ * message always remains available.
+ */
+enum class RefreshFailureReason {
+    /** The refresh token is past the validity that was in force when it was issued. */
+    Expired,
+
+    /** The refresh token was revoked, e.g. by RevokeToken or a global sign out. */
+    Revoked,
+
+    /**
+     * The token is bound to a device record that could not be validated: the device was forgotten,
+     * ConfirmDevice never completed, or the DeviceKey sent did not match the binding. Signing in
+     * again establishes a new device binding.
+     */
+    DeviceBindingFailed,
+
+    /** The token could not be validated for this app client at all. */
+    Invalid,
+
+    /** The user account is disabled. */
+    UserDisabled,
+
+    /** The user account was deleted. */
+    UserDeleted,
+
+    /** The service message was not recognised. Inspect the service message for detail. */
+    Unknown
+}
--- a/aws-auth-cognito/api/aws-auth-cognito.api
+++ b/aws-auth-cognito/api/aws-auth-cognito.api
@@ -267,6 +267,24 @@ public class com/amplifyframework/auth/cognito/exceptions/service/UsernameExists
 public final class com/amplifyframework/auth/cognito/exceptions/service/WebAuthnNotEnabledException : com/amplifyframework/auth/exceptions/ServiceException {
 }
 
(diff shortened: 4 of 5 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/RefreshFailureReason.kt
```diff
@@ -0,0 +1,49 @@
...
+    Invalid,
+
+    /** The user account is disabled. */
+    UserDisabled,
+
+    /** The user account was deleted. */
+    UserDeleted,
+
+    /** The service message was not recognised. Inspect the service message for detail. */
+    Unknown
```
mattcreaser: 🟡 **Warning — enum constant names diverge from the public-enum convention, and they are binary API.**

Every public enum in this module and in `core` uses `UPPER_SNAKE_CASE`: `AuthFlowType.USER_SRP_AUTH`, `AuthWebUIPrompt.CONSENT`, `MFAPreference.DISABLED`, `AWSCognitoAuthChannelEventName.FEDERATED_TO_IDENTITY_POOL`, `MFAType.SMS`, `AuthFactorType.PASSWORD_SRP`. The PascalCase ones in `aws-auth-cognito` — `UsernameAttribute`, `VerificationMechanism` in `AuthConfiguration.kt` — are both `@InternalAmplifyApi`. `RefreshFailureReason` is public, so these names land in `aws-auth-cognito.api` and cannot be renamed later without a breaking change. ktlint's `enum-entry-name-case` accepts both forms, so the passing lint run doesn't decide this.

```suggestion
enum class RefreshFailureReason {
    /** The refresh token is past the validity that was in force when it was issued. */
    EXPIRED,

    /** The refresh token was revoked, e.g. by RevokeToken or a global sign out. */
    REVOKED,

    /**
     * The token is bound to a device record that could not be validated: the device was forgotten,
     * ConfirmDevice never completed, or the DeviceKey sent did not match the binding. Signing in
     * again establishes a new device binding.
     */
    DEVICE_BINDING_FAILED,

    /** The token could not be validated for this app client at all. */
    INVALID,

    /** The user account is disabled. */
    USER_DISABLED,

    /** The user account was deleted. */
    USER_DELETED,

    /** The service message was not recognised. Inspect the service message for detail. */
    UNKNOWN
}
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
```diff
@@ -0,0 +1,58 @@
...
+            return when {
+                message.contains("has expired", ignoreCase = true) -> RefreshFailureReason.Expired
+                message.contains("has been revoked", ignoreCase = true) -> RefreshFailureReason.Revoked
+                message.contains("been deleted", ignoreCase = true) -> RefreshFailureReason.UserDeleted
+                message.contains("is disabled", ignoreCase = true) -> RefreshFailureReason.UserDisabled
+                // The trailing period distinguishes a device-binding failure from a token that
+                // could not be validated at all.
+                message.equals("Invalid Refresh Token.", ignoreCase = true) ->
+                    RefreshFailureReason.DeviceBindingFailed
+                message.contains("Invalid Refresh Token", ignoreCase = true) -> RefreshFailureReason.Invalid
```
mattcreaser: 🔴 **Critical — the recoverable/non-recoverable split is decided by a trailing period, matched with `equals`.**

`DeviceBindingFailed` is the one reason the PR wants apps to auto-recover from, and it is distinguished from `Invalid` only by punctuation in a string the service controls. Two concrete failure modes:

- Because this is `equals` on the trimmed message, any message that merely *starts* with `"Invalid Refresh Token."` — a suffix the service appends later, e.g. `"Invalid Refresh Token. Please sign in again."` — silently falls through to `Invalid` rather than to `Unknown`. So drift here doesn't surface as "we don't know", it surfaces as a confident wrong answer that sends the app down the non-recoverable path.
- The asymmetry matters: misclassifying `Invalid` as `DeviceBindingFailed` costs a failed silent re-auth attempt, while misclassifying `DeviceBindingFailed` as `Invalid` signs the user out unnecessarily — which is exactly the bug this PR fixes. Bias the matcher toward `DeviceBindingFailed`.

```suggestion
                // "Invalid Refresh Token." (with the trailing period) is what a device-binding
                // failure returns today; the same text without it means the token itself could not
                // be validated. This distinction is punctuation-level and undocumented, so match
                // on the prefix and prefer the recoverable reading when the service adds detail.
                message.startsWith("Invalid Refresh Token.", ignoreCase = true) ->
                    RefreshFailureReason.DeviceBindingFailed
                message.contains("Invalid Refresh Token", ignoreCase = true) -> RefreshFailureReason.Invalid
```

Even with that, please state in the `RefreshFailureReason` KDoc that `Invalid` and `DeviceBindingFailed` may be conflated by the service, so callers treat `Invalid` as "possibly recoverable" rather than as a definitive verdict.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
```diff
@@ -0,0 +1,58 @@
...
+         * a trailing period for a device-binding failure and without one when the token itself
+         * cannot be validated).
+         */
+        fun classify(serviceMessage: String?): RefreshFailureReason {
+            val message = serviceMessage?.trim() ?: return RefreshFailureReason.Unknown
+            return when {
+                message.contains("has expired", ignoreCase = true) -> RefreshFailureReason.Expired
+                message.contains("has been revoked", ignoreCase = true) -> RefreshFailureReason.Revoked
+                message.contains("been deleted", ignoreCase = true) -> RefreshFailureReason.UserDeleted
+                message.contains("is disabled", ignoreCase = true) -> RefreshFailureReason.UserDisabled
```
mattcreaser: 🔵 **Suggestion — these two substrings are broader than the reasons they map to.**

`"been deleted"` is checked before the token-specific branches, so a future message about a deleted *device* or app client would classify as `UserDeleted`. Anchoring both on the subject keeps the existing test strings matching (`"The user has been deleted for the associated refresh token"` and `"User is disabled."`) while narrowing the blast radius.

```suggestion
                message.contains("user has been deleted", ignoreCase = true) -> RefreshFailureReason.UserDeleted
                message.contains("user is disabled", ignoreCase = true) -> RefreshFailureReason.UserDisabled
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
```diff
@@ -0,0 +1,58 @@
...
+ * device-binding failure apart from a genuine expiry or a revocation instead of treating every
+ * rejection as an expired session.
+ */
+class AWSCognitoSessionExpiredException internal constructor(
+    /** Best-effort classification of the rejection. */
+    val reason: RefreshFailureReason,
+    /** The message returned by Cognito, unmodified. Null if the service supplied none. */
+    val serviceMessage: String?,
+    cause: Throwable? = null
+) : SessionExpiredException(cause = cause) {
```
mattcreaser: 🟡 **Warning — `recoverySuggestion` tells a disabled or deleted user to sign in again.**

The superclass defaults are inherited unchanged, so every reason reports `message = "Your session has expired."` and `recoverySuggestion = "Please sign in and reattempt the operation."`. For `UserDisabled` and `UserDeleted` that advice cannot succeed, and this string is what surfaces to app developers and often to end users. `recoverySuggestion` is already a constructor parameter on `SessionExpiredException`, so this is cheap to vary:

```suggestion
) : SessionExpiredException(
    recoverySuggestion = when (reason) {
        RefreshFailureReason.UserDisabled, RefreshFailureReason.UserDeleted ->
            "This account can no longer be used. Contact support if you believe this is an error."
        else -> "Please sign in and reattempt the operation."
    },
    cause = cause
) {
```

(Adjust the constant names if you take the `UPPER_SNAKE_CASE` rename.) Leaving `message` alone is the right call — changing it would break the assertion in your own test and anything matching on it.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Thrown when Cognito rejects a refresh token. Subclasses [SessionExpiredException] so existing
+ * handling keeps working, and adds [reason] plus [serviceMessage] so an app can tell a recoverable
+ * device-binding failure apart from a genuine expiry or a revocation instead of treating every
+ * rejection as an expired session.
+ */
+class AWSCognitoSessionExpiredException internal constructor(
+    /** Best-effort classification of the rejection. */
+    val reason: RefreshFailureReason,
+    /** The message returned by Cognito, unmodified. Null if the service supplied none. */
+    val serviceMessage: String?,
```
mattcreaser: 🔵 **Suggestion — the new fields are invisible to `equals`/`hashCode`.**

`AmplifyException` implements `equals`/`hashCode` over `message`, `recoverySuggestion` and `cause` only. Since `message` is the inherited constant and `recoverySuggestion` never varies, two `AWSCognitoSessionExpiredException` instances differing only in `reason` compare equal whenever their causes match. That's easy to trip over in tests that assert on captured state-machine errors. Either override both to include `reason` and `serviceMessage`, or add a line of KDoc saying equality deliberately ignores them.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/exceptions/session/AWSCognitoSessionExpiredException.kt
```diff
@@ -0,0 +1,58 @@
...
+
+package com.amplifyframework.auth.cognito.exceptions.session
+
+import com.amplifyframework.auth.exceptions.SessionExpiredException
+
+/**
+ * Thrown when Cognito rejects a refresh token. Subclasses [SessionExpiredException] so existing
+ * handling keeps working, and adds [reason] plus [serviceMessage] so an app can tell a recoverable
+ * device-binding failure apart from a genuine expiry or a revocation instead of treating every
+ * rejection as an expired session.
```
mattcreaser: 🔵 **Suggestion — say that not every `SessionExpiredException` is one of these.**

`HostedUIHttpHelper.kt:90` throws a plain `SessionExpiredException` for `invalid_grant` during the Hosted UI code exchange. That's a different path from token refresh so it isn't a gap in this change, but a reader of this KDoc would reasonably conclude that any `SessionExpiredException` from Auth can be cast to this type, and an unguarded cast there would throw.

```suggestion
/**
 * Thrown when Cognito rejects a refresh token. Subclasses [SessionExpiredException] so existing
 * handling keeps working, and adds [reason] plus [serviceMessage] so an app can tell a recoverable
 * device-binding failure apart from a genuine expiry or a revocation instead of treating every
 * rejection as an expired session.
 *
 * Only refresh-token rejections are reported as this subtype; a [SessionExpiredException] raised
 * elsewhere in Auth is still the base type, so check the type before reading [reason].
 */
```

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
```diff
@@ -79,7 +79,13 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
...
             } catch (notAuthorized: aws.sdk.kotlin.services.cognitoidentityprovider.model.NotAuthorizedException) {
                 logger.error("$id Failed to refresh user pool tokens: NotAuthorizedException", notAuthorized)
-                val error = SessionExpiredException(cause = notAuthorized)
+                // Cognito distinguishes refresh rejections only by message, so classify it and pass
+                // it through instead of reporting every rejection as a plain expired session.
+                val error = AWSCognitoSessionExpiredException(
+                    reason = AWSCognitoSessionExpiredException.classify(notAuthorized.message),
+                    serviceMessage = notAuthorized.message,
+                    cause = notAuthorized
+                )
```
mattcreaser: 🔵 **Suggestion — derive `reason` once, and log it.**

Two small things, one fix. `notAuthorized.message` is read twice, so the call site is structurally able to pass a `reason` classified from one string alongside a different `serviceMessage`; a single local removes that possibility. And since classification depends on service-controlled text, logging the reason next to the existing error makes drift observable in the field instead of only under a debugger.

```suggestion
                val reason = AWSCognitoSessionExpiredException.classify(notAuthorized.message)
                logger.error(
                    "$id Failed to refresh user pool tokens: NotAuthorizedException, reason=$reason",
                    notAuthorized
                )
                // Cognito distinguishes refresh rejections only by message, so classify it and pass
                // it through instead of reporting every rejection as a plain expired session.
                val error = AWSCognitoSessionExpiredException(
                    reason = reason,
                    serviceMessage = notAuthorized.message,
                    cause = notAuthorized
                )
```

Longer term, an `internal` factory taking the `NotAuthorizedException` and deriving both fields would make the mismatch unrepresentable, matching the `internal fun from(error: Throwable)` pattern the repo uses for sealed exception hierarchies.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
```diff
@@ -79,7 +79,13 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
...
                 AuthorizationEvent(AuthorizationEvent.EventType.ThrowError(error))
```
mattcreaser: 🟡 **Warning — this is the only production behaviour change in the PR, and no test asserts it.**

`AWSCognitoSessionExpiredExceptionTest` covers `classify` in isolation and constructs the exception by hand; it never executes the action. The existing `FetchAuthSessionCognitoActionsTest.kt:171` (`refreshUserPoolTokensAction handles NotAuthorizedException`) asserts only `capturedEvent.captured.shouldBeInstanceOf<AuthorizationEvent>()`, which would still pass if this whole block were reverted to `SessionExpiredException`, if `serviceMessage` were dropped, or if `classify` were never called.

That existing test already has the mocks in place, so tightening it is a few lines — capture the `AuthorizationEvent.EventType.ThrowError` and assert the error is an `AWSCognitoSessionExpiredException` with the expected `reason` and `serviceMessage`.

Worth noting while you're there: that test stubs `message = "Token expired"`, which classifies as `Unknown`, since `classify` matches `"has expired"` and not `"expired"`. Using a real service string makes the test meaningful and documents that the exact text is load-bearing.

PR #3288, opened 2026-04-17
fix(auth): Consistent device metadata storage key across auth flows

harsh62:
## Summary

Fixes device key persistence when switching between authentication flows (USER_PASSWORD_AUTH and USER_SRP_AUTH). Three root causes identified and fixed:

### 1. Device metadata storage key mismatch across auth flows
The device metadata storage key was derived from the JWT `username` claim (via `AuthHelper.getActiveUsername()`), which can differ between `USER_PASSWORD_AUTH` (returns internal UUID) and `USER_SRP_AUTH` (returns user alias like email). This caused device key lookup failures on subsequent logins when auth flows change, generating duplicate device IDs.

**Fix:** Thread the original user-typed username (`inputUsername`) through `SignedInData`, `AuthChallenge`, `SRPEvent`, and the `evaluateNextStep` helper chain so `ConfirmDevice` stores device metadata under the same key used for retrieval.

### 2. Case-sensitive storage key mismatch
Cognito normalizes usernames to lowercase in JWT claims and challenge parameters, but the user-typed input preserves original casing. This caused a storage key mismatch when storing with `inputUsername` (mixed case) and retrieving with the JWT username (lowercase).

**Fix:** Lowercase the username in all three device metadata storage operations (save, retrieve, delete) in `AWSCognitoAuthCredentialStore`.

### 3. USER_PASSWORD_AUTH flow missing DEVICE_KEY in InitiateAuth request
`MigrateAuthCognitoActions` (USER_PASSWORD_AUTH flow) never retrieved stored device metadata or sent `DEVICE_KEY` in the `InitiateAuth` request. This caused Cognito to issue a new device on every USER_PASSWORD_AUTH sign-in, even when a confirmed device already existed from a prior sign-in. `SRPCognitoActions` already had this logic.

**Fix:** Add `getDeviceMetadata()` retrieval and `DEVICE_KEY` parameter injection to `MigrateAuthCognitoActions`, matching the existing pattern in `SRPCognitoActions`.

## Changes

**Data classes** (3 files): Added `inputUsername: String? = null` to `SignedInData`, `AuthChallenge`, and `SRPEvent.RespondPasswordVerifier`/`RetryRespondPasswordVerifier`. All optional with null default for backwards-compatible serialization.

**Helper** (1 file): `SignInChallengeHelper.evaluateNextStep` accepts and propagates `inputUsername`.

[...]

Issue #3145, opened 2025-10-23
When using authenticationFlowType USER_PASSWORD_AUTH - device key saved in preferences under email but retrieved with username, causing refresh token failure


JEKMiles:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin - Coroutines

### Amplify Categories

Authentication

### Gradle script dependencies

<details>

```kotlin

amplify-cognito = "2.30.1"
amplify-cognito = { module = "com.amplifyframework:aws-auth-cognito", version.ref = "amplify-cognito" }
amplify-core-kotlin = { module = "com.amplifyframework:core-kotlin", version.ref = "amplify-cognito" }


```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 8.13
------------------------------------------------------------

Build time:    2025-02-25 09:22:14 UTC
Revision:      073314332697ba45c16c0a0ce1891fa6794179ff

Kotlin:        2.0.21
Groovy:        3.0.22
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  17.0.14 (JetBrains s.r.o. 17.0.14+1-b1367.22)
Daemon JVM:    ~/.sdkman/candidates/java/17.0.14-jbr (no JDK specified, using current Java home)
OS:            Mac OS X 15.7 aarch64



```

</details>

### Describe the bug

Using auth flow type USER_PASSWORD_AUTH with device tracking causes refresh tokens to fail.
During sign in, the device metadata is saved with the email address as the username to generate the shared preferences key
`generateKey("$username.$KEY_DEVICE_METADATA")`
However, on retrieval (after time has passed) it attempts to retrieve this key with a server side UUID as the username
As a result, the DeviceKey is missing from the refresh token request, which returns 
`{"__type":"NotAuthorizedException","message":"Invalid Refresh Token."}`

### Reproduction steps (if applicable)

Auth flow type USER_PASSWORD_AUTH
Device tracking enabled
Sign in
Wait some time
Attempt a token refresh

Expected result: 
Token refresh will succeed

Actual result:
Token refresh fails with 
`{"__type":"NotAuthorizedException","message":"Invalid Refresh Token."}`

### Configuration File
```
{
  "auth": {
[...]

PR #3393, opened 2026-08-24
fix(auth): Recover device metadata keyed by a pre-2.30.3 username

### LINKED ISSUE #3394: Signed out after app upgrade: device metadata from SDK <= 2.30.2 is unreachable on a device-tracking pool
**Before opening, please confirm:**

- [x] I have searched for duplicate or closed issues and discussions.

**Language and Async Model**
Kotlin - Coroutines

**Amplify Categories**
Authentication

**Gradle script dependencies**
com.amplifyframework:aws-auth-cognito 2.31.0 through current main (when upgrading from <= 2.30.2)

**Environment information**
Any. Requires a user pool with device tracking (ChallengeRequiredOnNewDevice=true).

**Please include any relevant guides or documentation you're referencing**
n/a

**Describe the bug**

SDK versions <= 2.30.2 stored device metadata in the credential store keyed
by the value the user typed at sign-in (an alias such as an email). #3149
(released in 2.30.3) changed the refresh path to look the metadata up by the
Cognito username. Installs that signed in on <= 2.30.2 keep the old entry,
so after an in-place app upgrade across that boundary the lookup misses,
DeviceKey is omitted from the refresh request, and Cognito rejects the
still-valid refresh token with NotAuthorizedException "Invalid Refresh
Token.". The SDK surfaces this as SessionExpiredException, so every affected
user is signed out even though their refresh token has remaining validity.

Related to #3145 (closed by #3149); this is the upgrade fallout of that fix,
not a recurrence of the original defect.

**Reproduction steps (if applicable)**

1. User pool with ChallengeRequiredOnNewDevice=true. App on Amplify 2.29.2,
   USER_PASSWORD_AUTH sign-in with an email alias, then
   fetchAuthSession(forceRefresh=true) once (succeeds).
2. Upgrade the app in place to any version >= 2.30.3 (credential store is
   preserved).
3. fetchAuthSession(forceRefresh=true) fails with SessionExpiredException
   caused by NotAuthorizedException "Invalid Refresh Token.".
   The same refresh token succeeds on the wire when sent with its DeviceKey.

If the session was never refreshed before the upgrade, the failure occurs on
the second post-upgrade refresh instead of the first.

cadivus:

*Issue #, if available:* fixes #3394 (upgrade fallout of #3149, originally reported via #3145)

*Description of changes:*

SDK versions <= 2.30.2 keyed device metadata by the typed sign-in alias
(e.g. email). #3149 changed the refresh-path key to the Cognito username but
left old entries behind. After an in-place app upgrade the refresh lookup
misses, omits DeviceKey, and on a device-tracking pool Cognito rejects the
still-valid token with "Invalid Refresh Token.". The SDK raises
SessionExpiredException and the user is logged out for no reason.

Fix (refresh path only): if the lookup under the current username misses,
try the alias claims from the id token (email, phone_number,
preferred_username). On a hit, copy the entry to the current username so the
primary lookup hits afterwards. The copy is best-effort (a failed write does
not fail the refresh; cancellation is rethrown) and the legacy entry is kept, because the sign-in
paths still read device metadata by the typed alias; deleting it would make
the next sign-in register a duplicate device. Keeping it also avoids
conflicting with #3288, which addresses the same key mismatch from the
sign-in side. No behavior change when the primary lookup succeeds. A
fruitless alias scan is remembered per username so sessions without any
device metadata do not pay extra credential-store round trips on every
refresh; a scan whose copy failed to persist is retried.

Known limitation: recovery only works when the stored key equals an alias
claim value. A key typed with different casing or phone formatting than the
stored attribute will not match; those users see the pre-fix behavior.

Reproduce: pool with ChallengeRequiredOnNewDevice=true, sign in on 2.29.2,
fetchAuthSession(forceRefresh=true) once, upgrade the app in place to main,
force refresh again -> SessionExpiredException / "Invalid Refresh Token.".
With this patch the refresh succeeds and logs the copy line once.

Unit tests for fallback, ordering, dedupe, best-effort copy, alias
extraction, and the FetchAuthSessionCognitoActions call-site wiring (module
suite: 560 pass). Emulator: 2.29.2 -> patched upgrade keeps the session
(copy runs once, repeat refreshes succeed); sign-out + sign-in after the
copy works and refresh keeps working; fresh install unaffected;
forgetDevice then refresh still fails (real failures are not masked).

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,14 +106,34 @@ internal class AuthEnvironment internal constructor(
...
-            deviceCredentials = AmplifyCredential.DeviceData(DeviceMetadata.Empty)
+    /**
+     * Loads device metadata for [username], falling back to [legacyUsernames] because SDK versions
+     * <= 2.30.2 keyed the metadata by the typed sign-in alias. A legacy hit is re-keyed to
+     * [username], so the fallback runs at most once.
+     */
+    suspend fun getDeviceMetadata(
+        username: String,
+        legacyUsernames: List<String> = emptyList()
+    ): DeviceMetadata.Metadata? {
```
mattcreaser, line 124: 🟡 **The fallback is ungated, so "runs at most once" only holds on a hit — and the miss case is the common one.**

The KDoc says "A legacy hit is re-keyed to `username`, so the fallback runs at most once," and the PR description says "No change when the primary lookup succeeds or no legacy entry exists." The first is true; the second is not.

Trace the three cases:

| Store state | `loadCredentials` calls per refresh |
|---|---|
| Metadata under canonical username | 1 (returns at the first key — unchanged ✅) |
| Legacy entry present | 2–4 once, then 1 forever ✅ |
| **No entry under any key** | **1 + one per alias claim, on every refresh, forever** ❌ |

The third row is the default state for every app that has not turned on device tracking in Cognito — `getDeviceMetadata` is called unconditionally here, not behind a device-tracking check. For a user whose id token carries `email`, `phone_number` and `preferred_username`, that is 4 lookups instead of 1 on every single refresh, with no possible benefit, because nothing is ever written and so nothing ever short-circuits the scan.

That is not just wasted work. Each `loadCredentials` is a full `CredentialStoreClient.listenForResult` round trip — send a `CredentialStoreEvent`, collect the state flow until `Idle` — over an `EncryptedSharedPreferences`-backed `KeyValueRepository`, so it involves a keystore decrypt. And it is a **new throw site on the critical refresh path**: `listenForResult` throws `InvalidStateException("Credential operation failed")` when the store errors, the loop does not guard against it, and `refreshUserPoolTokensAction` turns any escaped `Exception` into `AuthorizationEvent.ThrowError`. So a transient store failure that previously had one chance to break a refresh now has up to four — for users who were never affected by this bug.

Suggest gating the scan so it cannot run indefinitely. The cheapest version needs no new storage — `AuthEnvironment` lives for the plugin's lifetime, so an in-memory latch reduces it to once per process:

```suggestion
    private var legacyDeviceMetadataScanned = false

    /**
     * Loads device metadata for [username]. If nothing is stored under [username], retries the
     * [legacyUsernames] candidates once per process, because SDK versions <= 2.30.2 keyed the
     * metadata by the typed sign-in alias. A legacy hit is re-keyed to [username], so an affected
     * install self-heals permanently on its next refresh.
     */
    suspend fun getDeviceMetadata(
        username: String,
        legacyUsernames: List<String> = emptyList()
    ): DeviceMetadata.Metadata? {
```

…then skip the legacy portion of the loop when `legacyDeviceMetadataScanned` is already set, and set it once the scan completes. A persisted marker would be stronger still (it would survive process death), but per-process is enough to remove the unbounded cost, and it keeps the change inside this file.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,14 +106,34 @@ internal class AuthEnvironment internal constructor(
...
+        for (key in listOf(username) + legacyUsernames.filter { it != username }) {
```
mattcreaser: 🔵 The legacy candidates are filtered against `username` but not against **each other**. `signInAliases` returns `email`, `phone_number`, `preferred_username` verbatim, and `preferred_username` being equal to `email` is a common Cognito setup — that costs a second identical store round trip. One-word fix:

```suggestion
        for (key in (listOf(username) + legacyUsernames).distinct()) {
```

This also subsumes the existing `it != username` filter, so the behaviour asserted by `does not consult the store twice when the alias equals the current username` is preserved.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
```diff
@@ -106,14 +106,34 @@ internal class AuthEnvironment internal constructor(
...
+            }
+            val metadata = deviceCredentials.deviceMetadata as? DeviceMetadata.Metadata ?: continue
+            if (key != username) {
+                logger.info("Migrating device metadata stored by an earlier SDK version.")
+                credentialStoreClient.storeCredentials(
+                    CredentialType.Device(username),
+                    AmplifyCredential.DeviceData(metadata)
+                )
+                credentialStoreClient.clearCredentials(CredentialType.Device(key))
+            }
```
mattcreaser: 🟡 **The `clearCredentials` call is not required for the stated goal, and it can delete another user's device metadata.**

On the first point: once `storeCredentials` has written the entry under `username`, the next call returns at `key == username` on the first iteration and never reaches the legacy keys again. So the "fallback runs at most once" property is delivered entirely by the *copy* — the delete buys nothing beyond reclaiming one SharedPreferences entry, and it is the only destructive operation in the change.

On the second point: the key being deleted is an **alias claim value**, and there is no check that the entry found under it actually belongs to this user. I traced the delete — `clearCredentials(CredentialType.Device(key))` → `CredentialStoreCognitoActions` `is CredentialType.Device -> credentialStore.deleteDeviceKeyCredential(credentialType.username)` → `keyValue.remove("amplify.<pool>.<key>.deviceMetadata")` — so it removes exactly that entry, whoever owns it.

Device metadata for multiple usernames coexists in one store (sign-out does not clear it — that is the point of remembering a device), so on a shared device this collides whenever user B's `email` or `preferred_username` equals user A's Cognito username. When `email` is *not* configured as an alias attribute in the pool, Cognito does not enforce uniqueness on it and it is settable via `updateUserAttributes` without verification, so this is reachable deliberately, not only by accident. The stolen `deviceKey`/`deviceSecret` is scoped to user A and would be rejected for B, so the confidentiality impact is limited — but A's entry is destroyed, and A gets an unexpected new-device challenge on their next sign-in.

Dropping the delete removes both concerns at once:

```suggestion
            if (key != username) {
                logger.info("Migrating device metadata stored by an earlier SDK version.")
                credentialStoreClient.storeCredentials(
                    CredentialType.Device(username),
                    AmplifyCredential.DeviceData(metadata)
                )
            }
```

One related question I could not settle by reading: `SRPCognitoActions.initiateSRPAuthAction` calls `getDeviceMetadata(event.username)` — the **typed** username — and puts the result in `DEVICE_KEY` on `initiateAuth`. Deleting the alias-keyed entry makes that lookup miss on subsequent sign-ins. I *think* it recovers, because the later `getDeviceMetadata(AuthHelper.getActiveUsername(event.username, response))` now hits under the canonical key and feeds `parseResponseChallenge` — and that first lookup arguably already misses for these users post-#3149 anyway. Worth confirming on the emulator you already have set up, since it is the one interaction the unit tests cannot reach. Not deleting sidesteps the question entirely.

If you keep the delete, `migrates legacy metadata to the current username and clears the legacy entry` should stay; if you drop it, that test needs its second `coVerify` removed and `does not migrate when metadata already exists under the current username` still passes as written.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
```diff
@@ -87,6 +90,19 @@ internal class IdToken(override val tokenValue: String) : Jwt() {
...
+    // by the value the user typed at sign-in rather than by the Cognito username. Never throws: a
+    // token we cannot parse simply yields no candidates.
+    val signInAliases: List<String>
+        get() = runCatching {
+            listOfNotNull(
+                getClaim(Claim.Email),
+                getClaim(Claim.PhoneNumber),
+                getClaim(Claim.PreferredUsername)
+            )
+        }.getOrDefault(emptyList())
```
mattcreaser: 🔵 Two things about the `runCatching`.

The intent is right and I confirmed it is load-bearing — existing `FetchAuthSessionCognitoActionsTest` fixtures pass `idToken = "old_id"`, `Jwt.payload` throws `UnknownException` on a non-3-part token, and without the catch this would surface through `refreshUserPoolTokensAction`'s `catch (e: Exception)` as a failed refresh. So please keep the leniency.

But (a) `runCatching` catches `Throwable`, which includes `CancellationException`. Nothing under `getClaim` is suspending today so it cannot actually bite, but it is a pattern that becomes a real bug the moment someone makes the call chain suspend, and the codebase avoids it elsewhere. Catching the type that is actually thrown is both narrower and self-documenting:

```suggestion
    // Sign-in aliases. Used to locate device metadata written by older SDK versions, which keyed it
    // by the value the user typed at sign-in rather than by the Cognito username. A token we cannot
    // parse yields no candidates rather than failing the caller.
    val signInAliases: List<String>
        get() = try {
            listOfNotNull(
                getClaim(Claim.Email),
                getClaim(Claim.PhoneNumber),
                getClaim(Claim.PreferredUsername)
            )
        } catch (e: UnknownException) {
            emptyList()
        }
```

And (b) this is now the only property on `IdToken` that swallows a parse failure — `userSub`, `username` and `expiration` all propagate `UnknownException`. That asymmetry is defensible here, but it is worth a word in the comment saying *why* this one differs (a missing alias must not fail a refresh), so a later reader does not "fix" the inconsistency in the wrong direction. As a `get()` rather than a `by lazy`, it also re-parses on each access; only called once per refresh today, so not worth changing unless you are touching it anyway.

aws-auth-cognito/src/main/java/com/amplifyframework/statemachine/codegen/data/Tokens.kt
```diff
@@ -76,7 +76,10 @@ internal abstract class Jwt {
         UserSub("sub"),
         Username("username"), // username claim in the access token
         TokenRevocationId("origin_jti"),
-        CognitoUsername("cognito:username") // username claim in the identity token
+        CognitoUsername("cognito:username"), // username claim in the identity token
+        Email("email"),
+        PhoneNumber("phone_number"),
+        PreferredUsername("preferred_username")
```
mattcreaser: 🔵 The recovery keys on these claim values verbatim, but the legacy entry was keyed on **the string the user typed** at sign-in, and the two are not guaranteed to match byte-for-byte:

- **Email case.** The store key is a SharedPreferences key and therefore case-sensitive. A user who typed `User@Example.com` while the `email` claim reads `user@example.com` will not be recovered.
- **Phone format.** The `phone_number` claim is E.164 (`+15550100`); a user who typed `5550100` or `555-0100` will not be recovered.

Neither is a defect in the diff — the fallback is best-effort and a miss just means no self-heal. But it does mean the fix will not reach every user described in #3145, and that is worth stating in the KDoc on `getDeviceMetadata` so the next person debugging "the migration didn't run for me" does not have to rediscover it. If you want to widen the net cheaply, adding a lowercased variant of the email candidate covers the most likely of the two — though note that it interacts with the shared-device concern I raised on `AuthEnvironment.kt`, so it is more attractive once the destructive `clearCredentials` is gone.

Also worth noting: these three entries land on the shared `Jwt.Claim` enum, so they become reachable from `AccessToken` too. Harmless — `AccessToken` exposes no accessor for them — just flagging that the enum is shared, in case you would rather scope them.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
```diff
@@ -44,7 +44,9 @@ internal object FetchAuthSessionCognitoActions : FetchAuthSessionActions {
             val evt = try {
                 val username = signedInData.username
                 val tokens = signedInData.cognitoUserPoolTokens
-                val deviceMetadata: DeviceMetadata.Metadata? = getDeviceMetadata(username)
+                // Alias candidates let us recover device metadata keyed by an older SDK version.
+                val deviceMetadata: DeviceMetadata.Metadata? =
+                    getDeviceMetadata(username, tokens.idToken?.signInAliases ?: emptyList())
```
mattcreaser: The comment reads a little thin for how much it is carrying:

```suggestion
                // SDK <= 2.30.2 keyed device metadata by the typed sign-in alias rather than the
                // Cognito username, so pass the id token's aliases as recovery candidates. Without
                // this, an in-place upgrade drops DeviceKey and a device-tracking pool rejects the
                // still-valid refresh token. See #3145.
                val deviceMetadata: DeviceMetadata.Metadata? =
                    getDeviceMetadata(username, tokens.idToken?.signInAliases ?: emptyList())
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    private val environment = AuthEnvironment(
+        mockk<Context>(relaxed = true),
+        mockk(relaxed = true),
+        mockk(relaxed = true),
+        credentialStoreClient,
+        null,
+        null,
+        logger
+    )
```
mattcreaser, line 50: Please use named arguments to disambiguate these mocks.

```suggestion
    private val environment = AuthEnvironment(
        context = mockk<Context>(relaxed = true),
        configuration = mockk(relaxed = true),
        cognitoAuthService = mockk(relaxed = true),
        credentialStoreClient = credentialStoreClient,
        userContextDataProvider = null,
        hostedUIClient = null,
        logger = logger
    )
```

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        coVerify(exactly = 1) { credentialStoreClient.loadCredentials(CredentialType.Device(username)) }
+    }
+
+    @Test
+    fun `no legacy candidates behaves like a plain lookup`() = runTest {
+        storeContains(email to metadata)
+
+        environment.getDeviceMetadata(username) shouldBe null
+    }
+}
```
mattcreaser: 🔵 The suite is well shaped — Kotest assertions, MockK, `runTest`, backticked names, Arrange-Act-Assert throughout, all matching the module's conventions. Two branches are still uncovered:

1. **Duplicate legacy candidates.** `does not consult the store twice when the alias equals the current username` covers alias-equals-username, but not alias-equals-alias. A user whose `preferred_username` equals their `email` hits the store twice for the same key — see my `.distinct()` suggestion on `AuthEnvironment.kt:118`. Worth a test so the de-duplication does not regress:

```kotlin
    @Test
    fun `does not consult the store twice for duplicate legacy candidates`() = runTest {
        storeContains()

        environment.getDeviceMetadata(username, listOf(email, email)) shouldBe null

        coVerify(exactly = 1) { credentialStoreClient.loadCredentials(CredentialType.Device(email)) }
    }
```

2. **The non-`DeviceData` branch.** `loadCredentials` returning something that is not `AmplifyCredential.DeviceData` logs a warning and `continue`s. `storeContains` always returns `DeviceData`, so that path never executes. It is close to unreachable in production — `CredentialStoreCognitoActions` always answers a `CredentialType.Device` load with `DeviceData` — so if you agree it is dead, a one-line comment saying so is more valuable than a test. If it is reachable, note that the `continue` means the warning can now fire up to four times per refresh where it previously fired once.

Also a small observation on the sibling file: `IdTokenSignInAliasesTest` asserts with `shouldContainExactly`, which pins the **order** to email → phone → preferred_username. That order is meaningful — it decides which entry wins when two aliases both have stored metadata — so pinning it is correct. Just be aware the assertion is order-sensitive if the `listOfNotNull` is ever reordered.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/actions/FetchAuthSessionCognitoActions.kt
cadivus: Applied verbatim in 34886fad.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
cadivus: Gated in bfe3c2e7: a fruitless alias scan is recorded per username in an in-memory set, so the no-entry case costs the extra lookups once per process and 1 lookup after that.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
cadivus: Done in 179fd8e7 - `(listOf(username) + legacyUsernames).distinct()`.

aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AuthEnvironment.kt
cadivus: Dropped the delete in 8e3dfa7d for both reasons.

On your open question: I only traced it in code, I have not confirmed it on a device. `initiateSRPAuthAction` reads by the typed username, so for these users that first lookup hits the alias-keyed entry today and would start missing if we deleted it. Not deleting sidesteps it, as you say. The test now asserts `clearCredentials` is never called.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/AuthEnvironmentDeviceMetadataTest.kt
cadivus: Done in 179fd8e7.

mattcreaser (approved):

PR #3395, opened 2026-08-24
fix(all): resolve UnsafeOptInUsage and UseKtx errors from newer androidx lint

mattcreaser:
### Summary

Bumping `androidx.fragment` 1.2.4 → 1.9.0 (#3380) pulls in a newer `androidx.annotation-experimental` transitively, which **activates two Android Lint checks that were previously dormant** across the codebase. This PR resolves both so #3380 (and any future dependency bump that drags in the newer lint) passes `static-analysis`. The fragment bump itself stays in #3380 — this PR contains only the lint fixes.

### 1. `UnsafeOptInUsageError` — Amplify's own opt-in markers (88+ sites)

`@InternalAmplifyApi` (and the other Amplify markers) are `@RequiresOptIn` markers used pervasively for cross-module-internal APIs. Amplify already opts into them **globally for the Kotlin compiler** via `-opt-in=` (`optInAnnotations` in `build-logic/.../Util.kt`) — but Android Lint's `UnsafeOptInUsage` is a *separate* enforcement mechanism that does not read the Kotlin compiler flag, and also applies to **Java** sources.

Fix: add a root **`lint.xml`** configuring project-wide opt-in for the Amplify marker annotations, wired in via `lintConfig` in `AndroidLibraryConventionPlugin`. This is the lint-side mirror of the existing compiler opt-in. The published `@RequiresOptIn` markers still enforce opt-in for third-party consumers of the library (lint.xml only applies when building amplify-android itself).

### 2. `UnsafeOptInUsageError` — third-party smithy `InternalApi` (1 site)

`IamRequestDecorator.java` uses smithy's internal `Headers` builder. This third-party marker is **not** added to the global opt-in (so lint keeps catching unintended third-party opt-in usage). Instead it's opted into locally with `@SuppressLint("UnsafeOptInUsageError")`, matching the existing per-call-site pattern used for smithy `InternalApi` elsewhere in the repo (`AWS4Signer.kt`, `AWSS3StoragePlugin.java`, …).

### 3. `UseKtx` — idiomatic KTX conversions

The newer lint enables `UseKtx`, which flags `SharedPreferences.edit()` / `Uri.parse()` where androidx KTX extensions exist. Converted each site to the idiomatic `edit { }` / `String.toUri()` and added `androidx.core-ktx` where needed:
- `aws-connect/DeviceIdStore.kt` (×2)
- `aws-event-enrichment/SharedPreferencesClientIdProvider.kt`
- `aws-pinpoint-core/PinpointDatabase.kt` (×2), `SharedPreferencesUtil.kt`
- `core/TOTPSetupDetails.kt`

### Verification

Temporarily bumped `androidx.fragment` to 1.9.0 (reproducing #3380) and ran the full CI static-analysis gate:

```
./gradlew ktlintCheck checkstyle apiCheck lint --continue --rerun-tasks
BUILD SUCCESSFUL — 0 errors across all modules
```

(Confirmed the uncached `--rerun-tasks` sweep to avoid lint's incremental per-file caching hiding sites.) The temporary bump was reverted; this PR ships only the lint fixes.

### Next step

Once merged, comment `@dependabot rebase` on #3380 to re-run its `static-analysis` job green.

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/auth/IamRequestDecorator.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/auth/IamRequestDecorator.java
@@ -15,6 +15,8 @@
 
 package com.amplifyframework.api.aws.auth;
 
+import android.annotation.SuppressLint;
+
 import com.amplifyframework.api.ApiException.ApiAuthException;
 import com.amplifyframework.api.aws.AppSyncSigningException;
 import com.amplifyframework.api.aws.sigv4.AWS4Signer;
@@ -67,6 +69,8 @@ public class IamRequestDecorator implements RequestDecorator {
      * @return A new instance of the request containing the signature headers.
      * @throws ApiAuthException If the signing process fails.
      */
+    // Uses the smithy runtime's internal Headers builder API to construct the request to sign.
+    @SuppressLint("UnsafeOptInUsageError")
     public final okhttp3.Request decorate(okhttp3.Request req) throws ApiAuthException {
         //set the request body
         final byte[] bodyBytes = getBytes(req.body());
--- a/aws-connect/build.gradle.kts
+++ b/aws-connect/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
     implementation(project(":foundation-bridge"))
 
     implementation(libs.androidx.annotation)
+    implementation(libs.androidx.core.ktx)
     implementation(libs.kotlin.coroutines)
     implementation(libs.okhttp)
     implementation(libs.kotlin.serializationJson)
--- a/aws-connect/src/main/java/com/amplifyframework/connect/internal/DeviceIdStore.kt
+++ b/aws-connect/src/main/java/com/amplifyframework/connect/internal/DeviceIdStore.kt
@@ -16,6 +16,7 @@ package com.amplifyframework.connect.internal
 
 import android.content.Context
 import android.content.SharedPreferences
+import androidx.core.content.edit
 import java.util.UUID
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
@@ -58,7 +59,7 @@ internal class DeviceIdStore(
             existing
         } else {
             val newId = UUID.randomUUID().toString()
-            prefs.edit().putString(DEVICE_ID_KEY, newId).apply()
+            prefs.edit { putString(DEVICE_ID_KEY, newId) }
             newId
         }
     }
@@ -75,7 +76,7 @@ internal class DeviceIdStore(
      * Clears the persisted device ID.
      */
     suspend fun clear() = withContext(dispatcher) {
-        prefs.edit().remove(DEVICE_ID_KEY).apply()
+        prefs.edit { remove(DEVICE_ID_KEY) }
     }
 
     internal companion object {
--- a/aws-event-enrichment/build.gradle.kts
+++ b/aws-event-enrichment/build.gradle.kts
@@ -29,6 +29,7 @@ dependencies {
     api(project(":foundation"))
 
     implementation(libs.androidx.annotation)
+    implementation(libs.androidx.core.ktx)
     implementation(libs.kotlin.coroutines)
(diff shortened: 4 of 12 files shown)


thisisabhash (dismissed):

aws-api-appsync/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ dependencies {
     implementation(project(":aws-core"))
 
     api(libs.androidx.annotation)
+    implementation(libs.androidx.annotation.experimental)
```
mattcreaser: Two notes on this line — neither blocking.

**1. The convention plugin already half-does this, and that is the underlying bug.** `AndroidLibraryConventionPlugin.kt:107-111` has:

```kotlin
constraints {
    add("implementation", libs.findLibrary("androidx-annotation-experimental").get()) {
        because("Fixes a lint bug with RequiresOptIn")
    }
}
```

A `constraints { add(...) }` only pins the **version** if the module already resolves the dependency — it does not put it on the graph. So your explicit `implementation(...)` here is genuinely required, and the constraint is what makes it resolve to 1.4.1. (Worth confirming that is your understanding too, since the `because(...)` string reads like it was meant to be the whole fix.)

The consequence is that whether `UnsafeOptInUsageError` is active at all is decided ad hoc, per module, by transitive luck. That is why this defect sat latent, why `aws-storage-s3` already needed `@SuppressLint("UnsafeOptInUsageError")` in seven places while this module needed nothing, and why a Dependabot bump is what surfaced it. Promoting the constraint to a real `implementation` dependency in the convention plugin would make the check uniformly active and flush out every remaining gap at once. Larger blast radius, so reasonable to defer — but worth a follow-up issue rather than fixing these one bump at a time.

**2. Scope.** `androidx.annotation.OptIn` is CLASS-retention, so it is not needed at runtime, and `implementation` publishes it in this module's POM as a runtime dependency. `compileOnly` would be leaner, but AGP does not reliably process `lint.jar` from `compileOnly` AAR dependencies, and here you likely *want* the lint check active in this module. So `implementation` is probably right — just flagging it as a deliberate tradeoff rather than an oversight.

aws-api-appsync/src/main/java/com/amplifyframework/core/model/query/predicate/GsonPredicateAdapters.java
```diff
@@ -257,6 +261,7 @@ private JsonElement serializeQueryPredicateGroup(QueryPredicateGroup group, Json
          * @param jsonObject The JSON object containing the group data
          * @return QueryPredicateGroup instance with all nested predicates deserialized
          */
+        @OptIn(markerClass = InternalAmplifyApi.class)
```
mattcreaser: Correct placement, and the right mechanism. Verified by reading:

- The offending call is at line 280 (`return new QueryPredicateGroup(type, predicates);`), inside this method, and `QueryPredicateGroup`'s constructor is `@InternalAmplifyApi` at `core/.../QueryPredicateGroup.java:42`. Pre-change that is line 275, matching the error in the description.
- `@OptIn` here rather than marking this method `@InternalAmplifyApi` is the better choice: `GsonPredicateAdapters` is public API of this module, and propagating the marker outward would be wrong.
- Line 280 is the *only* unopted cross-module internal usage in this module's Java sources. `GraphQLRequestHelper.java` is already `@InternalAmplifyApi` at the class level, and the Kotlin sources are covered by the global `-opt-in` compiler arg from `KotlinConventionPlugin.kt:59`.

One correction to the PR description: it says this matches `aws-storage-s3`'s pattern via `AWSS3StorageListOperation.java`, but that file opts in to `InternalApiWarning`, a different (WARNING-level) marker. For `InternalAmplifyApi` the existing pattern in that module is actually `@SuppressLint("UnsafeOptInUsageError")` (`AWSS3StoragePlugin.java:119,140,143,234,300,988,1202`). Your `@OptIn` is the better of the two, so no change requested here — just worth correcting the claim, and worth a follow-up to converge the codebase on `@OptIn`.

cadivus (approved):

PR #3396, opened 2026-08-25
feat(predictions): report a distinct exception when a liveness session is interrupted

mattcreaser:

*Issue #, if available:* Addresses the predictions half of aws-amplify/amplify-ui-android#332

*Description of changes:*

When the websocket carrying a face liveness check dies mid-session — the user takes a call, or backgrounds the app long enough for the OS to close the socket, or connectivity drops — `LivenessWebSocket.onFailure` wrapped the transport throwable in a generic `PredictionsException("An unknown error occurred during the Liveness flow.")`. The *kind* of failure was never named, so by the time `LivenessCoordinator` in Amplify UI saw it, the only way to identify an interruption was to unwrap `cause` and match on `java.net.SocketException` — a JDK type that is no part of this SDK's contract.

This adds `FaceLivenessSessionInterruptedException` (`@InternalAmplifyApi`, `internal constructor`, matching the three internal siblings in that package — `AccessDeniedException`, the fourth, is public) and classifies a lost transport as that type:

```kotlin
private fun classifyConnectionFailure(t: Throwable): PredictionsException = when {
    socketOpened && t is IOException && t !is ProtocolException ->
        FaceLivenessSessionInterruptedException(cause = t)
    else -> PredictionsException(
        "An unknown error occurred during the Liveness flow.",
        t,
        "See attached exception for more details."
    )
}
```

Three conditions, each earning its place:

- **`socketOpened`** — set in `onOpen`. OkHttp routes both a lost transport *and* a connection that was never established through `onFailure`, and `ConnectException` / `SSLHandshakeException` / a connect-time `UnknownHostException` are all `IOException`. Without this gate a session that never started would be reported as interrupted with advice to retry, which is wrong for a pinning failure, an intercepting proxy, or a bad regional endpoint — all permanent. A connect-time failure keeps the generic message; distinguishing "could not connect" from "unknown" would need its own type and is not what #332 asked for.

  The flag is deliberately **not** cleared on the `ATTEMPT_RECONNECT` path, since a reconnect only follows a socket that did open and a server that did respond, so a failure on the retry really did interrupt a session that was under way.

- **`IOException`** — the family OkHttp reports for a dead transport: `SocketException` (the reported case), `SocketTimeoutException`, `UnknownHostException`, `SSLException`, `EOFException`.

- **`!is ProtocolException`** — a malformed frame is not an interruption. This also excludes a *rejected upgrade handshake*: OkHttp's `RealWebSocket.checkUpgradeSuccess` throws `ProtocolException("Expected HTTP 101 response but was ...")`, so an HTTP-level rejection of the handshake stays in the generic bucket too.

Two existing invariants are deliberately preserved:

- **`webSocketError ?:` keeps precedence**, so a server-sent typed error (validation, throttling, access-denied, session-not-found) can never be reclassified as an interruption. Covered by a test.
- **`clientStoppedSession` keeps its guard**, so a client-initiated close is still silent. A *backgrounded* app never disposes the composable, so the flag stays false and the classification applies exactly where intended. Covered by a test that specifically exercises a `SocketException` after a client stop.

`onClosed` is intentionally left alone. The abnormal-close path fires when the server closes with a non-1000 code without having sent a typed error — a server-side decision, not a lost connection. It is also moot here, since OkHttp reports connection loss (including 1006-style closure with no close frame) through `onFailure`.

The original throwable is still attached as `cause`, so no information is hidden — this adds a contract on top of it.

**Behaviour change:** callers that matched on the generic message string for a mid-session transport failure will no longer see it. The type is `@InternalAmplifyApi` and the public API dump is unchanged, so there is no public API impact.

**Follow-up:** Amplify UI maps this onto a public `FaceLivenessDetectionException.SessionInterruptedException` in aws-amplify/amplify-ui-android#334, which is what closes #332. That name matches the `FaceLivenessDetectionError.sessionInterrupted` iOS shipped in aws-amplify/amplify-ui-swift-liveness#240.

`./gradlew :aws-predictions:testDebugUnitTest :aws-predictions:ktlintCheck :aws-predictions:apiCheck` — passes. 78 tests across 8 classes in the module, every class reporting `failures="0"`; `apiCheck` reports no `.api` drift.

`LivenessWebSocketTest` goes from 21 to 31 tests:

- one test per lost-transport type — `SocketException`, `SocketTimeoutException`, `UnknownHostException`, `SSLException`, `EOFException` — each asserting the type and that `cause` is preserved
- `ProtocolException` → generic `PredictionsException`
- a non-`IOException` → generic `PredictionsException`
- a failure **before** the socket opens → generic `PredictionsException`
- a server error already recorded in `webSocketError` is **not** reclassified
- a `SocketException` after `clientStoppedSession` is **not** reported at all

Note that end-to-end confirmation on a device — background a live check until the socket dies and observe the typed error reach `onError`, and confirm a *brief* background still recovers — depends on the Amplify UI mapping and will be verified with #334.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/exceptions/FaceLivenessSessionInterruptedException.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/exceptions/FaceLivenessSessionInterruptedException.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amplifyframework.predictions.aws.exceptions
+
+import com.amplifyframework.annotations.InternalAmplifyApi
+import com.amplifyframework.predictions.PredictionsException
+
+/**
+ * Raised when the connection to the face liveness service is lost before the check completes, for example because the
+ * host app was backgrounded long enough for the socket to close, or because the device lost network connectivity.
+ */
+@InternalAmplifyApi
+class FaceLivenessSessionInterruptedException internal constructor(
+    message: String = "The face liveness session was interrupted.",
+    cause: Throwable? = null,
+    recoverySuggestion: String = "Retry the face liveness check."
+) : PredictionsException(message, cause, recoverySuggestion)
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/exceptions/FaceLivenessSessionInterruptedException.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/exceptions/FaceLivenessSessionInterruptedException.kt
@@ -19,8 +19,13 @@ import com.amplifyframework.annotations.InternalAmplifyApi
 import com.amplifyframework.predictions.PredictionsException
 
 /**
- * Raised when the connection to the face liveness service is lost before the check completes, for example because the
- * host app was backgrounded long enough for the socket to close, or because the device lost network connectivity.
+ * Raised when an established connection to the face liveness service is lost before the check completes, for example
+ * because the host app was backgrounded long enough for the socket to close, or because the device lost network
+ * connectivity mid-check.
+ *
+ * Any transport-level [java.io.IOException] other than a protocol violation is reported as this type, so it also
+ * covers read timeouts, DNS failures and TLS teardown. A connection that never opened is not reported as this type,
+ * since that indicates an unreachable or misconfigured endpoint rather than an interruption.
  */
 @InternalAmplifyApi
 class FaceLivenessSessionInterruptedException internal constructor(
--- a/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
+++ b/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
@@ -29,6 +29,7 @@ import com.amplifyframework.core.category.CategoryType
 import com.amplifyframework.predictions.PredictionsException
 import com.amplifyframework.predictions.aws.BuildConfig
 import com.amplifyframework.predictions.aws.exceptions.AccessDeniedException
+import com.amplifyframework.predictions.aws.exceptions.FaceLivenessSessionInterruptedException
 import com.amplifyframework.predictions.aws.exceptions.FaceLivenessSessionNotFoundException
 import com.amplifyframework.predictions.aws.exceptions.FaceLivenessUnsupportedChallengeTypeException
 import com.amplifyframework.predictions.aws.models.liveness.BoundingBox
@@ -48,6 +49,8 @@ import com.amplifyframework.predictions.models.Challenge
 import com.amplifyframework.predictions.models.FaceLivenessChallengeType
 import com.amplifyframework.predictions.models.FaceLivenessSessionInformation
 import com.amplifyframework.util.UserAgent
+import java.io.IOException
+import java.net.ProtocolException
 import java.net.URI
 import java.net.URLDecoder
 import java.nio.ByteBuffer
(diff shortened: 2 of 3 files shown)


aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -247,16 +250,26 @@ internal class LivenessWebSocket(
...
+    not an interruption. Anything else is not something we can attribute, so it keeps the generic message.
+     */
+    private fun classifyConnectionFailure(t: Throwable): PredictionsException = when {
+        t is IOException && t !is ProtocolException -> FaceLivenessSessionInterruptedException(cause = t)
+        else -> PredictionsException(
+            "An unknown error occurred during the Liveness flow.",
+            t,
+            "See attached exception for more details."
+        )
+    }
```
mattcreaser: 🟡 **`onFailure` also fires when the socket never opened, so a connect-time failure is now reported as an *interruption*.**

OkHttp uses `onFailure` for both a lost transport and a connection that was never established. `ConnectException`, `SSLHandshakeException` and a connect-time `UnknownHostException` are all `IOException` and none is a `ProtocolException`, so they take the first branch and the caller gets `"The face liveness session was interrupted."` / `"Retry the face liveness check."` for a session that never started.

For no-network-at-launch that reads fine. For a TLS-intercepting proxy, a pinning failure, or a bad regional endpoint it presents a permanent misconfiguration as transient — and the follow-up UI mapping will surface it to users as `SessionInterruptedException`, so the wrong advice reaches the screen.

The class already knows whether the socket ever opened: `webSocket` is assigned in `onOpen` (line 143).

```suggestion
    private fun classifyConnectionFailure(t: Throwable): PredictionsException = when {
        t is IOException && t !is ProtocolException && webSocket != null ->
            FaceLivenessSessionInterruptedException(cause = t)
        else -> PredictionsException(
            "An unknown error occurred during the Liveness flow.",
            t,
            "See attached exception for more details."
        )
    }
```

One caveat if you take this: `webSocket` is never nulled, so on the `ATTEMPT_RECONNECT` path (`onClosed`, line 236) it still holds the previous session's socket, and a reconnect that fails to connect would classify as interrupted. That is arguably correct there — a session really was in flight — but it is worth deciding deliberately rather than inheriting it.

If you would rather keep the current single-bucket behaviour, then the class name and message are the thing to reconsider: something like "the connection to the liveness service failed" would cover both without asserting a mid-session interruption the code cannot confirm.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/http/LivenessWebSocket.kt
```diff
@@ -247,16 +250,26 @@ internal class LivenessWebSocket(
             LOG.debug("WebSocket onFailure")
             super.onFailure(webSocket, t, response)
             if (!clientStoppedSession) {
-                val faceLivenessException = webSocketError ?: PredictionsException(
-                    "An unknown error occurred during the Liveness flow.",
-                    t,
-                    "See attached exception for more details."
-                )
-                onErrorReceived.accept(faceLivenessException)
+                onErrorReceived.accept(webSocketError ?: classifyConnectionFailure(t))
```
mattcreaser: 🔵 The `clientStoppedSession` invariant this PR claims to preserve is not exercised against the new classification path. The only test covering the guard, `onFailure does not call onError if client stopped` (`LivenessWebSocketTest.kt:182-196`), passes a bare `mockk()` as the throwable, which lands in the `else` branch. So nothing asserts that a `SocketException` arriving after a client-initiated stop stays silent — which is precisely the regression a future edit here would introduce. A one-line variant of the existing test with `SocketException(...)` in place of `mockk()` closes that.

aws-predictions/src/main/java/com/amplifyframework/predictions/aws/exceptions/FaceLivenessSessionInterruptedException.kt
```diff
@@ -0,0 +1,30 @@
...
+
+package com.amplifyframework.predictions.aws.exceptions
+
+import com.amplifyframework.annotations.InternalAmplifyApi
+import com.amplifyframework.predictions.PredictionsException
+
+/**
+ * Raised when the connection to the face liveness service is lost before the check completes, for example because the
+ * host app was backgrounded long enough for the socket to close, or because the device lost network connectivity.
+ */
```
mattcreaser: 🔵 The KDoc names two causes, but the classification rule admits every non-`ProtocolException` `IOException` — read timeouts, DNS failures, TLS errors, `EOFException`. A reader who matches the doc against the rule will find it narrower than the behaviour.

```suggestion
/**
 * Raised when the connection to the face liveness service is lost before the check completes — for example because
 * the host app was backgrounded long enough for the socket to close, or because the device lost network
 * connectivity. Any transport-level [java.io.IOException] other than a protocol violation is reported as this type.
 */
```

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -649,6 +750,14 @@ internal class LivenessWebSocketTest {
         assertEquals("AWS4-HMAC-SHA256", reconnectRequest.url.queryParameter("X-Amz-Algorithm"))
     }
 
+    // Response does not like to be mockk
+    private fun failureResponse() = Response.Builder()
+        .code(200)
+        .request(Request.Builder().url(URL("https://amazon.com")).build())
+        .protocol(Protocol.HTTP_2)
+        .message("Response")
+        .build()
```
mattcreaser: 🔵 Good helper, but it leaves the file with three copies of the same builder instead of one. `onFailure calls onError` (lines 169-174) and `onFailure does not call onError if client stopped` (lines 186-191) still build it inline, each under a typo'd copy of this comment ("Response does **noted** like to be mockk"). Replacing both call sites with `failureResponse()` removes twelve lines and the typo along with them.

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -184,6 +195,96 @@ internal class LivenessWebSocketTest {
...
+            verify {
+                onErrorReceived.accept(
+                    withArg {
+                        it.shouldBeInstanceOf<FaceLivenessSessionInterruptedException>()
+                        it.cause shouldBe transportFailure
+                    }
+                )
+            }
+        }
+    }
```
mattcreaser: 🔵 Four independent cases share one `@Test`, so a failure reports only `onFailure reports an interrupted session for any lost transport` and you have to read the loop to learn which exception broke. It also stops at the first failure, hiding the rest, and the `clearMocks(onErrorReceived)` on line 223 exists only to work around the sharing (it is a no-op on the first iteration).

A parameterized test, or `withData` from Kotest, would name the failing case for you. If you would rather keep it a loop, wrapping the body in `assertSoftly` at least reports all four.

thisisabhash (approved):

PR #3400, opened 2026-08-27
chore: Skip issue triage labeling for pull request comments

cadivus:
`adjust-labels` ran on PR comments, where the token only has `issues: write`, so it exited 1 with `Resource not accessible by integration`. Latest failure: #3399, [run 32994778127](https://github.com/aws-amplify/amplify-android/actions/runs/32994778127).

Restores the job-level `!github.event.issue.pull_request` guard that #2892 dropped. Issue behavior unchanged.

### DIFF
--- a/.github/workflows/issue_comment.yml
+++ b/.github/workflows/issue_comment.yml
@@ -23,13 +23,16 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       issues: write
+    # These triage labels only apply to issues. The GITHUB_TOKEN also lacks
+    # pull-requests: write, so labeling a PR would fail.
+    if: ${{ !github.event.issue.pull_request }}
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       ISSUE_NUMBER: ${{ github.event.issue.number }}
       REPOSITORY_NAME: ${{ github.event.repository.full_name }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/issue_comment.yml
```diff
@@ -23,13 +23,16 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       issues: write
+    # These triage labels only apply to issues. The GITHUB_TOKEN also lacks
+    # pull-requests: write, so labeling a PR would fail.
```
mattcreaser: 🔵 **Minor precision on the second sentence.** The token does not inherently lack `pull-requests: write` — it lacks it because the `permissions:` block three lines above is an explicit allowlist, so any scope not listed is dropped. The runner log confirms this: `GITHUB_TOKEN Permissions / Issues: write / Metadata: read`.

As written, a future reader could conclude that granting the scope is impossible, when in fact adding `pull-requests: write` here would make the labeling succeed. Skipping is still the right call — but the reason is the first sentence, not the token.

```suggestion
    # These triage labels only apply to issues, so there is nothing to do for a
    # PR comment. The permissions block above also withholds pull-requests: write,
    # so attempting to label a PR fails rather than no-oping.
```

Non-blocking; the comment is accurate enough as-is if you would rather keep it short.

mattcreaser (approved):

PR #3401, opened 2026-08-27
fix(all): make AndroidX Core upgrade compatible

skrcode:
This completes #3398 without changing the existing Dependabot workflow: this PR targets the Dependabot branch, so merging it updates the original dependency PR.

## What changed

- use AndroidX Core 1.18.0, the highest published stable release compatible with this repository's compile SDK 36 and Android Gradle Plugin 8.11.1
- declare `POST_NOTIFICATIONS` in the common module that posts notifications
- reuse the existing `PushNotificationPermission` check before dispatching notification work
- handle the permission-revocation race at the final `notify` call
- add three Robolectric tests covering pre-Android-13, denied, and granted permission states

AndroidX Core 1.18.0's AAR metadata requires compile SDK 36 and AGP 8.9.1. Version 1.19.0 requires compile SDK 37 and AGP 9.1.0, so accepting 1.19.0 would require a separate toolchain migration rather than a bounded dependency update.

## Reproduced failures

At the exact Dependabot head `e74847b789c234adf0c9443059109648e524accf`:

- `:apollo-appsync-amplify:checkDebugUnitTestAarMetadata` fails with four metadata errors because Core/Core KTX 1.19.0 require compile SDK 37 and AGP 9.1.0
- `:aws-push-notifications-pinpoint-common:lintDebug` fails with two `MissingPermission` errors at the notification call

## Verification

- focused compatibility check, common-module lint, formatting, API check, and all 22 common-module tests pass (19 existing + 3 new)
- Gradle dependency insight resolves both `androidx.core:core` and `androidx.core:core-ktx` to 1.18.0
- `./gradlew --no-daemon clean build -x :aws-kinesis:testDebugUnitTest -x :aws-kinesis:testReleaseUnitTest` passes from a clean tree: 3,483 tasks
- `./gradlew --no-daemon koverXmlReport -x :aws-kinesis:testDebugUnitTest -x :aws-kinesis:testReleaseUnitTest` passes: 1,875 tasks

The two Kinesis exclusions are a local Apple Silicon environment limitation, not a candidate regression. The untouched pre-upgrade commit reproduces 34/58 `:aws-kinesis:testDebugUnitTest` failures from the SQLite native loader. The exact Dependabot head cannot reach those tests because its AAR metadata check fails first. The final build still runs Kinesis compilation, assembly, API checks, Checkstyle, formatting, lint, license checks, and coverage wiring; GitHub's Linux CI remains authoritative for the two Kinesis test variants.


aws-push-notifications-pinpoint-common/src/main/java/com/amplifyframework/pushnotifications/pinpoint/PushNotificationsUtils.kt
```diff
@@ -107,6 +108,8 @@ class PushNotificationsUtils(
     @Suppress("DEPRECATION")
     @SuppressLint("NewApi")
     fun showNotification(notificationId: Int, payload: PinpointNotificationPayload, targetClass: Class<*>?) {
+        if (!PushNotificationPermission(context).hasRequiredPermission) return
```
mattcreaser: 🟡 **This guard is probably unreachable — and if it isn't, the caller now reports the wrong result.**

The only caller is `AWSPinpointPushNotificationsPlugin.handleNotificationReceived`, which gates on `canShowNotification()` (line 431):

```kotlin
private fun canShowNotification(payload: PinpointNotificationPayload): Boolean {
    val notificationsEnabled = pushNotificationsUtils.areNotificationsEnabled()
    ...
}
```

`areNotificationsEnabled()` delegates to `NotificationManagerCompat.areNotificationsEnabled()`, which on API 33+ returns `false` when `POST_NOTIFICATIONS` is denied — the permission gates the app's notification-enabled state. So `canShowNotification()` should already be `false` in exactly the case this new check catches, making this line dead at the only call site. Could you confirm which of the two actually rejected the notification in your testing?

If it *is* reachable (targetSdk < 33 on an API 33+ device is the case I'd look at), there's a correctness problem, because the caller can't tell that nothing happened:

```kotlin
canShowNotification(pinpointPayload) -> {
    pushNotificationsUtils.showNotification(...)   // silently returns
    PushNotificationResult.NotificationPosted     // ...but we claim success
}
```

The plugin then records the Pinpoint event with `"isOptedOut" to "false"` and hands `NotificationPosted` to `onSuccess`. Pinpoint gets told a notification was displayed when it wasn't, which corrupts campaign/journey delivery metrics — and that's silent, so nobody would notice.

Cleanest fix is to put the permission check where the decision is already being made, so the result stays truthful:

```kotlin
private fun canShowNotification(payload: PinpointNotificationPayload): Boolean {
    val notificationsEnabled = pushNotificationsUtils.areNotificationsEnabled()
    val hasPermission = PushNotificationPermission(context).hasRequiredPermission
    return notificationsEnabled && hasPermission && !payload.silentPush && deviceRegistered
}
```

That routes the denied case to `PushNotificationResult.OptedOut` and `isOptedOut=true`, which is what actually happened. Keeping a defensive check here too is fine — just not as the only signal.

Minor, while you're here: this allocates a `PushNotificationPermission` on every notification. A `private val permission by lazy { PushNotificationPermission(context) }` field would read better, since the object is stateless and just wraps `context`.

aws-push-notifications-pinpoint-common/src/main/java/com/amplifyframework/pushnotifications/pinpoint/PushNotificationsUtils.kt
```diff
@@ -136,8 +139,12 @@ class PushNotificationsUtils(
...
 
-            with(NotificationManagerCompat.from(context)) {
-                notify(notificationId, builder.build())
+            try {
+                with(NotificationManagerCompat.from(context)) {
+                    notify(notificationId, builder.build())
+                }
+            } catch (_: SecurityException) {
+                // The notification permission can be revoked after the pre-launch check.
             }
```
mattcreaser: 🟡 **Silent catch, and it's load-bearing for lint in a non-obvious way.**

Two things worth recording here.

**1. This `try/catch` — not the check on line 111 — is almost certainly what silences `MissingPermission`.** Lint recognises a `SecurityException` catch around a `@RequiresPermission` call, but it will not connect the line-111 guard to this call: the guard is in the enclosing function while `notify` is inside a `launch { }` lambda, which lint's dataflow won't trace across. So if someone later "cleans up" this empty catch because line 111 appears to make it redundant, `lintDebug` breaks again and the reason will be hard to find. Please state that in the comment.

**2. The catch discards the exception entirely.** If it ever does fire, a dropped notification leaves no trace anywhere — no log, no analytics, nothing. That's the worst failure mode to debug in the field. Also note the premise in the comment is doubtful: when `POST_NOTIFICATIONS` is revoked, `NotificationManagerService` drops the notification silently rather than throwing, so I'd expect this branch to be effectively dead. Either way, log it.

```suggestion
            try {
                with(NotificationManagerCompat.from(context)) {
                    notify(notificationId, builder.build())
                }
            } catch (error: SecurityException) {
                // Required: lint recognizes this SecurityException catch as the guard for
                // notify()'s @RequiresPermission(POST_NOTIFICATIONS). The check in
                // showNotification() does not satisfy lint, because it sits outside this
                // coroutine lambda. Removing this catch will fail :lintDebug.
                // Reached if the permission is revoked between that check and this call.
                logger.warn("Failed to post notification $notificationId: permission denied.", error)
            }
```

`PushNotificationsUtils` has no logger today, so this needs one injected or an `Amplify.Logging.forNamespace(...)` field — happy to defer on the exact mechanism, but please don't leave it fully silent.

aws-push-notifications-pinpoint-common/src/test/java/com/amplifyframework/pushnotifications/pinpoint/PushNotificationsUtilsTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+import org.robolectric.util.ReflectionHelpers
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.P])
+class PushNotificationsUtilsTest {
```
mattcreaser: 🟡 **This file tests `PushNotificationPermission`, not `PushNotificationsUtils` — and the code this PR adds has no coverage.**

All three tests assert on `PushNotificationPermission(context).hasRequiredPermission`. `PushNotificationsUtils` is never instantiated. Two consequences:

**The name and location are wrong.** `AGENTS.md` states: *"Test files: `<ClassName>Test.kt` — placed in the same package under the test source tree."* The class under test is `PushNotificationPermission`, which lives in `...pinpoint.permissions`. This should be `src/test/java/com/amplifyframework/pushnotifications/pinpoint/permissions/PushNotificationPermissionTest.kt`. As it stands, the next person who needs to test `PushNotificationsUtils` finds the obvious filename taken by unrelated tests.

**More importantly, these tests cover pre-existing code, not this PR's change.** `hasRequiredPermission` is unmodified on this branch — its `SDK_INT < 33 || checkSelfPermission(...) == GRANTED` logic is exactly as it was. So the description's *"add three Robolectric tests covering pre-Android-13, denied, and granted permission states"* is literally true but characterises the coverage more favourably than it is: the new behaviour — the early return in `showNotification`, and the `SecurityException` catch — is entirely untested.

Given the reporting concern I raised on `PushNotificationsUtils.kt:111`, the test that would actually protect this change asserts on the observable outcome: with `SDK_INT = 33` and the permission denied, `showNotification` posts nothing. Something like:

```kotlin
@Test
fun `does not post notification when permission is denied`() {
    setSdk(Build.VERSION_CODES.TIRAMISU)
    shadowOf(context).denyPermissions(PERMISSION_NAME)

    PushNotificationsUtils(context).showNotification(1, payload, null)

    shadowOf(NotificationManagerCompat.from(context)).allNotifications.shouldBeEmpty()
}
```

That needs the coroutine on line 113 to be observable from a test, which today it isn't — `CoroutineScope(Dispatchers.IO).launch` is created inline, so there's no way to await it. Injecting the dispatcher/scope into the constructor would make this testable and is worth doing regardless. If you'd rather not restructure that in this PR, please keep the rename and say plainly in the description that the new guard is untested.

aws-push-notifications-pinpoint-common/src/test/java/com/amplifyframework/pushnotifications/pinpoint/PushNotificationsUtilsTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    fun `accepts granted notification permission on API 33`() {
+        setSdk(Build.VERSION_CODES.TIRAMISU)
+        shadowOf(context).grantPermissions(PERMISSION_NAME)
+
+        PushNotificationPermission(context).hasRequiredPermission shouldBe true
+    }
+
+    private fun setSdk(sdk: Int) = ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", sdk)
```
mattcreaser: 🔵 Please add a comment explaining why this reflection exists.

I worked out the reason and it's a good one: the catalog pins `robolectric = "4.7"`, which has no SDK 33 sandbox, so `@Config(sdk = [Build.VERSION_CODES.TIRAMISU])` — the idiomatic approach — cannot work here. Patching `Build.VERSION.SDK_INT` under an SDK 28 sandbox is a reasonable workaround, and it's safe *in this specific case* because `hasRequiredPermission` reads nothing but `SDK_INT` and `checkSelfPermission`.

But that reasoning is invisible in the code. The next reader sees a reflective static-field write next to a `@Config` annotation and "simplifies" it to `@Config(sdk = [33])`, which fails on Robolectric 4.7. One line prevents that:

```suggestion
    // Robolectric 4.7 has no SDK 33 sandbox, so @Config(sdk = [TIRAMISU]) is not an option.
    // hasRequiredPermission only reads SDK_INT and checkSelfPermission, so patching SDK_INT
    // under the SDK 28 sandbox is sufficient here. Always restored in @After — the value is
    // static on a sandbox classloader that Robolectric reuses across test classes.
    private fun setSdk(sdk: Int) = ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", sdk)
```

The `@After` restore correctly handles the normal path (JUnit runs it even when a test fails), so I'm not flagging a leak — just noting in the comment that the restore is load-bearing, since the field is static on a sandbox that other Robolectric classes in this module's JVM would share.

gradle/libs.versions.toml
```diff
@@ -7,7 +7,7 @@ androidx-appcompat = "1.2.0"
 androidx-browser = "1.9.0"
 androidx-concurrent = "1.1.0"
 androidx-coordinatorlayout = "1.1.0"
-androidx-core = "1.19.0"
+androidx-core = "1.18.0"
```
mattcreaser: 🔵 **This pin is correct but undefended — Dependabot will re-propose 1.19.0.**

I confirmed the pin holds today: the highest transitive `androidx.core` requirement across this catalog's androidx dependencies is 1.12.0 (`work-runtime` 2.11.2), so Gradle's highest-wins resolution lands on your declared 1.18.0. Your dependency-insight claim checks out.

The problem is that nothing here records *why* 1.18.0 is deliberately one minor behind latest. Two predictable consequences:

1. Dependabot reopens the identical 1.19.0 PR on its next run, and whoever triages it repeats this whole investigation from scratch.
2. If any future dependency starts requiring `androidx.core >= 1.19.0`, highest-wins silently overrides this line and the `checkDebugUnitTestAarMetadata` failure returns — with no hint that 1.18.0 was intentional.

At minimum, leave the reason in the file:

```suggestion
# Pinned below latest deliberately: core 1.19.0's AAR metadata requires compile SDK 37 and
# AGP 9.1.0; this repo is on compile SDK 36 / AGP 8.11.1. 1.18.0 requires compile SDK 36 /
# AGP 8.9.1. Do not bump until the SDK 37 + AGP 9 toolchain migration lands.
androidx-core = "1.18.0"
```

Beyond that, worth considering in this PR or a follow-up:
- an `ignore` entry in `.github/dependabot.yml` for `androidx.core:*` until the toolchain migration, so the bot stops proposing it;
- a resolution constraint with `because("...")` if you want the failure to be a loud conflict rather than a silent override.

Also flagging for the reviewer, since the description doesn't mention it: `main` currently declares `androidx-core = "1.5.0"`, so the net effect of merging #3398 with this fix is 1.5.0 → 1.18.0, not a small step. Transitive requirements meant ~1.12.0 was already being resolved in practice, so the real change is narrower than 13 minors — but this is what newly exposes `NotificationManagerCompat.notify`'s `@RequiresPermission` to lint on the module's own compile classpath, which is the actual root cause of the lint failure this PR is fixing. Might be worth stating explicitly.

aws-push-notifications-pinpoint-common/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,8 @@
 -->
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
mattcreaser: 🔵 `POST_NOTIFICATIONS` is now declared in two manifests.

`aws-push-notifications-pinpoint/src/main/AndroidManifest.xml` already has it:

```xml
<uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
```

Adding it here is still the right call — lint's `MissingPermission` is evaluated per module against that module's merged manifest, and this is the module that actually calls `notify`, so the declaration belongs with the code that needs it. The manifest merger deduplicates `uses-permission`, so there's no build or APK impact.

Two small notes:
- The description says *"declare `POST_NOTIFICATIONS` in the common module that posts notifications"* without mentioning it already existed downstream. Worth a word, so a reviewer doesn't read this as a previously-missing permission.
- The downstream declaration is now redundant, since `aws-push-notifications-pinpoint` depends on this module and inherits it through the merge. Removing it would keep the permission declared in exactly one place — the module that requires it. Your call whether that belongs in this PR.

PR #3403, opened 2026-08-28
fix(auth): Return signed-out session for User Pool-only apps

cadivus:
## What
Return a stable signed-out session for User Pool-only apps. User Pool fields report `SignedOutException`, identity fields report `InvalidAccountTypeException`, and repeated calls no longer refresh empty credentials.

## Why
No Identity Pool is valid for User Pool-only apps. The result split follows Flutter and Swift's distinction between missing Identity Pool data and signed-out User Pool data.

## Testing
- `:aws-auth-cognito:testDebugUnitTest` — 564 tests, 0 failures
- `:aws-auth-cognito:ktlintCheck` and `:aws-auth-cognito:apiCheck`
- Android emulator: first, repeated, and post-restart fetches

### DIFF
--- a/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.kt
+++ b/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.kt
@@ -20,6 +20,7 @@ import com.amplifyframework.auth.AWSCognitoUserPoolTokens
 import com.amplifyframework.auth.AWSCredentials
 import com.amplifyframework.auth.AuthException
 import com.amplifyframework.auth.cognito.exceptions.configuration.InvalidUserPoolConfigurationException
+import com.amplifyframework.auth.cognito.exceptions.service.InvalidAccountTypeException
 import com.amplifyframework.auth.cognito.helpers.SessionHelper
 import com.amplifyframework.auth.exceptions.ConfigurationException
 import com.amplifyframework.auth.exceptions.InvalidStateException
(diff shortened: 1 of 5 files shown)


aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthSession.kt
```diff
@@ -164,12 +165,26 @@ internal fun AmplifyCredential.getCognitoSession(exception: AuthException? = nul
...
+                        "Identity Pool not configured.",
+                        "Please check amplifyconfiguration.json file."
+                    )
+                )
+            } else {
+                exception
+            }
+            AWSCognitoAuthSession(
+                false,
+                identityIdResult = AuthSessionResult.failure(identityException),
```
mattcreaser: 🟡 **The identity axis now reports a different exception type depending on sign-in state, within the same app.**

For a User Pool-only app that is *signed in*, the `AmplifyCredential.UserPool` branch above (lines 124-140) fails `identityIdResult` / `awsCredentialsResult` with `ConfigurationException("Could not retrieve Identity ID", "Cognito Identity not configured. Please check amplifyconfiguration.json file.")`. After this change, the *same app* while signed out fails the same two fields with `InvalidAccountTypeException` — a `ServiceException`. Nothing about the configuration changed between those two calls; only auth state did.

Flutter does not have this seam: `InvalidAccountTypeException.noIdentityPool()` is applied whenever `!hasIdentityPool`, independent of whether user pool tokens are present (`fetch_auth_session_state_machine.dart:232-244`).

Either switch the `AmplifyCredential.UserPool` branch to `InvalidAccountTypeException` as well so one app reports one type, or add a line to the PR description explaining why signed-in and signed-out intentionally differ. As it stands a consumer writing `when (identityIdResult.error)` has to handle both.

aws-auth-cognito/src/test/java/com/amplifyframework/auth/cognito/usecases/FetchAuthSessionUseCaseTest.kt
```diff
@@ -410,4 +417,40 @@ class FetchAuthSessionUseCaseTest {
...
+        result.isSignedIn shouldBe false
+        result.userSubResult.error.shouldBeInstanceOf<SignedOutException>()
+        result.userPoolTokensResult.error.shouldBeInstanceOf<SignedOutException>()
+        result.identityIdResult.error.shouldBeInstanceOf<InvalidAccountTypeException>()
+        result.awsCredentialsResult.error.shouldBeInstanceOf<InvalidAccountTypeException>()
+        verify(exactly = 0) { stateMachine.send(any()) }
+    }
+
+    @Test
+    fun `forceRefresh fetch returns aligned signed-out results without refreshing`() = runTest {
```
mattcreaser: 🔵 **The two new tests are a matched pair but assert different amounts.**

`repeated fetch ...` above asserts all four typed results; this one asserts only `userSubResult` and `identityIdResult`. Since the whole point of the pair is “`forceRefresh` changes nothing”, asserting the identical four here makes that claim directly rather than by implication — and it would catch a regression that only affects `awsCredentialsResult` or `userPoolTokensResult` under `forceRefresh`.

Missing coverage worth adding while you are here: nothing pins that a **non**-`SignedOutException` `SessionError` carrying `AmplifyCredential.Empty` still refreshes. `sends RefreshSession for SessionError with non-federated credentials` (line 211) uses `userPoolCredential`, so the `Empty` + other-exception combination — the nearest neighbour to the new branch — is currently unguarded in `execute()`.

mattcreaser (approved):

PR #3405, opened 2026-08-31
feat(api): add AppSyncException hierarchy and endpoint parser

mattcreaser:

*Issue #, if available:*

n/a

*Description of changes:*

### Sealed `AppSyncException` hierarchy

Base extends **foundation's** `AmplifyException` — not `core`'s, and not `ApiException`. `ApiException` is slated for removal in V3, so inheriting it would have made this hierarchy a blocker to deleting it.

Two levels. Five sealed intermediate groups so a caller can catch a category, with flat prefixed top-level leaves:

```kotlin
is Result.Failure -> when (val e = result.error) {
    is AppSyncAuthException -> reauthenticate()      // any of 6 auth leaves
    is AppSyncNetworkException -> retryLater()
    else -> report(e)
}
```

### Partition-aware endpoint parser

Replaces `Configuration.inferRegion`'s regex, which assumed a `.amazonaws.com` suffix. Being unanchored, it extracted the right *region* for China endpoints by accident — but it produced no realtime host and no partition, and a realtime host built on that assumption does not resolve for `.amazonaws.com.cn`.

The parser returns `region`, `host`, `realtimeHost` and `dnsSuffix`. It swaps only the `appsync-api` label for `appsync-realtime-api` and preserves everything after the region, which is why China and GovCloud need no special-casing. Region matching is structural (`^[a-z]{2,}(-[a-z]+)+-\d+$`) rather than an allowlist, so a region launched after this ships still parses. A custom domain carries no region, so it fails and the caller must set `region` explicitly — which `Configuration` already supports.

n/a

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests — not applicable; every client method body is still `TODO()`

### DIFF
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -18,6 +18,7 @@ import com.amplifyframework.annotations.ExperimentalAmplifyApi
 import com.amplifyframework.api.graphql.GraphQLRequest
 import com.amplifyframework.api.graphql.GraphQLResponse
 import com.amplifyframework.foundation.result.Result
+import com.amplifyframework.foundation.result.getOrThrow
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.SharedFlow
 import okhttp3.OkHttpClient
@@ -143,9 +144,6 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
                 require(::endpoint.isInitialized) { "endpoint is required" }
                 require(::authorization.isInitialized) { "authorization is required" }
                 val resolvedRegion = region ?: inferRegion(endpoint)
-                requireNotNull(resolvedRegion) {
-                    "region is required. Either set it explicitly or use a standard AppSync endpoint URL."
-                }
                 return Configuration(
                     endpoint = endpoint,
                     authorization = authorization,
@@ -167,7 +165,8 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
              * Expected format: `https://{id}.appsync-api.{region}.{dnsSuffix}/graphql`, across the
              * commercial, China and GovCloud partitions.
              */
-            internal fun inferRegion(endpoint: String): String? = AppSyncEndpointParser.regionOrNull(endpoint)
+            internal fun inferRegion(endpoint: String): String =
+                AppSyncEndpointParser.parse(endpoint).getOrThrow().region
         }
     }
 }
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -15,7 +15,6 @@
 package com.amazonaws.appsync
 
 import com.amplifyframework.annotations.ExperimentalAmplifyApi
-import com.amplifyframework.api.ApiException
 import com.amplifyframework.api.graphql.GraphQLRequest
 import com.amplifyframework.api.graphql.GraphQLResponse
 import com.amplifyframework.foundation.result.Result
@@ -59,18 +58,18 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * Execute a GraphQL query.
      *
      * @param request The GraphQL request. Use model helpers or construct manually.
-     * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [ApiException].
+     * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [AppSyncException].
      */
-    suspend fun <T> query(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, ApiException> =
+    suspend fun <T> query(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> =
         TODO("Query implementation will be added in a follow-up PR")
 
     /**
      * Execute a GraphQL mutation.
      *
      * @param request The GraphQL request. Use model helpers or construct manually.
-     * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [ApiException].
+     * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [AppSyncException].
      */
-    suspend fun <T> mutate(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, ApiException> =
+    suspend fun <T> mutate(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> =
         TODO("Mutation implementation will be added in a follow-up PR")
 
     /**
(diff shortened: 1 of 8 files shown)


aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -165,12 +164,10 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
...
              * Infer the AWS region from an AppSync endpoint URL.
-             * Expected format: `https://{id}.appsync-api.{region}.amazonaws.com/graphql`
+             * Expected format: `https://{id}.appsync-api.{region}.{dnsSuffix}/graphql`, across the
+             * commercial, China and GovCloud partitions.
              */
-            internal fun inferRegion(endpoint: String): String? {
-                val regex = Regex("""\.appsync-api\.([a-z0-9-]+)\.amazonaws\.com""")
-                return regex.find(endpoint)?.groupValues?.get(1)
-            }
+            internal fun inferRegion(endpoint: String): String? = AppSyncEndpointParser.regionOrNull(endpoint)
```
mattcreaser: 🟡 **`AppSyncEndpointResolutionException` is dead at the one place it was built for.**

`regionOrNull` discards the typed failure the parser just constructed, and `Configuration.Builder.build()` (line 147 at this ref) then re-derives the same diagnosis by hand:

```kotlin
val resolvedRegion = region ?: inferRegion(endpoint)
requireNotNull(resolvedRegion) {
    "region is required. Either set it explicitly or use a standard AppSync endpoint URL."
}
```

That `requireNotNull` message is a paraphrase of `AppSyncEndpointResolutionException`'s own default recovery suggestion — *"Provide a standard AppSync endpoint URL, or set region explicitly."* — and its KDoc says the type is for exactly this: *"The endpoint could not be parsed, or a region could not be resolved from it."* So the PR adds a precise, catchable error, throws it away one call later, and surfaces an `IllegalArgumentException` that loses which of the six failure modes actually fired. A caller who passed `https://abc.appsync-api.notaregion.amazonaws.com/graphql` gets told to use a standard URL, with no indication that `notaregion` was the problem.

It also cuts against AGENTS.md §6, *"Kotlin code SHOULD NOT throw exceptions as part of their API — use `Result<T, E>` instead."* I recognise `build()` already throws for `endpoint` and `authorization`, so that's pre-existing style rather than something this PR introduced — but this is a newly-typed error being deliberately downgraded to an untyped throw.

Two ways out, either fine:
- Have `build()` propagate the parser's failure, e.g. `throw parse(endpoint).errorOrNull()` when region is absent, so the message and recovery suggestion survive even inside an exception.
- Or leave it and say in the description that wiring this up is Phase 3, so the type isn't read as unused-by-oversight.

What I'd avoid is merging it silently unused, because the next reader can't tell whether `regionOrNull` is a deliberate lossy convenience or a missed hookup.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
```diff
@@ -0,0 +1,132 @@
...
+            ?: return failure("Could not read a host from the endpoint URL '$endpoint'.")
+
+        // {apiId}.appsync-api.{region}.{dnsSuffix...}
+        val labels = host.split('.')
+        val apiLabelIndex = labels.indexOf(API_LABEL)
+
+        if (apiLabelIndex < 1) {
+            return failure(
+                "The endpoint host '$host' is not a standard AppSync endpoint: expected a " +
+                    "'$API_LABEL' label, as in {apiId}.$API_LABEL.{region}.amazonaws.com."
```
mattcreaser: 🔵 Two small things in this failure.

**The message hardcodes `amazonaws.com`** — in the parser whose entire reason for existing is that assuming that suffix is wrong. A China user pasting a `.amazonaws.com.cn` endpoint that fails for an unrelated reason gets told the expected form ends in `amazonaws.com`, which is the opposite of the fix. Use `{dnsSuffix}` in the exemplar instead, matching the KDoc at line 44 and the `inferRegion` KDoc, which both already say `{dnsSuffix}`.

**`apiLabelIndex < 1` conflates two distinct failures.** `indexOf` returns `-1` for "no `appsync-api` label anywhere" and `0` for "`appsync-api` is the first label, so there's no apiId" — and both produce a message asserting the label is expected. For the second case that reads as a contradiction: the label is right there. Your test `an endpoint with no apiId label before appsync-api fails` hits exactly this path and only asserts `Result.Failure`, so the misleading text isn't caught.

```suggestion
        if (apiLabelIndex < 0) {
            return failure(
                "The endpoint host '$host' is not a standard AppSync endpoint: expected a " +
                    "'$API_LABEL' label, as in {apiId}.$API_LABEL.{region}.{dnsSuffix}."
            )
        }
        if (apiLabelIndex == 0) {
            return failure(
                "The endpoint host '$host' has no API id label before '$API_LABEL'."
            )
        }
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
```diff
@@ -0,0 +1,132 @@
...
+            .substringBefore('#')
+            .substringBefore(':') // strip any port
+            .lowercase()
+        return host.ifEmpty { null }
+    }
+
+    // Region labels are of the form {partition}-{area}-{number}, e.g. us-east-1, cn-north-1,
+    // us-gov-west-1, ap-southeast-4. Deliberately structural rather than an allowlist, so a region
+    // launched after this ships still parses.
+    private fun isRegionLike(candidate: String): Boolean = Regex("""^[a-z]{2,}(-[a-z]+)+-\d+$""").matches(candidate)
```
mattcreaser: 🔵 The `Regex` is compiled on every call. `parse` is not hot, so this is cosmetic — but this is an `object` that already hoists `API_LABEL` and `REALTIME_LABEL` into `private const val`, so the pattern is the odd one out.

```suggestion
    private val REGION_PATTERN = Regex("""^[a-z]{2,}(-[a-z]+)+-\d+$""")

    private fun isRegionLike(candidate: String): Boolean = REGION_PATTERN.matches(candidate)
```

(Keep the explanatory comment above it — the "structural, not an allowlist" rationale is the most useful comment in the file.)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Resolves just the region from [endpoint], for callers that only need that.
+     *
+     * @return the region, or null if the endpoint is not a standard AppSync URL.
+     */
+    fun regionOrNull(endpoint: String): String? = (parse(endpoint) as? Result.Success)?.data?.region
+
+    private fun hostOf(endpoint: String): String? {
+        val withoutScheme = endpoint.substringAfter("://", missingDelimiterValue = endpoint)
```
mattcreaser: 🔵 **The scheme is stripped and discarded without ever being looked at**, so `http://abc123.appsync-api.us-east-1.amazonaws.com/graphql` — or `ftp://`, or any other scheme — parses as a perfectly good endpoint. Your `tolerates a missing scheme` test locks in that a bare host is accepted, which is fine, but nothing constrains what a *present* scheme may be.

That's harmless in this PR, since nothing builds a URL from the result yet. It stops being harmless in Phase 3: SigV4 signing over cleartext transmits derived credentials in the clear, and the checklist on this PR asserts security best practices are followed. Given the parser is the single chokepoint every endpoint passes through, rejecting a non-`https` scheme here is far cheaper than auditing each later call site.

Suggest either rejecting a scheme other than `https` (and `wss` if realtime hosts are ever fed back through), or, if you'd rather not add a failure mode in this PR, noting it as Phase 3 work in the KDoc so it isn't assumed handled.

Minor, same function: `substringAfter("://")` takes the *first* occurrence, so an endpoint with no scheme but a `://` later in a query string resolves the wrong host. Contrived enough to ignore — mentioning it only because it fails toward a rejected endpoint rather than a wrong one, which is the safe direction.

mattcreaser:
Note: I'm aware the endpoint parser does not handle custom endpoints. This is will be fixed in the subscription PR.

harsh62 (approved):

PR #3406, opened 2026-08-31
feat(all): add CloudWatch client

thisisabhash:

*Issue #, if available:*

N/A

*Description of changes:*

Adds the v3 standalone CloudWatch logging client: a lightweight `AmplifyCloudWatchClient` (`:aws-cloudwatch`, package `com.amplifyframework.cloudwatch`) that depends only on `AmplifyFoundation` / `foundation-bridge` and the `aws-cloudwatchlogs` SDK (no `:core` / `:aws-core`), following the `:aws-kinesis` standalone-client pattern (Kotlin, coroutines, `Result<T, E>`, builder + DSL options). The public API is marked `@ExperimentalAmplifyApi`.

**`:aws-cloudwatch` (new client)**
- `AmplifyCloudWatchClient` — implements `LogSink`. `isEnabledFor(level)` reports the client's enabled state; `emit(message)` filters against `LoggingConstraints` then persists; `enable()`/`disable()` toggle capture + automatic flushing (disable preserves already-buffered entries); `setUserIdentifier(id)`; `setLoggingConstraints(constraints)`; `suspend flushLogs(): FlushResult`; `getCloudWatchLogsClient()`; `events: SharedFlow<LoggingEvent>` (write/flush failures, `DROP_OLDEST` buffered).
- `CloudWatchLogManager` — the batching/upload engine (per-request limits: 10,000 events / 1,048,576 bytes / +26 bytes per event / 24h span; events over the per-event limit are dropped so they can't block the buffer), log-stream creation, delete-on-success, too-new-event retry, and coroutine + WorkManager auto-flush (`CloudWatchRouterWorker` / `CloudWatchLogsSyncWorker` / `CloudWatchWorkerFactory`).
- Persistence: the SQLCipher-backed layer (`CloudWatchDatabase`, `CloudWatchDatabaseHelper`, `LogEventTable`, `LogEvent` in `com.amplifyframework.cloudwatch.db`; `CloudWatchLogEvent` in `.models`) plus `CloudWatchPreferences`, with the database passphrase persisted via foundation's `AmplifyKeyValueRepository`.
- `CloudWatchLoggingFilter` — namespace-based filtering with user > namespace > default precedence (namespace keys case-insensitive, user keys case-sensitive; `None` blocks all).
- `AmplifyCloudWatchClientOptions` (builder + `invoke { }` DSL): `logGroupName`, `localStoreMaxSizeInMB` = 5, `flushStrategy`, `loggingConstraints`, `configureClient`, plus the `CloudWatchLogsClientConfigurationProvider` typealias.
- `FlushStrategy` — `Interval(Duration = 60.seconds)` and `None`; `FlushData` (empty result) + `FlushResult` typealias.
- `LoggingConstraints` / `UserLogLevel` — `defaultLogLevel` (default `Error`), `namespaceLogLevel`, `userLogLevel`.
- `LoggingEvent` — sealed `WriteLogFailure` / `FlushLogFailure` (`context`, `cause`).
- `AmplifyCloudWatchException` — sealed hierarchy (`Storage` / `Service` / `Configuration` / `Unknown`) over foundation `AmplifyException`, with an internal `from(...)` mapper.

**Multi-instance isolation** — the worker factory, the WorkManager unique-work name, and the on-device store (database + passphrase preferences) are all keyed by log group, so two clients targeting different log groups can't hijack each other's delegate factory, clobber each other's scheduled flush, or share a buffer.

**`:foundation` — `KeyValueRepository` moved out of `:core`**
- Moved the `AmplifyKeyValueRepository` cluster (`KeyValueRepository`, `AmplifyKeyValueRepository`, `EncryptedKeyValueRepository`, `InMemoryKeyValueRepository`(`Provider`)) from `:core` into `foundation/androidMain`, so standalone clients can persist a database passphrase without depending on `:core`. `KeyValueRepository` and `EncryptedKeyValueRepository` stay public (their public API moves from `core.api` to `foundation.api`); `AmplifyKeyValueRepository` remains `@InternalAmplifyApi` and `InMemoryKeyValueRepository` `@InternalApiWarning`.
- Enabled Robolectric-based host tests in the KMP module (`jvmToolchain(17)` + host-test Android resources) so the relocated keystore / `EncryptedSharedPreferences` tests run in `:foundation`.
- `LogLevel.allows` promoted to `@InternalAmplifyApi` and now encapsulates all `None` handling (a `None` message is never emitted; a `None` threshold blocks everything); the CloudWatch filter reuses it.

**`:core` and consumers**
- Removed the `KeyValueRepository` cluster from `:core`'s public API and repointed `:aws-auth-cognito`, `:aws-pinpoint-core`, `:aws-analytics-pinpoint`, `:aws-push-notifications-pinpoint`, and `:aws-auth-plugins-core` to the `:foundation` types (behavior unchanged). `apiDump` updated for `:core` and `:foundation`.
- Note on the shared keystore key: all Amplify components (Cognito, Pinpoint, this client, …) share the single `"amplify_master_key"` alias managed by the one `EncryptedKeyValueRepository`; on keystore corruption the recovery path resets those Amplify-encrypted stores together. This is documented in the class as an accepted trade-off — consolidating removes the duplicate keystore-management implementation but does not isolate the components' key material.

**`:aws-logging-cloudwatch` (v2 plugin)**
- Now depends on `:aws-cloudwatch` and reuses its persistence layer (`db/*`, `models/CloudWatchLogEvent`, `CloudWatchPreferences`) instead of its own copies. Public API and behavior are unchanged (apiCheck passes); it keeps its own engine, Hub events, remote config, and `CategoryType` constraints.

- Unit tests in `:aws-cloudwatch` (Robolectric + MockK): filter precedence + `None` handling, exception mapping, options builder/DSL, client behavior (emit filtering, disabled-drop, flush `Result` mapping, enable/disable, `setUserIdentifier`/`setLoggingConstraints`), and the manager engine — batching boundary cases (crossing the 10,000-event / 1 MB / 24h limits plus oversized-event drop), stream creation/reuse, too-new retry, failure callbacks, WorkManager scheduling, and per-log-group instance isolation.
- Connected integration tests in `:aws-cloudwatch` (`AmplifyCloudWatchClientInstrumentationTest`, `DeviceFarmTestBase`) exercise the real client against a provisioned CloudWatch backend — escape hatch, flushing emitted messages, and verbose-after-enable (sent) / verbose-after-disable (not sent) — verifying delivery via `filterLogEvents`. They use guest Cognito identity-pool credentials and read two gitignored, CI-injected resources from `src/androidTest/res/raw/`: `amplify_outputs.json` (Auth) and `amplifyconfiguration_logging.json` (region, log group, etc.). Verified 10/10 connected tests pass on an emulator.
- The relocated `KeyValueRepository` / `EncryptedKeyValueRepository` unit tests run as `:foundation` Robolectric host tests.
- Full `build` (compile, ktlint, checkstyle, Android lint, `apiCheck`, unit tests, coverage) passes for `:aws-cloudwatch`, `:foundation`, `:core`, `:aws-logging-cloudwatch`, and the repointed consumer modules; v2's existing unit tests still pass with its public API unchanged.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [x] Added Integration Tests


aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,325 @@
...
+                                logGroupName = this@CloudWatchLogManager.logGroupName
+                                logStreamName = streamName
+                            }
+                        )
+                        // Everything up to tooNewLogEventStartIndex was accepted; the rest stays for retry.
+                        var acceptedIds = batch.sendableIds
+                        response.rejectedLogEventsInfo?.tooNewLogEventStartIndex?.let {
+                            acceptedIds = acceptedIds.slice(IntRange(0, it - 1))
+                        }
+                        lastAttemptedIds = acceptedIds
```
mattcreaser: 🔴 **The cache-full recovery valve no longer drops the batch that failed — regression vs v2.**

`lastAttemptedIds` is assigned *after* `putLogEvents` returns. So when `putLogEvents` throws, `lastAttemptedIds` still holds either `emptyList()` (first batch) or the ids of the *previous, already-deleted* batch. The catch block's `bulkDelete(lastAttemptedIds)` is therefore a no-op exactly when it matters.

v2 assigned before the call, which is what made the recovery work:

```kotlin
// aws-logging-cloudwatch/.../CloudWatchLogManager.kt on main
val nextBatch = getNextBatch(queriedEvents)
val inputLogEvents = nextBatch.first
inputLogEventsIdToBeDeleted = nextBatch.second   // <-- before putLogEvents
if (inputLogEvents.isEmpty()) { return@withContext }
createLogStreamIfNotCreated(streamName, groupName, client)
client.putLogEvents(...)
```

Consequence: with a full local store and a batch CloudWatch permanently rejects, the buffer never drains. `getNextBatch` removes the events from the in-memory list but they stay in the DB, so the next `saveLogEvent` sees `isCacheFull()`, triggers another sync, picks the same rows, fails again — indefinitely, with every subsequent log write paying for a failed flush.

Suggested fix (also drops the misleading name — the variable holds *accepted* ids after the `tooNew` slice):

```suggestion
                        lastAttemptedIds = batch.sendableIds
                        val response = client.putLogEvents(
                            PutLogEventsRequest {
                                logEvents = batch.events
                                logGroupName = this@CloudWatchLogManager.logGroupName
                                logStreamName = streamName
                            }
                        )
                        // Everything up to tooNewLogEventStartIndex was accepted; the rest stays for retry.
                        var acceptedIds = batch.sendableIds
                        response.rejectedLogEventsInfo?.tooNewLogEventStartIndex?.let {
                            acceptedIds = acceptedIds.slice(IntRange(0, it - 1))
                        }
```

There is no test for "putLogEvents fails while the cache is full", which is why this slipped — worth adding alongside the fix.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,237 @@
...
+    private val logManager = logManagerFactory(cloudWatchLogsClient, eventsFlow)
+
+    init {
+        scope.launch { logManager.startSync() }
+    }
+
+    // region LogSink
+
+    /** Returns true if the client is enabled. */
+    override fun isEnabledFor(level: LogLevel): Boolean = isEnabled
```
mattcreaser: 🟡 **`isEnabledFor` ignoring `level` breaks the `LogSink` contract and disables lazy message evaluation framework-wide.**

`LogSink` documents it as "Returns true if this sink will emit logs at the given level", and `AmplifyLogging` relies on that to avoid materializing lazy messages:

```kotlin
// foundation/src/commonMain/.../AmplifyLogging.kt
override fun log(level: LogLevel, cause: Throwable?, message: () -> String) {
    // Only evaluate the message if at least one sink will emit it
    if (sinks.any { it.isEnabledFor(level) }) {
        log(level, message(), cause)
    }
}
```

With the default `LoggingConstraints(defaultLogLevel = Error)`, registering this client via `AmplifyLogging.addSink` makes `sinks.any { … }` true for *every* level, so every `logger.verbose { … }` / `logger.debug { … }` string in the whole framework gets built and then discarded by `emit`'s filter. Correctness is fine; the cost is not.

`isEnabledFor` has no namespace, but it can still answer conservatively using the most permissive configured threshold — never under-reporting, so per-namespace filtering in `emit` stays authoritative:

```kotlin
override fun isEnabledFor(level: LogLevel): Boolean = isEnabled && filter.couldLog(level)
```

where `couldLog` checks `level` against `defaultLogLevel`, `namespaceLogLevel.values` and the `userLogLevel` thresholds.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Disable logging and automatic flushing. Messages emitted while disabled are dropped, but any
+     * log entries already buffered locally are preserved and will be sent on the next [flushLogs] or
+     * once re-[enable]d.
+     */
+    fun disable() {
+        logger.info { "Disabling CloudWatch logging and automatic flushing" }
+        isEnabled = false
+        logManager.stopSync()
+    }
```
mattcreaser, line 215: 🟡 **`enable()`/`disable()` are asymmetric, so a quick `enable()` → `disable()` can leave auto-flush running.**

`enable()` defers the state flip into `scope.launch { logManager.startSync() }` while `disable()` calls `logManager.stopSync()` inline. If `disable()` wins the race, the queued `startSync()` runs afterwards, sets `stopSync = false` and re-enqueues the WorkManager job — the client reports `isEnabled == false` but keeps flushing on the interval.

`CloudWatchLogManager.startSync()` is `suspend` but does nothing suspending (`stopSync = false` then `enqueueSync()`, which is a plain function). Making it a normal `fun` and calling it inline from `enable()` removes the race and the coroutine, and makes the two methods symmetric:

```suggestion
    /** Enable logging and automatic flushing. */
    fun enable() {
        logger.info { "Enabling CloudWatch logging and automatic flushing" }
        isEnabled = true
        logManager.startSync()
    }

    /**
     * Disable logging and automatic flushing. Messages emitted while disabled are dropped, but any
     * log entries already buffered locally are preserved and will be sent on the next [flushLogs] or
     * once re-[enable]d.
     */
    fun disable() {
        logger.info { "Disabling CloudWatch logging and automatic flushing" }
        isEnabled = false
        logManager.stopSync()
    }
```

(The `init` block's `scope.launch { logManager.startSync() }` can then become a direct call too.)

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,325 @@
...
+        cancelSync()
+        logger.debug { "Stopping sync" }
+    }
+
+    suspend fun syncLogEventsWithCloudwatch() {
+        // Atomic guard so overlapping triggers (manual flush, cache-full write, WorkManager) can't
+        // run concurrently and delete each other's rows.
+        if (!isSyncInProgress.compareAndSet(false, true)) {
+            return
+        }
```
mattcreaser: 🟡 **`flushLogs()` reports `Success` when the flush was skipped, not performed.**

When a sync is already in flight (interval worker, or a cache-full write) the `compareAndSet` guard returns immediately, `flushLogs()` sees no exception and returns `Result.Success(FlushData())`. A caller that awaits `flushLogs()` and then reads CloudWatch has no way to tell "flushed" from "skipped" — the integration test's `awaitEvents` helper asserts `shouldBeSuccess()` on every attempt and would be satisfied by a run that flushed nothing.

The guard itself is right (and an improvement over v2's racy `get()`/`set(true)`); it's the reporting that loses information. `FlushData` was introduced precisely so "the operation can grow its result shape without a breaking change" — a `flushed: Boolean` or an events-sent count would make this observable. Alternatively await the in-flight sync (e.g. a `Mutex`) so `flushLogs()` genuinely means "pending entries have been attempted".

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/db/CloudWatchDatabase.kt
```diff
@@ -94,25 +85,24 @@ internal class CloudWatchLoggingDatabase(
         }
     }
 
-    internal fun isCacheFull(cacheSizeInMB: Int): Boolean {
-        val path = context.getDatabasePath(CloudWatchDatabaseHelper.DATABASE_NAME)
+    fun isCacheFull(cacheSizeInMB: Int): Boolean {
+        val path = context.getDatabasePath(databaseName)
         return if (path.exists()) {
             path.length() >= cacheSizeInMB * mb
```
mattcreaser: 🟡 **`cacheSizeInMB * mb` overflows `Int`, and `localStoreMaxSizeInMB` is now settable from a public builder.**

`mb` is an `Int`, so the multiplication happens in `Int` before widening for the comparison against `path.length()`. At `localStoreMaxSizeInMB = 2048` the product is exactly `Int.MIN_VALUE`; at 4096 it is `0`. Either way `path.length() >= <negative-or-zero>` is always true, so `isCacheFull()` returns true for an empty database and **every single `saveLogEvent` triggers a full flush**.

This expression is inherited from v2, but there the value came from internal plugin config. `AmplifyCloudWatchClientOptions.Builder.localStoreMaxSizeInMB(Int)` is public API with no validation, so it is now reachable.

```suggestion
        val path = context.getDatabasePath(databaseName)
        return if (path.exists()) {
            path.length() >= cacheSizeInMB.toLong() * mb
```

Also worth a `require(localStoreMaxSizeInMB > 0)` in `AmplifyCloudWatchClientOptions.build()` — a zero or negative value silently means "flush on every write" today.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/worker/CloudWatchRouterWorker.kt
```diff
@@ -0,0 +1,86 @@
...
+
+            override val entries: MutableSet<MutableMap.MutableEntry<String, CloudWatchWorkerFactory>>
+                get() = backingWorkerMap.entries
+        }
+    }
+
+    override fun startWork(): ListenableFuture<Result> {
+        // Route to the factory registered for this work's log group (see CloudWatchLogManager), so multiple
+        // clients don't hijack each other. Falls back to the shared key if no id was provided.
+        val factoryKey = parameter.inputData.getString(WORKER_ID) ?: WORKER_FACTORY_KEY
```
mattcreaser, line 67: 🟡 **This future never completes, so the "retry" path hangs the worker instead of retrying.**

`CallbackToFutureAdapter.getFuture` takes a `Resolver<T>` whose return value is only a *debug tag*; the result must be delivered through the `Completer`. Here the lambda returns `Result.retry()` as the tag and never calls `completer.set(...)`, so the returned `ListenableFuture` stays unresolved — WorkManager waits until it times out (~10 min) rather than rescheduling. The comment above says this exists to handle the factory-not-yet-registered race, but as written the race resolves by stalling.

```suggestion
            // this is to prevent a race condition where workManager starts work before worker factory is initialized
            if (!isWorkerFactoriesInitialized) {
                return CallbackToFutureAdapter.getFuture { completer ->
                    completer.set(Result.retry())
                    "CloudWatchRouterWorker.retry"
                }
            } else {
```

This is copied verbatim from v2's `CloudwatchRouterWorker`, so it is a pre-existing bug rather than a new one — but this is a new file in a new module, and the fix is two lines.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/worker/CloudWatchRouterWorker.kt
```diff
@@ -0,0 +1,86 @@
...
+    ListenableWorker(appContext, parameter) {
+
+    private val workerClassName =
+        parameter.inputData.getString(WORKER_CLASS_NAME)
+            ?: throw IllegalArgumentException("Worker class name is missing")
+    private var delegateWorker: CoroutineWorker? = null
+
+    companion object {
+        internal const val WORKER_CLASS_NAME = "WORKER_CLASS_NAME"
+        internal const val WORKER_ID = "WORKER_ID"
```
mattcreaser, line 41: 🟡 **App-global mutable state without thread safety, and entries are never removed.**

Two issues in this companion:

1. `backingWorkerMap` is a plain `mutableMapOf` and `isWorkerFactoriesInitialized` a non-`@Volatile` `var`. Writes happen in `CloudWatchLogManager.init` (whatever thread constructs the client — typically main) and reads happen on a WorkManager background thread in `startWork()`. That's an unsynchronized happens-before gap; `ConcurrentHashMap` + `@Volatile` costs nothing here.
2. Nothing ever removes a registration. Every `CloudWatchLogManager` ever constructed — hence every `AmplifyCloudWatchClient`, and each one holds an SDK client, a `CoroutineScope` and a SQLCipher handle — is reachable from a static map for the process lifetime. The integration test constructs a fresh client per `@Before` and this is where they accumulate.

A `remove(workerFactoryKey)` hook paired with client teardown (see the separate lifecycle comment) would close both.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * Base names for the client's own encrypted on-device store (database + passphrase preferences), kept
+ * distinct from any other CloudWatch logging store on the device. The target log group is appended (see
+ * [storeToken]) so that clients targeting different log groups keep separate stores and never mix events.
+ */
+private const val CLIENT_DATABASE_NAME_PREFIX = "amplify.cloudwatch.client"
+private const val CLIENT_PASSPHRASE_PREFERENCES_PREFIX = "awscloudwatchclientdb"
+
+/** A filesystem-safe, stable token derived from the log group, used to name the client's on-device store. */
+private fun storeToken(logGroupName: String): String = logGroupName.hashCode().toUInt().toString(16)
```
mattcreaser: 🔵 **`hashCode()` is a collision-prone basis for a claim of store isolation.**

The KDoc just above states that appending this token means clients targeting different log groups "keep separate stores and never mix events", and the PR description makes the same guarantee. `String.hashCode()` is 32-bit and trivially collidable, so the guarantee is probabilistic rather than structural — two different log groups can land on the same database file and passphrase-prefs name.

In practice an app has a couple of log groups, so this is very unlikely to bite. But since the cost is one line, a truncated SHA-256 (or sanitising the log group name itself into a filesystem-safe form) would make the doc comment literally true.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/AmplifyCloudWatchClient.kt
```diff
@@ -0,0 +1,237 @@
...
+ *
+ */
+@ExperimentalAmplifyApi
+@OptIn(InternalAmplifyApi::class)
+class AmplifyCloudWatchClient internal constructor(
+    private val cloudWatchLogsClient: CloudWatchLogsClient,
+    initialConstraints: LoggingConstraints,
+    private val scope: CoroutineScope,
+    logManagerFactory: (client: CloudWatchLogsClient, events: MutableSharedFlow<LoggingEvent>) -> CloudWatchLogManager
+) : LogSink {
```
mattcreaser, line 105: 🔵 **No teardown API — everything a client owns outlives it.**

The public constructor creates a `CoroutineScope(SupervisorJob() + Dispatchers.IO)`, a `CloudWatchLogsClient` (which owns an HTTP engine and is `Closeable`), a `CloudWatchLogManager` with a second scope, a SQLCipher connection, and a permanent entry in `CloudWatchRouterWorker.workerFactories`. There is no `close()`, and `disable()` only stops the scheduled flush.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,325 @@
...
+                        val streamName = "$streamDate.$deviceId.${userIdentityId ?: "guest"}"
+
+                        val batch = getNextBatch(queriedEvents)
+                        // Events too large to ever be accepted are dropped so they can't block the buffer.
+                        if (batch.oversizedIds.isNotEmpty()) {
+                            cloudWatchLoggingDatabase.bulkDelete(batch.oversizedIds)
+                        }
+                        if (batch.events.isEmpty()) continue
+
+                        createLogStreamIfNotCreated(streamName, logGroupName, client)
```
mattcreaser, line 205: 🔵 **`describeLogStreams` is issued once per batch.**

`streamName` is recomputed each pass through the inner loop but only actually changes when `userIdentityId` changes (the date is captured once outside). So a flush of N batches makes N `describeLogStreams` calls that all return the same answer, doubling the request count against the CloudWatch API for large backlogs — and after the first batch the stream is guaranteed to exist.

A `MutableSet<String>` of already-ensured stream names on the manager (checked before the call) keeps it to one request per stream per process.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -0,0 +1,325 @@
...
+            cloudWatchLoggingDatabase.saveLogEvent(event)
+            if (isCacheFull()) {
+                syncLogEventsWithCloudwatch()
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            logger.error(e) { "Failed to save event" }
+            onWriteLogFailure(event.message, e)
+        }
```
mattcreaser, line 136: 🔵 **A cache-full flush failure is reported twice, once under the wrong name.**

`syncLogEventsWithCloudwatch()` already calls `onFlushLogFailure(null, exception)` and then rethrows. Called from the `isCacheFull()` branch above, that exception lands here and is re-reported as `onWriteLogFailure(event.message, e)` — so a single upload failure surfaces on `events` as both a `FlushLogFailure` and a `WriteLogFailure`, the latter attributing an upload problem to the write of an event that was in fact persisted successfully. It also puts the log message body into the write-failure `context`.

Scoping the `try` to `cloudWatchLoggingDatabase.saveLogEvent(event)` and letting the cache-full flush report itself would keep the two channels meaning what they say.

aws-cloudwatch/src/test/java/com/amplifyframework/cloudwatch/CloudWatchLogManagerTest.kt
```diff
@@ -0,0 +1,386 @@
...
+    @Test
+    fun `sync reuses an existing log stream`() = runTest {
+        coEvery { database.queryAllEvents() } returns listOf(LogEvent(1L, "hi", 1L)) andThen emptyList()
+        stubSuccessfulUpload(existingStream = true)
+
+        manager.syncLogEventsWithCloudwatch()
+
+        coVerify(exactly = 0) { cloudWatchLogsClient.createLogStream(any()) }
+    }
+
```
mattcreaser, line 256: 🔵 **Coverage gap that maps directly onto the blocker above.**

The 20 tests here do cover every batching claim in the description — the 10,000-event, 1 MB and 24h boundaries, oversized-event drop, stream create/reuse, too-new retention, per-log-group isolation, and auto-flush scheduling. Two paths are missing:

1. **`putLogEvents` fails while `isCacheFull()` is true.** This is the scenario the `lastAttemptedIds` catch block exists for, and the regression flagged in `CloudWatchLogManager.kt` would be caught by asserting `bulkDelete` is called with the ids of the batch that failed.
2. **`enable()` racing `disable()`**, per the asymmetry comment on `AmplifyCloudWatchClient`.

This test covers the failure callback and rethrow, but not the cache-full branch inside the same `catch`.

mattcreaser (approved):

PR #3407, opened 2026-08-31
feat(api): implement query and mutate over HTTP

mattcreaser:

*Issue #, if available:*

n/a

*Description of changes:*

The single-auth HTTP path. Four internal files, **none of which change the public API surface** — `apiCheck` reports no diff, because this only fills in bodies that were `TODO()` and everything new is `internal`.

Per #3377 the client reimplements the plugin internals it needs privately rather than depending on `aws-api`. What that came to:

### `AppSyncRequestDecorator`

Collapses the plugin's six-class `RequestDecorator` hierarchy into a single suspend `when`. The plugin needs that hierarchy because its call sites are synchronous — `IamRequestDecorator` reaches the smithy signer through `AWS4Signer.signBlocking`, which is `runBlocking` under the hood. Every credential source here is already a `suspend` function, so this calls `DefaultAwsSigner.sign` directly and there is nothing to collapse a hierarchy around.

All five auth modes are covered. A credential supplier is customer code and can throw anything, so failures become `AppSyncTokenFetchException` or `AppSyncSigningException` rather than escaping as whatever the lambda raised.

This is the first consumer of the Phase 2 endpoint parser's region output, which is why the parser was in Foundations rather than here.

### `AppSyncHttpTransport`

Bridges OkHttp to a coroutine with `suspendCancellableCoroutine`, so cancelling the calling coroutine cancels the in-flight call instead of orphaning it. The plugin's equivalent takes `Consumer` callbacks and an `ExecutorService`.

### `AppSyncResponseDeserializer` + `AppSyncGson`

Reimplement response parsing. Seven of the nine adapters `GsonFactory` registers already live in `aws-api-appsync`, so they are reused directly; the two that do not — `ModelListDeserializer` and `ModelPageDeserializer` — are lazy-loading and belong to Phase 7.

### `query` / `mutate` / `close`

Both public functions delegate to one private `send()`. AppSync distinguishes a query from a mutation by the operation in the document, not by transport, so there is genuinely one code path; they stay separate public functions for call-site clarity and parity with Swift and Flutter.

`CancellationException` is deliberately rethrown rather than converted to a `Result.Failure` — swallowing it would break structured concurrency. Everything else becomes a `Failure`, so a caller never needs a try/catch.

`close()` cancels in-flight calls, evicts the connection pool, and is idempotent. A closed client fails subsequent requests instead of sending them.

### One deliberate improvement over the plugin

`AppSyncGraphQLOperation` collapses every 4xx into a generic `ApiException("OkHttp client request failed.")`. But AppSync returns validation and authorization failures as a **4xx carrying a GraphQL error body**, so the actionable detail is right there in the response and the plugin discards it. This deserializes it into `AppSyncGraphQLErrorException` with the real errors, and falls back to `AppSyncValidationException` only when the body cannot be parsed.

This is `AppSyncException.from()` gaining its first real mappings — Phase 2 shipped it thin precisely because there were no throw sites yet.

**Tests go from 29 to 76**, all passing. Run with `--rerun-tasks`:

- `:aws-appsync:testDebugUnitTest` — **76 tests, 0 failures**
  - `AmplifyAppSyncClientTest` (18) — end-to-end over `MockWebServer`: success, partial success (errors *alongside* data stay a `Success`, since reporting a `Failure` would discard data the service returned), 4xx with and without a parseable error body, 5xx, unreachable endpoint, empty body, malformed body, a failing token supplier (asserting **zero** requests were sent), `close()` semantics, and region inference including a China endpoint.
  - `AppSyncRequestDecoratorTest` (14) — a header assertion per auth mode, that the API key supplier is invoked per request rather than cached, and for SigV4: the `AWS4-HMAC-SHA256` prefix, the `x-amz-content-sha256` body header AppSync requires, credential scoping to the configured region and the `appsync` service, the session token for temporary credentials, body preservation across the signing rebuild, and a China host keeping its `.amazonaws.com.cn` suffix in the signed credential. Plus the four supplier-failure paths.
  - `AppSyncResponseDeserializerTest` (15) — data, errors, both together, error locations/path/extensions, the list shapes, and pagination.
- `:aws-appsync:apiCheck` — **no diff**, as expected for an all-`internal` change.
- `:aws-appsync:ktlintCheck`, `:aws-appsync:compileReleaseKotlin`, `:aws-appsync:licensee`.
- `:aws-api:compileDebugKotlin`, `:aws-api-appsync:compileDebugKotlin`, `:aws-datastore:compileDebugKotlin` as downstream consumers.

**Two test-infrastructure notes worth a reviewer's attention**, because both were bugs I hit rather than choices I planned:

1. **The client tests need Robolectric.** The transport sets a `User-Agent`, and `UserAgent.string()` constructs a `UserAgent` whose constructor reads `android.os.Build.MANUFACTURER` — a `Stub!` throw on a plain JVM. `aws-connect` uses Robolectric for the same reason.
2. **`MockWebServer.takeRequest()` is now called with a timeout.** Before Robolectric was added, the `android.os.Build` throw meant no request was ever sent, and the bare `takeRequest()` blocked the test thread forever. Because it *blocks* rather than suspends, `runTest`'s timeout cannot interrupt it — so the build hung instead of failing. The helper now fails in 5s with a message naming the problem. This guards the failure mode, not just the instance of it.

**One test was passing for the wrong reason** and is worth flagging as a correctness note about the response shape. `GsonResponseAdapters.shouldSkipQueryLevel` strips the query-name level for any `Iterable` response type, so `data` must be an object with exactly one field. My original list fixture was `{"data":{"items":[...]}}`, which passed only because `items` was being consumed *as the query name*. The real shape is `{"data":{"listTodos":{"items":[...]}}}`. Corrected, with the constraint documented in the test file and pagination coverage added.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests — Phase 6 ports the e2e suite; there is no subscription path yet to test against a real endpoint

### DIFF
--- a/aws-appsync/build.gradle.kts
+++ b/aws-appsync/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
     implementation(libs.gson)
     implementation(libs.kotlin.coroutines)
     implementation(project(":foundation-bridge"))
+    implementation(platform(libs.aws.bom))
     implementation(libs.aws.signing)
     implementation(libs.aws.smithy.http)
 
--- a/aws-appsync/build.gradle.kts
+++ b/aws-appsync/build.gradle.kts
@@ -31,6 +31,11 @@ dependencies {
     implementation(libs.gson)
     implementation(libs.kotlin.coroutines)
     implementation(project(":foundation-bridge"))
+    implementation(libs.aws.signing)
+    implementation(libs.aws.smithy.http)
 
     testImplementation(libs.bundles.test.unit)
+    testImplementation(libs.bundles.test.unit.android)
+    testImplementation(libs.test.mockwebserver)
+    testImplementation(project(":testutils"))
 }
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -138,8 +138,12 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
         TODO("Subscription implementation will be added in a follow-up PR")
 
     /**
-     * Close the client, cancelling in-flight requests and releasing pooled connections.
+     * Close the client, cancelling enqueued requests and releasing pooled connections.
      * The client cannot be reused after closing.
+     *
+     * A request that has already passed its own closed check can still be dispatched after this
+     * returns, because closing does not lock the send path. Such a request completes normally rather
+     * than being cancelled.
      *
      * TODO: terminate active subscriptions once subscriptions are supported.
      */
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -54,11 +54,13 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
 
     private val closed = AtomicBoolean(false)
 
-    private val httpClient: OkHttpClient by lazy {
+    private val lazyHttpClient = lazy {
         OkHttpClient.Builder()
             .apply { configuration.httpClientConfigurator?.invoke(this) }
             .build()
     }
+
+    private val httpClient: OkHttpClient by lazyHttpClient
 
     private val transport: AppSyncHttpTransport by lazy {
         AppSyncHttpTransport(
@@ -103,8 +105,10 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      */
     private suspend fun <T> send(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> {
         if (closed.get()) {
+            // A lifecycle misuse, not a bad request: AppSyncRequestException would tell the caller to
+            // correct a request that was fine.
             return Result.Failure(
-                AppSyncValidationException(
+                AppSyncInvalidConfigException(
                     message = "This client has been closed and cannot be reused.",
                     recoverySuggestion = "Create a new AmplifyAppSyncClient."
                 )
@@ -115,7 +119,7 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
             Result.Success(withContext(Dispatchers.IO) { transport.execute(request) })
         } catch (cancellation: CancellationException) {
             throw cancellation
-        } catch (error: Throwable) {
+        } catch (error: Exception) {
             Result.Failure(AppSyncException.from(error))
         }
     }
@@ -134,11 +138,16 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
         TODO("Subscription implementation will be added in a follow-up PR")
 
     /**
-     * Close the client. Terminates all active subscriptions and releases resources.
+     * Close the client, cancelling in-flight requests and releasing pooled connections.
      * The client cannot be reused after closing.
+     *
+     * TODO: terminate active subscriptions once subscriptions are supported.
      */
     fun close() {
         if (!closed.compareAndSet(false, true)) return
+        // Nothing to release if no request was ever issued, and touching the client here would build
+        // one — running the caller's configurator — purely to cancel an empty dispatcher.
+        if (!lazyHttpClient.isInitialized()) return
         httpClient.dispatcher.cancelAll()
         httpClient.connectionPool.evictAll()
     }
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -19,8 +19,12 @@ import com.amplifyframework.api.graphql.GraphQLRequest
 import com.amplifyframework.api.graphql.GraphQLResponse
 import com.amplifyframework.foundation.result.Result
 import com.amplifyframework.foundation.result.getOrThrow
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 
 /**
@@ -48,6 +52,23 @@ import okhttp3.OkHttpClient
 @ExperimentalAmplifyApi
 class AmplifyAppSyncClient(val configuration: Configuration) {
 
+    private val closed = AtomicBoolean(false)
+
+    private val httpClient: OkHttpClient by lazy {
+        OkHttpClient.Builder()
+            .apply { configuration.httpClientConfigurator?.invoke(this) }
+            .build()
+    }
+
+    private val transport: AppSyncHttpTransport by lazy {
+        AppSyncHttpTransport(
+            endpoint = configuration.endpoint,
+            client = httpClient,
+            authorization = configuration.authorization,
+            decorator = AppSyncRequestDecorator(configuration.region)
+        )
+    }
+
     /**
      * Per-client connection state flow.
      * Emits [ConnectionState] changes for the shared WebSocket connection.
@@ -61,8 +82,7 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * @param request The GraphQL request. Use model helpers or construct manually.
      * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [AppSyncException].
      */
-    suspend fun <T> query(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> =
-        TODO("Query implementation will be added in a follow-up PR")
+    suspend fun <T> query(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> = send(request)
 
     /**
      * Execute a GraphQL mutation.
@@ -70,8 +90,35 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * @param request The GraphQL request. Use model helpers or construct manually.
      * @return [Result.Success] with the typed GraphQL response, or [Result.Failure] with an [AppSyncException].
      */
-    suspend fun <T> mutate(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> =
-        TODO("Mutation implementation will be added in a follow-up PR")
+    suspend fun <T> mutate(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> = send(request)
+
+    /**
+     * Queries and mutations are the same HTTP exchange — AppSync distinguishes them by the operation
+     * in the document, not by transport. They are separate public functions for call-site clarity and
+     * for parity with Swift and Flutter.
+     *
+     * Failures arrive as [Result.Failure] rather than being thrown, so a caller never needs a
+     * try/catch. Coroutine cancellation is deliberately not caught: it must propagate for structured
+     * concurrency to work.
+     */
+    private suspend fun <T> send(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> {
+        if (closed.get()) {
+            return Result.Failure(
+                AppSyncValidationException(
+                    message = "This client has been closed and cannot be reused.",
+                    recoverySuggestion = "Create a new AmplifyAppSyncClient."
+                )
+            )
+        }
+
+        return try {
+            Result.Success(withContext(Dispatchers.IO) { transport.execute(request) })
+        } catch (cancellation: CancellationException) {
+            throw cancellation
+        } catch (error: Throwable) {
+            Result.Failure(AppSyncException.from(error))
+        }
+    }
 
     /**
      * Subscribe to a GraphQL subscription. Returns a [Flow] of [SubscriptionEvent].
@@ -91,7 +138,9 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * The client cannot be reused after closing.
      */
     fun close() {
-        TODO("Close implementation will be added in a follow-up PR")
+        if (!closed.compareAndSet(false, true)) return
+        httpClient.dispatcher.cancelAll()
+        httpClient.connectionPool.evictAll()
     }
 
     // ── Configuration ───────────────────────────────────────────────────
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -25,15 +25,12 @@ import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 
 /**
- * The client's own Gson instance, deliberately private to this module rather than shared with the API
- * plugin's `GsonFactory`: this module does not depend on `aws-api`, and reimplements what it needs
- * instead.
+ * The Gson instance the client serializes requests and deserializes responses with.
  *
- * It registers the same adapters as the plugin except the two lazy-loading deserializers
- * (`ModelListDeserializer`, `ModelPageDeserializer`), which live in `aws-api` and are therefore out of
- * reach. TODO: register equivalents once lazy model loading is supported.
+ * Private to the client rather than shared, so neither the adapter set nor the null handling below can
+ * be altered from outside.
  *
- * Every adapter used here is in `aws-api-appsync`, which this module already exposes as `api`.
+ * TODO: register deserializers for lazily-loaded model lists and pages, which this set does not cover.
  */
 internal object AppSyncGson {
 
@@ -48,8 +45,8 @@ internal object AppSyncGson {
                 SerializedModelAdapter.register(it)
                 SerializedCustomTypeAdapter.register(it)
             }
-            // Matches the plugin's factories: a mutation that clears a field needs an explicit
-            // `"field": null` in the payload, since AppSync reads an absent field as "leave unchanged".
+            // A mutation that clears a field needs an explicit `"field": null` in the payload, because
+            // AppSync reads an absent field as "leave unchanged" rather than "set to null".
             .serializeNulls()
             .create()
     }
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -25,12 +25,14 @@ import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 
 /**
- * The client's own Gson instance, deliberately private to this module rather than shared with the
- * API plugin's `GsonFactory` — per #3377 the client reimplements what it needs instead of depending
- * on `aws-api`.
+ * The client's own Gson instance, deliberately private to this module rather than shared with the API
+ * plugin's `GsonFactory`: this module does not depend on `aws-api`, and reimplements what it needs
+ * instead.
  *
  * It registers the same adapters as the plugin except the two lazy-loading deserializers
- * (`ModelListDeserializer`, `ModelPageDeserializer`), which live in `aws-api` and are Phase 7 work.
+ * (`ModelListDeserializer`, `ModelPageDeserializer`), which live in `aws-api` and are therefore out of
+ * reach. TODO: register equivalents once lazy model loading is supported.
+ *
  * Every adapter used here is in `aws-api-appsync`, which this module already exposes as `api`.
  */
 internal object AppSyncGson {
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.appsync
+
+import com.amplifyframework.api.graphql.GsonResponseAdapters
+import com.amplifyframework.core.model.query.predicate.GsonPredicateAdapters
+import com.amplifyframework.core.model.temporal.GsonTemporalAdapters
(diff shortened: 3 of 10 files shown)


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncRequestDecorator.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    private fun Request.withHeader(name: String, value: String) = newBuilder().header(name, value).build()
+
+    /**
+     * Invokes a credential supplier, translating any failure into a typed exception. A supplier is
+     * customer code, so it can throw anything.
+     */
+    private suspend fun (suspend () -> String).fetch(description: String): String = try {
+        this()
+    } catch (error: Exception) {
```
mattcreaser: 🟡 `catch (error: Exception)` catches `CancellationException`.

On JVM `kotlinx.coroutines.CancellationException` is a typealias for `java.util.concurrent.CancellationException`, which extends `IllegalStateException` → `RuntimeException` → `Exception`. So if the caller's job is cancelled while a customer's suspend supplier is suspended, the cancellation is caught here and rethrown as `AppSyncTokenFetchException`.

That then defeats the guard in `AmplifyAppSyncClient.send()`: `catch (cancellation: CancellationException)` no longer matches, so it falls into `catch (error: Throwable)` and the caller gets `Result.Failure(AppSyncTokenFetchException("Fetching the User Pools token failed."))` for what was actually a cancellation. That is exactly the structured-concurrency break the KDoc on `send()` says is deliberately avoided.

How bad in practice: the outer `withContext` and the subsequent `await()` will usually still surface the cancellation, since the calling coroutine is already cancelled and fails at its next suspension point — so this is latent rather than a live hang. But the misattributed failure is real, and `AppSyncRequestDecoratorTest` calls `decorate()` directly, where nothing downstream rescues the contract.

```suggestion
    private suspend fun (suspend () -> String).fetch(description: String): String = try {
        this()
    } catch (cancellation: CancellationException) {
        throw cancellation
    } catch (error: Exception) {
```

(needs `import kotlinx.coroutines.CancellationException`)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncRequestDecorator.kt
```diff
@@ -0,0 +1,166 @@
...
+        val signed = try {
+            val config = AwsSigningConfig {
+                region = this@AppSyncRequestDecorator.region
+                service = APPSYNC_SERVICE_NAME
+                credentials = authorizer.credentialsProvider.toSmithyProvider().resolve()
+                useDoubleUriEncode = true
+                signedBodyHeader = AwsSignedBodyHeader.X_AMZ_CONTENT_SHA256
+            }
+            DefaultAwsSigner.sign(signable, config).output
+        } catch (error: Exception) {
```
mattcreaser: 🟡 Same `CancellationException` problem as `fetch()` — both `toSmithyProvider().resolve()` and `DefaultAwsSigner.sign()` are suspend calls inside this `try`, so a cancellation here becomes an `AppSyncSigningException`. Rethrow cancellation first:

```suggestion
        } catch (cancellation: CancellationException) {
            throw cancellation
        } catch (error: Exception) {
```

🔵 Separately: this `try` spans credential *resolution* (line 114) as well as signing, so a failing `AwsCredentialsProvider` — which is customer code, the same category as the token suppliers on line 77 — reports as "Signing the request with SigV4 failed." `AppSyncSigningException`'s default recovery suggestion happens to fit, and `a throwing credentials provider becomes a signing exception` asserts this on purpose, so it may well be intentional. Worth deciding explicitly though: a customer whose Cognito Identity call fails gets told signing broke. Hoisting `resolve()` above the `try` and wrapping it in `AppSyncTokenFetchException` would make the two credential-source paths consistent.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
```diff
@@ -0,0 +1,51 @@
...
+            .also {
+                GsonTemporalAdapters.register(it)
+                GsonJavaTypeAdapters.register(it)
+                GsonPredicateAdapters.register(it)
+                GsonResponseAdapters.register(it)
+                ModelWithMetadataAdapter.register(it)
+                SerializedModelAdapter.register(it)
+                SerializedCustomTypeAdapter.register(it)
+            }
+            .create()
```
mattcreaser: 🟡 `serializeNulls()` is missing.

Both `GsonFactory`s call it as the last step before `create()` — `com.amplifyframework.api.aws.GsonFactory` (aws-api) and `com.amplifyframework.util.GsonFactory` (aws-api-appsync, the one `GsonVariablesSerializer()` defaults to). The class KDoc says this registers "the same adapters as the plugin except the two lazy-loading deserializers," which is accurate about *adapters* and is why the gap is easy to miss: `serializeNulls()` isn't an adapter.

Harmless today, because `AppSyncGson.instance` is only ever read by `AppSyncResponseDeserializer` and `serializeNulls` affects serialization only. It stops being harmless the moment this instance serializes anything — a mutation that clears a field to `null` needs the explicit `"field": null` in the payload, and without this it is omitted, which AppSync reads as "leave unchanged" rather than "set to null."

Cheapest fix is to add it now while the divergence is still free:

```suggestion
    val instance: Gson by lazy {
        GsonBuilder()
            .also {
                GsonTemporalAdapters.register(it)
                GsonJavaTypeAdapters.register(it)
                GsonPredicateAdapters.register(it)
                GsonResponseAdapters.register(it)
                ModelWithMetadataAdapter.register(it)
                SerializedModelAdapter.register(it)
                SerializedCustomTypeAdapter.register(it)
            }
            .serializeNulls()
            .create()
```

If it's deliberately left out until a serializing consumer exists, a line in the KDoc saying so would stop the next reader re-deriving this.

🔵 Also: the KDoc says "Every adapter used here is in `aws-api-appsync`." `GsonResponseAdapters` and the three `datastore.appsync` adapters are; `GsonTemporalAdapters`, `GsonJavaTypeAdapters` and `GsonPredicateAdapters` are in `core`. Both are `api` dependencies so it compiles either way — the sentence just isn't quite true.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -0,0 +1,146 @@
...
+                    errors.joinToString("; ") { it.message },
+                errors = errors
+            )
+        } else {
+            AppSyncValidationException(
+                message = "The request was rejected with HTTP status ${response.code}.",
+                recoverySuggestion = "Check the request document and variables, and the API's auth configuration."
+            )
+        }
+    }
```
mattcreaser: 🟡 No 4xx maps to `AppSyncAuthException`, so authorization failures can't be caught as auth failures.

The base `AppSyncException` KDoc leads with exactly this usage:

```kotlin
is AppSyncAuthException -> reauthenticate()
```

But every 4xx leaving this function is either `AppSyncGraphQLErrorException` (an `AppSyncResponseException`) or `AppSyncValidationException` (an `AppSyncRequestException`). AppSync answers an expired token or a rejected API key with a 401 carrying `UnauthorizedException`, and `@auth`-rule denials with a 4xx whose GraphQL error has `extensions.errorType` of `Unauthorized`. Both land in the first branch. So the single most common reason to re-authenticate arrives as a *response* error, and `AppSyncTokenExpiredException` — which exists on the base and has the right recovery suggestion — has no throw site.

This is squarely in scope because unpacking the 4xx body is the deliberate improvement the PR is built around: the errors are already parsed on line 99, so the information needed to classify is in hand. Something like — on a 401, or when an error's `extensions["errorType"]` is `UnauthorizedException`/`Unauthorized`, return `AppSyncTokenExpiredException` (or a new `AppSyncUnauthorizedException` under `AppSyncAuthException`) and keep `AppSyncGraphQLErrorException` for genuine validation and resolver errors.

If classifying auth is intended as Phase 4 work alongside `@auth`-rule resolution, that's reasonable — but the test `a 401 with no parseable body becomes a validation exception` currently *pins* the 401→`AppSyncValidationException` mapping, so it will need changing later. A `// Phase 4` note here and on that test would signal it.

🔵 Minor, same function: `AppSyncValidationException`'s recovery suggestion covers both request shape *and* auth config, which reads like a hedge because the type can't express the auth case.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -0,0 +1,146 @@
...
+        if (response.code in CLIENT_ERROR_CODES) {
+            throw clientError(request, response, body)
+        }
+
+        if (!response.isSuccessful) {
+            throw AppSyncNetworkException(
+                message = "The request failed with HTTP status ${response.code}.",
+                recoverySuggestion = "This is usually transient. Retry the request."
+            )
+        }
```
mattcreaser: 🔵 Because the 4xx branch above returns first, this branch catches 5xx *and* any 1xx/3xx OkHttp didn't resolve — and tells the caller "This is usually transient. Retry the request."

That's right for 5xx and wrong for 3xx. It's reachable: `httpClientConfigurator` is public API, and `followRedirects(false)` or a cross-protocol redirect leaves the 3xx here. A caller acting on the recovery suggestion retries a request that will redirect identically forever.

Narrowing the transient claim to `response.code >= 500` and giving anything else a non-retry message would close it. Low priority — nobody points an AppSync client at a redirect on purpose.

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -60,17 +81,43 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
...
+                    recoverySuggestion = "Create a new AmplifyAppSyncClient."
+                )
+            )
+        }
+
+        return try {
+            Result.Success(withContext(Dispatchers.IO) { transport.execute(request) })
+        } catch (cancellation: CancellationException) {
+            throw cancellation
+        } catch (error: Throwable) {
```
mattcreaser: 🔵 `catch (error: Throwable)` also catches `Error` — `OutOfMemoryError`, `StackOverflowError`, `LinkageError` — and `AppSyncException.from()` turns each into an `AppSyncUnknownException`. A caller then sees a `Result.Failure` that looks like an API problem while the VM is in an unrecoverable state, and the process keeps running.

`catch (error: Exception)` covers everything `from()` can meaningfully classify. Note the ordering with the `CancellationException` branch above stays correct either way, since `CancellationException` is an `Exception`.

```suggestion
        } catch (error: Exception) {
```

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -60,17 +81,43 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
...
+     */
+    private suspend fun <T> send(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> {
+        if (closed.get()) {
+            return Result.Failure(
+                AppSyncValidationException(
+                    message = "This client has been closed and cannot be reused.",
+                    recoverySuggestion = "Create a new AmplifyAppSyncClient."
+                )
+            )
+        }
```
mattcreaser: 🔵 A closed client is a lifecycle/misuse error, but it's reported as `AppSyncValidationException`, which sits under `AppSyncRequestException` — "the request could not be built or was rejected before being sent." A caller matching `is AppSyncRequestException -> fixTheRequest()` will be told to correct a request that was fine; the client was the problem.

Nothing in the existing hierarchy fits well (`AppSyncConfigurationException` is the closest in spirit — "programming errors, surfaced at construction or on first use rather than silently"), so this may be the least-bad choice available. Two options if you'd rather not add a type: use `AppSyncInvalidConfigException`, or note here why `AppSyncValidationException` was chosen so the next reader doesn't re-litigate it.

Also worth noting the `closed.get()` check races with `close()`: a request that passes the check and is then closed underneath fails via OkHttp's cancel as `AppSyncNetworkException`, not with the "closed" message. That's fine — just not what the message on line 107 promises, and no test pins either side of the race.

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -90,7 +137,9 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * The client cannot be reused after closing.
      */
     fun close() {
-        TODO("Close implementation will be added in a follow-up PR")
+        if (!closed.compareAndSet(false, true)) return
+        httpClient.dispatcher.cancelAll()
+        httpClient.connectionPool.evictAll()
     }
```
mattcreaser: 🔵 `httpClient` is `by lazy`, so `close()` on a client that never issued a request *constructs* an `OkHttpClient` — running the customer's `httpClientConfigurator` — purely to cancel an empty dispatcher and evict an empty pool. `close is idempotent` and `a closed client fails subsequent requests` both take this path today.

Guarding on initialization keeps close cheap and side-effect-free:

```kotlin
private val lazyHttpClient = lazy { … }
private val httpClient by lazyHttpClient

fun close() {
    if (!closed.compareAndSet(false, true)) return
    if (!lazyHttpClient.isInitialized()) return
    httpClient.dispatcher.cancelAll()
    httpClient.connectionPool.evictAll()
}
```

Two related notes:

- The KDoc says "Terminates all active subscriptions and releases resources," but subscriptions are Phase 5 and nothing here touches a WebSocket. Fine as a forward-looking contract; just currently untrue.
- `dispatcher.executorService` is not shut down. OkHttp's own guidance is that this isn't required — idle threads exit after 60s — so I'd leave it, but if a future `close()` is meant to be a hard release, that's the missing piece.

aws-appsync/build.gradle.kts
```diff
@@ -31,6 +31,11 @@ dependencies {
     implementation(libs.gson)
     implementation(libs.kotlin.coroutines)
     implementation(project(":foundation-bridge"))
+    implementation(libs.aws.signing)
+    implementation(libs.aws.smithy.http)
```
mattcreaser: 🟡 Both aliases are versionless in the catalog and there is no BOM declared here.

```toml
aws-signing    = { module = "aws.smithy.kotlin:aws-signing-default" }   # no version.ref
aws-smithy-http = { module = "aws.smithy.kotlin:http-client" }          # no version.ref
```

The other two consumers in the repo pair them with an explicit platform — `aws-api` has `api(platform(libs.aws.bom))` immediately above `api(libs.aws.signing)`, and `aws-logging-cloudwatch` does the same. This module resolves instead through `implementation(project(":foundation-bridge"))`, because foundation-bridge exports the BOM as `api(project.dependencies.platform(libs.aws.bom))`.

So it builds, but the smithy versions are inherited from an unrelated bridge module's transitive platform. If foundation-bridge ever demotes that BOM to `implementation`, or this module stops needing foundation-bridge, the build breaks with a bare "no version specified for aws.smithy.kotlin:aws-signing-default" that points nowhere near the cause. One line makes the dependency on the platform explicit and matches the two existing precedents:

```suggestion
    implementation(platform(libs.aws.bom))
    implementation(libs.aws.signing)
    implementation(libs.aws.smithy.http)
```

(Unverifiable without running: whether `:aws-appsync:licensee` needs new allowlist entries. Both artifacts already ship via `aws-api`, so I'd expect the existing allowlist to cover them, and the description reports the task passing.)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Bridges an OkHttp [Call] to a coroutine. Cancelling the coroutine cancels the call, so a
+     * cancelled `query` does not leave a request in flight.
+     */
+    private suspend fun Call.await(): Response = suspendCancellableCoroutine { continuation ->
+        enqueue(
+            object : Callback {
+                override fun onResponse(call: Call, response: Response) = continuation.resume(response)
```
thisisabhash: **Response can leak on a cancel/response race**

```kotlin
override fun onResponse(call: Call, response: Response) = continuation.resume(response)
```

If the coroutine is cancelled just as the response arrives, `resume(response)` is dropped by the coroutines machinery and the `Response` body is never closed — a connection leak. `invokeOnCancellation { cancel() }` only cancels the *call*, not an already-delivered response. Use the resource-aware overload:

```kotlin
override fun onResponse(call: Call, response: Response) =
    continuation.resume(response) { _, _, _ -> response.closeQuietly() }
```

Low severity (cancellation race only), but it's the one genuine defect here.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        // AppSync reports GraphQL-level problems with a 4xx and a GraphQL error body. Deserializing it
+        // yields the errors themselves, which is far more useful than the status code alone. This is
+        // where the plugin collapses everything into a generic ApiException.
+        if (response.code in CLIENT_ERROR_CODES) {
+            throw clientError(request, response, body)
+        }
+
+        if (response.code >= SERVER_ERROR_MIN) {
```
thisisabhash, line 86: **Every status ≥ 500 is treated as transient**

`response.code >= SERVER_ERROR_MIN` maps 501/505/etc. to a retryable `AppSyncNetworkException`, so a non-transient 5xx like 501 Not Implemented advises a pointless retry. Minor — flagging in case you want to narrow what counts as retryable.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncRequestDecorator.kt
```diff
@@ -0,0 +1,175 @@
...
+            throw AppSyncSigningException(
+                message = "Signing the request with SigV4 failed.",
+                cause = error
+            )
+        }
+
+        // Rebuild the OkHttp request from the signed one. The signed Content-Type is read back out of
+        // the headers and reapplied through the body's MediaType, because OkHttp takes Content-Type
+        // from the body rather than from a header set on the builder.
+        val builder = Request.Builder().url(url)
```
thisisabhash, line 141: **Content-Type read-back is unreachable**

```kotlin
if (name.equals(CONTENT_TYPE_HEADER, ignoreCase = true)) {
    values.firstOrNull()?.let { contentType = it }
}
```

Content-Type rides on the OkHttp *body* (`toRequestBody(CONTENT_TYPE.toMediaType())`), not in `request.headers`, so it's never in the smithy headers built for signing, never in `signed.headers`, and this branch never fires — `contentType` always stays `DEFAULT_CONTENT_TYPE`. Not a correctness bug (the default is right), but the branch and its comment imply a reachable path. Consider simplifying to the default, or confirm a future auth mode will populate it.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncRequestDecorator.kt
```diff
@@ -0,0 +1,175 @@
...
+            body = HttpBody.fromBytes(bodyBytes),
+            trailingHeaders = DeferredHeaders.Empty
+        )
+
+        val signed = try {
+            val config = AwsSigningConfig {
+                region = this@AppSyncRequestDecorator.region
+                service = APPSYNC_SERVICE_NAME
+                credentials = authorizer.credentialsProvider.toSmithyProvider().resolve()
+                useDoubleUriEncode = true
```
thisisabhash, line 122: **`useDoubleUriEncode = true` deserves a comment**

Correct for AppSync (unlike S3, which uses `false`), but it's easy to get wrong and silently break signing. A one-line comment noting the AppSync/API-Gateway-style choice would help future readers.

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -61,17 +84,45 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
...
+    /**
+     * Queries and mutations are the same HTTP exchange — AppSync distinguishes them by the operation
+     * in the document, not by transport. They are separate public functions for call-site clarity and
+     * for parity with Swift and Flutter.
+     *
+     * Failures arrive as [Result.Failure] rather than being thrown, so a caller never needs a
+     * try/catch. Coroutine cancellation is deliberately not caught: it must propagate for structured
+     * concurrency to work.
+     */
+    private suspend fun <T> send(request: GraphQLRequest<T>): Result<GraphQLResponse<T>, AppSyncException> {
```
thisisabhash, line 106: **`close()` vs in-flight `send()` race**

A request that passes the `closed.get()` check at line 107 can still build the transport and send after `close()` returns; `close()`'s `cancelAll()` won't catch a call that hasn't been enqueued yet. The KDoc "cancelling in-flight requests" slightly oversells it. Acceptable for an experimental API, but consider tightening the doc wording or noting the race.

aws-appsync/src/test/java/com/amazonaws/appsync/AmplifyAppSyncClientTest.kt
```diff
@@ -0,0 +1,335 @@
...
+import io.kotest.matchers.nulls.shouldBeNull
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+import io.kotest.matchers.string.shouldNotContain
+import io.kotest.matchers.types.shouldBeInstanceOf
+import java.net.HttpURLConnection
+import java.util.concurrent.TimeUnit
+import kotlinx.coroutines.test.runTest
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
```
thisisabhash: **Uses the deprecated `MockWebServer`**

`okhttp3.mockwebserver.MockWebServer` is deprecated in favor of `mockwebserver3`. Prefer the newer artifact if the repo has standardized on it.

aws-appsync/src/test/java/com/amazonaws/appsync/AmplifyAppSyncClientTest.kt
```diff
@@ -0,0 +1,335 @@
...
+ * End-to-end tests for [AmplifyAppSyncClient.query] and [AmplifyAppSyncClient.mutate] over a
+ * [MockWebServer]. These cover the whole HTTP path — request construction, auth decoration, transport
+ * and deserialization — and are the only place the error mapping is exercised the way a caller sees
+ * it, as a `Result.Failure` carrying a specific [AppSyncException] subtype.
+ *
+ * Robolectric is required because the transport sets a User-Agent, and building it reads
+ * `android.os.Build`, which is a stub on a plain JVM.
+ */
+@RunWith(RobolectricTestRunner::class)
+class AmplifyAppSyncClientTest {
```
thisisabhash, line 52: **No coverage for cancellation mid-flight**

No test asserts response-body closure / no connection leak when the collecting coroutine is cancelled while a request is in flight, so the `onResponse` leak noted in `AppSyncHttpTransport` would go undetected. A test that cancels during a slow `MockResponse` and asserts the OkHttp dispatcher drains would cover it.

thisisabhash (approved):

PR #3410, opened 2026-08-31
chore: fix cloudwatch client review comments

thisisabhash:

*Issue #, if available:*

N/A

*Description of changes:*

Addresses review feedback on the standalone v3 CloudWatch client (`AmplifyCloudWatchClient`), covering a correctness regression, several robustness fixes, and a few smaller cleanups.

**Correctness**
- **Cache-full recovery no longer drops the failed batch (regression vs v2).** `lastAttemptedIds` was assigned *after* `putLogEvents` returned, so on failure the catch block's `bulkDelete` was a no-op — a full local store with a permanently-rejected batch would never drain. Now the batch ids are recorded *before* the call so the recovery valve drops the batch that actually failed (`CloudWatchLogManager`).
- **`isCacheFull` integer overflow.** `cacheSizeInMB * mb` overflowed `Int` (e.g. `Int.MIN_VALUE` at 2048 MB, `0` at 4096 MB), making `isCacheFull()` always true and flushing on every write. The multiplication is now done in `Long`, and `AmplifyCloudWatchClientOptions.build()` rejects a non-positive `localStoreMaxSizeInMB` (`CloudWatchDatabase`, `AmplifyCloudWatchClientOptions`).
- **`CloudWatchRouterWorker` retry future never completed.** The `CallbackToFutureAdapter` lambda returned `Result.retry()` as its debug tag but never called `completer.set(...)`, so the worker hung until timeout instead of rescheduling. It now delivers the result through the completer.

**Robustness / API**
- **`isEnabledFor` now honors the log level.** It previously ignored `level` and returned `isEnabled`, breaking the `LogSink` contract and defeating `AmplifyLogging`'s lazy-message evaluation framework-wide. It now returns `isEnabled && filter.couldLog(level)` — namespace-agnostic and conservative (never under-reporting), so per-namespace filtering in `emit` stays authoritative (`AmplifyCloudWatchClient`, `CloudWatchLoggingFilter.couldLog`).
- **`enable()`/`disable()` are now symmetric.** `enable()` deferred its state flip into `scope.launch { startSync() }` while `disable()` ran inline, so a quick `enable()` → `disable()` could leave auto-flush running. `startSync()` is now a plain (non-suspend) function called inline from both `enable()` and `init` (`AmplifyCloudWatchClient`, `CloudWatchLogManager`).
- **`flushLogs()` distinguishes "flushed" from "skipped".** When a sync was already in flight the guard returned immediately and `flushLogs()` still reported `Success`, so callers couldn't tell a real flush from a skipped one. `syncLogEventsWithCloudwatch()` now returns a `Boolean` surfaced as `FlushData.flushed` (`CloudWatchLogManager`, `FlushData`).
- **Thread-safe worker-factory registry + teardown.** `CloudWatchRouterWorker.workerFactories` is now backed by a `ConcurrentHashMap` with a `@Volatile` initialized flag (writes happen on the constructing thread, reads on a WorkManager thread), and gained a `remove` hook. An `@InternalAmplifyApi close()` on the client stops sync, drops the factory registration, cancels the scope, and closes the SDK client so a client no longer outlives its resources (`CloudWatchRouterWorker`, `CloudWatchLogManager.close`, `AmplifyCloudWatchClient.close`). `close()` is kept internal for now and can be promoted to public without a breaking change.

**Cleanups**
- **Store isolation token uses a truncated SHA-256** instead of `String.hashCode()` (32-bit, trivially collidable), making the "different log groups never share a store" guarantee structural rather than probabilistic (`AmplifyCloudWatchClient.storeToken`).
- **`describeLogStreams` is issued at most once per stream per process** instead of once per batch, halving the CloudWatch request count for large backlogs (`CloudWatchLogManager`).
- **A cache-full flush failure is no longer reported twice.** The `saveLogEvent` `try` is scoped to the DB write; the cache-full flush reports its own failure via `onFlushLogFailure` and its rethrow is swallowed, so an upload problem is no longer also surfaced as a `WriteLogFailure` for an event that was in fact persisted (`CloudWatchLogManager`).

- Added unit tests for the cache-full flush-failure recovery, the "flush failure while cache is not full keeps the batch" path, single-report behavior, the single-`describeLogStreams` optimization, `isEnabledFor` under constraints, `enable()`/`disable()` ordering, `close()` teardown, and `localStoreMaxSizeInMB` validation.
- `./gradlew :aws-cloudwatch:testDebugUnitTest :aws-cloudwatch:ktlintCheck :aws-cloudwatch:apiCheck` — all pass.
- Ran the connected integration tests on an emulator against the provisioned CloudWatch backend — 10/10 pass.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
+++ b/aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
@@ -48,6 +48,7 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 
@@ -94,6 +95,10 @@ internal class CloudWatchLogManager(
     private val isSyncInProgress = AtomicBoolean(false)
     private val logger = AmplifyLogging.logger<CloudWatchLogManager>()
 
+    // Log streams already confirmed to exist this process, so a multi-batch flush issues at most one
+    // describeLogStreams per stream instead of one per batch. Only touched while the sync guard is held.
+    private val ensuredStreams = mutableSetOf<String>()
+
     // Key the worker-factory registration and the WorkManager unique work by log group so that two clients
     // targeting different log groups don't hijack each other's delegate factory or scheduled auto-flush (both
     // the factory map and unique-work names are app-global).
@@ -109,14 +114,24 @@ internal class CloudWatchLogManager(
     suspend fun saveLogEvent(event: CloudWatchLogEvent) = withContext(coroutineDispatcher) {
         try {
             cloudWatchLoggingDatabase.saveLogEvent(event)
-            if (isCacheFull()) {
-                syncLogEventsWithCloudwatch()
-            }
         } catch (e: CancellationException) {
             throw e
         } catch (e: Exception) {
             logger.error(e) { "Failed to save event" }
             onWriteLogFailure(event.message, e)
+            return@withContext
+        }
+        // A cache-full flush reports its own failures via onFlushLogFailure and rethrows; swallow that
+        // here so an upload problem isn't also surfaced as a WriteLogFailure for an event that was in
+        // fact persisted successfully.
+        if (isCacheFull()) {
+            try {
+                syncLogEventsWithCloudwatch()
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                logger.warn(e) { "Cache-full flush failed" }
+            }
         }
     }
 
@@ -138,7 +153,7 @@ internal class CloudWatchLogManager(
         }
     }
 
-    suspend fun startSync() {
+    fun startSync() {
         stopSync = false
         enqueueSync()
     }
@@ -149,11 +164,22 @@ internal class CloudWatchLogManager(
         logger.debug { "Stopping sync" }
     }
 
-    suspend fun syncLogEventsWithCloudwatch() {
+    /** Cancels the scope and drops the worker-factory registration. */
+    fun close() {
+        stopSync()
+        CloudWatchRouterWorker.workerFactories.remove(workerFactoryKey)
+        coroutineScope.cancel()
+    }
+
+    /**
+     * Attempts to flush all buffered events to CloudWatch. Returns `true` if this call performed the
+     * flush, or `false` if it was skipped because another flush was already in progress.
+     */
+    suspend fun syncLogEventsWithCloudwatch(): Boolean {
         // Atomic guard so overlapping triggers (manual flush, cache-full write, WorkManager) can't
         // run concurrently and delete each other's rows.
         if (!isSyncInProgress.compareAndSet(false, true)) {
-            return
+            return false
         }
         withContext(coroutineDispatcher) {
             var lastAttemptedIds: List<Long> = emptyList()
@@ -176,6 +202,9 @@ internal class CloudWatchLogManager(
                         if (batch.events.isEmpty()) continue
 
                         createLogStreamIfNotCreated(streamName, logGroupName, client)
+                        // Record the ids we're about to send *before* the call, so the cache-full
+                        // recovery valve in the catch block drops this batch if putLogEvents throws.
+                        lastAttemptedIds = batch.sendableIds
                         val response = client.putLogEvents(
                             PutLogEventsRequest {
                                 logEvents = batch.events
@@ -188,7 +217,6 @@ internal class CloudWatchLogManager(
                         response.rejectedLogEventsInfo?.tooNewLogEventStartIndex?.let {
                             acceptedIds = acceptedIds.slice(IntRange(0, it - 1))
                         }
-                        lastAttemptedIds = acceptedIds
                         // Nothing accepted (all too new): re-querying would return the same rows forever.
                         if (acceptedIds.isEmpty()) return@withContext
                         cloudWatchLoggingDatabase.bulkDelete(acceptedIds)
@@ -206,6 +234,7 @@ internal class CloudWatchLogManager(
                 isSyncInProgress.set(false)
             }
         }
+        return true
     }
 
     private suspend fun createLogStreamIfNotCreated(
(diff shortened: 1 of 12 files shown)


aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLoggingFilter.kt
```diff
@@ -45,6 +45,25 @@ internal class CloudWatchLoggingFilter(initialConstraints: LoggingConstraints) {
...
 
+    /**
+     * Returns true if [level] could pass under *any* configured threshold (default, per-namespace, or
+     * per-user). Namespace-agnostic and deliberately permissive — [canLog] stays authoritative per
+     * message. Used by [AmplifyCloudWatchClient.isEnabledFor] so callers can skip materializing lazy
+     * log messages that no threshold would ever emit.
+     */
+    fun couldLog(level: LogLevel): Boolean {
+        val constraints = loggingConstraints
+        val thresholds = buildList {
```
mattcreaser: 🟡 **`couldLog` allocates and rescans every threshold on every single log statement.**

`isEnabledFor` is not a cold path. In `foundation`, `LoggerImpl` calls it for every registered sink on *both* overloads:

```kotlin
override fun log(level: LogLevel, message: String, cause: Throwable?) {
    sinks.forEach { if (it.isEnabledFor(level)) { it.emit(logMessage) } }
}
override fun log(level: LogLevel, cause: Throwable?, message: () -> String) {
    if (sinks.any { it.isEnabledFor(level) }) { log(level, message(), cause) }
}
```

So every framework log call now builds an `ArrayList` and walks `namespaceLogLevel` plus every `userLogLevel`'s default *and* namespace map — O(users × namespaces) per log statement, on the path whose entire purpose is to be cheap enough to skip building a message. With a non-trivial `userLogLevel` map this costs more than the lazy evaluation it's saving.

Since `allows` is `this <= level`, "*any* threshold allows `level`" is exactly "the *minimum* threshold allows `level`", so this collapses to an O(1) check precomputed whenever constraints change:

```suggestion
    fun couldLog(level: LogLevel): Boolean = mostPermissiveThreshold allows level
```

with, alongside the existing `loggingConstraints` property:

```kotlin
@Volatile
var loggingConstraints: LoggingConstraints = initialConstraints
    set(value) {
        field = value
        mostPermissiveThreshold = value.mostPermissive()
    }

@Volatile
private var mostPermissiveThreshold: LogLevel = initialConstraints.mostPermissive()

private fun LoggingConstraints.mostPermissive(): LogLevel = sequence {
    yield(defaultLogLevel)
    yieldAll(namespaceLogLevel.values)
    userLogLevel.values.forEach {
        yield(it.defaultLogLevel)
        yieldAll(it.namespaceLogLevel.values)
    }
}.min()
```

Behaviour is identical for every input, including `LogLevel.None` thresholds and `couldLog(None)`.

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/CloudWatchLogManager.kt
```diff
@@ -206,27 +234,35 @@ internal class CloudWatchLogManager(
...
     ) {
-        client.describeLogStreams(
-            DescribeLogStreamsRequest {
-                logGroupName = groupName
-                logStreamNamePrefix = logStream
-            }
-        ).apply {
-            if (this.logStreams == null || this.logStreams?.isEmpty() == true) {
+        // add() returns false if we've already ensured this stream this process; skip the describe call.
+        if (!ensuredStreams.add(logStream)) return
```
mattcreaser, line 253: 🟡 **`ensuredStreams` is never invalidated on a `putLogEvents` failure, so a server-side stream deletion becomes unrecoverable until process restart.**

The `catch` below correctly un-caches when `describeLogStreams`/`createLogStream` itself fails. What isn't handled is the stream disappearing *after* it was cached — log group deleted and recreated, a stream removed by a retention/lifecycle action, or an operator cleaning up. Before this change every batch re-ran `describeLogStreams`, so the client self-healed on the next flush. Now `putLogEvents` will fail with `ResourceNotFoundException` on every subsequent flush for the life of this `CloudWatchLogManager`, and because the failure is per-batch and the cache-full valve only drops the *current* batch, the buffer will churn without ever recovering.

Suggest clearing the entry from the `catch` in `syncLogEventsWithCloudwatch` when the cause indicates the stream is gone:

```kotlin
} catch (exception: Exception) {
    if (exception is ResourceNotFoundException) ensuredStreams.clear()
    onFlushLogFailure(null, exception)
    ...
}
```

(`clear()` rather than removing one name, since the group itself may have been recreated.)

aws-cloudwatch/src/main/java/com/amplifyframework/cloudwatch/worker/CloudWatchRouterWorker.kt
```diff
@@ -70,8 +77,9 @@ internal class CloudWatchRouterWorker(appContext: Context, private val parameter
...
             // this is to prevent a race condition where workManager starts work before worker factory is initialized
             if (!isWorkerFactoriesInitialized) {
-                return CallbackToFutureAdapter.getFuture {
-                    Result.retry()
+                return CallbackToFutureAdapter.getFuture { completer ->
+                    completer.set(Result.retry())
+                    "CloudWatchRouterWorker.retry"
                 }
             } else {
                 throw IllegalStateException("Failed to find delegate for $workerClassName")
```
mattcreaser: 🟡 **Adding `remove` makes this `throw` branch reachable through the normal lifecycle for the first time.**

`isWorkerFactoriesInitialized` is only ever *set* — `remove` doesn't clear it (correctly, since clearing it would reintroduce unbounded retries). Previously that was harmless because the map only ever grew, so "no factory found" implied "not yet initialized" and the retry branch was the only reachable one.

`close()` breaks that invariant. It calls `stopSync()` → `cancelUniqueWork`, which is **asynchronous**, and then drops the registration synchronously. A worker WorkManager has already dequeued therefore finds no factory *with the flag still true*, and takes this `throw` instead. The same happens if two clients share a log group (the KDoc says don't, but the factory key is derived from the log group, so `A.close()` unregisters B's routing).

I can't execute WorkManager to confirm the exact downstream handling, but throwing out of `startWork()` is strictly worse than reporting failure explicitly — it surfaces as an uncaught worker exception rather than a clean terminal state:

```suggestion
            } else {
                return CallbackToFutureAdapter.getFuture { completer ->
                    completer.set(Result.failure())
                    "CloudWatchRouterWorker.noDelegate"
                }
            }
```

If you'd rather keep the throw as a genuine programming-error signal, that's defensible — but then `close()` should await `cancelUniqueWork(...).result` before dropping the registration.

mattcreaser (approved):

PR #3411, opened 2026-08-31
fix(api): fail fast on AppSync subscription connection failure

thisisabhash:

*Issue #, if available:*

Internal

*Description of changes:*

Makes AppSync subscription setup fail fast when the underlying WebSocket has already gone away, so that pending subscriptions no longer block for the full acknowledgement timeouts during network changes (WiFi ↔ mobile data, brief connectivity loss). Previously a subscription could stall for up to 10s (waiting for a `start_ack`) or 30s (waiting for the connection ack) on a socket that was already dead, which amplified app-level reconnect loops into overlapping/duplicate subscriptions.

Both changes are in `SubscriptionEndpoint`.

**Fail pending subscriptions on connection failure**
- On a transport failure (`onFailure`, e.g. `EOFException`/broken pipe from a network change) the listener already broadcast the error, but any subscription already sitting in `awaitSubscriptionReady()` kept blocking for the full 10s `start_ack` timeout because its readiness latch was never released. A new `failAllPendingSubscriptions()` helper now releases every pending subscription (via `acknowledgeSubscriptionFailure()`) so they fail immediately instead of waiting for a `start_ack` that can never arrive. It is invoked from both the `onFailure` path and the `CONNECTION_ERROR` message path.

**Honor `WebSocket.send()` return value**
- `WebSocket.send()` returns `false` when the socket is closed/closing. The subscription `start` send previously ignored this and proceeded to a 10s `start_ack` wait; it now fast-fails the subscription with an `AppSyncSubscriptionConnectionException`.
- The `connection_init` send similarly ignored the return value, which would leave the request blocked for the full 30s connection timeout; it now marks the connection failed and releases waiters immediately.

**Testing seam (no production behavior change)**
- Added a `@VisibleForTesting` constructor that injects the `OkHttpClient`; the public constructor delegates to it via a `buildOkHttpClient` helper. This lets the unit tests supply a mock WebSocket and drive the listener callbacks directly.

Note: these are targeted fixes for the current V2 `aws-api` behavior. The durable improvement — SDK-managed reconnect with connectivity awareness, automatic re-subscription, backoff, and a single-flight guard (as amplify-swift already implements) — is tracked separately.

- Added `SubscriptionEndpointFastFailTest` (Kotlin, MockK + Kotest + Robolectric) with three tests, each of which would block for the full timeout (and fail its <5s latch assertion) without the fix:
  - pending subscription fails fast when the connection fails (`onFailure` while awaiting `start_ack`);
  - subscription fails fast when the `start` message send is rejected;
  - connection fails fast when the `connection_init` send is rejected.
- `./gradlew :aws-api:testDebugUnitTest :aws-api:checkstyle :aws-api:ktlintCheck` — all pass (new tests: 3/3, 0 skipped).

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -18,6 +18,7 @@ package com.amplifyframework.api.aws;
 import android.net.Uri;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import androidx.core.util.ObjectsCompat;
 
 import com.amplifyframework.AmplifyException;
@@ -86,6 +87,17 @@ final class SubscriptionEndpoint {
             @NonNull SubscriptionAuthorizer authorizer,
             @Nullable String apiName
     ) {
+        this(apiConfiguration, responseFactory, authorizer, apiName, buildOkHttpClient(configurator));
+    }
+
+    @VisibleForTesting
+    SubscriptionEndpoint(
+            @NonNull ApiConfiguration apiConfiguration,
+            @NonNull GraphQLResponse.Factory responseFactory,
+            @NonNull SubscriptionAuthorizer authorizer,
+            @Nullable String apiName,
+            @NonNull OkHttpClient okHttpClient
+    ) {
         this.apiConfiguration = Objects.requireNonNull(apiConfiguration);
         this.subscriptions = new ConcurrentHashMap<>();
         this.responseFactory = Objects.requireNonNull(responseFactory);
@@ -93,7 +105,10 @@ final class SubscriptionEndpoint {
         this.timeoutWatchdog = new TimeoutWatchdog();
         this.pendingSubscriptionIds = Collections.synchronizedSet(new HashSet<>());
         this.apiName = apiName;
+        this.okHttpClient = Objects.requireNonNull(okHttpClient);
+    }
 
+    private static OkHttpClient buildOkHttpClient(@Nullable OkHttpConfigurator configurator) {
         OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder()
                 .retryOnConnectionFailure(true);
 
@@ -101,7 +116,7 @@ final class SubscriptionEndpoint {
             configurator.applyConfiguration(okHttpClientBuilder);
         }
 
-        this.okHttpClient = okHttpClientBuilder.build();
+        return okHttpClientBuilder.build();
     }
 
     <T> void requestSubscription(
@@ -170,7 +185,18 @@ final class SubscriptionEndpoint {
                 .put("authorization", authorizer.createHeadersForSubscription(request, authType))))
                 .toString();
 
-            socket.send(jsonMessage);
+            if (!socket.send(jsonMessage)) {
+                // The WebSocket rejected the message because it is closed or closing (for example,
+                // the connection dropped during a network change). Fail fast instead of waiting the
+                // full acknowledgement timeout for a start_ack that will never arrive.
+                if (pendingSubscriptionIds.remove(subscriptionId)) {
+                    onSubscriptionError.accept(new AppSyncSubscriptionConnectionException(
+                        "Failed to send subscription registration message; the connection is closed or closing.",
+                        null,
+                        "Check your Internet connection. Is your device online?"));
+                }
+                return;
+            }
         } catch (JSONException | ApiException exception) {
             // If the subscriptionId was still pending, then we can call the onSubscriptionError
             if (pendingSubscriptionIds.remove(subscriptionId)) {
@@ -217,6 +243,18 @@ final class SubscriptionEndpoint {
         Subscription<?> subscription = subscriptions.get(subscriptionId);
         if (subscription != null && pendingSubscriptionIds.remove(subscriptionId)) {
             subscription.acknowledgeSubscriptionFailure();
+        }
+    }
+
+    // Release every subscription still awaiting a start_ack so they fail fast rather than blocking
+    // for the full acknowledgement timeout. Used when the underlying connection fails, since a
+    // transport-level failure means no start_ack will ever be delivered for pending subscriptions.
+    private void failAllPendingSubscriptions() {
+        for (String subscriptionId : new HashSet<>(pendingSubscriptionIds)) {
+            Subscription<?> subscription = subscriptions.get(subscriptionId);
+            if (subscription != null && pendingSubscriptionIds.remove(subscriptionId)) {
+                subscription.acknowledgeSubscriptionFailure();
+            }
         }
     }
 
@@ -564,6 +602,9 @@ final class SubscriptionEndpoint {
             connectionResponse.countDown();
             // This will broadcast the error to all subscriptions
             notifyError(failure);
+            // Release any subscriptions still awaiting a start_ack so they fail fast instead of
+            // blocking for the full acknowledgement timeout now that the connection has failed.
+            failAllPendingSubscriptions();
         }
 
         @Override
@@ -599,7 +640,15 @@ final class SubscriptionEndpoint {
                     .put("type", "connection_init")
                     .toString();
 
-                webSocket.send(jsonMessage);
+                if (!webSocket.send(jsonMessage)) {
+                    // The socket is closed or closing, so the connection can never be acknowledged.
+                    // Mark the connection failed and release waiters instead of letting them block
+                    // for the full connection acknowledgement timeout.
+                    endpointStatus.set(EndpointStatus.CONNECTION_FAILED);
+                    connectionFailureCause = new IOException(
+                        "Failed to send connection_init message; the connection is closed or closing.");
+                    connectionResponse.countDown();
+                }
             } catch (JSONException jsonException) {
                 notifyError(jsonException);
             }
@@ -645,6 +694,7 @@ final class SubscriptionEndpoint {
                         }
                         
                         connectionResponse.countDown();
+                        failAllPendingSubscriptions();
                         break;
                     case SUBSCRIPTION_ACK:
                         notifySubscriptionAcknowledged(jsonMessage.getString("id"));
--- a/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
+++ b/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
@@ -41,7 +41,6 @@ import java.io.IOException;
 import java.lang.reflect.Type;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
@@ -103,7 +102,7 @@ final class SubscriptionEndpoint {
         this.responseFactory = Objects.requireNonNull(responseFactory);
         this.authorizer = Objects.requireNonNull(authorizer);
         this.timeoutWatchdog = new TimeoutWatchdog();
-        this.pendingSubscriptionIds = Collections.synchronizedSet(new HashSet<>());
+        this.pendingSubscriptionIds = ConcurrentHashMap.newKeySet();
         this.apiName = apiName;
         this.okHttpClient = Objects.requireNonNull(okHttpClient);
     }
@@ -175,6 +174,16 @@ final class SubscriptionEndpoint {
             }
         }
 
+        // Register the subscription before sending "start" so that a connection failure arriving
+        // mid-setup (for example during IAM/Cognito auth header creation below) can fail it fast via
+        // failAllPendingSubscriptions() instead of leaving this thread blocked for the full
+        // acknowledgement timeout waiting for a start_ack that will never arrive.
+        Subscription<T> subscription = new Subscription<>(
+            onNextItem, onSubscriptionError, onSubscriptionComplete,
+            responseFactory, request.getResponseType(), request, apiName
+        );
+        subscriptions.put(subscriptionId, subscription);
+
         try {
             String jsonMessage = new JSONObject()
                 .put("id", subscriptionId)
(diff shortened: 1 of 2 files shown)


aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -645,6 +694,7 @@ private void processJsonMessage(WebSocket webSocket, String message) throws ApiE
                         }
                         
                         connectionResponse.countDown();
+                        failAllPendingSubscriptions();
```
mattcreaser, line 713: 🔴 **Critical — this call site releases the latch without dispatching any error, so the caller gets no callback at all.**

`acknowledgeSubscriptionFailure()` sets `failed = true`, and `awaitSubscriptionReady()` then returns `false` **without dispatching**, on the documented assumption at L441-443:

> `// An error was already dispatched at the time of failure, so don't dispatch a second one.`

Every other caller honors that contract:
- `onFailure` calls `notifyError(failure)` (L604) *before* `failAllPendingSubscriptions()`.
- `notifySubscriptionFailure` (SUBSCRIPTION_ERROR) is immediately followed by `notifySubscriptionData(...)`, which dispatches the error payload (L709-710).

This `CONNECTION_ERROR` branch dispatches nothing. A subscription already registered in `subscriptions` (past L222) and awaiting `start_ack` when a `connection_error` frame arrives now receives **neither `onSubscriptionStarted` nor `onSubscriptionError`, ever**. Before this PR it received `AppSyncSubscriptionTimeoutException` after 10s. That converts a slow-but-reported failure into a silent one, which is worse for the app than the behavior being fixed.

```suggestion
                        connectionResponse.countDown();
                        notifyError(connectionFailureCause);
                        failAllPendingSubscriptions();
```

`notifyError` also reaches already-established subscriptions, which matches what `onFailure` does for a connection-level failure. If that is too broad, pass the cause into `failAllPendingSubscriptions` and dispatch per pending subscription instead.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -220,6 +246,18 @@ private void notifySubscriptionFailure(final String subscriptionId) {
...
+    // for the full acknowledgement timeout. Used when the underlying connection fails, since a
+    // transport-level failure means no start_ack will ever be delivered for pending subscriptions.
+    private void failAllPendingSubscriptions() {
+        for (String subscriptionId : new HashSet<>(pendingSubscriptionIds)) {
+            Subscription<?> subscription = subscriptions.get(subscriptionId);
+            if (subscription != null && pendingSubscriptionIds.remove(subscriptionId)) {
+                subscription.acknowledgeSubscriptionFailure();
+            }
+        }
+    }
```
mattcreaser, line 266: 🔴 **Critical — `new HashSet<>(pendingSubscriptionIds)` traverses a `Collections.synchronizedSet` without holding its lock.**

`pendingSubscriptionIds` is `Collections.synchronizedSet(new HashSet<>())` (L106). The `HashSet(Collection)` constructor delegates to `addAll`, which iterates the source with a plain iterator. `SynchronizedCollection.iterator()` is explicitly **not** synchronized — its javadoc requires the caller to synchronize manually while traversing. `requestSubscription` adds to this set from arbitrary caller threads (L159), so a concurrent `add` during a transport failure — exactly the concurrency this PR targets — can throw `ConcurrentModificationException` on the OkHttp reader thread from inside `onFailure` / `onMessage`.

The pre-existing `notifyError` and `notifyAllSubscriptionsCompleted` copy `subscriptions.values()`, a `ConcurrentHashMap` view with a weakly-consistent iterator, so they are safe. This helper is the first place the pattern is applied to the synchronized set.

```suggestion
    private void failAllPendingSubscriptions() {
        final Set<String> pendingSnapshot;
        synchronized (pendingSubscriptionIds) {
            pendingSnapshot = new HashSet<>(pendingSubscriptionIds);
        }
        for (String subscriptionId : pendingSnapshot) {
            Subscription<?> subscription = subscriptions.get(subscriptionId);
            if (subscription != null && pendingSubscriptionIds.remove(subscriptionId)) {
                subscription.acknowledgeSubscriptionFailure();
            }
        }
    }
```

Alternatively change the field to `ConcurrentHashMap.newKeySet()`, which removes the hazard for every other use site too.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -564,6 +602,9 @@ public void onFailure(@NonNull WebSocket webSocket, @NonNull Throwable failure,
             connectionResponse.countDown();
             // This will broadcast the error to all subscriptions
             notifyError(failure);
+            // Release any subscriptions still awaiting a start_ack so they fail fast instead of
+            // blocking for the full acknowledgement timeout now that the connection has failed.
+            failAllPendingSubscriptions();
```
mattcreaser, line 614: 🟡 **The fix has a hole: an ID becomes "pending" long before its `Subscription` object exists, and `failAllPendingSubscriptions()` skips exactly those.**

`pendingSubscriptionIds.add(subscriptionId)` happens at L159, but `subscriptions.put(subscriptionId, subscription)` only at L222. The helper requires `subscription != null`, so for any ID in that window it does nothing — and does not even remove the ID from `pendingSubscriptionIds`.

The window is not negligible: it contains `authorizer.createHeadersForSubscription(request, authType)` (L185), which for IAM/Cognito auth performs token fetch and SigV4 signing. If `onFailure` lands there, the request thread proceeds to L222-223 and blocks the full 10s — precisely the behavior this PR sets out to remove. So the fix works for subscriptions that already sent `start`, but not for ones caught mid-setup, which is a likely shape for the WiFi↔mobile-data transition described in the PR.

The first new test masks this by sleeping 300ms to push the request thread past L222 before triggering `onFailure`.

Either fix closes it:
- register the `Subscription` in `subscriptions` *before* sending `start`, so the helper can always find it; or
- have `requestSubscription` re-check `socketListener.isDisconnectedState()` immediately before `awaitSubscriptionReady()` and fail fast there — `endpointStatus` is already `CONNECTION_FAILED` by the time `onFailure` returns.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -170,7 +185,18 @@ <T> void requestSubscription(
...
+                // the connection dropped during a network change). Fail fast instead of waiting the
+                // full acknowledgement timeout for a start_ack that will never arrive.
+                if (pendingSubscriptionIds.remove(subscriptionId)) {
+                    onSubscriptionError.accept(new AppSyncSubscriptionConnectionException(
+                        "Failed to send subscription registration message; the connection is closed or closing.",
+                        null,
+                        "Check your Internet connection. Is your device online?"));
+                }
+                return;
+            }
```
mattcreaser, line 210: 🔵 Neither new fast-fail path logs anything, unlike every other failure path in this class (`LOG.warn` at L597, L623, L666, L682). These paths fire during exactly the network-change scenario the PR targets, so a `LOG.warn` here and in `sendConnectionInit` (L643) would make field diagnosis meaningfully easier.

aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java
```diff
@@ -599,7 +640,15 @@ private void sendConnectionInit(WebSocket webSocket) {
...
-                webSocket.send(jsonMessage);
+                if (!webSocket.send(jsonMessage)) {
+                    // The socket is closed or closing, so the connection can never be acknowledged.
+                    // Mark the connection failed and release waiters instead of letting them block
+                    // for the full connection acknowledgement timeout.
+                    endpointStatus.set(EndpointStatus.CONNECTION_FAILED);
+                    connectionFailureCause = new IOException(
+                        "Failed to send connection_init message; the connection is closed or closing.");
+                    connectionResponse.countDown();
+                }
```
mattcreaser, line 662: 🔵 Two small observations on this branch.

1. Unlike `onFailure` (L600), it does not `webSocket.cancel()`. The socket is left for OkHttp to tear down, and the next `requestSubscription` overwrites the `webSocket` field (L151) because `CONNECTION_FAILED.isDisconnectedState()` is `true`. Adding `webSocket.cancel()` would be symmetric with the other failure path.

2. Setting `CONNECTION_FAILED` here races `onClosed` (L611-613), which sets `DISCONNECTED`. If `onClosed` lands first, the `CONNECTION_FAILED.equals(...)` check at L631 is `false` and `waitForConnectionReady()` returns a *successful* `Connection`, so the request thread goes on to send `start` on a dead socket. Not a correctness bug — the new `send()` check at L188 catches it — but the caller then sees the vaguer "closed or closing" message instead of the `connection_init` one. Using a dedicated failure flag rather than the shared status enum would make this deterministic.

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionEndpointFastFailTest.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun `pending subscription fails fast when the connection fails`() {
+        setup()
+        val callbacks = requestSubscriptionAsync()
+        val listener = awaitListener()
+
+        driveConnected(listener)
+        // Give the request thread a moment to send "start" and begin awaiting the start_ack.
+        Thread.sleep(300)
+        // A transport failure (e.g. EOFException on a network change) arrives while awaiting start_ack.
+        listener.onFailure(webSocket, EOFException("connection reset"), null)
```
mattcreaser: 🟡 `Thread.sleep(300)` is the only thing ordering this test, and it will be flaky on a loaded CI machine: if the request thread has not yet reached `subscriptions.put(...)` (`SubscriptionEndpoint.java` L222), `failAllPendingSubscriptions()` finds nothing, the thread blocks the full 10s, and the 5s latch assertion at L134 fails.

Signal the ordering explicitly from the mocked `send` instead:

```suggestion
        driveConnected(listener)
        // Wait until the request thread has actually sent "start" and is awaiting the start_ack.
        startSent.assertAwait(5.seconds)
        // A transport failure (e.g. EOFException on a network change) arrives while awaiting start_ack.
        listener.onFailure(webSocket, EOFException("connection reset"), null)
```

with a `startSent` latch counted down inside the `send` answer in `setup` when the message contains `"type":"start"`.

To be clear, that narrows the race but does not close it — `send` returns just *before* `subscriptions.put`. Closing it fully requires the production-side change described on `SubscriptionEndpoint.java` L605-607.

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionEndpointFastFailTest.kt
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Verifies the fast-fail behavior of [SubscriptionEndpoint] so that pending subscriptions do not
+ * block for the full acknowledgement timeouts when the underlying connection fails or a message
+ * cannot be sent. Without these behaviors, each of these tests would block for the 10s (start_ack)
+ * or 30s (connection) timeout and the [Callbacks.done] latch would not count down in time.
+ */
+@RunWith(RobolectricTestRunner::class)
+class SubscriptionEndpointFastFailTest {
+    private val executor = Executors.newCachedThreadPool()
```
mattcreaser: 🟡 Two convention issues with the threading here.

`AGENTS.md` §7 states new code MUST use Kotlin coroutines and avoid `Executors`, `CountDownLatch`, and futures. `requestSubscription` is a blocking Java API so *something* must run it off-thread, but `runTest { ... }` with `withContext(Dispatchers.IO)` / `async` would match the convention, and `kotlinx-coroutines-test` is already on the classpath via `libs.bundles.test.unit`.

Separately, and regardless of that: this `newCachedThreadPool()` is never shut down. JUnit constructs a fresh instance per test method, so three pools leak threads across this class. At minimum:

```kotlin
@After
fun tearDown() {
    executor.shutdownNow()
}
```

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionEndpointFastFailTest.kt
```diff
@@ -0,0 +1,173 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mattcreaser: 🔵 Copyright year should be 2026. The sibling files added in this cycle — `AppSyncErrorPropagationTest.kt`, `AppSyncExceptionHierarchyTest.kt`, `GraphQLResponseExceptionTest.kt` — all use 2026.

```suggestion
 * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

aws-api/src/test/java/com/amplifyframework/api/aws/SubscriptionEndpointFastFailTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+/**
+ * Verifies the fast-fail behavior of [SubscriptionEndpoint] so that pending subscriptions do not
+ * block for the full acknowledgement timeouts when the underlying connection fails or a message
+ * cannot be sent. Without these behaviors, each of these tests would block for the 10s (start_ack)
+ * or 30s (connection) timeout and the [Callbacks.done] latch would not count down in time.
+ */
```
mattcreaser, line 55: 🔵 The three tests cover `onFailure` and the two rejected `send()` calls, but not the `CONNECTION_ERROR` message path — the third new `failAllPendingSubscriptions()` call site (`SubscriptionEndpoint.java` L697). Given the silent-failure problem flagged there, a fourth test would be worth adding: `driveConnected(listener)`, then deliver `{"type":"connection_error","payload":{...}}` while a subscription is awaiting `start_ack`, and assert `errorRef` is non-null. That assertion fails against the current implementation.

mattcreaser (approved):

PR #3412, opened 2026-09-01
feat(api): resolve auth mode from @auth rules and retry on failure

mattcreaser:

*Issue #, if available:*

n/a

*Description of changes:*

Multi-auth over HTTP: `@auth` rule inspection, priority ordering, and retry-on-auth-failure.

### The ordering logic is reused, not reimplemented

`MultiAuthModeStrategy` and `MultiAuthorizationTypeIterator` already live in **`aws-api-appsync`**, which this module exposes as `api`. They already filter rules by operation, sort by auth strategy then provider precedence, and de-duplicate. So `AppSyncAuthModeResolver` is thin by design — it bridges their `AuthorizationType` output to the client's `AppSyncAuthMode` and drops modes with no configured authorizer.

### Precedence

Highest first: **per-request override** → **model `@auth` rules** in priority order → **configured default**.

Two deliberate choices in there:

- An override is the **only** candidate, not merely the first. The caller named a mode; falling back would authorize as a different identity than they asked for.
- An override naming a mode with **no configured authorizer is ignored**, not honoured. Honouring it would send a request with no credentials attached.

### Retry

Two retryable signals, matching the plugin: credentials for a mode could not be obtained, and AppSync returned its `Unauthorized` error type. Unauthorized detection reuses `AppSyncExtensions`, so the client and the plugin agree on which error types count rather than drifting.

**Single-candidate behaviour is unchanged from Phase 3.** With one mode there is nothing to fall back to, so the underlying failure surfaces directly and a 200-carrying-errors stays a `Success`. `AppSyncAuthExhaustedException` appears only when more than one mode was genuinely tried, and it carries `attemptedAuthModes`. There is a regression test pinning this.

### One new public type

`AppSyncAuthExhaustedException`, added to the `AppSyncAuthException` group. Multi-auth needs it and the plugin has an equivalent with three throw sites. It is the only public API change: +6 lines in the dump.

**Tests go from 76 to 96**, all passing, with `--rerun-tasks`:

- `AppSyncAuthModeResolverTest` (12, new) — single-auth resolution; priority ordering; dropping unconfigured modes; falling back to the default when no rule matches; a raw request having no schema; de-duplicating providers; excluding rules that don't cover the operation; both override cases; and the full `AuthorizationType` → `AppSyncAuthMode` mapping including `NONE` → null.
- `AmplifyAppSyncClientTest` (18 → 25) — end-to-end over `MockWebServer`: retry-then-succeed; **that the retry actually switches identity** (asserting the first request carries the User Pools header and the second the API key, so a passing test can't just be a resend); exhaustion reporting the attempted modes; retry after a failing token supplier with only one HTTP call made; a non-auth error *not* retrying; an override not falling back; and the single-auth regression guard.
- `apiCheck` (+6), `ktlintCheck`, `compileReleaseKotlin`, `licensee`, and `aws-api` / `aws-api-appsync` / `aws-datastore`.

The resolver tests mock `AppSyncGraphQLRequest`'s three relevant properties rather than building a real one. A real request runs selection-set generation, which needs a code-generated model class and is unrelated to auth resolution — I tried the real-request route first and it NPE'd in `SelectionSet$Builder.modelClass`.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests — Phase 6 ports the e2e suite


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -45,27 +46,95 @@ internal class AppSyncHttpTransport(
...
+            val isLastAttempt = index == authModes.lastIndex
+
+            val authorizer = authorization.authorizerFor(authMode)
+                ?: throw AppSyncProviderNotConfiguredException(
+                    message = "No authorizer is configured for auth mode $authMode."
+                )
+
+            val decorated = try {
+                decorate(request, authorizer)
+            } catch (error: AppSyncAuthException) {
```
mattcreaser: 🔴 **Cancellation is swallowed here and converted into an auth retry.**

`AppSyncRequestDecorator.fetch()` wraps *any* `Exception` from a customer credential supplier into `AppSyncTokenFetchException`. `kotlinx.coroutines.CancellationException` is a typealias for `java.util.concurrent.CancellationException`, which is a `RuntimeException` — so it is caught there, rethrown as an `AppSyncAuthException`, and now caught here as a retryable signal.

The effect is that cancelling a request while a token supplier is suspended does not cancel it. The loop advances to the next candidate and issues a **real HTTP request under a different identity**, up to once per remaining mode, then reports `AppSyncAuthExhaustedException`. `AmplifyAppSyncClient.send()` states "Coroutine cancellation is deliberately not caught: it must propagate for structured concurrency to work" — but by the time the exception reaches that `catch (cancellation: CancellationException)` it is no longer one, so the guard never fires. This also cuts against AGENTS.md §7.

The root cause is Phase 3 code, so the tidier fix is in `fetch()` rather than here:

```kotlin
private suspend fun (suspend () -> String).fetch(description: String): String = try {
    this()
    throw cancellation
} catch (error: Exception) {
    throw AppSyncTokenFetchException(
        message = "Fetching the $description failed.",
        cause = error
    )
}
```

Ordering matters: `java.util.concurrent.CancellationException extends IllegalStateException`, so the rethrow has to come first. `a failing token supplier retries with the next auth mode instead of failing` keeps passing — `error("session expired")` raises a plain `IllegalStateException`, which is not a `CancellationException`. That test is also why this is invisible today: it asserts exactly the retry behaviour that cancellation now triggers by accident.

---

Separate point about the same `catch`, since it is cheap to fix together: `AppSyncAuthException` is wider than this comment's "credentials for this mode could not be obtained". `AppSyncSigningException` is also an `AppSyncAuthException`, and `AppSyncRequestDecorator` raises it for two things — a SigV4 signing failure *and* a request body that could not be read (`RequestBody?.toByteArray()`). A body that cannot be read is not a credential problem, and if the next candidate is `API_KEY` (no signing, no body read) the retry will succeed and hide it entirely.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -45,27 +46,95 @@ internal class AppSyncHttpTransport(
...
+                if (isLastAttempt) throw exhausted(authModes, error)
+                lastAuthFailure = error
+                return@forEachIndexed
+            }
+
+            val response = client.newCall(decorated).await().use { deserialize(request, it) }
+
+            // An unauthorized response is the other retryable signal: AppSync accepted the request but
+            // rejected the identity, which a different mode may satisfy.
+            if (canFallBack && response.isUnauthorized()) {
```
mattcreaser: 🟡 **A retry here re-POSTs the document, which for `mutate()` means executing the mutation twice.**

`canFallBack && response.isUnauthorized()` fires on any 200 whose `errors` contain an `Unauthorized` type — including one where AppSync returned `data` alongside a field-level authorization error. In that case this discards `response.data` and sends the same document again under the next identity. `query()` and `mutate()` share this path, so a mutation that partially applied can apply a second time.

`GraphQLResponse` already exposes `hasData()`, so restricting the response-path retry to wholly-rejected requests is a one-line change:

```suggestion
            if (canFallBack && !response.hasData() && response.isUnauthorized()) {
```

Two caveats on this one, in fairness:

- The plugin has the same shape (`graphQLResponse.hasErrors() && hasAuthRelatedErrors(...)` with no data check), so this is inherited rather than introduced. A new client is just a good moment to narrow it.
- **Unverified:** I could not confirm by reading whether AppSync actually returns 200 + partial `data` + `Unauthorized` for a mutation with a denied field — that needs a live endpoint. If it never does, the guard is a no-op and the concern is moot. Either way the guard costs nothing.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -45,27 +46,95 @@ internal class AppSyncHttpTransport(
...
+        return decorator.decorate(httpRequest, authorizer)
+    }
+
+    /**
+     * Whether the response carries an AppSync `Unauthorized` error. Reuses [AppSyncExtensions] so the
+     * client agrees with the plugin on which error types count as unauthorized.
+     */
+    private fun GraphQLResponse<*>.isUnauthorized(): Boolean = errors.any { error ->
+        val extensions = error.extensions
+        !extensions.isNullOrEmpty() && AppSyncExtensions(extensions).isUnauthorizedErrorType
```
mattcreaser: 🔵 **Two small things about this predicate.**

**1. It does not quite "agree with the plugin".** The *error-type set* is shared via `isUnauthorizedErrorType`, which is the valuable half of the claim. But the response-level predicate differs. The plugin's `hasAuthRelatedErrors` returns on the **first** error carrying non-empty extensions and never inspects the rest:

```java
for (GraphQLResponse.Error error : response.getErrors()) {
    if (!Empty.check(error.getExtensions())) {
        AppSyncExtensions extensions = new AppSyncExtensions(error.getExtensions());
        return extensions.isUnauthorizedErrorType();   // returns, not continues
    }
}
```

`errors.any { … }` inspects all of them. So for `errors = [<non-auth error with extensions>, <Unauthorized>]` the plugin does not retry and this does. Your version looks like the better one — the plugin's early `return` reads like a bug — but it is a behavioural divergence, so the description would be more accurate calling it an intentional improvement rather than agreement.

**2. `AppSyncExtensions(extensions)` can throw on server-controlled input.** Its constructor does `(String) extensions.get("errorType")` with no type check. Gson deserialises a JSON number to `Double`, so a response containing `"extensions": {"errorType": 1}` raises `ClassCastException` right here. It does not crash the app — `AmplifyAppSyncClient.send` funnels it into `AppSyncUnknownException` — but the whole response is lost and the message is unhelpful, because one malformed error object poisoned the check. Cheap to contain:

```kotlin
private fun GraphQLResponse<*>.isUnauthorized(): Boolean = errors.any { error ->
    val extensions = error.extensions
    !extensions.isNullOrEmpty() &&
        runCatching { AppSyncExtensions(extensions).isUnauthorizedErrorType }.getOrDefault(false)
}
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -45,27 +46,95 @@ internal class AppSyncHttpTransport(
...
+        // With one candidate there is nothing to fall back to, so auth failures are reported as they
+        // were before multi-auth existed: the real error, or a response carrying its own errors.
+        val canFallBack = authModes.size > 1
+        var lastAuthFailure: AppSyncException? = null
+
+        authModes.forEachIndexed { index, authMode ->
+            val isLastAttempt = index == authModes.lastIndex
+
+            val authorizer = authorization.authorizerFor(authMode)
+                ?: throw AppSyncProviderNotConfiguredException(
```
mattcreaser: 🔵 **This throw is unreachable, and if it ever became reachable it would do the wrong thing.**

Every mode that reaches this loop is already known to have an authorizer:

- `AppSyncAuthorization.Multi.init` does `require(authorizers.any { it.authMode == defaultAuthMode })`, so the resolver's `listOf(default)` fallback is always satisfiable.
- `Single` resolves the default to `authorizer.authMode`, which matches by construction.
- Both the override path and the `@auth`-rule path in `AppSyncAuthModeResolver` filter on `authorizerFor(it) != null`.

So there is no input that gets here with a null authorizer. Worth noticing because it also sits **outside** the `try`: if it somehow did fire on a non-final candidate it would abandon the remaining modes rather than moving to the next one, which is the opposite of what the surrounding loop exists to do.

Since the PR argues (correctly) against declaring exception types with no reachable throw site, this is the same shape one level down. Having `resolve()` return the mode *and* its authorizer would delete the impossible branch rather than documenting it:

```kotlin
fun resolve(request: GraphQLRequest<*>): List<Pair<AppSyncAuthMode, AppSyncClientAuthorizer>>
```

The resolver already calls `authorizerFor` on every path, so it has the value in hand.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthModeResolver.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * The modes the model's `@auth` rules allow, in priority order. Empty when the request carries no
+     * schema — a raw request has no rules to inspect.
+     */
+    private fun authRuleModes(request: GraphQLRequest<*>): List<AppSyncAuthMode> {
+        val appSyncRequest = request as? AppSyncGraphQLRequest<*> ?: return emptyList()
+        val schema = appSyncRequest.modelSchema ?: return emptyList()
+        val operation = appSyncRequest.authRuleOperation ?: return emptyList()
```
mattcreaser: 🟡 **These two `?:` guards are inert, and the failure that can actually happen is unguarded.**

`getModelSchema()` and `getAuthRuleOperation()` are Java with no `@Nullable`, so Kotlin sees `ModelSchema!` / `ModelOperation!`. The `?:` compiles, but neither value is ever null in practice:

- `AppSyncGraphQLRequest.Builder.build()` rejects a request where both `modelSchema` and `modelClass` are null, so anything actually built has a schema.
- `getAuthRuleOperation()` is **computed, not a field**. It maps QUERY/SUBSCRIPTION → `READ` and MUTATION → `ModelOperation.valueOf(mutationType.name())`, and its `default` branch throws:

```java
default:
    throw new IllegalArgumentException("Invalid graphql operation type:" + ...);
```

It never returns null — it throws, or NPEs if `operation` was never set. That propagates out of `resolve()` and `execute()` to `AppSyncException.from`, which has no `IllegalArgumentException` branch, so it arrives as `AppSyncUnknownException` — outside the typed hierarchy this module otherwise keeps. Phase 3 never touched this accessor; Phase 4 reads it on every request, so the path is new.

Two related things:

- The KDoc on `authRuleModes` says "Empty when the request carries no schema — a raw request has no rules to inspect", but a raw request already returned one line above at the `as?` cast. The schema branch is a *different*, unreachable case. `a raw request has no schema, so multi auth falls back to the default` is therefore testing the cast, not the schema check, despite its name.
- Both `modelRequest()` helpers stub `authRuleOperation`, so nothing in the suite calls the real accessor.

Suggest dropping the two inert guards, and if unmapped operation types are a genuine concern, converting that throw into a typed `AppSyncValidationException` at this call site rather than letting it become `AppSyncUnknownException`.

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncAuthModeResolverTest.kt
```diff
@@ -0,0 +1,268 @@
...
+
+    @Test
+    fun `a raw request has no schema, so multi auth falls back to the default`() {
+        val resolver = AppSyncAuthModeResolver(multi())
+
+        resolver.resolve(rawRequest()) shouldContainExactly listOf(AppSyncAuthMode.API_KEY)
+    }
+
+    @Test
+    fun `duplicate providers across rules yield one mode each`() {
```
mattcreaser: 🟡 **This test does not exercise the `.distinct()` it is named for — it would pass with `.distinct()` deleted.**

`MultiAuthorizationTypeIterator` de-duplicates by loading the rules into a `TreeSet` ordered by `(authStrategy.priority, authProvider.priority)`. Two `OWNER`/`USER_POOLS` rules compare equal, so the `TreeSet` discards one before the iterator ever runs and `USER_POOLS` is yielded once. The resolver's `.distinct()` is a no-op for this input, so the test pins the iterator's behaviour rather than the resolver's.

The case `.distinct()` actually carries is **different strategies sharing one provider**, which that comparator treats as distinct (per `AuthStrategy`):

| strategy | priority | default provider |
|---|---|---|
| `OWNER` | 2 | `USER_POOLS` |
| `GROUPS` | 3 | `USER_POOLS` |
| `PRIVATE` | 4 | `USER_POOLS` |

So `@auth(rules: [{allow: owner}, {allow: private}])` — an entirely ordinary schema — makes the iterator yield `USER_POOLS` twice. Without `.distinct()` the transport would issue two HTTP calls with identical credentials and report `attemptedAuthModes = [USER_POOLS, USER_POOLS]`. The existing tests all miss it because `privateRule()` overrides the provider to `IAM`.

`.distinct()` is the right call and is load-bearing — it just needs a test that fails without it:

```kotlin
@Test
fun `rules with different strategies but the same provider yield one mode`() {
    // OWNER(2) and PRIVATE(4) both default to userPools, so the rule-level dedup in
    // MultiAuthorizationTypeIterator keeps both and the iterator yields USER_POOLS twice.
    val resolver = AppSyncAuthModeResolver(multi())

    val privateUserPools = AuthRule.builder()
        .authStrategy(AuthStrategy.PRIVATE)
        .authProvider(AuthStrategy.PRIVATE.defaultAuthProvider)
        .operations(ALL_OPERATIONS)
        .build()

    resolver.resolve(modelRequest(ownerRule(), privateUserPools)) shouldContainExactly
        listOf(AppSyncAuthMode.USER_POOLS)
}
```

Renaming the existing test to something like `duplicate rules yield one mode` would also make it clearer that it covers the iterator's dedup, not the resolver's.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthModeResolver.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        val iterator = MultiAuthModeStrategy.getInstance().authTypesFor(schema, operation)
```
mattcreaser: 🔵 **Field-level `@auth` rules widen the candidate set, and nothing covers that.**

`authTypesFor` collects applicable rules from the model *and* from every field:

```java
filterAuthRules.accept(modelSchema.getAuthRules());
for (ModelField field : modelSchema.getFields().values()) {
    filterAuthRules.accept(field.getAuthRules());
}
```

Both `modelRequest()` helpers build `ModelSchema.builder().name("Todo").authRules(...)` with no fields, so every test here drives the model-level branch only. A field-level rule can introduce a mode the model-level rules do not allow, which changes both the candidate list and the retry order — the two things this class owns. One test with a `ModelField` carrying its own rule would close it.

harsh62 (approved):

PR #3419, opened 2026-09-02
feat(api): add GraphQL subscriptions over a shared WebSocket

mattcreaser:

*Issue #, if available:*

n/a

*Description of changes:*

Implements `subscribe()`, the connection-state flow and `close()`. Every remaining `TODO()` in the module is gone, so the client surface is complete.

**Four commits, reviewable in order** — each is self-contained:

| Commit | What |
|---|---|
| `9947129` | Subscriptions over a shared WebSocket |
| `dcc2ac0` | Retry a subscription with the next eligible auth mode |
| `ff132cc` | Send the owner claim for owner-restricted subscriptions |
| `d030ff4` | Report the API's concurrent subscription limit |

The only public API change across all four is `AppSyncLimitExceededException` (+5 lines in the dump). Everything else is `internal`, and `subscribe()` / `events` / `close()` already existed as signatures.

---

### 1. Subscriptions over a shared WebSocket

The plugin's `SubscriptionEndpoint` is 720 lines of Java built around locks, latches and a callback registry. This exposes a single `messages` flow instead, which is what lets a subscription be a **filter keyed by id** rather than a registry entry.

Two details are load-bearing rather than stylistic:

- **`connect()` starts collecting before opening the socket.** The flow has no replay, so open-then-listen loses a fast `connection_ack` and waits forever. A test delivers the ack *synchronously* from inside `newWebSocket` — harsher than a real server — to pin exactly that. `subscribe()` sends `start` from `onStart` for the same reason.
- **`onFailure` also closes out**, because OkHttp does not call `onClosed` after a failure. Without it a dead socket never reports closure and every subscriber hangs.

`AppSyncWebSocketProvider` opens on first use and shares one connection, since AppSync multiplexes every subscription over one socket. A mutex plus a shared `Deferred` means concurrent first-subscribers join a single attempt; a naive check-then-open has a window where each sees "no connection" and each opens one. A failed attempt is not cached and a closed connection is replaced, which is what lets a client recover from an idle timeout without the caller managing connection state.

**Terminality is asymmetric and easy to implement backwards, so each row has a test:**

| | Behaviour |
|---|---|
| Not terminal | GraphQL errors carried *inside* a data message — delivered on the response, since the service considered the message deliverable |
| Not terminal | A data message that fails to deserialize — dropped, stream continues. Losing one message beats ending a working subscription |
| Terminal | An `error` frame for this id; an `error` with **no** id (connection-level); `connection_error`; a socket close **with** a cause |
| Normal completion | A `complete` frame; a socket close with **no** cause — which is what makes `close()` graceful rather than an error for every subscriber |

Also in this commit: `realtimeUrl()` on the endpoint parser, deliberately separate from `parse()`. `parse()` needs a region in the host so it cannot handle a custom domain at all — but a custom-domain API is configured with an explicit region and must still be able to subscribe. A standard host swaps the `appsync-api` label; a custom one appends `/realtime` to the path.

### 2. Multi-auth retry

Brings subscriptions in line with the HTTP path. Registration is retried under a fresh id rather than the whole subscription: the connection is already established, so only `start` needs replaying.

Two triggers, matching the plugin: an `error` frame carrying an unauthorized error type, **and** a `data` message carrying one — AppSync can reject an identity in a data message rather than an error frame, and that response was previously handed straight to the consumer.

A mode whose credentials cannot be obtained is skipped rather than failing the subscription. Non-auth errors stay terminal even when modes remain, since retrying a validation failure would fail identically and hide the reason.

**Retry only engages for a request carrying a model schema.** A raw request has no `@auth` rules to inspect, so it resolves to the default mode alone and there is nothing to fall back to — same as the HTTP path.

This also fixed a real defect: with a *single* candidate mode the code reported `AppSyncProviderNotConfiguredException` — "no auth mode was available" — for what was actually the service rejecting the one identity you have. It now reports what the service said.

### 3. Owner claim injection

A model whose `@auth` rules restrict reads to an owner produces a subscription AppSync expects to carry that owner as a variable. The value comes from a claim in the caller's token, so only the client can supply it, and without it the service rejects the subscription. Subscriptions only — AppSync resolves the owner server-side for queries and mutations.

The rule logic is worth reviewing because getting it wrong is silent **in both directions**: injecting unnecessarily narrows what a subscriber receives, while failing to inject makes the subscription fail. Each branch is tested — a public-read rule excuses the owner but only when the subscription actually uses the API key; group membership takes precedence, because a user who can already read via a group would receive less with an owner filter; more than one owner read rule is refused rather than guessed at, since AppSync generates a variable per rule.

This gives `AppSyncTokenParsingException` and `AppSyncAuthorizationClaimException` their first throw sites, and they are deliberately distinct: an unreadable token is a parsing failure, a readable token missing the required claim is a claim failure. Conflating them points a caller at the wrong problem.

Claims are read **without verifying the signature** — the token is read for its content rather than trusted, since the claim populates a variable AppSync then authorizes for itself. Decoding uses okio rather than `android.util.Base64`, so this needs no Android runtime to exercise.

### 4. Concurrent subscription limit

Recognises `MaxSubscriptionsReachedError` as `AppSyncLimitExceededException`, reintroducing the type that #3412 removed for having no throw site. It has one now, and its recovery suggestion says to close existing subscriptions, because neither re-subscribing nor different credentials will help.

Checked **before** the auth retry — the limit belongs to the API, not the identity, so another auth mode would consume attempts and fail identically. Terminal, with a test for a frame carrying both a limit error and an unauthorized error.

**This commit also fixes error classification more broadly.** Errors on the wire carry their type in two places depending on the frame: a subscription `error` frame has a GraphQL response shape, nesting it under `extensions`, while connection-level frames put it directly on the error object. Only `extensions` was being read, so a top-level `errorType` was silently dropped — which affected the unauthorized detection from commit 2, not just this. Both locations are now read, with a test for each.

`LimitExceededError` is a sibling meaning a *request-rate* limit rather than a subscription count. It needs different recovery advice, so it is deliberately not folded in.

**199 unit tests, all passing.** Also green with `--rerun-tasks`: `apiCheck`, `ktlintCheck`, `compileReleaseKotlin`, `licensee`.

| Class | Tests |
|---|---|
| `AppSyncSubscriberTest` | 29 |
| `AmplifyAppSyncClientTest` | 25 |
| `AppSyncEndpointParserTest` | 21 |
| `AppSyncWebSocketMessageTest` | 20 |
| `AppSyncWebSocketTest` | 19 |
| `AppSyncExceptionTest` | 16 |
| `AppSyncResponseDeserializerTest` | 15 |
| `AppSyncClaimInjectorTest` | 14 |
| `AppSyncRequestDecoratorTest` | 14 |
| `AppSyncWebSocketProviderTest` | 14 |
| `AppSyncAuthModeResolverTest` | 12 |

**On test approach, since it is unusual:** WebSocket behaviour is tested by driving the `WebSocketListener` callbacks directly against a mocked OkHttp socket, not through `MockWebServer`. That matches the rest of the repo — `EventsWebSocketTest` mocks `OkHttpClient` and `WebSocket` outright, and `LivenessWebSocketTest`'s server-side listener is *entirely empty* because it invokes the client listener itself. I tried `MockWebServer`'s `withWebSocketUpgrade` first and server-to-client frames never arrive: the upgrade succeeds with a 101, then the connection resets.

### One known gap

**No instrumented coverage of the real socket**, for the reason above — it needs a device or emulator test. Everything else in the protocol and flow layers is covered by unit tests.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests — needed for real-socket coverage; see the known gap above

### DIFF
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -21,9 +21,14 @@ import com.amplifyframework.foundation.result.Result
 import com.amplifyframework.foundation.result.getOrThrow
 import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 
@@ -54,6 +59,10 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
 
     private val closed = AtomicBoolean(false)
 
+    // close() is not suspend but the WebSocket teardown is, so it needs somewhere to run. Deliberately
+    // not cancelled: cancelling it would abort the very teardown it was created to perform.
+    private val teardownScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+
     private val lazyHttpClient = lazy {
         OkHttpClient.Builder()
             .apply { configuration.httpClientConfigurator?.invoke(this) }
@@ -71,12 +80,38 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
         )
     }
 
+    private val webSocketClient: OkHttpClient by lazy {
+        OkHttpClient.Builder()
+            .apply { configuration.webSocketClientConfigurator?.invoke(this) }
+            .build()
+    }
+
+    private val subscriber: AppSyncSubscriber by lazy {
+        val decorator = AppSyncRequestDecorator(configuration.region)
+        val realtimeUrl = AppSyncEndpointParser.realtimeUrl(configuration.endpoint).getOrThrow()
+
+        AppSyncSubscriber(
+            provider = AppSyncWebSocketProvider {
+                AppSyncWebSocket(
+                    realtimeUrl = realtimeUrl,
+                    httpEndpoint = configuration.endpoint,
+                    authorizer = configuration.authorization.defaultAuthorizer,
+                    decorator = decorator,
+                    client = webSocketClient
+                )
+            },
+            authorization = configuration.authorization,
+            decorator = decorator,
+            httpEndpoint = configuration.endpoint
+        )
+    }
+
     /**
      * Per-client connection state flow.
      * Emits [ConnectionState] changes for the shared WebSocket connection.
      */
     val events: SharedFlow<ConnectionState>
-        get() = TODO("Connection state will be implemented with subscriptions")
+        get() = subscriber.events
 
     /**
      * Execute a GraphQL query.
@@ -134,26 +169,40 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * @param request The GraphQL subscription request. Use model helpers or construct manually.
      * @return A cold [Flow] of [SubscriptionEvent].
      */
-    fun <T> subscribe(request: GraphQLRequest<T>): Flow<SubscriptionEvent<T>> =
-        TODO("Subscription implementation will be added in a follow-up PR")
+    fun <T> subscribe(request: GraphQLRequest<T>): Flow<SubscriptionEvent<T>> = flow {
+        // Checked at collection rather than call time, so a closed client fails the collector rather
+        // than throwing from a function that only builds a cold flow.
+        if (closed.get()) {
+            throw AppSyncValidationException(
+                message = "This client has been closed and cannot be reused.",
+                recoverySuggestion = "Create a new AmplifyAppSyncClient."
+            )
+        }
+        emitAll(subscriber.subscribe(request))
+    }
 
     /**
-     * Close the client, cancelling enqueued requests and releasing pooled connections.
-     * The client cannot be reused after closing.
+     * Close the client, cancelling enqueued requests, terminating active subscriptions and releasing
+     * pooled connections. The client cannot be reused after closing.
+     *
+     * Subscription teardown is asynchronous: this returns before the WebSocket has finished closing.
      *
      * A request that has already passed its own closed check can still be dispatched after this
      * returns, because closing does not lock the send path. Such a request completes normally rather
      * than being cancelled.
-     *
-     * TODO: terminate active subscriptions once subscriptions are supported.
      */
     fun close() {
         if (!closed.compareAndSet(false, true)) return
         // Nothing to release if no request was ever issued, and touching the client here would build
         // one — running the caller's configurator — purely to cancel an empty dispatcher.
-        if (!lazyHttpClient.isInitialized()) return
-        httpClient.dispatcher.cancelAll()
-        httpClient.connectionPool.evictAll()
+        if (lazyHttpClient.isInitialized()) {
+            httpClient.dispatcher.cancelAll()
+            httpClient.connectionPool.evictAll()
+        }
+        // The WebSocket teardown is suspending — it waits for closure to be observed so subscription
+        // flows complete normally rather than being cut off. close() is not suspend, so it is launched
+        // on a scope that deliberately outlives this call.
+        teardownScope.launch { subscriber.close() }
     }
 
     // ── Configuration ───────────────────────────────────────────────────
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
@@ -86,7 +86,7 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
             .build()
     }
 
-    private val subscriber: AppSyncSubscriber by lazy {
+    private val lazySubscriber = lazy {
         val decorator = AppSyncRequestDecorator(configuration.region)
         val realtimeUrl = AppSyncEndpointParser.realtimeUrl(configuration.endpoint).getOrThrow()
 
@@ -105,6 +105,8 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
             httpEndpoint = configuration.endpoint
         )
     }
+
+    private val subscriber: AppSyncSubscriber by lazySubscriber
 
     /**
      * Per-client connection state flow.
@@ -199,10 +201,16 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
             httpClient.dispatcher.cancelAll()
             httpClient.connectionPool.evictAll()
         }
-        // The WebSocket teardown is suspending — it waits for closure to be observed so subscription
-        // flows complete normally rather than being cut off. close() is not suspend, so it is launched
-        // on a scope that deliberately outlives this call.
-        teardownScope.launch { subscriber.close() }
+        // Same reasoning as the HTTP client: nothing to tear down if nothing ever subscribed, and
+        // touching the subscriber here would build it — resolving a realtime URL and standing up two
+        // scopes — only to cancel them. Resolving that URL can also fail, and it would fail on the
+        // teardown scope below, off the caller's thread, where nothing is watching.
+        if (lazySubscriber.isInitialized()) {
+            // The WebSocket teardown is suspending — it waits for closure to be observed so subscription
+            // flows complete normally rather than being cut off. close() is not suspend, so it is launched
+            // on a scope that deliberately outlives this call.
+            teardownScope.launch { subscriber.close() }
+        }
     }
 
     // ── Configuration ───────────────────────────────────────────────────
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
@@ -55,6 +55,9 @@ internal object AppSyncEndpointParser {
 
     private const val API_LABEL = "appsync-api"
     private const val REALTIME_LABEL = "appsync-realtime-api"
+    private const val WSS_SCHEME = "wss://"
+    private const val DEFAULT_PATH = "/graphql"
+    private const val REALTIME_PATH_SUFFIX = "/realtime"
     private val REGION_REGEX = """^[a-z]{2,}(-[a-z]+)+-\d+$""".toRegex()
 
     /**
@@ -113,6 +116,38 @@ internal object AppSyncEndpointParser {
         )
     }
 
+    /**
+     * Derives the WebSocket URL that subscriptions connect to.
+     *
+     * Unlike [parse] this does **not** require a region, so it works for a custom domain — which
+     * carries no region in its host and therefore cannot be parsed at all. That matters because a
+     * custom-domain API is configured with an explicit region and must still be able to subscribe.
+     *
+     * The two domain shapes derive differently, which is why this cannot be one string substitution:
+     * - **Standard** (`{apiId}.appsync-api.{region}.{dnsSuffix}`) swaps the `appsync-api` label for
+     *   `appsync-realtime-api` and keeps the path.
+     * - **Custom** (anything else) keeps the host and appends `/realtime` to the path.
+     *
+     * @return the `wss://` URL, or a failure if no host could be read.
+     */
+    fun realtimeUrl(endpoint: String): Result<String, AppSyncEndpointResolutionException> {
+        val host = hostOf(endpoint)
+            ?: return failure("Could not read a host from the endpoint URL '$endpoint'.")
+        val path = pathOf(endpoint)
+
+        val labels = host.split('.')
+        val apiLabelIndex = labels.indexOf(API_LABEL)
+
+        // apiLabelIndex > 0 rather than >= 0: a bare `appsync-api.…` host with no API id is not a
+        // standard endpoint, and is treated as custom rather than silently mangled.
+        return if (apiLabelIndex > 0) {
+            val realtimeLabels = labels.toMutableList().also { it[apiLabelIndex] = REALTIME_LABEL }
+            Result.Success("$WSS_SCHEME${realtimeLabels.joinToString(".")}$path")
+        } else {
+            Result.Success("$WSS_SCHEME$host$path$REALTIME_PATH_SUFFIX")
+        }
+    }
+
     private fun hostOf(endpoint: String): String? {
         val withoutScheme = endpoint.substringAfter("://", missingDelimiterValue = endpoint)
         val host = withoutScheme
@@ -122,6 +157,17 @@ internal object AppSyncEndpointParser {
             .substringBefore(':') // strip any port
             .lowercase()
         return host.ifEmpty { null }
+    }
+
+    /**
+     * The path portion of the endpoint, defaulting to `/graphql`. Query and fragment are dropped —
+     * neither is meaningful on a WebSocket handshake URL.
+     */
+    private fun pathOf(endpoint: String): String {
+        val withoutScheme = endpoint.substringAfter("://", missingDelimiterValue = endpoint)
+        val authorityAndPath = withoutScheme.substringBefore('?').substringBefore('#')
+        val path = authorityAndPath.substringAfter('/', missingDelimiterValue = "")
+        return if (path.isEmpty()) DEFAULT_PATH else "/${path.trimEnd('/')}"
     }
 
     // Region labels are of the form {partition}-{area}-{number}, e.g. us-east-1, cn-north-1,
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
@@ -148,7 +148,12 @@ internal object AppSyncEndpointParser {
         }
     }
 
-    private fun hostOf(endpoint: String): String? {
+    /**
+     * The host portion of [endpoint], lowercased, or null if there is none.
+     *
+     * String surgery rather than a URL type, so it runs on a plain JVM as well as on a device.
+     */
+    fun hostOf(endpoint: String): String? {
         val withoutScheme = endpoint.substringAfter("://", missingDelimiterValue = endpoint)
         val host = withoutScheme
             .substringBefore('/')
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.appsync
+
+import com.amplifyframework.api.graphql.GraphQLRequest
+import java.util.UUID
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.flow.transformWhile
+import kotlinx.coroutines.launch
+
+/**
+ * Turns a GraphQL subscription request into a [Flow] of [SubscriptionEvent], multiplexed over the
+ * client's shared WebSocket.
+ *
+ * A private reimplementation of the plugin's `SubscriptionOperation` and the registry inside
+ * `SubscriptionEndpoint`. Because the socket exposes one flow of messages, a subscription is a
+ * *filter* over it keyed by subscription id rather than an entry in a callback registry.
+ *
+ * @param provider Supplies the shared connection, opening it on first subscribe.
+ * @param authorization The client's authorizer configuration.
+ * @param decorator Produces the per-subscription authorization headers.
+ * @param httpEndpoint The HTTP endpoint. SigV4 signs the `start` message against it.
+ * @param authModeResolver Chooses the auth mode, honouring `@auth` rules and per-request overrides.
+ */
+internal class AppSyncSubscriber(
+    private val provider: AppSyncWebSocketProvider,
+    private val authorization: AppSyncAuthorization,
+    private val decorator: AppSyncRequestDecorator,
+    private val httpEndpoint: String,
+    private val authModeResolver: AppSyncAuthModeResolver = AppSyncAuthModeResolver(authorization),
+    ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    private val connectionEvents = MutableSharedFlow<ConnectionState>(replay = 1)
+    private val scope = CoroutineScope(ioDispatcher + SupervisorJob())
+
+    /** Connection state for the shared socket. Replays the latest so a late observer is not blind. */
+    val events: SharedFlow<ConnectionState> = connectionEvents.asSharedFlow()
+
+    /**
+     * A cold flow for one subscription. Nothing happens until it is collected; cancelling the
+     * collector sends `stop` and releases the subscription.
+     *
+     * Lifecycle and terminality:
+     * - `Connecting` → `Connected` → `Data`*, then normal completion or a terminal throw.
+     * - Registration failures and connection failures are **terminal**.
+     * - GraphQL errors carried *inside* a data message are **not** terminal — they are delivered on
+     *   the response, because the service considered the message deliverable.
+     * - A data message that fails to deserialize is **not** terminal either; it is dropped and the
+     *   stream continues. Losing one message is better than tearing down a working subscription.
+     * - `complete` from the service, and a clean socket close, both end the flow normally.
+     */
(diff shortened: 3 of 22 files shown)


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,263 @@
...
+
+    /** Set before the socket is torn down so the emitted [AppSyncWebSocketMessage.Closed] can explain why. */
+    @Volatile
+    private var pendingCloseCause: AppSyncException? = null
+
+    /**
+     * Opens the socket and completes the AppSync handshake.
+     *
+     * @throws AppSyncConnectionException if the connection is refused or closes during the handshake.
+     * @throws AppSyncTimeoutException if no acknowledgement arrives.
```
mattcreaser: 🔴 **`connect()` can hang forever, and this `@throws` cannot happen.**

Nothing in `connect()` produces an `AppSyncTimeoutException`. The only site that constructs one is the watchdog, and the watchdog is armed *after* the ack (`resetWatchdog(result.connectionTimeoutMs)`, line 110) — `watchdogTimeoutMs` is 0 until then, so `restartWatchdog()` returns early on line 228 for every frame that arrives before the ack.

That leaves no bound on `handshake.await()`. OkHttp does not supply one either: it sets the socket read timeout to infinite once a WebSocket is upgraded, and relies on `pingInterval` for liveness — which the `webSocketClient` in `AmplifyAppSyncClient` does not set (`OkHttpClient.Builder().apply { configuration.webSocketClientConfigurator?.invoke(this) }.build()`). So a server that completes the 101 upgrade and then goes silent suspends `connect()` indefinitely. Worse, `AppSyncWebSocketProvider` caches the in-flight `Deferred`, so every later subscriber joins the same hung attempt and the client never recovers.

The plugin bounds exactly this: `SubscriptionEndpoint.java:66` — `CONNECTION_ACKNOWLEDGEMENT_TIMEOUT = 30 /* seconds */`, applied at line 581 via `connectionResponse.await(...)`.

The same gap exists one level up for registration: `AppSyncSubscriber` sends `start` and waits for `StartAck` with no bound, so a subscription can sit on `SubscriptionEvent.Connecting` forever. The plugin bounds that too — `ACKNOWLEDGEMENT_TIMEOUT = 10 /* seconds */` (line 352, awaited at 397).

A `withTimeout` around the handshake await, throwing `AppSyncTimeoutException`, would both close the hole and make the KDoc true. If the intent is instead to inherit the timeout from `webSocketClientConfigurator`, the KDoc should say so, since the default configuration has none.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocketProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+        inProgress?.takeUnless { it.isCompleted }?.let {
+            return@coroutineScope it.await().getOrElse { error -> throw error }
+        }
+
+        val attempt = mutex.withLock {
+            // Re-check under the lock: another coroutine may have finished between the checks above
+            // and acquiring it. Without this, both would start an attempt.
+            existing?.let { return@coroutineScope it }
+            inProgress?.takeUnless { it.isCompleted }
+                ?: async { attemptConnection() }.also { inProgress = it }
```
mattcreaser: 🔴 **The shared attempt is a child of whichever caller created it, so cancelling that caller cancels every joiner.**

`async` here is scoped to the `coroutineScope { }` on line 56, i.e. to the *calling* coroutine's job. The `Deferred` is then published to `inProgress` and awaited by other callers on line 63 — but it is still owned by caller A.

If caller A is cancelled while the connect is in flight (a `viewModelScope` closing, a `take(1)`, a screen going away — all ordinary for a subscription), A's `coroutineScope` cancels its children, the shared `Deferred` is cancelled, and callers B..H get a `CancellationException` instead of the connection they were promised. That inverts the class's stated contract: "a caller arriving while a connection is still being established must wait for that attempt rather than starting a second one."

It also compounds with `AppSyncSubscriber.connectedSocket()`, which catches `Throwable` — so each joiner additionally publishes a spurious `ConnectionState.Disconnected(AppSyncUnknownException)` to every `events` observer.

The 8-way concurrency test is good but cannot catch this: `awaitAll` never cancels anyone. A test where the first of several callers is cancelled mid-connect would fail today.

Fix is to give the provider (or the client) its own scope and launch the attempt there, so the attempt's lifetime is the provider's rather than the first arriving caller's.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+
+        is AppSyncWebSocketMessage.Complete -> message.id != id
+
+        is AppSyncWebSocketMessage.Error -> {
+            // A null id is a connection-level error, which is terminal for every subscription on it.
+            if (message.id == id || message.id == null) {
+                throw AppSyncGraphQLErrorException(
+                    message = "The subscription failed: " +
+                        message.errors.joinToString("; ").ifEmpty { "no reason given" },
+                    errors = emptyList()
```
mattcreaser: 🟡 **`errors = emptyList()` discards the errors that were just parsed.**

`message.errors` is a `List<String>` and is being read on the line above to build the message, then thrown away. `AppSyncGraphQLErrorException`'s own default recovery suggestion is "Inspect the errors list for the failures the service reported" — which will never be actionable for a subscription, since the list is always empty.

```suggestion
                    errors = message.errors.map { GraphQLResponse.Error(it, null, null, null) }
```

Separately, worth a second look at the category: an `error` frame terminating a subscription is surfaced as an `AppSyncResponseException`, but `AppSyncSubscriptionException` is the branch documented as "A subscription could not be established or was terminated. All subscription errors are terminal" — so a consumer catching that category to handle subscription death will miss this one, while `ConnectionError` two cases below does land there. If reusing `AppSyncGraphQLErrorException` is deliberate (its KDoc does say "this is for the case where errors are terminal"), that is fine, but the asymmetry between the two adjacent throws is not obvious from the code.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,263 @@
...
+    // callback thread, which cannot suspend to wait for a slow collector.
+    private val messageFlow = MutableSharedFlow<AppSyncWebSocketMessage.Inbound>(
+        extraBufferCapacity = Int.MAX_VALUE
+    )
+
+    /** Every inbound message, including the synthetic [AppSyncWebSocketMessage.Closed]. */
+    val messages = messageFlow.asSharedFlow()
+
+    private val scope = CoroutineScope(ioDispatcher + SupervisorJob())
+    private var socket: WebSocket? = null
```
mattcreaser: 🟡 **`socket` is the one mutable field here that is not `@Volatile`.**

`isClosed` (line 80), `pendingCloseCause` (line 85) and `watchdogTimeoutMs` (line 242) are all `@Volatile`; `socket` is written from the `connect()` coroutine (line 107) and read from `send()` (line 136, called from arbitrary threads via `onStart`), from the watchdog coroutine (line 236), from `disconnect()` (line 152) and from `teardown()` (line 252). Given the rest of the class is explicit about cross-thread visibility, this looks like an omission rather than a decision.

```suggestion
    @Volatile
    private var socket: WebSocket? = null
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+        val socket = connectedSocket()
+        val id = UUID.randomUUID().toString()
+        val start = startMessage(id, request)
+
+        emitAll(
+            socket.messages
+                // onStart runs once this collector is subscribed. Sending `start` any earlier would
+                // race the reply: the socket's flow has no replay, so a fast start_ack would be lost
+                // and the subscription would appear to hang.
+                .onStart { socket.send(start) }
```
mattcreaser: 🟡 **A subscriber that joins after `Closed` has been emitted waits forever, and `send`'s return value is ignored here.**

This is the same `replay = 0` hazard `connect()` defends against, one level up. `connectedSocket()` checks liveness through `provider.existing` (which filters on `isClosed`), but the socket can die in the window between that check and `emitAll` subscribing on line 84. When it does:

- `Closed` was already `tryEmit`ted with no subscriber present, so it is gone — `messages` has no replay.
- `socket.send(start)` returns `false` (OkHttp rejects sends on a closed socket), and that `false` is discarded here.
- The flow then sits on `SubscriptionEvent.Connecting` indefinitely: no further frames will arrive on a dead socket, and the one message that would have terminated it has passed.

`send()` documents its `false` return precisely so a caller can react to this. Cheapest guard is to fail the flow when the send did not land or the socket has since closed:

```suggestion
                .onStart {
                    if (!socket.send(start) || socket.isClosed) {
                        throw AppSyncConnectionException(
                            message = "The subscription connection closed before the subscription could be registered."
                        )
                    }
                }
```

(`isClosed` is still checked after the send because the send can succeed into a socket that closes immediately after.)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+     * message, so emitting from the per-subscription path would report the same disconnect N times.
+     */
+    private fun watchForDisconnect(socket: AppSyncWebSocket) {
+        scope.launch {
+            // Already gone — the Closed message has been and passed, so there is nothing left to await.
+            if (socket.isClosed) {
+                connectionEvents.emit(ConnectionState.Disconnected(null))
+                return@launch
+            }
+            val closed = socket.messages.first { it is AppSyncWebSocketMessage.Closed }
```
mattcreaser: 🟡 **Same `replay = 0` race, and here the consequence is a disconnect that is never reported.**

The `isClosed` check on line 193 narrows the window but does not close it: if the socket closes between that check and `first { }` subscribing, the `Closed` message is emitted to nobody and this coroutine waits forever, so `events` stays on `Connected` for a connection that is gone. Since `events` replays, a late observer then learns the wrong state — which is the opposite of what the replay is there for.

Giving `messages` `replay = 1` would not be right (subscribers would re-see stale frames), but having `AppSyncWebSocket` expose the terminal state directly — e.g. a `CompletableDeferred<AppSyncException?>` completed inside `handleClosed()` — would make this await race-free, and would also let the `onStart` guard above and `disconnect()`'s `awaitClosed` stop depending on subscribe-before-emit ordering.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    @Volatile
+    private var watchdogTimeoutMs: Long = 0
+
+    private fun handleClosed() {
+        if (isClosed) return
+        isClosed = true
+        watchdog?.cancel()
+        messageFlow.tryEmit(AppSyncWebSocketMessage.Closed(pendingCloseCause))
```
mattcreaser: 🟡 **`scope` is not cancelled on the paths that actually happen most often.**

`handleClosed()` cancels the watchdog job but leaves `scope` alive. `scope` is only cancelled in `teardown()` (line 254) and at the end of `disconnect()` (line 157) — and `disconnect()` returns early on line 145 when `isClosed` is already true, which is exactly the state after a natural `onFailure`/`onClosed`. So for a socket that dies on its own (idle timeout, network drop — the common case, and the one the class says the consumer recovers from by re-subscribing), the `CoroutineScope` is never cancelled.

No job is left running, so this is not a live leak, but AGENTS.md §7 is explicit: "Every `CoroutineScope` MUST have a defined lifetime and be canceled when done." Cancelling in `handleClosed()` (which every closure path already funnels through) would make the scope's lifetime exactly the socket's.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Returns the shared socket, emitting connection state around the attempt.
+     */
+    private suspend fun connectedSocket(): AppSyncWebSocket {
+        provider.existing?.let { return it }
+
+        connectionEvents.emit(ConnectionState.Connecting)
+        val socket = try {
+            provider.connection()
+        } catch (error: Throwable) {
```
mattcreaser: 🟡 **`catch (error: Throwable)` catches `CancellationException`, so ordinary cancellation is published as a disconnect.**

When a collector is cancelled while the connection is being established, `provider.connection()` raises `CancellationException`; this catch wraps it via `AppSyncException.from` (which has no `CancellationException` case, so it becomes `AppSyncUnknownException`) and emits `ConnectionState.Disconnected` to every `events` observer before rethrowing. Nothing was disconnected — the caller went away.

`AmplifyAppSyncClient.send()` handles this explicitly and documents why: "Coroutine cancellation is deliberately not caught: it must propagate for structured concurrency to work." Worth the same treatment here.

```suggestion
        } catch (cancellation: CancellationException) {
            throw cancellation
        } catch (error: Throwable) {
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+        is AppSyncWebSocketMessage.StartAck -> {
+            if (message.id == id) emit(SubscriptionEvent.Connected)
+            true
+        }
+
+        is AppSyncWebSocketMessage.Data -> {
+            if (message.id == id) {
+                // Deserialization failure is deliberately swallowed: it is non-terminal, so one
+                // unreadable message must not end a healthy subscription.
+                runCatching { AppSyncResponseDeserializer.deserialize(request, message.payload) }
```
mattcreaser: 🟡 **The dropped message leaves no trace, and there is no logging anywhere in this module.**

Dropping rather than terminating is the right call and I agree with the reasoning. But `runCatching { }.getOrNull()` discards the `AppSyncDeserializationException` entirely, so a subscription that silently delivers nothing — a response-type mismatch, the most likely cause — is undiagnosable from the outside: no event, no error, no log.

This is the same gap that makes `AppSyncWebSocketMessage.Unknown`'s KDoc claim untrue today: "Kept rather than dropped so an unexpected wire change is visible in logs instead of looking like silence." I grepped all five of `AppSyncWebSocket.kt`, `AppSyncSubscriber.kt`, `AppSyncWebSocketProvider.kt`, `AppSyncHttpTransport.kt` and `AmplifyAppSyncClient.kt` for any logger reference — zero hits. Nothing in the module logs anything, so `Unknown` is currently dropped just as silently as it would be if it were not modelled.

Adding a `Logger` (foundation has one, with the lazy-supplier form `logger.debug { }` the conventions call for) would make both this drop and `Unknown` frames visible, and would let the KDoc's stated rationale hold. If logging is intentionally deferred, the `Unknown` KDoc should stop citing logs as the reason it exists.

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -141,6 +185,10 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
         if (!closed.compareAndSet(false, true)) return
         httpClient.dispatcher.cancelAll()
         httpClient.connectionPool.evictAll()
+        // The WebSocket teardown is suspending — it waits for closure to be observed so subscription
+        // flows complete normally rather than being cut off. close() is not suspend, so it is launched
+        // on a scope that deliberately outlives this call.
+        teardownScope.launch { subscriber.close() }
```
mattcreaser: 🟡 **The client-level behaviour added here has no test.**

`AmplifyAppSyncClientTest` covers `a closed client fails subsequent requests instead of sending them` and `close is idempotent`, but both go through `send()` — the query/mutate path. Nothing covers:

- `subscribe()` on a closed client failing the collector (the new check on line 172, which is a deliberately different mechanism from `send()`'s `Result.Failure` and worth pinning as such).
- `close()` reaching `subscriber.close()` at all, i.e. that subscriptions actually terminate. This launch is the one piece of the teardown chain that nothing exercises.
- `webSocketClientConfigurator`, which this PR makes live for the first time. There is a `the http client configurator is applied` test; the WebSocket one has no counterpart, so a wiring mistake here (e.g. passing `httpClientConfigurator` by accident) would not be caught.

The 94→175 growth is all in the four new/extended classes, which is a fair reflection of where the substance is — but the client is where the new code is publicly reachable.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -0,0 +1,210 @@
...
+            authorizationHeaders = decorator.authorizationHeaders(authorizer, httpEndpoint, request.content)
+        )
+    }
+
+    /**
+     * Decides what one inbound message means for this subscription.
+     *
+     * @return true to keep the flow open, false to complete it normally. Throws to terminate it.
+     */
+    private suspend fun <T> kotlinx.coroutines.flow.FlowCollector<SubscriptionEvent<T>>.handle(
```
mattcreaser: 🔵 **Fully-qualified receiver type where everything else in the file is imported.**

Every other coroutines type here is imported (`Flow`, `MutableSharedFlow`, `SharedFlow`, `emitAll`, `onStart`, `onCompletion`, `transformWhile`, …), so this one stands out.

```suggestion
    private suspend fun <T> FlowCollector<SubscriptionEvent<T>>.handle(
```

with `import kotlinx.coroutines.flow.FlowCollector` added to the block.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,263 @@
...
+                message = "No traffic on the subscription connection for ${timeout}ms."
+            )
+            // cancel(), not close(): a timed-out socket is not expected to complete a closing handshake.
+            socket?.cancel()
+            handleClosed()
+        }
+    }
+
+    @Volatile
+    private var watchdogTimeoutMs: Long = 0
```
mattcreaser: 🔵 **`watchdogTimeoutMs` is declared below the functions that read it.**

It is safe — both readers run after construction — but it is the only field declared down in the private-function section, away from `socket`, `watchdog`, `isClosed` and `pendingCloseCause` at lines 75-85. Moving it up with them makes the initialization order self-evident rather than something a reader has to reason about.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocketMessage.kt
```diff
@@ -0,0 +1,188 @@
...
+                addProperty("id", id)
+                addProperty("type", TYPE_START)
+                add("payload", payload)
+            }.toString()
+        }
+    }
+
+    /** Unregisters one subscription. Sent on flow cancellation. */
+    data class Stop(val id: String) : Outbound {
+        override fun toJson() = """{"id":"$id","type":"$TYPE_STOP"}"""
```
mattcreaser: 🔵 **`Stop` interpolates into raw JSON while `Start` builds a `JsonObject`.**

Safe today — `id` is always a `UUID.randomUUID().toString()` from `AppSyncSubscriber` — but the two sibling messages serialize by different mechanisms, and this one would produce invalid JSON for any id containing a quote or backslash. `ConnectionInit` has the same shape but no interpolated value, so it is fine as-is.

Either build this one with `JsonObject` too, or note on the `id` parameter that it is trusted to be JSON-safe.

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncSubscriberTest.kt
```diff
@@ -0,0 +1,395 @@
...
+        every { this@mockk.messages } returns this@AppSyncSubscriberTest.messages
+        every { isClosed } returns false
+        every { send(capture(sentSlot)) } returns true
+    }
+
+    private fun subscriber(
+        authorization: AppSyncAuthorization = AppSyncAuthorization.Single(
+            AppSyncClientAuthorizer.ApiKey("da2-fakekey")
+        ),
+        ioDispatcher: CoroutineDispatcher = StandardTestDispatcher()
```
mattcreaser: 🔵 **`StandardTestDispatcher()` with no scheduler is unlinked from `runTest`'s virtual time.**

This creates its own `TestCoroutineScheduler`, so nothing ever advances it and `AppSyncSubscriber.scope` — hence `watchForDisconnect` — never runs in most of these tests. That appears to be the point: the disconnect test on line 355 opts into `UnconfinedTestDispatcher(testScheduler)` precisely to make the watcher run.

If so it is worth saying, because the usual reading of an unlinked `StandardTestDispatcher()` is a mistake rather than a choice — and a reader who "fixes" it to `StandardTestDispatcher(testScheduler)` would start seeing `events` emissions in tests that currently assert `expectNoEvents()`.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncEndpointParser.kt
```diff
@@ -113,6 +116,38 @@ internal object AppSyncEndpointParser {
...
+        val host = hostOf(endpoint)
+            ?: return failure("Could not read a host from the endpoint URL '$endpoint'.")
+        val path = pathOf(endpoint)
+
+        val labels = host.split('.')
+        val apiLabelIndex = labels.indexOf(API_LABEL)
+
+        // apiLabelIndex > 0 rather than >= 0: a bare `appsync-api.…` host with no API id is not a
+        // standard endpoint, and is treated as custom rather than silently mangled.
+        return if (apiLabelIndex > 0) {
```
mattcreaser: 🔵 **This classifies standard-vs-custom differently from the plugin, which is probably an improvement but is worth recording.**

The plugin uses `DomainType.from()`, whose `STANDARD_ENDPOINT_REGEX` requires an exactly-26-character id, a region label, `amazonaws.com` or `amazonaws.com.cn`, and a path of exactly `/graphql`. Anything else is CUSTOM. This uses "has an `appsync-api` label at index > 0", which is looser in both directions:

- `https://<id>.appsync-api.us-east-1.amazonaws.com/graphql/` — CUSTOM to the plugin (so it appends `/realtime` to the standard host), standard here (label swap). Different URLs for the same input.
- A custom domain that happens to contain an `appsync-api` label, e.g. `https://foo.appsync-api.example.com/graphql`, is treated as standard here.

I think the looser rule is the better one — the plugin's 26-character literal is brittle and its trailing-slash behaviour looks accidental rather than intended. But the description says the asymmetry "is how the plugin does it too", and on classification that is only approximately true. The `a trailing slash does not produce a doubled separator` test covers the custom shape; the standard-host-with-trailing-slash case is the one where the two implementations actually disagree.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,263 @@
...
+    fun send(message: AppSyncWebSocketMessage.Outbound): Boolean = socket?.send(message.toJson()) ?: false
+
+    /**
+     * Closes the socket and waits for closure to be observed, so a caller can rely on the connection
+     * being gone once this returns. Idempotent.
+     *
+     * @param cause Why it is being closed, or null for a clean shutdown.
+     */
+    suspend fun disconnect(cause: AppSyncException? = null): Unit = withContext(ioDispatcher) {
+        if (isClosed) return@withContext
```
mattcreaser: 🔵 **"Idempotent" holds for sequential calls, not concurrent ones.**

Two concurrent `disconnect()` calls can both pass this check before either sets `isClosed`. The first closes and `handleClosed()` emits `Closed`; the second has by then subscribed via `awaitClosed`, but `handleClosed()` is guarded by `if (isClosed) return`, so no second `Closed` is ever emitted and the second caller's `closed.await()` never returns — inside `withContext(ioDispatcher)`, so it hangs rather than failing.

Latent today: the only caller is `AppSyncWebSocketProvider.close()`, which nulls `connected` under the mutex first, so a second concurrent call passes `null` and never reaches here. Worth either a mutex around the body or narrowing the KDoc to "Idempotent when called sequentially", so the guarantee does not get relied on later.

aws-appsync/src/main/java/com/amazonaws/appsync/AmplifyAppSyncClient.kt
```diff
@@ -134,26 +169,40 @@ class AmplifyAppSyncClient(val configuration: Configuration) {
      * @param request The GraphQL subscription request. Use model helpers or construct manually.
      * @return A cold [Flow] of [SubscriptionEvent].
      */
-    fun <T> subscribe(request: GraphQLRequest<T>): Flow<SubscriptionEvent<T>> =
-        TODO("Subscription implementation will be added in a follow-up PR")
+    fun <T> subscribe(request: GraphQLRequest<T>): Flow<SubscriptionEvent<T>> = flow {
+        // Checked at collection rather than call time, so a closed client fails the collector rather
+        // than throwing from a function that only builds a cold flow.
+        if (closed.get()) {
```
mattcreaser: Should consolidate this logic - note the exception here is different from in send.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocketProvider.kt
```diff
@@ -0,0 +1,117 @@
...
+            connected.also {
+                connected = null
+                inProgress = null
+            }
+        }
+        socket?.disconnect(cause)
+        scope.cancel()
+    }
+
+    private suspend fun attemptConnection(): Result<AppSyncWebSocket> = try {
```
mattcreaser: use resultCatching to avoid swallowing CancellationException

aws-appsync/src/test/java/com/amazonaws/appsync/AmplifyAppSyncClientTest.kt
```diff
@@ -342,6 +347,91 @@ class AmplifyAppSyncClientTest {
...
+    // ── Subscriptions ───────────────────────────────────────────────────
+
+    @Test
+    fun `subscribe on a closed client fails the collector`() = runTest {
+        // Deliberately a different mechanism from query/mutate: subscribe returns a cold flow, so the
+        // check belongs at collection rather than at the call that merely builds it.
+        val client = client()
+        client.close()
+
+        val error = runCatching { client.subscribe(request()).first() }.exceptionOrNull()
```
mattcreaser: Use shouldThrow instead

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -0,0 +1,308 @@
...
+    private val authorizer: AppSyncClientAuthorizer,
+    private val decorator: AppSyncRequestDecorator,
+    private val client: OkHttpClient,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) : WebSocketListener() {
+
+    // extraBufferCapacity keeps tryEmit from ever dropping a frame: emission happens on OkHttp's
+    // callback thread, which cannot suspend to wait for a slow collector.
+    private val messageFlow = MutableSharedFlow<AppSyncWebSocketMessage.Inbound>(
+        extraBufferCapacity = Int.MAX_VALUE
```
mattcreaser: Reconcile this with the original plugin implementation's behaviour. Although MutableSharedFlow uses lazy allocation it still seems like a bad idea to have an effectively unbounded buffer here.

harsh62 (approved):

Issue #2793, opened 2024-04-25
Build release get: Timed out waiting for connection acknowledgement


sontnfabbi:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Kotlin, Kotlin - Coroutines

### Amplify Categories

GraphQL API

### Gradle script dependencies

<details>

    def amplify_version = "2.14.11"
    implementation "com.amplifyframework:core:$amplify_version"
    implementation "com.amplifyframework:aws-api:$amplify_version"
    implementation "com.amplifyframework:core-kotlin:$amplify_version"

</details>

### Environment information

<details>

Build time:   2023-06-30 18:02:30 UTC                                  
Revision:     5f4a070a62a31a17438ac998c2b849f4f6892877                 
                                                                       
Kotlin:       1.8.20                                                   
Groovy:       3.0.17                                                   
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          17.0.6 (JetBrains s.r.o. 17.0.6+0-b2043.56-10027231)     
OS:           Windows 10 10.0 amd64          

</details>

### Describe the bug

When build release with minifyEnabled true, Amplify.API.subscribe get timeout eror
Still that project build debug with minifyEnabled true, has no problem..

ApiException{message=Timed out waiting for connection acknowledgement., cause=null, recoverySuggestion=Sorry, we don't have a suggested fix for this error yet.}
proguard-rules.pro:

# Amazon
-keep class com.amazon.** { *; }
-keep class com.amazonaws.** { *; }
-keep class com.amplifyframework.** { *; }
# Class names are needed in reflection
-keepnames class com.amazonaws.**
-keepnames class com.amazon.**
# Request handlers defined in request.handlers
-keep class com.amazonaws.services.**.*Handler
# The following are referenced but aren't required to run
-dontwarn com.fasterxml.jackson.**
-dontwarn org.apache.commons.logging.**
# Android 6.0 release removes support for the Apache HTTP client
-dontwarn org.apache.http.**
# The SDK has several references of Apache HTTP client
-dontwarn com.amazonaws.http.**
-dontwarn com.amazonaws.metrics.**
-dontwarn com.amazonaws.mobile.auth.facebook.FacebookSignInProvider
[...]

PR #3423, opened 2026-09-03
fix(api): keep org.json under R8 to prevent subscription connection acknowledgement timeout

VarshithaPamisetty:
## Description

On R8 release builds, GraphQL subscriptions fail with "Timed out waiting for connection acknowledgement". Amplify reads the AppSync `connection_ack` using `org.json`. When another dependency bundles its own `org.json` (Google IMA via mux-stats-sdk-media3), R8 renames those classes, so the ack is never read and the connection times out. Debug builds work because R8 is off. This ships a consumer ProGuard rule in `aws-api` that keeps `org.json`, so R8 leaves it alone. Same approach as `aws-auth-cognito`.

## Testing

- Reproduced e2e on an emulator: built an app with Amplify subscriptions + mux/IMA against a live AppSync backend. The minified release build hit the 30s timeout; the debug build worked.
- With the fix, the release build connected in ~2s and received subscription events.
- Confirmed the rule ships in `aws-api-release.aar` and reaches R8 in the app; `apiCheck`/`ktlintCheck` clean.
- No unit test: R8 bugs can't be reproduced in a JVM unit test, so the e2e repro is the proof.

Refs #2793

### DIFF
--- a/aws-api/build.gradle.kts
+++ b/aws-api/build.gradle.kts
@@ -22,6 +22,9 @@ apply(from = rootProject.file("configuration/checkstyle.gradle"))
 
 android {
     namespace = "com.amplifyframework.api.aws"
+    defaultConfig {
+        consumerProguardFiles += file("consumer-rules.pro")
+    }
 }
 
 dependencies {
--- a/aws-api/consumer-rules.pro
+++ b/aws-api/consumer-rules.pro
@@ -1,0 +1,2 @@
+# Keep org.json; R8 can obfuscate a dependency-bundled copy in release builds and break connection_ack parsing.
+-keep class org.json.** { *; }


aws-api/consumer-rules.pro
```diff
@@ -0,0 +1,2 @@
+# Keep org.json; R8 can obfuscate a dependency-bundled copy in release builds and break connection_ack parsing.
```
mattcreaser: 🔵 Suggestion — traceability. Worth naming the issue in the file so the next person to touch these rules knows whether the workaround is still needed.

```suggestion
# Keep org.json; R8 can obfuscate a dependency-bundled copy in release builds and break connection_ack parsing.
# See https://github.com/aws-amplify/amplify-android/issues/2793
```

aws-api/consumer-rules.pro
```diff
@@ -0,0 +1,2 @@
...
+-keep class org.json.** { *; }
```
mattcreaser: 🟡 Placement — consider the shared rules file instead of `aws-api`.

`build-logic/plugins/src/main/kotlin/AndroidLibraryConventionPlugin.kt` already does this in `defaultConfig` for every Android library module:

```kotlin
consumerProguardFiles += rootProject.file("configuration/consumer-rules.pro")
```

So adding the keep rule to `configuration/consumer-rules.pro` would ship it from every published module and collapse this PR to a single added line — no new file, no `defaultConfig` block in `aws-api`.

That matters because `org.json` is not specific to `aws-api`. Modules that parse JSON with `org.json` include `core` (`AmplifyConfiguration`, `CategoryConfiguration`, `Resources` — i.e. Gen1 config parsing itself), `aws-core` (`DocumentExtensions`), `aws-auth-cognito`, `aws-datastore`, `aws-storage-s3`, `aws-pinpoint-core`, `aws-geo-location`, `aws-predictions`, `aws-push-notifications-pinpoint`, and `aws-logging-cloudwatch`. An app on, say, Auth + Storage with the same `org.json`-bundling dependency has the same exposure and would not get this rule, because it has no `aws-api` dependency.

Counterpoint, and the reason this is 🟡 and not 🔴: scoping the rule to the module with the confirmed repro is a defensible, lower-blast-radius choice for a targeted fix. If you keep it here, that is fine — just a deliberate decision rather than an accident.

mattcreaser, line 2: 🔵 Follow-up (separate PR) — the keep rule mitigates the symptom; the fragile line is still in our source.

Not in this diff, but it is the likely first failure point. `SubscriptionEndpoint.java` L684-L686:

```java
Integer.parseInt(
    jsonMessage.getJSONObject("payload").getString("connectionTimeoutMs")
)
```

and the ack fixture `aws-api/src/test/resources/websocket-connection_ack.json` has `"connectionTimeoutMs":300000` — a JSON **number**. Android's platform `org.json` coerces a number to a String in `getString`; the reference JSON-java implementation (the flavour such dependencies bundle) throws `JSONException("...is not a string")` instead. That explains why the failure lands specifically on the ack rather than on config parsing or on data frames, which matches the reported symptom precisely.

`getInt("connectionTimeoutMs")` coerces in **both** implementations, so switching to it makes the parse independent of which `org.json` wins the class-loader race. It would also cover the case a keep rule cannot reach: a dependency that *relocates* its bundled copy, where obfuscation is not what breaks us.

To be explicit about the limits of this review: the above is reasoning from source, not a reproduction — I did not execute anything.

aws-api/build.gradle.kts
```diff
@@ -22,6 +22,9 @@ apply(from = rootProject.file("configuration/checkstyle.gradle"))
 
 android {
     namespace = "com.amplifyframework.api.aws"
+    defaultConfig {
+        consumerProguardFiles += file("consumer-rules.pro")
```
mattcreaser: ✅ Verified, no change requested — recording the checks so a reader does not have to redo them.

- `+=` is an append, not a replace: the convention plugin sets `consumerProguardFiles` during plugin application, and this `android { }` block runs afterwards, so the module ends up with `[configuration/consumer-rules.pro, aws-api/consumer-rules.pro]`. The repo-wide rules are not clobbered.
- "Same approach as `aws-auth-cognito`" checks out — `aws-auth-cognito/build.gradle.kts` has the identical `defaultConfig { consumerProguardFiles += file("consumer-rules.pro") }`.
- Placing it in `defaultConfig` of a library module is the correct spot for AAR-packaged consumer rules.

I could not verify that the rule actually lands in `aws-api-release.aar` without building; the mechanism is right and CI builds the AAR.

If you take the suggestion to move the rule into `configuration/consumer-rules.pro`, this whole `defaultConfig` block goes away again.

mattcreaser (approved):

aws-api/consumer-rules.pro
VarshithaPamisetty: Yes, will create a follow up PR. Thanks!

PR #3425, opened 2026-09-03
fix(api): classify rate limiting, and log what the AppSync client drops silently

mattcreaser:

### Description

**A throttled request was reported as a request defect.** Any 4xx that was not an auth rejection became `AppSyncValidationException`, whose advice is to correct the request — which sends a caller to fix a document and variables that are already correct, when the only thing wrong was the timing. `AppSyncRateLimitExceededException` is now raised for a 429 and for any 4xx carrying a `LimitExceededError` type, since the service does not always pair the two. The subscription path recognises it as well, beside the concurrent-subscription check and before the auth retry — a rate limit belongs to the API, not the identity, so another auth mode would consume an attempt and fail identically.

It is kept deliberately distinct from `AppSyncLimitExceededException`, the cap on how many subscriptions may be open at once. The two read alike but are released by different actions — backing off versus closing something already open — so collapsing them would give one of the two the wrong advice.

**Nothing in the module logged**, so three paths that deliberately continue rather than fail left no trace. The significant one is a subscription message that cannot be deserialized: it is dropped by design, but the consumer is told nothing, so a message lost to a schema mismatch is indistinguishable from one the service never sent. That now warns. An unrecognized frame logs at debug — keep-alives and acks deliberately do not, since they shared a branch with the unknown case and logging all three would bury the one that matters. The socket also warns when the watchdog closes a connection for silence and when OkHttp reports a failure; both end every subscription on that connection and neither previously said why.

Loggers are created internally per class, so there is no public API and no configuration for this. A customer enables output by registering a sink during application startup.

### Not included

The third TODO, registering deserializers for lazily-loaded model lists and pages, is left in place. It reads like an adapter registration but the counterpart is a family of types plus a way to issue follow-up queries on demand, which needs the client itself — registering adapters without that would produce values that deserialize and then fail on access. It belongs with the rest of the lazy-loading work rather than here.

### Testing

220 unit tests, 0 failures. `apiCheck`, `ktlintCheck`, `compileReleaseKotlin` and `licensee` all pass. `apiDump` regenerated: the only public change is `AppSyncRateLimitExceededException`.

New coverage: a 429, a `LimitExceededError` type under a different status, and a subscription error frame carrying it.

The logging is not unit tested, deliberately. Sinks are process-global, a logger captures the sink set when it is constructed, and foundation keeps sink removal internal to itself — so a test registering a fake sink could not remove it afterwards and would collect output from every other test in the module.

### DIFF
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
@@ -21,8 +21,9 @@ import com.amplifyframework.datastore.appsync.AppSyncExtensions
  * Recognises the errors that mean "this identity was rejected", as opposed to errors that would recur
  * with any identity. Only the former is worth retrying with a different auth mode.
  *
- * Both the HTTP and WebSocket paths need this, and both defer the classification to
- * [AppSyncExtensions] rather than maintaining a list of error-type strings here.
+ * Both the HTTP and WebSocket paths need this. Authorization failures defer to [AppSyncExtensions],
+ * which already knows which types mean "rejected identity". The two limit checks below cannot: its
+ * error-type enum does not model either of them, so they match the wire string directly.
  */
 
 /**
@@ -61,6 +62,9 @@ internal fun List<AppSyncWebSocketMessage.WireError>.hasSubscriptionLimitError()
  *
  * A cap on how fast requests may arrive, not on how many subscriptions may be open, so the remedy is to
  * back off rather than to release anything.
+ *
+ * Matched against the wire string for the same reason as [hasSubscriptionLimitError]: [AppSyncExtensions]
+ * does not model this error type either.
  */
 internal fun List<AppSyncWebSocketMessage.WireError>.hasRateLimitError(): Boolean =
     any { it.errorType == LIMIT_EXCEEDED }
@@ -83,6 +87,6 @@ internal fun List<AppSyncWebSocketMessage.WireError>.toGraphQLErrors(): List<Gra
         it.message,
         null,
         null,
-        it.errorType?.let { type -> mapOf("errorType" to type) } ?: emptyMap()
+        it.errorType?.let { type -> mapOf(ERROR_TYPE_KEY to type) } ?: emptyMap()
     )
 }
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
@@ -50,13 +50,28 @@ internal fun List<AppSyncWebSocketMessage.WireError>.hasUnauthorizedError(): Boo
  * Checked against the wire string directly rather than through [AppSyncExtensions], whose error-type
  * enum does not model this case.
  *
- * TODO: `LimitExceededError` is a sibling meaning a request-rate limit rather than a subscription
- *  count. It needs different recovery advice, so it is not folded in here.
+ * Deliberately narrower than [hasRateLimitError]: the two limits are released by different actions, so
+ * they must not be folded together.
  */
 internal fun List<AppSyncWebSocketMessage.WireError>.hasSubscriptionLimitError(): Boolean =
     any { it.errorType == MAX_SUBSCRIPTIONS_REACHED }
 
+/**
+ * Whether any of these errors says the API's request rate limit was exceeded.
+ *
+ * A cap on how fast requests may arrive, not on how many subscriptions may be open, so the remedy is to
+ * back off rather than to release anything.
+ */
+internal fun List<AppSyncWebSocketMessage.WireError>.hasRateLimitError(): Boolean =
+    any { it.errorType == LIMIT_EXCEEDED }
+
+/** Whether any error on this response says the API's request rate limit was exceeded. */
+internal fun GraphQLResponse<*>.hasRateLimitError(): Boolean =
+    errors.any { it.extensions?.get(ERROR_TYPE_KEY) == LIMIT_EXCEEDED }
+
 private const val MAX_SUBSCRIPTIONS_REACHED = "MaxSubscriptionsReachedError"
+private const val LIMIT_EXCEEDED = "LimitExceededError"
+private const val ERROR_TYPE_KEY = "errorType"
 
 /**
  * Converts wire errors into the response type callers already handle, so an exception carrying them can
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
@@ -283,6 +283,22 @@ class AppSyncValidationException(
 
 // ── Ungrouped leaves ────────────────────────────────────────────────────
 
+/**
+ * The API's request rate limit was exceeded.
+ *
+ * Ungrouped for the same reason [AppSyncNetworkException] is: nothing about the request is wrong and
+ * no credential will change the outcome. The same request is likely to succeed once the rate falls.
+ *
+ * Distinct from [AppSyncLimitExceededException], which is the cap on how many subscriptions may be
+ * open at once — that one is released by closing a subscription, this one by slowing down.
+ */
+@ExperimentalAmplifyApi
+class AppSyncRateLimitExceededException(
+    message: String,
+    recoverySuggestion: String = "Retry with exponential backoff, and reduce the rate of requests.",
+    cause: Throwable? = null
+) : AppSyncException(message, recoverySuggestion, cause)
+
 /** The request could not reach the service. */
 @ExperimentalAmplifyApi
 class AppSyncNetworkException(
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
@@ -39,7 +39,8 @@ import java.io.IOException
  *
  * The groups are [AppSyncAuthException], [AppSyncConfigurationException],
  * [AppSyncResponseException], [AppSyncSubscriptionException] and [AppSyncRequestException].
- * [AppSyncNetworkException] and [AppSyncUnknownException] are leaves directly under this base.
+ * [AppSyncNetworkException], [AppSyncRateLimitExceededException] and [AppSyncUnknownException] are
+ * leaves directly under this base.
  *
  * @param message Error message describing what went wrong
  * @param recoverySuggestion Suggested action to resolve the error
@@ -257,7 +258,7 @@ class AppSyncTimeoutException(
  * something already open has to be released first.
  */
 @ExperimentalAmplifyApi
-class AppSyncLimitExceededException(
+class AppSyncSubscriptionLimitExceededException(
     message: String,
     recoverySuggestion: String = "Close subscriptions that are no longer needed before opening more.",
     cause: Throwable? = null
@@ -289,8 +290,8 @@ class AppSyncValidationException(
  * Ungrouped for the same reason [AppSyncNetworkException] is: nothing about the request is wrong and
  * no credential will change the outcome. The same request is likely to succeed once the rate falls.
  *
- * Distinct from [AppSyncLimitExceededException], which is the cap on how many subscriptions may be
- * open at once — that one is released by closing a subscription, this one by slowing down.
+ * Only raised for a rejected request. A throttle reported on a successful response is delivered as an
+ * error on that response, as any other GraphQL error is.
  */
 @ExperimentalAmplifyApi
 class AppSyncRateLimitExceededException(
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
@@ -189,6 +189,15 @@ internal class AppSyncHttpTransport(
         // rather than having to recognise a status code or an error string. Either signal is enough:
         // AppSync answers a rejected identity with a 401, and a rejected operation with an
         // Unauthorized error type that can arrive under any 4xx.
+        // A throttle is not a defect in the request, so it must not arrive as one — the advice to check
+        // the document and variables would send a caller to correct something that is already correct.
+        if (response.code == HTTP_TOO_MANY_REQUESTS || parsed?.hasRateLimitError() == true) {
+            return AppSyncRateLimitExceededException(
+                message = "The request was throttled (HTTP status ${response.code})" +
+                    (errors?.joinToString("; ") { it.message }?.let { ": $it" } ?: ".")
+            )
+        }
+
         if (response.code == HTTP_UNAUTHORIZED || parsed?.hasUnauthorizedError() == true) {
             return AppSyncUnauthorizedException(
                 message = "The request was not authorized (HTTP status ${response.code})" +
@@ -244,6 +253,7 @@ internal class AppSyncHttpTransport(
         const val USER_AGENT_HEADER = "User-Agent"
         val CLIENT_ERROR_CODES = 400..499
         const val HTTP_UNAUTHORIZED = 401
+        const val HTTP_TOO_MANY_REQUESTS = 429
         const val SERVER_ERROR_MIN = 500
 
         // Internal error, bad gateway, service unavailable and gateway timeout. Each can clear without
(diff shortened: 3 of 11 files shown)


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -188,6 +188,15 @@ internal class AppSyncHttpTransport(
         // rather than having to recognise a status code or an error string. Either signal is enough:
         // AppSync answers a rejected identity with a 401, and a rejected operation with an
         // Unauthorized error type that can arrive under any 4xx.
+        // A throttle is not a defect in the request, so it must not arrive as one — the advice to check
```
mattcreaser: 🟡 The new two lines land at the *end* of the comment block that explains the unauthorized branch, so the whole block now sits above the throttle check. A reader arriving at line 193 reads "Reported as an auth failure so a caller can match the whole category to re-authenticate" immediately before an `if` that returns a rate-limit exception.

Move the throttle rationale above the auth block, or move the auth block down to sit on line 200 where it belongs.

```suggestion
        // A throttle is not a defect in the request, so it must not arrive as one — the advice to check
        // the document and variables would send a caller to correct something that is already correct.
```
(and relocate the four `Reported as an auth failure…` lines to directly above the `HTTP_UNAUTHORIZED` check)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncHttpTransport.kt
```diff
@@ -188,6 +188,15 @@ internal class AppSyncHttpTransport(
...
+        // the document and variables would send a caller to correct something that is already correct.
+        if (response.code == HTTP_TOO_MANY_REQUESTS || parsed?.hasRateLimitError() == true) {
```
mattcreaser: 🔵 Scoping question, not a defect. `hasRateLimitError()` is only consulted from `clientError()`, so it is reachable only for a 4xx. `hasUnauthorizedError()` is *also* consulted on a 200 in `execute()` (line 97). If AppSync can answer a throttle with a 200 carrying `LimitExceededError` — the same way it answers a rejected identity with a 200 carrying `Unauthorized` — then that case still reaches the caller as a successful `GraphQLResponse` with errors rather than as `AppSyncRateLimitExceededException`.

Not a regression either way, and the description does scope the change to 4xx deliberately. Worth a sentence in the KDoc saying a throttle on a 200 is delivered on the response, so the asymmetry with the auth path is a stated decision rather than something a later reader has to rediscover.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
```diff
@@ -283,6 +283,22 @@ class AppSyncValidationException(
 
 // ── Ungrouped leaves ────────────────────────────────────────────────────
 
+/**
+ * The API's request rate limit was exceeded.
+ *
+ * Ungrouped for the same reason [AppSyncNetworkException] is: nothing about the request is wrong and
```
mattcreaser: 🟡 This makes the class KDoc on `AppSyncException` stale. Lines 40–42 read:

> The groups are [AppSyncAuthException], [AppSyncConfigurationException], [AppSyncResponseException], [AppSyncSubscriptionException] and [AppSyncRequestException]. **[AppSyncNetworkException] and [AppSyncUnknownException] are leaves directly under this base.**

There are now three ungrouped leaves. That sentence is the only place a caller writing an exhaustive `when` over the hierarchy is told what sits outside the groups, so it needs the new type added.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
```diff
@@ -283,6 +283,22 @@ class AppSyncValidationException(
...
+ * no credential will change the outcome. The same request is likely to succeed once the rate falls.
+ *
+ * Distinct from [AppSyncLimitExceededException], which is the cap on how many subscriptions may be
```
mattcreaser: 🔵 The KDoc has to spend two sentences distinguishing this from `AppSyncLimitExceededException`, and the commit message spends a paragraph on it, which is the tell that the *names* are doing the wrong work rather than the docs.

The whole hierarchy is `@ExperimentalAmplifyApi`, so renaming the older one is free right now and expensive after release. `AppSyncSubscriptionLimitExceededException` says what it is, matches its `AppSyncSubscriptionException` parent, and would let both KDocs drop the mutual disambiguation. Optional, but this is the cheapest moment it will ever be.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncException.kt
```diff
@@ -283,6 +283,22 @@ class AppSyncValidationException(
...
+ * open at once — that one is released by closing a subscription, this one by slowing down.
+ */
+@ExperimentalAmplifyApi
+class AppSyncRateLimitExceededException(
```
mattcreaser: 🟡 `AppSyncExceptionTest` was not updated for this leaf, and it has two tests that should have caught it:

1. `every leaf carries a non-blank default recovery suggestion` (line 109) enumerates all 16 leaves and asserts `leaves.size shouldBe 16`. That guard is meant to force an update when a leaf is added, but it counts the literal list rather than the hierarchy, so it passes unchanged and the new default suggestion is untested.
2. `network and unknown are leaves directly under the base, in no group` (line 90) asserts exactly the invariant this class's KDoc claims — "Ungrouped for the same reason `AppSyncNetworkException` is". The new type should be in that list, otherwise the ungrouped placement is asserted only in prose and a later refactor could quietly park it under `AppSyncRequestException` with every test still green.

Both are one-line additions (plus bumping 16 → 17).

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -211,6 +214,7 @@ internal class AppSyncWebSocket(
     }
 
     override fun onFailure(webSocket: WebSocket, t: Throwable, response: Response?) {
+        logger.warn(t) { "The subscription connection failed." }
```
mattcreaser: 🟡 This warns on deliberate teardown, not just on genuine failure.

`restartWatchdog` (line 279) and `teardown` (line 301) both call `socket?.cancel()`, and an OkHttp `cancel()` delivers `onFailure` with the resulting `IOException`. So the watchdog path now emits two warnings in sequence — the accurate "Closing the subscription connection: no traffic for Nms." followed by "The subscription connection failed.", whose cause is the client's own cancellation. The handshake-timeout and `ConnectionError` paths in `connect()` produce the second line alone.

`handleClosed()` is idempotent so nothing breaks, but the log now contradicts itself precisely in the case the PR set out to explain. The existing `pendingCloseCause` check already distinguishes the two situations:

```suggestion
    override fun onFailure(webSocket: WebSocket, t: Throwable, response: Response?) {
        // onClosed is not called after a failure, so this has to close things out itself.
        if (pendingCloseCause == null) {
            logger.warn(t) { "The subscription connection failed." }
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -89,6 +90,8 @@ internal class AppSyncWebSocket(
     @Volatile
     private var watchdogTimeoutMs: Long = 0
 
+    private val logger = AmplifyLogging.logger<AppSyncWebSocket>()
```
mattcreaser: 🔵 This splits the `@Volatile` field group — `socket`, `watchdog`, `watchdogTimeoutMs`, then the logger, then `isClosed`. The logger is not volatile and not mutable state; grouping it with the other non-volatile `val`s near `messageFlow`/`scope` (or immediately after the constructor, as `AppSyncSubscriber` does) keeps the volatile block scannable as one thing.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -240,12 +243,21 @@ internal class AppSyncSubscriber(
             if (message.id == registration.id) {
                 // Deserialization failure is deliberately swallowed: it is non-terminal, so one
                 // unreadable message must not end a healthy subscription.
-                val response = runCatching {
+                val failure = runCatching {
```
mattcreaser: 🔵 `failure` holds a `Result`, which is a success on the normal path, so `failure.getOrNull()` on the next line reads as though a failure could yield a value. `attempt`, or `deserialization`, keeps both call sites honest:

```suggestion
                val deserialization = runCatching {
```
(then `deserialization.getOrNull()` and `deserialization.exceptionOrNull()`)

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -310,10 +328,16 @@ internal class AppSyncSubscriber(
         // subscription streams normally rather than throwing.
         is AppSyncWebSocketMessage.Closed -> message.cause?.let { throw it } ?: false
 
-        // Keep-alives, unknown frames, and anything addressed to another subscription.
+        // Expected traffic that this subscription has nothing to do.
```
mattcreaser: 🔵 Missing word.

```suggestion
        // Expected traffic that this subscription has nothing to do with.
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncSubscriber.kt
```diff
@@ -310,10 +328,16 @@ internal class AppSyncSubscriber(
...
         is AppSyncWebSocketMessage.ConnectionAck,
-        is AppSyncWebSocketMessage.KeepAlive,
-        is AppSyncWebSocketMessage.Unknown -> true
+        is AppSyncWebSocketMessage.KeepAlive -> true
+
+        // Logged, unlike the two above, because an unmodelled frame means the wire carried something
+        // this client does not understand — worth a trace, where a keep-alive would only be noise.
+        is AppSyncWebSocketMessage.Unknown -> {
+            logger.debug { "Ignoring an unrecognized message of type ${message.type}." }
```
mattcreaser: 🟡 This fires once per active subscription, not once per frame.

`handle()` is invoked from every `subscribe()` collector — that is the design the class KDoc states ("a subscription is a *filter* over that flow keyed by subscription id"). `Unknown` carries no id, so unlike `Data`, `Error` and `Complete` there is no `message.id == registration.id` gate to drop it: with ten open subscriptions, one unmodelled frame produces ten identical debug lines. The `Closed` branch has the same shape, but that one genuinely concerns every subscription; an unknown frame concerns none of them.

The commit message says "the component that receives a frame is what logs it", and the single component that receives each frame exactly once is `AppSyncWebSocket.onMessage`. Logging there — where `parse()` is called, and where the `raw` payload is still in scope for a more useful message — gives one line per frame and lets this branch go back to a bare `true`.

Secondary: `message.type` is nullable, so a frame with no `type` field logs "of type null".

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
```diff
@@ -42,13 +42,28 @@ internal fun List<AppSyncWebSocketMessage.WireError>.hasUnauthorizedError(): Boo
...
 internal fun List<AppSyncWebSocketMessage.WireError>.hasSubscriptionLimitError(): Boolean =
     any { it.errorType == MAX_SUBSCRIPTIONS_REACHED }
 
+/**
+ * Whether any of these errors says the API's request rate limit was exceeded.
+ *
+ * A cap on how fast requests may arrive, not on how many subscriptions may be open, so the remedy is to
+ * back off rather than to release anything.
+ */
+internal fun List<AppSyncWebSocketMessage.WireError>.hasRateLimitError(): Boolean =
```
mattcreaser: 🔵 I checked `AppSyncErrorType` — it models only `ConflictUnhandled`, `OperationDisabled`, `UnauthorizedException` and `Unauthorized`, so matching the wire string here is correct and unavoidable. But `hasSubscriptionLimitError` states that reason explicitly ("Checked against the wire string directly rather than through [AppSyncExtensions], whose error-type enum does not model this case") and this one does not, which leaves it looking like an oversight rather than a decision.

Related: the file header at lines 20–26 still asserts that "both defer the classification to [AppSyncExtensions] rather than maintaining a list of error-type strings here", while the file now maintains two such strings. That header was already half-stale before this PR; adding the second string is a good moment to correct it.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncAuthErrors.kt
```diff
@@ -42,13 +42,28 @@ internal fun List<AppSyncWebSocketMessage.WireError>.hasUnauthorizedError(): Boo
...
+    any { it.errorType == LIMIT_EXCEEDED }
+
+/** Whether any error on this response says the API's request rate limit was exceeded. */
+internal fun GraphQLResponse<*>.hasRateLimitError(): Boolean =
+    errors.any { it.extensions?.get(ERROR_TYPE_KEY) == LIMIT_EXCEEDED }
+
 private const val MAX_SUBSCRIPTIONS_REACHED = "MaxSubscriptionsReachedError"
+private const val LIMIT_EXCEEDED = "LimitExceededError"
+private const val ERROR_TYPE_KEY = "errorType"
```
mattcreaser: 🔵 `ERROR_TYPE_KEY` is introduced here but `toGraphQLErrors()` at line 78 still writes the literal `"errorType"` into the extensions map it builds. Those two are the write and the read of the same key — `hasRateLimitError()` on a `GraphQLResponse` matches against exactly what `toGraphQLErrors()` produces — so they should not be able to drift. Use the constant in both.

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncSubscriberTest.kt
```diff
@@ -526,6 +526,28 @@ class AppSyncSubscriberTest {
         }
     }
 
+    @Test
+    fun `a rate limit error is terminal and distinct from the subscription limit`() = runTest {
```
mattcreaser: 🟡 This does not test what its name and the PR description claim. The title says "distinct from the subscription limit" and the description says the check sits "beside the concurrent-subscription check and before the auth retry" — but the frame carries only `LimitExceededError`, so `rejectedIdentity` is false and the auth-retry branch could never have run regardless of ordering. Move the new `when` branch below `rejectedIdentity && registration.canRetry` and this test still passes.

The file already has the right pattern one test down — `a subscription limit error wins over an unauthorized error in the same frame` emits `listOf(unauthorized(), wireError("limit", "MaxSubscriptionsReachedError"))`. Same shape here pins the ordering:

```kotlin
messages.emit(
    AppSyncWebSocketMessage.Error(
        firstId,
        listOf(unauthorized(), wireError("Rate exceeded", "LimitExceededError"))
    )
)

awaitError().shouldBeInstanceOf<AppSyncRateLimitExceededException>()
```

A second frame carrying both `MaxSubscriptionsReachedError` and `LimitExceededError` would pin the other half of the ordering — which of the two limits wins — since "distinct from the subscription limit" is what the test name promises and nothing currently asserts it.

Minor: `val firstId = startedId` is captured but only used once, immediately; the neighbouring tests just use `startedId` inline.

aws-appsync/src/test/java/com/amazonaws/appsync/AmplifyAppSyncClientTest.kt
```diff
@@ -186,6 +186,31 @@ class AmplifyAppSyncClientTest {
...
+        server.enqueue(jsonResponse("Too Many Requests", code = 429))
+
+        val error = client().query(request()).shouldBeFailure().error
+
+        error.shouldBeInstanceOf<AppSyncRateLimitExceededException>()
+        error.recoverySuggestion shouldContain "backoff"
+    }
+
+    @Test
+    fun `a LimitExceededError type under another status is still a rate limit`() = runTest {
```
mattcreaser: 🔵 Between the two new tests, the message `AppSyncRateLimitExceededException` actually carries is never asserted. The 429 test sends a non-JSON body, so `parsed` is null, `errors` is null, and the message takes the `"."` branch; this test sends real errors but asserts only the type. So the `": Rate exceeded"` suffix — the part that tells a caller which limit the service reported — is unexercised.

The adjacent auth tests already assert on the message (`error.message shouldContain "Valid authorization header not provided"`). One line here would match:

```kotlin
client().query(request()).shouldBeFailure().error
    .shouldBeInstanceOf<AppSyncRateLimitExceededException>()
    .message shouldContain "Rate exceeded"
```

harsh62 (approved):

PR #3426, opened 2026-09-04
chore: upgrade MockWebServer to 5.4.0 and migrate off the removed package

mattcreaser:

### Why

MockWebServer had its own version pin at `5.0.0-alpha.11` while OkHttp moved to `5.4.0`. The two are published from the same release, so a separate pin only creates the opportunity for them to drift — and they had.

The alpha's WebSocket support reaches into `okhttp3.internal.ws.RealWebSocket$Streams`, which no longer exists, so serving a WebSocket upgrade fails at `start()` with a `NoClassDefFoundError` on an Android runtime. That is how this surfaced: it makes instrumented coverage of any WebSocket client impossible. JVM unit tests were unaffected, which is why it went unnoticed.

### What changed

**The separate version pin is gone.** The library entry now takes `version.ref = "okhttp"`, so the two cannot diverge again. This is the actual fix — the version bump alone would leave the same trap set for the next OkHttp upgrade.

**The coordinate moves from `com.squareup.okhttp3:mockwebserver` to `:mockwebserver3`.** The former still ships in 5.4.0, but only as a compatibility shim: 13 classes holding the deprecated `okhttp3.mockwebserver` package and delegating to `mockwebserver3`. Worth dropping now that nothing needs the old package.

That is what requires the test migration — the deprecated package is no longer on the classpath:

- `MockResponse` is immutable in 5.x, so every construction moves to its `Builder` (`setResponseCode` → `code`, `setBody` → `body`, `withWebSocketUpgrade` → `webSocketUpgrade`).
- `RecordedRequest` changed: `getHeader(name)` → `headers[name]`, `getPath()` → `target`, and `body` is a nullable `ByteString` rather than a `Buffer`.
- `server.shutdown()` → `server.close()`.

Two behaviour changes broke tests until handled, and are worth knowing if you maintain tests here:

- **`url()` no longer starts the server implicitly.** Tests calling it from a field initializer, or relying on lazy startup, need an explicit `start()`. That alone accounted for every failure in `aws-predictions` and in the Events REST tests.
- **The `SocketPolicy` enum is replaced by `SocketEffect`**, applied at a named point in the exchange. The single `DISCONNECT_AT_START` usage becomes `onRequestStart(SocketEffect.CloseSocket())`, which preserves its meaning: accept the connection, then kill the socket before replying. `ShutdownConnection` would be a graceful FIN and `Stall` would turn a disconnect test into a write-timeout test, so neither is equivalent.

### Testing

419 tests across `aws-api` (214), `aws-connect` (49), `aws-predictions` (78) and `aws-sdk-appsync-events` (78) — 0 failures. `ktlintCheck` and `checkstyle` pass.

Every declared test still runs: per-class counts were compared against the `@Test` counts at `HEAD` to confirm nothing was silently dropped, and no assertion was weakened to accommodate the new API. Verified that only `mockwebserver3` resolves on the test classpath, so nothing still reaches the shim.

### Noted, not fixed here

`LivenessWebSocketTest.web socket detects clock skew from server response` waits out its full 3s timeout both before and after this change — its `CountDownLatch(2)` never reaches zero, and it passes only because `webSocket` was already assigned by an earlier manually-invoked `onOpen`. Pre-existing and unrelated to this migration, but it does not currently assert what its name suggests.

### DIFF
--- a/appsync/aws-sdk-appsync-events/src/test/java/com/amazonaws/sdk/appsync/events/EventsRestClientTest.kt
+++ b/appsync/aws-sdk-appsync-events/src/test/java/com/amazonaws/sdk/appsync/events/EventsRestClientTest.kt
@@ -29,14 +29,14 @@ import kotlinx.coroutines.test.runTest
 import kotlinx.serialization.json.JsonPrimitive
 import mockwebserver3.MockResponse
 import mockwebserver3.MockWebServer
-import mockwebserver3.SocketPolicy
+import mockwebserver3.SocketEffect
 import okio.Buffer
 import org.junit.After
-import org.junit.Before
 import org.junit.Test
 
 class EventsRestClientTest {
-    private val mockWebServer = MockWebServer()
+    // Server must be started before url() is called in the interceptor initializer below
+    private val mockWebServer = MockWebServer().apply { start() }
     private val expectedEndpoint = "https://abc.appsync-api.us-east-1.amazonaws.com/event"
     private val expectedHost = "abc.appsync-api.us-east-1.amazonaws.com"
     private val expectedStandardHeaders = mapOf(
@@ -58,14 +58,9 @@ class EventsRestClientTest {
         )
     )
 
-    @Before
-    fun setUp() {
-        mockWebServer.start()
-    }
-
     @After
     fun tearDown() {
-        mockWebServer.shutdown()
+        mockWebServer.close()
     }
 
     @Test
@@ -78,9 +73,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
@@ -122,10 +118,11 @@ class EventsRestClientTest {
             .readText()
 
         mockWebServer.enqueue(
-            MockResponse()
-                .setSocketPolicy(SocketPolicy.DISCONNECT_AT_START)
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .onRequestStart(SocketEffect.CloseSocket())
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         val response = client.publish(expectedChannel, JsonPrimitive(1))
@@ -160,9 +157,10 @@ class EventsRestClientTest {
             .readText()
 
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(400)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(400)
+                .body(responseBody)
+                .build()
         )
 
         val response = client.publish(expectedChannel, JsonPrimitive(1))
@@ -199,9 +197,10 @@ class EventsRestClientTest {
         val responseBody = "uh-oh"
 
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(400)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(400)
+                .body(responseBody)
+                .build()
         )
 
         val response = client.publish(expectedChannel, JsonPrimitive(1))
@@ -241,9 +240,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
@@ -284,9 +284,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
@@ -327,9 +328,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
@@ -368,9 +370,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
@@ -411,9 +414,10 @@ class EventsRestClientTest {
             .bufferedReader()
             .readText()
         mockWebServer.enqueue(
-            MockResponse()
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .code(200)
+                .body(responseBody)
+                .build()
         )
 
         // WHEN
--- a/aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
+++ b/aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
@@ -53,14 +53,14 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.StreamSupport;
 
+import mockwebserver3.MockResponse;
+import mockwebserver3.MockWebServer;
+import mockwebserver3.RecordedRequest;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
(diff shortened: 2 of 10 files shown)


gradle/libs.versions.toml
```diff
@@ -45,7 +45,7 @@ material = "1.8.0"
 mockito = "3.9.0"
 mockito-inline = "3.11.2"
 mockk = "1.14.5"
-mockwebserver = "5.0.0-alpha.11"
+mockwebserver = "5.4.0"
```
mattcreaser: 🟡 Two things about this catalog entry.

**1. The stated rationale for the package migration does not hold.** I unpacked `com.squareup.okhttp3:mockwebserver:5.4.0` and the deprecated package is still very much there:

```
okhttp3/mockwebserver/MockWebServer.class
okhttp3/mockwebserver/MockResponse.class
okhttp3/mockwebserver/RecordedRequest.class
okhttp3/mockwebserver/SocketPolicy.class
okhttp3/mockwebserver/DeprecationBridgeKt.class
```

And every member the six `aws-api` files were using still exists on it — `MockResponse.setBody(String)`, `setResponseCode(int)`, `setSocketPolicy(...)`, `withWebSocketUpgrade(...)`, `MockWebServer.shutdown()`, `RecordedRequest.getHeader(String)`, `getPath()`, `getBody(): Buffer`, and `SocketPolicy.DISCONNECT_AT_START`. The class is a delegating bridge (`getDelegate(): mockwebserver3.MockWebServer`).

So "the deprecated `okhttp3.mockwebserver` package is gone" and "not a version bump alone" are both inaccurate: a bump-only change would have compiled, and the `NoClassDefFoundError` fix comes entirely from the bump — the alpha's `mockwebserver3` classes reference `okhttp3/internal/ws/RealWebSocket$Streams`, which `okhttp-jvm-5.4.0.jar` no longer contains (confirmed both ways). The migration is still the right call — the bridge is deprecated and its `url()` has the same no-implicit-start behaviour anyway — but please fix the rationale in the description and commit message, since it is the load-bearing justification for the diff's size.

**2. Point the coordinate at the artifact you actually use.** With this PR, every `MockWebServer` usage in the repo imports `mockwebserver3.*`, which is only on the compile classpath transitively (`mockwebserver` declares `mockwebserver3` in its `apiElements`). Line 149 should name it directly:

```suggestion:-0+0
test-mockwebserver = { module = "com.squareup.okhttp3:mockwebserver3", version.ref = "mockwebserver" }
```

`mockwebserver3` is published at `5.4.0` (jar + pom + module metadata all resolve). That drops the deprecated package and its `junit:junit` dependency from four test classpaths, and makes the declared coordinate match the imports — so nobody accidentally reintroduces the bridge. (I could not put a suggestion directly on line 149 as it is outside the diff.)

appsync/aws-sdk-appsync-events/src/test/java/com/amazonaws/sdk/appsync/events/EventsRestClientTest.kt
```diff
@@ -29,14 +29,14 @@ import kotlinx.coroutines.test.runTest
...
+import mockwebserver3.SocketEffect
 import okio.Buffer
 import org.junit.After
-import org.junit.Before
 import org.junit.Test
 
 class EventsRestClientTest {
-    private val mockWebServer = MockWebServer()
+    // Server must be started before url() is called in the interceptor initializer below
+    private val mockWebServer = MockWebServer().apply { start() }
```
mattcreaser: 🔵 This works, and the comment earns its place, but it makes the fix depend on Kotlin property initialisation order: `interceptor` on line 48 calls `mockWebServer.url("/event")`, so correctness now rests on `mockWebServer` being declared above it. Reordering these declarations — a change that looks purely cosmetic — breaks the class, and the failure surfaces during construction rather than in a `@Before`, so JUnit reports it without naming a test method.

Making the dependency explicit rather than positional would be more durable, e.g. keep `@Before fun setUp() { mockWebServer.start() }` and make the dependents lazy:

```kotlin
private val mockWebServer = MockWebServer()
private val interceptor by lazy { ConvertToMockRequestInterceptor(mockWebServer.url("/event")) }
private val client by lazy { ... }
```

Your call — the current form is defensible given the comment.

appsync/aws-sdk-appsync-events/src/test/java/com/amazonaws/sdk/appsync/events/EventsRestClientTest.kt
```diff
@@ -122,10 +118,11 @@ class EventsRestClientTest {
             .readText()
 
         mockWebServer.enqueue(
-            MockResponse()
-                .setSocketPolicy(SocketPolicy.DISCONNECT_AT_START)
-                .setResponseCode(200)
-                .setBody(responseBody)
+            MockResponse.Builder()
+                .onRequestStart(SocketEffect.CloseSocket())
```
mattcreaser: ✅ Verified this mapping, and your reasoning in the description is right. `SocketEffect.CloseSocket()` defaults to `closeSocket = true, shutdownInput = false, shutdownOutput = false`, and the upstream KDoc confirms the alternatives are not equivalent — `ShutdownConnection` sends `GOAWAY` on HTTP/2, `Stall` means "stop processing this". `onRequestStart` is applied before the exchange is served, which preserves `DISCONNECT_AT_START`.

🔵 While this line is being rewritten: `.code(200)` and `.body(responseBody)` are unreachable on this response — the socket is closed before anything is written — and `responseBody` above is read out of `publish_single_success.json` purely to feed them. Dropping both (and the resource read) would make the test say what it means. Dead before this PR too, so entirely optional.

aws-predictions/src/test/java/com/amplifyframework/predictions/aws/http/LivenessWebSocketTest.kt
```diff
@@ -117,11 +117,13 @@ internal class LivenessWebSocketTest {
     fun setUp() {
         Dispatchers.setMain(Dispatchers.Unconfined)
         server = MockWebServer()
+        // url() no longer starts the server implicitly, so it must be started before it can be addressed
+        server.start()
```
mattcreaser: ✅ Verified: this is the only `server.start()` left in the file, so no double-start, and it is genuinely required — `createLivenessWebSocket()` calls `server.url("/")` on line 790, which no longer starts the server implicitly.

✅ Your "noted, not fixed here" observation on `web socket detects clock skew from server response` is consistent with the code as written. `LatchingWebSocketResponseListener.onOpen` counts down exactly once per open, and the first `onOpen` is invoked manually on line 686 *before* the latching listener is installed on line 698 — so only one countdown can ever reach a `CountDownLatch(2)` and `await(3, SECONDS)` must time out. The subsequent assertions pass because the reconnect reassigns `livenessWebSocket.webSocket` regardless of the latch. I could not confirm the 3s wall-clock cost without running it, but the reasoning holds by inspection. Worth a follow-up issue so it does not get lost.

aws-api/src/test/java/com/amplifyframework/api/aws/AppSyncErrorPropagationTest.kt
```diff
@@ -75,12 +75,12 @@ class AppSyncErrorPropagationTest {
 
     /**
      * Stop the [MockWebServer] that was started in [setup].
-     * @throws IOException On failure to shutdown the MockWebServer
+     * @throws IOException On failure to close the MockWebServer
      */
     @After
     @Throws(IOException::class)
```
mattcreaser: 🔵 Now vestigial. The old `okhttp3.mockwebserver.MockWebServer.shutdown()` was declared `throws IOException`; `mockwebserver3.MockWebServer.close()` is `public override fun close()` with no checked exception. So `@Throws(IOException::class)` here can no longer fire, and the KDoc line above it — which this PR reworded rather than removed — documents an outcome that cannot occur.

Same pattern in `RestQuerySemicolonEncodingTest.cleanup()` (line 65), and as `throws IOException` on `cleanup()`/`tearDown()` in `AWSApiPluginTest`, `AWSApiPluginUserAgentTest`, `OwnerBasedAuthTest` and `AWSRestOperationTest`. Harmless, but since you touched every one of these signatures' documentation it is a natural place to tidy up. The `java.io.IOException` imports stay needed either way — `setup()` still throws it from `start()`.

aws-api/src/test/java/com/amplifyframework/api/aws/AWSApiPluginTest.java
```diff
@@ -325,7 +328,7 @@ public void headerInterceptorsAreConfigured() throws ApiException, InterruptedEx
 
         RecordedRequest recordedRequest = webServer.takeRequest(5, TimeUnit.MILLISECONDS);
         assertNotNull(recordedRequest);
-        assertEquals("specialValue", recordedRequest.getHeader("specialKey"));
+        assertEquals("specialValue", recordedRequest.getHeaders().get("specialKey"));
```
mattcreaser: ✅ Correct replacement: `mockwebserver3.RecordedRequest` genuinely drops `getHeader(String)`, and `headers` is a plain `okhttp3.Headers`, so `getHeaders().get(name)` is the right translation and preserves the null-returning semantics the `assertNull` on line 355 depends on.

🔵 Unrelated and pre-existing, but two lines up: `takeRequest(5, TimeUnit.MILLISECONDS)` is a 5 *millisecond* budget for a request to traverse the interceptor chain and reach the server — every other call site in these tests uses seconds. It is a latent flake on a loaded CI host, and the `assertNotNull` on line 330 is what turns it into a confusing failure. Since this hunk is already open, widening it to seconds is nearly free. Not a blocker.

thisisabhash (approved):

Issue #2898, opened 2024-08-19
io.reactivex.rxjava3.exceptions.UndeliverableException in DataStore


tobias-feldmann:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
    //Amplify
    implementation("com.amplifyframework:core:2.21.0")
    implementation("com.amplifyframework:aws-api:2.21.0")
    implementation("com.amplifyframework:aws-auth-cognito:2.21.0")
    implementation("com.amplifyframework:aws-datastore:2.21.0")
    implementation("com.amplifyframework:aws-storage-s3:2.21.0")
```

</details>

### Environment information

<details>

```
------------------------------------------------------------
Gradle 8.2
------------------------------------------------------------

Build time:   2023-06-30 18:02:30 UTC
Revision:     5f4a070a62a31a17438ac998c2b849f4f6892877

Kotlin:       1.8.20
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          17.0.1 (Oracle Corporation 17.0.1+12-LTS-39)
OS:           Mac OS X 14.4 aarch64



```

</details>

### Describe the bug

We see this error on many devices via the Google Play logs.
But I can't reproduce it on our test devices. 

Occurs with various Android SDKs and devices:
Android 9 (SDK 28)
Android 10 (SDK 29)
Android 11 (SDK 30)
Android 12 (SDK 31)
Android 13 (SDK 33)
Android 14 (SDK 34)
Android 15 Beta (SDK 35)

Same issue as in the ticket: #2654

> com.amplifyframework.datastore.storage.sqlite.PersistentModelVersion$$ExternalSyntheticLambda2.accept
Exception io.reactivex.rxjava3.exceptions.UndeliverableException:
at io.reactivex.rxjava3.plugins.RxJavaPlugins.onError (RxJavaPlugins.java:367)
at io.reactivex.rxjava3.internal.operators.single.SingleCreate$Emitter.onError (SingleCreate.java:82)
at com.amplifyframework.datastore.storage.sqlite.PersistentModelVersion$$ExternalSyntheticLambda2.accept
[...]

Issue #2654, opened 2023-12-14
Fatal Exception: io.reactivex.rxjava3.exceptions.UndeliverableException


mvishnureddy007:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-android/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-android/discussions).

### Language and Async Model

Java

### Amplify Categories

DataStore

### Gradle script dependencies

<details>

```groovy
// Put output below this line

//Amplify Framework
    implementation 'com.amplifyframework:core:2.13.0'
    implementation 'com.amplifyframework:aws-api:2.13.0'
    implementation 'com.amplifyframework:aws-datastore:2.13.0'
    implementation 'com.amplifyframework:aws-auth-cognito:2.13.0'
    implementation 'com.amplifyframework:rxbindings:2.13.0'
```

</details>

### Environment information

<details>

```
# Put output below this line

------------------------------------------------------------
Gradle 7.2
------------------------------------------------------------

Build time:   2021-08-17 09:59:03 UTC
Revision:     a773786b58bb28710e3dc96c4d1a7063628952ad

Kotlin:       1.5.21
Groovy:       3.0.8
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.21 (Ubuntu 11.0.21+9-post-Ubuntu-0ubuntu122.04)
OS:           Linux 6.2.0-37-generic amd64

```

</details>

### Describe the bug

The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | DataStoreException{message=Wanted 1 metadata for item with id = b73c344a-ce0a-44b4-9893-293510a1b992, but had 0., cause=null, recoverySuggestion=This is likely a bug. please report to AWS.}

### Code Snippet

```java
// Put your code below this line.

```

### Log output

<details>

```
// Put your logs below this line


```

</details>

### GraphQL Schema

<details>

```graphql
// Put your schema below this line


```

</details>

gpanshu:
Hello,

Can you provide an example schema and replication steps to help reproduce the issue ?

mvishnureddy007:
Hello,
I have integrated amplify aws for chat functionality in my application

These are the following steps
[...]

PR #3433, opened 2026-09-09
fix(datastore): Use tryOnError in base sync to prevent UndeliverableException crash

VarshithaPamisetty:
## Description

DataStore's base sync wraps the AppSync query in an RxJava `Single` in `SyncProcessor`. If sync is stopped while a request is still running, the OkHttp callback comes back after the subscriber is already disposed. The error is then delivered to a dead emitter, so RxJava reports it as an `UndeliverableException` and the app crashes. This is the crash seen across devices in Play logs, the same crash class as the earlier #2654 report.

`SyncProcessor` now emits these errors through `tryOnError` on both the failure and no-data paths. On a disposed emitter it is a no-op, so the error is logged rather than crashing the app, while active subscribers are unaffected. This follows the existing convention in `Orchestrator` and the rxbindings adapters.

## Testing

- Reproduced e2e on 2.21.0: started sync, dropped the network, and stopped DataStore mid-flight. The `UndeliverableException` appeared in logcat matching the reported stack.
- With the fix, re-ran the same race on API 28, 31, and 34 (~380 attempts): zero crashes, sync unaffected.
- Added unit tests for the failure and no-data branches. Each fails on the old code and passes with the fix. Unit tests, ktlint, checkstyle, and apiCheck all pass.

Refs #2898

### DIFF
--- a/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
+++ b/aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
@@ -22,6 +22,8 @@ import com.amplifyframework.AmplifyException;
 import com.amplifyframework.api.graphql.GraphQLRequest;
 import com.amplifyframework.api.graphql.GraphQLResponse;
 import com.amplifyframework.api.graphql.PaginatedResult;
+import com.amplifyframework.core.Consumer;
+import com.amplifyframework.core.async.NoOpCancelable;
 import com.amplifyframework.core.model.Model;
 import com.amplifyframework.core.model.ModelProvider;
 import com.amplifyframework.core.model.ModelSchema;
@@ -79,12 +81,16 @@ import java.util.Random;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.exceptions.UndeliverableException;
 import io.reactivex.rxjava3.observers.TestObserver;
+import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 
 import static com.amplifyframework.datastore.appsync.TestModelWithMetadataInstances.BLOGGER_ISLA;
 import static com.amplifyframework.datastore.appsync.TestModelWithMetadataInstances.BLOGGER_JAMESON;
@@ -92,10 +98,12 @@ import static com.amplifyframework.datastore.appsync.TestModelWithMetadataInstan
 import static com.amplifyframework.datastore.appsync.TestModelWithMetadataInstances.DRUM_POST;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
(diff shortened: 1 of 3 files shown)


aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+     * on a disposed emitter instead of routing to the global RxJava error handler.
+     * @throws AmplifyException On failure to build the SyncProcessor.
+     * @throws InterruptedException If interrupted while awaiting the captured sync callback.
+     */
+    @Test
+    public void syncFailureAfterDisposeDoesNotThrowUndeliverable() throws AmplifyException, InterruptedException {
+        isSyncRetryEnabled = false;
+        initSyncProcessor(SYNC_MAX_RECORDS);
+
+        // Capture the failure callback but never invoke it, so the sync stays in flight.
```
mattcreaser: [WARNING] Comment contradicts the test: the failure callback is invoked

Current code:
```
        // Capture the failure callback but never invoke it, so the sync stays in flight.
```

The comment says the callback is never invoked, but line 927 calls capturedOnFailure.get().accept(...) after disposal, and that invocation is the whole point of the test. Describe the deferral, not a non-invocation.

Suggested change:
```suggestion
        // Capture the failure callback without invoking it yet, so the sync stays in flight.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+     * resulting error must be delivered via emitter.tryOnError and never surface as an RxJava
+     * UndeliverableException.
+     * @throws AmplifyException On failure to build the SyncProcessor.
+     */
+    @Test
+    public void syncNoDataErrorAfterDisposeDoesNotThrowUndeliverable() throws AmplifyException {
+        isSyncRetryEnabled = false;
+        initSyncProcessor(SYNC_MAX_RECORDS);
+
+        // Capture the response callback but never invoke it, so the sync stays in flight.
```
mattcreaser: [WARNING] Comment contradicts the test: the response callback is invoked

Current code:
```
        // Capture the response callback but never invoke it, so the sync stays in flight.
```

Same inaccuracy as the failure-callback test. Line 975 calls capturedOnResponse.get().accept(noDataResponse) after disposal.

Suggested change:
```suggestion
        // Capture the response callback without invoking it yet, so the sync stays in flight.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+    /**
+     * Regression test: when the base-sync subscriber is disposed (for example, via DataStore.stop())
+     * while a sync network call is in flight, a subsequent failure callback must not crash the app with
+     * an RxJava UndeliverableException. The fix delivers the error via emitter.tryOnError, which no-ops
+     * on a disposed emitter instead of routing to the global RxJava error handler.
+     * @throws AmplifyException On failure to build the SyncProcessor.
+     * @throws InterruptedException If interrupted while awaiting the captured sync callback.
+     */
+    @Test
+    public void syncFailureAfterDisposeDoesNotThrowUndeliverable() throws AmplifyException, InterruptedException {
```
mattcreaser: [WARNING] New tests are Java, but AGENTS.md requires Kotlin

Current code:
```
    @Test
    public void syncFailureAfterDisposeDoesNotThrowUndeliverable() throws AmplifyException, InterruptedException {
```

AGENTS.md section 9 states "New tests MUST be written in Kotlin". Both new regression cases were added to this Java class. The counter-argument is real: they reuse initSyncProcessor, AppSyncMocking, and the class fixture, and a sibling Kotlin class would duplicate that setup. Either move them to Kotlin or state in the PR description why an exception is warranted, so the next contributor does not treat this file as precedent.

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
     }
 
+    /**
+     * Regression test: when the base-sync subscriber is disposed (for example, via DataStore.stop())
+     * while a sync network call is in flight, a subsequent failure callback must not crash the app with
+     * an RxJava UndeliverableException. The fix delivers the error via emitter.tryOnError, which no-ops
+     * on a disposed emitter instead of routing to the global RxJava error handler.
+     * @throws AmplifyException On failure to build the SyncProcessor.
+     * @throws InterruptedException If interrupted while awaiting the captured sync callback.
+     */
```
mattcreaser: [SUGGESTION] Javadoc narrates the implementation it is testing

Current code:
```
    /**
     * Regression test: when the base-sync subscriber is disposed (for example, via DataStore.stop())
     * while a sync network call is in flight, a subsequent failure callback must not crash the app with
     * an RxJava UndeliverableException. The fix delivers the error via emitter.tryOnError, which no-ops
     * on a disposed emitter instead of routing to the global RxJava error handler.
     * @throws AmplifyException On failure to build the SyncProcessor.
     * @throws InterruptedException If interrupted while awaiting the captured sync callback.
     */
```

The method name and body already show the disposal setup, and naming emitter.tryOnError couples the test doc to an implementation detail that may change. Keep the observable guarantee and the required @throws tags.

Suggested change:
```suggestion
    /**
     * Verifies that a late AppSync failure after disposal is not routed to RxJava's global error handler.
     * @throws AmplifyException On failure to build the SyncProcessor.
     * @throws InterruptedException If interrupted while awaiting the captured sync callback.
     */
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+        }
+    }
+
+    /**
+     * Companion to the failure-callback regression above, covering the success-callback error branch:
+     * when AppSync returns a no-data (errors-only) response after the subscriber has been disposed, the
+     * resulting error must be delivered via emitter.tryOnError and never surface as an RxJava
+     * UndeliverableException.
+     * @throws AmplifyException On failure to build the SyncProcessor.
+     */
```
mattcreaser: [SUGGESTION] Javadoc opens with a cross-reference instead of the guarantee

Current code:
```
    /**
     * Companion to the failure-callback regression above, covering the success-callback error branch:
     * when AppSync returns a no-data (errors-only) response after the subscriber has been disposed, the
     * resulting error must be delivered via emitter.tryOnError and never surface as an RxJava
     * UndeliverableException.
     * @throws AmplifyException On failure to build the SyncProcessor.
     */
```

"Companion to the failure-callback regression above" breaks if either method moves, and the tryOnError detail repeats the production code. One sentence naming the errors-only branch is enough to distinguish this case.

Suggested change:
```suggestion
    /**
     * Verifies that a late errors-only AppSync response after disposal is not routed to RxJava's
     * global error handler.
     * @throws AmplifyException On failure to build the SyncProcessor.
     */
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
cadivus: Remove the nonexistent InterruptedException contract

This method performs no interruptible wait. Remove this tag and the matching throws declaration.

```suggestion
    /**
     * Regression test: when the base-sync subscriber is disposed (for example, via DataStore.stop())
     * while a sync network call is in flight, a subsequent failure callback must not crash the app with
     * an RxJava UndeliverableException. The fix delivers the error via emitter.tryOnError, which no-ops
     * on a disposed emitter instead of routing to the global RxJava error handler.
     * @throws AmplifyException On failure to build the SyncProcessor.
     */
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
cadivus: Do not claim the failure callback is never invoked

The test invokes this callback later; the mock only delays invocation.

```suggestion
        // Capture the failure callback without invoking it yet, keeping the sync in flight.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
cadivus: Do not claim the response callback is never invoked

The test invokes this callback later; the mock only delays invocation.

```suggestion
        // Capture the response callback without invoking it yet, keeping the sync in flight.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
cadivus: I think we write new tests in Kotlin. @mattcreaser what is your take here?

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+            return new NoOpCancelable();
+        }).when(appSync).sync(any(), any(), any());
+
+        // Record errors routed to RxJava's global handler.
+        final List<Throwable> undelivered = new CopyOnWriteArrayList<>();
+        final io.reactivex.rxjava3.functions.Consumer<? super Throwable> previousHandler =
+                RxJavaPlugins.getErrorHandler();
+        RxJavaPlugins.setErrorHandler(undelivered::add);
+        try {
+            // Subscribe, then wait for the sync callback to be captured.
```
cadivus: Suggestion: Label the subscription step as the Act phase

Expose the Arrange-Act-Assert transition instead of narrating the adjacent statements.

```suggestion
            // Act: Subscribe and wait for the sync callback to be captured.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+            assertNotNull("AppSync.sync was never invoked", capturedOnFailure.get());
+
+            // Dispose (models DataStore.stop()), then deliver the failure.
+            testObserver.dispose();
+            capturedOnFailure.get().accept(
+                    new DataStoreException("Sync failed after stop.", "Expected in this race."));
+
+            // tryOnError drops it on the disposed emitter.
+            assertTrue("Expected no undeliverable errors, but got: " + undelivered, undelivered.isEmpty());
+            // Contained: the disposed subscriber gets no error.
```
cadivus: Suggestion: Use an assertion phase label

Contained is unclear; label the assertion and state the expected subscriber behavior.

```suggestion
            // Assert: The disposed subscriber receives no error.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
```diff
@@ -883,6 +890,99 @@ public void run() {
...
+            // Dispose, then deliver a no-data (errors-only) response.
+            testObserver.dispose();
+            GraphQLResponse<PaginatedResult<ModelWithMetadata<BlogOwner>>> noDataResponse =
+                    new GraphQLResponse<>(null, Arrays.asList(
+                            new GraphQLResponse.Error("Errors from AppSync after stop.", null, null, null)));
+            capturedOnResponse.get().accept(noDataResponse);
+
+            // tryOnError drops it on the disposed emitter.
+            assertTrue("Expected no undeliverable errors, but got: " + undelivered, undelivered.isEmpty());
+            // Contained: the disposed subscriber gets no error.
```
cadivus: Suggestion: Use an assertion phase label

Contained is unclear; label the assertion and state the expected subscriber behavior.

```suggestion
            // Assert: The disposed subscriber receives no error.
```

aws-datastore/src/test/java/com/amplifyframework/datastore/syncengine/SyncProcessorTest.java
cadivus: Ah yeah, for sure: https://github.com/aws-amplify/amplify-android/blob/main/AGENTS.md#9-testing-patterns

VarshithaPamisetty: Yes wrote the tests in Kotlin. Thanks for pointing this out.

cadivus (approved):

PR #3434, opened 2026-09-09
feat(api): lazy loading for related models

mattcreaser:

### Description

Fifth in the AppSync client stack, on top of #3425. Lazy loading, in two commits.

**Related lists and pages that arrive complete.** A response carrying one had no adapter registered, so the client could not read it. A full list becomes a `LoadedModelList`; a page becomes a `ModelPage` with an opaque token, absent on the final page — which is what terminates a caller looping on `hasNextPage`.

**A reference that arrives as just its primary key.** It can now be fetched on access. The implementation satisfies the core lazy contract, so a generated model's field accepts it unchanged: those fields are typed as the interface, and the only implementation codegen ever constructs is the loaded one. Nothing generated is touched. The follow-up query comes from the existing key-map request factory rather than being assembled here.

Three behaviours are load-bearing rather than incidental:

- An empty key map means there is no related model, so it resolves to null without issuing a request that could not succeed.
- The cached value is wrapped, because a bare nullable reference cannot distinguish *loaded, and the value is null* from *not loaded yet*.
- Concurrent access serializes on a mutex that re-reads the cache inside the lock, so ten callers arriving together issue one query.

### The exception wrapping, and why

The contract being implemented declares a checked exception from a **different hierarchy** than the client's own — `core`'s `AmplifyException` versus the `foundation` one this client is built on. Both are checked. Throwing the client's type directly would be an undeclared checked exception: legal on the JVM, but invisible to a Java caller's `catch (AmplifyException e)`, and the compiler would never have forced them to handle it.

So failures are wrapped, with the typed exception carried as `cause` and its recovery suggestion preserved. The KDoc states that the cause is always an `AppSyncException`, so a Kotlin caller can still match on the category — which matters here, since a lazy fetch failing on authorisation is exactly a re-authenticate case.

This is worth flagging beyond this PR: the same signature sits on `LazyModelList.fetchPage`, and every V2-era `core` interface the client implements will impose `core`'s exception type the same way. The durable fix is foundation equivalents for these interfaces, which is a spec decision rather than a PR one.

### Known gap

**Lazy lists are implemented and tested but not yet reachable from a response.** A response leaves such a field null, and filling it in needs a post-deserialization pass that derives the foreign-key map from the child schema's associations. That pass is not here, so the type is currently reachable only from tests. The class carries a TODO saying so. The reference path is fully wired.

Flagging it rather than burying it: if you would rather this PR carried only what a response can actually produce, I can drop the list implementation and land it together with its wiring.

### Other notes

Locating the generated model provider gives `AppSyncInvalidConfigException` its first real throw sites — it previously had none. The schema lookup is lazy, so a consumer with no lazy relationships never triggers the class-path search.

The Gson instance is no longer a singleton, because it needs the per-client seam that reaches the query path; the response deserializer stops being one for the same reason. Both take their collaborator with no default, since a default would silently produce a client whose lazy relationships cannot load.

### Testing

266 unit tests, 0 failures — up from 221 at the base. `apiCheck`, `ktlintCheck` and `compileReleaseKotlin` pass; no public API change, as everything added is `internal`.

Two claims were mutation-tested rather than trusted: removing the `return` after a cache-hit callback makes the single-delivery test fail, and removing the re-check inside the lock makes the single-query test fail. Both were reintroduced deliberately, observed failing, and reverted.

Adapter registration is exercised through the client's own Gson instance rather than by constructing adapters directly — one written correctly but never registered would pass any test that instantiated it itself.

### Stack

```
feat/appsync-client
  └─ #3412 multi-auth
       └─ #3419 subscriptions
            └─ #3425 rate limits + logging
                 └─ this PR
```

### DIFF
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -53,6 +53,9 @@ internal class AppSyncGson(
                 AppSyncModelListDeserializer.register(it)
                 AppSyncModelPageDeserializer.register(it)
                 AppSyncModelReferenceDeserializer.register(it, loader, schemaRegistry)
+                // Registered after the adapters above so that it wraps whichever of them reads a model:
+                // they do the reading, and this fills in the relationship fields they left null.
+                AppSyncRelationshipTypeAdapterFactory.register(it, loader, schemaRegistry)
             }
             // A mutation that clears a field needs an explicit `"field": null` in the payload, because
             // AppSync reads an absent field as "leave unchanged" rather than "set to null".
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -30,7 +30,8 @@ import com.google.gson.GsonBuilder
  * Private to the client rather than shared, so neither the adapter set nor the null handling below can
  * be altered from outside.
  *
- * TODO: register deserializers for lazily-loaded model lists and pages, which this set does not cover.
+ * Related lists and pages that arrive in full are covered. TODO: a lazily-loaded ModelReference still
+ * is not — it needs a way to issue the follow-up query, which this instance has no handle on.
  */
 internal object AppSyncGson {
 
@@ -44,6 +45,8 @@ internal object AppSyncGson {
                 ModelWithMetadataAdapter.register(it)
                 SerializedModelAdapter.register(it)
                 SerializedCustomTypeAdapter.register(it)
+                AppSyncModelListDeserializer.register(it)
+                AppSyncModelPageDeserializer.register(it)
             }
             // A mutation that clears a field needs an explicit `"field": null` in the payload, because
             // AppSync reads an absent field as "leave unchanged" rather than "set to null".
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
@@ -28,14 +28,19 @@ import com.google.gson.GsonBuilder
  * The Gson instance the client serializes requests and deserializes responses with.
  *
  * Private to the client rather than shared, so neither the adapter set nor the null handling below can
- * be altered from outside.
+ * be altered from outside. One instance per client, because a lazily-loaded relationship needs to issue
+ * its own request and so the adapters have to carry that client's [AppSyncModelLoader].
  *
- * Related lists and pages that arrive in full are covered. TODO: a lazily-loaded ModelReference still
- * is not — it needs a way to issue the follow-up query, which this instance has no handle on.
+ * [gson] is built on first use, which is what lets a client hand in a loader that routes back through
+ * the client itself: the loader is captured while the client is still being constructed, but is not
+ * invoked until a relationship is actually loaded.
  */
-internal object AppSyncGson {
+internal class AppSyncGson(
+    private val loader: AppSyncModelLoader,
+    private val schemaRegistry: AppSyncSchemaRegistry = AppSyncSchemaRegistry()
+) {
 
-    val instance: Gson by lazy {
+    val gson: Gson by lazy {
         GsonBuilder()
             .also {
                 GsonTemporalAdapters.register(it)
@@ -47,6 +52,7 @@ internal object AppSyncGson {
                 SerializedCustomTypeAdapter.register(it)
                 AppSyncModelListDeserializer.register(it)
                 AppSyncModelPageDeserializer.register(it)
+                AppSyncModelReferenceDeserializer.register(it, loader, schemaRegistry)
             }
             // A mutation that clears a field needs an explicit `"field": null` in the payload, because
             // AppSync reads an absent field as "leave unchanged" rather than "set to null".
--- a/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
+++ b/aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
@@ -14,16 +14,90 @@
  */
 package com.amazonaws.appsync
 
+import com.amplifyframework.core.model.LoadedModelReferenceImpl
 import com.amplifyframework.core.model.Model
 import com.amplifyframework.core.model.ModelList
 import com.amplifyframework.core.model.ModelPage
+import com.amplifyframework.core.model.ModelReference
 import com.google.gson.GsonBuilder
 import com.google.gson.JsonDeserializationContext
 import com.google.gson.JsonDeserializer
 import com.google.gson.JsonElement
 import com.google.gson.JsonObject
+import com.google.gson.JsonPrimitive
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
+
+/**
+ * Deserializes a related model, which may or may not have arrived with the response.
+ *
+ * A selection set that requested the related model's fields yields a
+ * [com.amplifyframework.core.model.LoadedModelReference] holding it. One that requested only its
+ * primary key yields a reference that fetches the model when it is first accessed.
+ *
+ * A relationship the parent does not have is never seen here: a JSON null becomes a null field before
+ * a deserializer is consulted.
+ */
+internal class AppSyncModelReferenceDeserializer<M : Model>(
+    private val loader: AppSyncModelLoader,
+    private val schemaRegistry: AppSyncSchemaRegistry
+) : JsonDeserializer<ModelReference<M>> {
+
+    override fun deserialize(json: JsonElement, typeOfT: Type, context: JsonDeserializationContext): ModelReference<M> {
+        val parameterized = typeOfT as? ParameterizedType
+            ?: throw AppSyncDeserializationException(
+                message = "A related model was requested as ${typeOfT.typeName}, which carries no model type.",
+                recoverySuggestion = "Request the relationship as ModelReference<T> so the model type is known."
+            )
+
+        @Suppress("UNCHECKED_CAST")
+        val modelClass = parameterized.actualTypeArguments.first() as Class<M>
+        val jsonObject = json.asJsonObjectOrThrow()
+        val keyFields = schemaRegistry.schemaFor(modelClass).primaryIndexFields
+
+        // More fields than the key means the selection set asked for the model itself, so it is already
+        // here and there is nothing to defer. A failure to read it is not fatal: the key is still
+        // present, so the model can be fetched instead.
+        if (jsonObject.size() > keyFields.size) {
+            runCatching { context.deserialize<M>(json, modelClass) }
+                .onSuccess { return LoadedModelReferenceImpl(it) }
+        }
+
+        return AppSyncLazyModelReference(modelClass, jsonObject.keyValues(keyFields), loader)
+    }
+
+    /**
+     * Reads the primary key values, which are what identify the model in the follow-up query.
+     *
+     * A key that is incomplete identifies nothing, so it is reported as no key at all — the reference
+     * then resolves to null rather than issuing a request that cannot succeed.
+     */
+    private fun JsonObject.keyValues(keyFields: List<String>): Map<String, Any> {
+        val values = keyFields.mapNotNull { field ->
+            (get(field) as? JsonPrimitive)?.let { field to it.keyValue() }
+        }.toMap()
+        return if (values.size == keyFields.size) values else emptyMap()
+    }
+
+    /**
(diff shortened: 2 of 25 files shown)


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
```diff
@@ -0,0 +1,103 @@
...
+    val elementType = parameterized.actualTypeArguments.first()
+
+    val items = json.asJsonObjectOrThrow().get(ITEMS_KEY)
+        ?.takeIf { it.isJsonArray }
+        ?: throw AppSyncDeserializationException(
+            message = "A related list was expected to carry an \"$ITEMS_KEY\" array, but did not.",
+            recoverySuggestion = "Verify the selection set requests the list's items."
+        )
+
+    return items.asJsonArray.map { context.deserialize(it.asJsonObject, elementType) }
```
mattcreaser: [WARNING] Non-object list entries surface as unknown failures

Current code:
```
    return items.asJsonArray.map { context.deserialize(it.asJsonObject, elementType) }
```

asJsonObject throws IllegalStateException for a null or scalar items entry. AppSyncResponseDeserializer.deserialize only catches JsonParseException, so that escapes and AppSyncException.from maps it to AppSyncUnknownException instead of the typed AppSyncDeserializationException this file uses for every other shape problem. No test covers a non-object item.

Suggested change:
```suggestion
    return items.asJsonArray.mapIndexed { index, item ->
        if (!item.isJsonObject) {
            throw AppSyncDeserializationException(
                message = "A related list item at index $index was expected to be a JSON object."
            )
        }
        context.deserialize(item, elementType)
    }
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncGson.kt
```diff
@@ -30,7 +30,8 @@ import com.google.gson.GsonBuilder
  * Private to the client rather than shared, so neither the adapter set nor the null handling below can
  * be altered from outside.
  *
- * TODO: register deserializers for lazily-loaded model lists and pages, which this set does not cover.
+ * Related lists and pages that arrive in full are covered. TODO: a lazily-loaded ModelReference still
+ * is not — it needs a way to issue the follow-up query, which this instance has no handle on.
```
mattcreaser: [SUGGESTION] Keep the TODO to the missing dependency

Current code:
```
 * Related lists and pages that arrive in full are covered. TODO: a lazily-loaded ModelReference still
 * is not — it needs a way to issue the follow-up query, which this instance has no handle on.
```

The first sentence repeats the adapter registrations below. The TODO can state the missing client handle directly.

Suggested change:
```suggestion
 * TODO: Give lazy ModelReference deserialization access to the client so it can issue follow-up queries.
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.google.gson.JsonObject
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+/**
+ * Deserializes a related list that arrived in full, as `{"items": [...]}`.
+ *
+ * Produces a [com.amplifyframework.core.model.LoadedModelList], never a lazy one: everything the
+ * caller asked for is already in the payload, so there is nothing to defer.
+ */
```
mattcreaser: [SUGGESTION] Collapse the loaded-list contract

Current code:
```
/**
 * Deserializes a related list that arrived in full, as `{"items": [...]}`.
 *
 * Produces a [com.amplifyframework.core.model.LoadedModelList], never a lazy one: everything the
 * caller asked for is already in the payload, so there is nothing to defer.
 */
```

The second paragraph restates the meaning of LoadedModelList and why a complete payload needs no deferral.

Suggested change:
```suggestion
/** Deserializes a complete related list into a [com.amplifyframework.core.model.LoadedModelList]. */
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
```diff
@@ -0,0 +1,103 @@
...
+internal class AppSyncModelPageDeserializer<M : Model> : JsonDeserializer<ModelPage<M>> {
+    override fun deserialize(json: JsonElement, typeOfT: Type, context: JsonDeserializationContext) = AppSyncModelPage(
+        items = deserializeItems<M>(json, typeOfT, context),
+        nextToken = deserializeNextToken(json)
+    )
+
+    companion object {
+        fun register(builder: GsonBuilder) {
+            // A hierarchy adapter, unlike the list above: a response type may be a subtype of ModelPage,
+            // and an exact-type adapter would not be consulted for one.
```
mattcreaser: [SUGGESTION] State the hierarchy-adapter reason in one sentence

Current code:
```
            // A hierarchy adapter, unlike the list above: a response type may be a subtype of ModelPage,
            // and an exact-type adapter would not be consulted for one.
```

The subtype constraint is the useful part; the contrast with the list adapter and the follow-up clause are not.

Suggested change:
```suggestion
            // Use a hierarchy adapter because response types may be ModelPage subtypes.
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncLazyDeserializers.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+}
+
+/**
+ * Reads the `items` array, deserializing each entry as the list's element type.
+ *
+ * The element type comes from the requested type's parameter, so the caller's `ModelList<Todo>` is what
+ * decides how each item is read.
+ */
```
mattcreaser: [SUGGESTION] Trim the item-deserialization doc

Current code:
```
/**
 * Reads the `items` array, deserializing each entry as the list's element type.
 *
 * The element type comes from the requested type's parameter, so the caller's `ModelList<Todo>` is what
 * decides how each item is read.
 */
```

Both paragraphs narrate a private function whose ParameterizedType lookup is immediately visible.

Suggested change:
```suggestion
/** Deserializes each item using the requested list element type. */
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncModelListTypes.kt
```diff
@@ -0,0 +1,44 @@
...
+package com.amazonaws.appsync
+
+import com.amplifyframework.core.model.LoadedModelList
+import com.amplifyframework.core.model.Model
+import com.amplifyframework.core.model.ModelPage
+import com.amplifyframework.core.model.PaginationToken
+
+/**
+ * A list of models that arrived complete in a response, so there is nothing left to fetch.
+ */
```
mattcreaser: [SUGGESTION] Drop the self-describing comment

Current code:
```
/**
 * A list of models that arrived complete in a response, so there is nothing left to fetch.
 */
```

The class name and the LoadedModelList supertype already say the list arrived complete.

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncModelListTypes.kt
```diff
@@ -0,0 +1,44 @@
...
+internal class AppSyncLoadedModelList<out M : Model>(override val items: List<M>) : LoadedModelList<M>
+
+/**
+ * One page of models, and the token for the next page when the service reported one.
+ *
+ * [ModelPage.hasNextPage] is derived from [nextToken], so a null token is what tells a caller the
+ * pagination is finished.
+ */
```
mattcreaser: [SUGGESTION] Keep only the derived pagination contract

Current code:
```
/**
 * One page of models, and the token for the next page when the service reported one.
 *
 * [ModelPage.hasNextPage] is derived from [nextToken], so a null token is what tells a caller the
 * pagination is finished.
 */
```

The opening sentence repeats the two properties; the non-obvious fact is how a null nextToken drives hasNextPage.

Suggested change:
```suggestion
/** A null [nextToken] makes [ModelPage.hasNextPage] false. */
```

aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncModelListTypes.kt
```diff
@@ -0,0 +1,44 @@
...
+    override val items: List<M>,
+    override val nextToken: AppSyncPaginationToken?
+) : ModelPage<M>
+
+/**
+ * An opaque cursor into a paginated result.
+ *
+ * [PaginationToken] carries no members: the value is meaningful only to AppSync, and a caller is
+ * expected to hand it back rather than interpret it.
+ */
```
mattcreaser: [SUGGESTION] Condense the token contract

Current code:
```
/**
 * An opaque cursor into a paginated result.
 *
 * [PaginationToken] carries no members: the value is meaningful only to AppSync, and a caller is
 * expected to hand it back rather than interpret it.
 */
```

The marker-interface detail does not help callers. The opaque and hand-back constraints fit one sentence.

Suggested change:
```suggestion
/** An opaque AppSync cursor that callers must return unchanged. */
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncModelListDeserializerTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+import io.kotest.matchers.types.shouldBeInstanceOf
+import org.junit.Test
+
+/**
+ * Tests the related-list and page deserializers, through [AppSyncGson] rather than in isolation, so the
+ * registration is covered too — an adapter that is written correctly but never registered would pass any
+ * test that instantiated it directly.
+ */
```
mattcreaser: [SUGGESTION] Shorten the test-scope doc

Current code:
```
/**
 * Tests the related-list and page deserializers, through [AppSyncGson] rather than in isolation, so the
 * registration is covered too — an adapter that is written correctly but never registered would pass any
 * test that instantiated it directly.
 */
```

The final clause elaborates a failure mode already implied by testing through AppSyncGson.

Suggested change:
```suggestion
/** Verifies related-list and page adapter registration through [AppSyncGson]. */
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncModelListDeserializerTest.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    @Test
+    fun `a final page has no next token, which is what ends pagination`() {
+        val page: ModelPage<Todo> = gson.fromJson(
+            """{"items":[{"id":"1","name":"first"}]}""",
+            object : TypeToken<ModelPage<Todo>>() {}.type
+        )
+
+        page.nextToken.shouldBeNull()
+        // hasNextPage is derived, so a caller looping on it terminates without inspecting the token.
```
mattcreaser: [SUGGESTION] Remove narration before the assertion

Current code:
```
        // hasNextPage is derived, so a caller looping on it terminates without inspecting the token.
```

The adjacent null-token and false hasNextPage assertions already express this relationship.

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncModelListDeserializerTest.kt
```diff
@@ -0,0 +1,127 @@
...
+            )
+        }
+
+        error.message shouldContain "items"
+    }
+
+    @Test
+    fun `a list requested without an element type is reported rather than silently empty`() {
+        // The element type is what tells the deserializer how to read each item, so a raw ModelList
+        // cannot be honoured — failing loudly beats returning items deserialized as the wrong type.
```
mattcreaser: [SUGGESTION] Shorten the raw-list rationale

Current code:
```
        // The element type is what tells the deserializer how to read each item, so a raw ModelList
        // cannot be honoured — failing loudly beats returning items deserialized as the wrong type.
```

The exception assertion already establishes the failure policy; only the reason a raw list is invalid needs stating.

Suggested change:
```suggestion
        // A raw ModelList lacks the element type required to deserialize its items.
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncModelListDeserializerTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        val error = shouldThrow<AppSyncDeserializationException> {
+            gson.fromJson<ModelList<*>>("""{"items":[]}""", ModelList::class.java)
+        }
+
+        error.message shouldContain "element type"
+    }
+
+    /** A minimal model: these tests exercise the list wrapper, not model deserialization. */
```
mattcreaser: [SUGGESTION] Drop the obvious fixture comment

Current code:
```
    /** A minimal model: these tests exercise the list wrapper, not model deserialization. */
```

A two-field private model inside a list-deserializer test already reads as a minimal fixture.

PR #3437, opened 2026-09-10
test: make TargetingClient endpoint update tests deterministic

mattcreaser:
`TargetingClientTest.testUpdateEndpointProfileOptsIn` failed on `main` in
[run 34491135422](https://github.com/aws-amplify/amplify-android/actions/runs/34491135422)
with `Verification failed: ... updateEndpoint(matcher<UpdateEndpointRequest>(), any())) was not called`,
then passed on a re-run of the identical commit with no code change.

## Cause

`TargetingClient.executeUpdate` dispatches the SDK call asynchronously:

```kotlin
coroutineScope.launch {
    pinpointClient.updateEndpoint(updateEndpointRequest)
}
```

`coroutineScope` is built from the constructor's `coroutineDispatcher` parameter, which defaults to
`Dispatchers.Default`. `constructTargetingClient()` never supplied one, so the coroutine ran on a
real background thread pool. `runTest` only controls coroutines on the *test* dispatcher, so it
neither advanced nor awaited that work.

Three tests call `updateEndpointProfile()` and then `coVerify` the call with no synchronisation
between them, so they passed only when the background coroutine happened to win the race:

- `testUpdateEndpointProfile`
- `testUpdateEndpointProfileOptsIn`
- `testUpdateEndpointProfileOptOutNotTouched`

## Fix

Supply a dispatcher the test controls. `UnconfinedTestDispatcher` runs the launched coroutine
eagerly on the calling thread, so the SDK call has happened by the time `updateEndpointProfile()`
returns — deterministic by construction rather than by timing.

This is the pattern `AnalyticsClientTest` in this same module already uses for the same reason;
`TargetingClientTest` had simply never adopted it. No production code changes: the constructor
already accepted the parameter.

## Verification, and its limits

- `TargetingClientTest` passes; 8 consecutive `--rerun-tasks` runs all green.
- **The stress run does not prove the fix.** The unpatched code also passed 8/8 locally, so an idle
  workstation cannot reproduce a race that surfaces on a loaded 4-vCPU CI runner. The justification
  is the mechanism plus the observed CI failure, not a local reproduction.
- `ktlintCheck` clean.

### DIFF
--- a/aws-pinpoint-core/src/test/java/com/amplifyframework/pinpoint/core/ConstructTargetingClasses.kt
+++ b/aws-pinpoint-core/src/test/java/com/amplifyframework/pinpoint/core/ConstructTargetingClasses.kt
@@ -29,6 +29,9 @@ import io.mockk.every
 import io.mockk.mockk
 import io.mockk.mockkStatic
 import java.util.Locale
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
 
 internal fun constructSharedPreferences(): SharedPreferences =
     ApplicationProvider.getApplicationContext<Context>().getSharedPreferences(
@@ -80,7 +83,13 @@ internal fun constructPinpointClient(): PinpointClient {
     return pinpointClient
 }
 
-internal fun constructTargetingClient(): TargetingClient {
+// TargetingClient dispatches its updateEndpoint call with coroutineScope.launch, so a caller that
+// verifies the call has to supply a dispatcher the test controls. UnconfinedTestDispatcher runs the
+// launched coroutine eagerly, so the call has happened by the time the client method returns.
+@OptIn(ExperimentalCoroutinesApi::class)
+internal fun constructTargetingClient(
+    coroutineDispatcher: CoroutineDispatcher = UnconfinedTestDispatcher()
+): TargetingClient {
     setup()
     val prefs = constructSharedPreferences()
     return TargetingClient(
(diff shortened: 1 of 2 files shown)


thisisabhash (dismissed):

cadivus (dismissed):

aws-pinpoint-core/src/test/java/com/amplifyframework/pinpoint/core/ConstructTargetingClasses.kt
```diff
@@ -80,7 +83,13 @@ internal fun constructPinpointClient(): PinpointClient {
     return pinpointClient
 }
 
-internal fun constructTargetingClient(): TargetingClient {
+// TargetingClient dispatches its updateEndpoint call with coroutineScope.launch, so a caller that
+// verifies the call has to supply a dispatcher the test controls. UnconfinedTestDispatcher runs the
+// launched coroutine eagerly, so the call has happened by the time the client method returns.
```
mattcreaser: Suggestion: Collapse the dispatcher rationale to one sentence

The first two sentences narrate the launch call and the new dispatcher parameter, which a reader can see directly. Only the eager-execution guarantee explains why the helper takes a dispatcher, so one sentence keeps the useful rationale.

```suggestion
// UnconfinedTestDispatcher lets tests verify updateEndpoint before updateEndpointProfile returns.
```

cadivus: Comment contradicts the default dispatcher

The helper already defaults to UnconfinedTestDispatcher, so callers do not have to supply a dispatcher. The long explanation also obscures the useful race-prevention rationale.

```suggestion
// Run launched updateEndpoint calls eagerly so they complete before verification.
```

cadivus (approved):

PR #3439, opened 2026-09-11
chore: add custom Android Lint rule infrastructure

mattcreaser:

*Issue #, if available:* N/A

*Description of changes:*

Adds the infrastructure for repo-specific Android Lint rules, so that a future rule requires writing only a detector and its test — no build wiring. Ships one real rule to prove the pipeline end to end.

**New `:lint-rules` module** holds detectors plus an `IssueRegistry`, which Lint discovers via a `Lint-Registry-v2` jar manifest attribute (rather than `@AutoService`, which would pull KSP in for one annotation). The module is deliberately unpublished and stays out of the API and coverage surfaces by not applying `amplify.publishing`, `amplify.api`, or `amplify.kover`.

**New `amplify.lint` convention plugin** makes Lint's configuration uniform across every module type and attaches the checks via `lintChecks`. This also fixes three pre-existing inconsistencies:

| Module type | Before | After |
|---|---|---|
| Android library | Lint configured inline in `AndroidLibraryConventionPlugin` | shared config |
| KMP (`foundation`, `foundation-bridge`) | bare `com.android.lint`, **no** config — no `lint.xml`, no `warningsAsErrors` | shared config |
| Kotlin-JVM (`annotations`, `apollo-appsync`, `aws-sdk-appsync-core`) | **no Lint at all** | shared config |

Only two convention plugins apply it: `KotlinConventionPlugin` and `KotlinMultiplatformConventionPlugin`. Android library modules inherit it because `amplify.android.library` applies `amplify.kotlin`.

**The rule — `AmplifyPrintln`** (ERROR): reports `print`/`println` that resolve to `kotlin.io.ConsoleKt` or a `java.io.PrintStream` subclass. It resolves types rather than matching names, so a user-defined function called `println` is not flagged.

### Reviewer notes — deliberate decisions you may want to question

**The rule found a real defect on its first run.** `AppSyncProtocol.kt` had a stray debug `println` in a subscription-error path in a published library. This PR deletes it. The failure is still reported through `listener.operationError(subscriptionId, null)` on the following line, so no error *signal* is lost — but the exception *detail* is, because that call passes a `null` payload. Fixing that properly means either giving `apollo-appsync` a logging facility (it has none; its only dependency is `apollo-runtime`) or changing what `operationError` reports to consumers. Both are product decisions outside lint plumbing, so this is left as a **follow-up**.

**`JavaLogger` is suppressed, not fixed.** It writes to stdout by design — that is the whole purpose of that logging plugin — so it carries a method-scoped `@SuppressLint("AmplifyPrintln")` with a comment. Method-scoped rather than line-scoped means a future print added to that same 14-line method would be silently allowed; accepted as a reasonable trade for discoverability.

**Two behaviour changes worth knowing about:**
1. `warningsAsErrors = true` and `abortOnError = true` now reach the KMP modules and the three Kotlin-JVM modules for the first time. A future AGP bump can therefore introduce a built-in check that breaks modules which have never had a Lint gate. `AGENTS.md` documents the escape hatch: a per-issue severity override in the root `lint.xml`, or a Lint baseline — **not** turning `warningsAsErrors` off.
2. `check` (and so `build`) in `annotations`, `apollo-appsync`, and `aws-sdk-appsync-core` now depends on `lint`, which depends on `:lint-rules:jar`. Local builds in those modules get slightly slower and can now fail on Lint.

**Rule scope, to pre-empt "it missed X":** `AmplifyPrintln` intentionally covers only `print`/`println` on `kotlin.io.ConsoleKt` and `PrintStream` subclasses. `System.out.printf`, `PrintWriter.println` (not a `PrintStream` subclass), and `printStackTrace` are *not* caught.

**Caching:** every module's `lintAnalyze*` task now takes the `:lint-rules` jar as a classpath input, so editing a detector invalidates Lint analysis repo-wide. That only triggers on detector changes, not ordinary source edits.

**Version coupling is enforced, not documented.** Lint's version tracks AGP's with a major offset of 23 (AGP 8.11.1 → Lint 31.11.1). This does not fail loudly on its own: because `lint-api` is `compileOnly` and the runtime copy comes from AGP, bumping `agp` and forgetting `lint` leaves the registry compiled against an older Lint, which Lint reports as "usually fine". A `check()` in `lint-rules/build.gradle.kts` derives the expected version from `agp` and fails configuration on a mismatch.

- **9 unit tests** for the detector via `TestLintTask`. Coverage includes both flagged forms in both languages, three clean cases (user-defined `println`, Java `@SuppressLint`, Kotlin `@Suppress`), a test-source case, and a registry-membership assertion. One test uses a golden `expect()` locking the message, line, and column; the rest use `expectErrorCount` to stay robust across Lint versions.
- **Mutation-tested the load-bearing assertions.** Forcing the detector's type check to always pass fails exactly the user-defined-`println` test, confirming that test proves type resolution rather than name matching. Deleting the `isTestSource` guard fails exactly the test-source test.
- **Proved the rule fires through Gradle on all three module types**, which the unit tests structurally cannot do since they run the detector in-process. A temporary `println` was added to `annotations` (Kotlin-JVM), `foundation` (KMP `commonMain`), and confirmed against the pre-existing violation in `core` (Android library); each failed with `[AmplifyPrintln from com.amplifyframework:lint-rules]`, then was reverted.
- `./gradlew lint --continue` green across all 35 modules; `ktlintCheck`, `checkstyle`, `apiCheck`, `:apollo-appsync:test` all green.
- Verified `:lint-rules` has no `apiCheck` task and appears in no published POM.

CI note: `:lint-rules:test` is added to the static-analysis job, because `koverXmlReport` is the only test execution in CI and Kover is applied solely by `PublishingConventionPlugin` — so an unpublished module's tests would otherwise never run. The workflow's own comment predicted exactly this case.

- [x] No

*General Checklist*
- [x] Added Unit Tests
- [ ] Added Integration Tests

### DIFF
--- a/lint-rules/src/main/java/com/amplifyframework/lint/NoPrintlnDetector.kt
+++ b/lint-rules/src/main/java/com/amplifyframework/lint/NoPrintlnDetector.kt
@@ -34,6 +34,9 @@ class NoPrintlnDetector : Detector(), SourceCodeScanner {
     override fun getApplicableMethodNames() = listOf("print", "println")
 
     override fun visitMethodCall(context: JavaContext, node: UCallExpression, method: PsiMethod) {
+        // Tests may print freely; this rule is about library code.
+        if (context.isTestSource) return
+
         val evaluator = context.evaluator
         val printsToConsole = evaluator.isMemberInClass(method, KOTLIN_CONSOLE_CLASS) ||
             evaluator.isMemberInSubClassOf(method, PRINT_STREAM_CLASS, false)
@@ -54,9 +57,11 @@ class NoPrintlnDetector : Detector(), SourceCodeScanner {
         val ISSUE: Issue = Issue.create(
             id = "AmplifyPrintln",
             briefDescription = "Console printing in library code",
+            // The trailing backslashes are Lint markup meaning "join with the next line". Without
+            // them Lint treats each source newline as a hard break and wraps the text mid-sentence.
             explanation = """
-                Writing to stdout or stderr bypasses the logging configuration, so consumers of
-                the SDK cannot filter, redirect, or disable the output. Emit the message through a
+                Writing to stdout or stderr bypasses the logging configuration, so consumers of \
+                the SDK cannot filter, redirect, or disable the output. Emit the message through a \
                 `Logger` instead, which routes it to the configured logging plugin.
             """,
             category = Category.CORRECTNESS,
--- a/lint-rules/src/main/java/com/amplifyframework/lint/NoPrintlnDetector.kt
+++ b/lint-rules/src/main/java/com/amplifyframework/lint/NoPrintlnDetector.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amplifyframework.lint
+
+import com.android.tools.lint.detector.api.Category
+import com.android.tools.lint.detector.api.Detector
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import com.intellij.psi.PsiMethod
+import org.jetbrains.uast.UCallExpression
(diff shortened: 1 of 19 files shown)


lint-rules/src/main/java/com/amplifyframework/lint/NoPrintlnDetector.kt
```diff
@@ -0,0 +1,73 @@
...
+    // Lint consults its method name index for these, rather than visiting every call in the source.
+    override fun getApplicableMethodNames() = listOf("print", "println")
+
+    override fun visitMethodCall(context: JavaContext, node: UCallExpression, method: PsiMethod) {
+        // Tests may print freely; this rule is about library code.
+        if (context.isTestSource) return
+
+        val evaluator = context.evaluator
+        val printsToConsole = evaluator.isMemberInClass(method, KOTLIN_CONSOLE_CLASS) ||
+            evaluator.isMemberInSubClassOf(method, PRINT_STREAM_CLASS, false)
```
mattcreaser: Suggestion: No test covers the documented PrintStream subclass behavior

isMemberInSubClassOf makes subclass print/println calls part of the detector contract, but every current test invokes a method declared directly on java.io.PrintStream or on kotlin.io.ConsoleKt, so this subtype branch is untested. Consider adding a fixture that extends PrintStream, overrides println(String), invokes it from production source, and asserts one AmplifyPrintln error, so the documented behavior cannot regress unnoticed.

harsh62 (approved):

PR #3441, opened 2026-09-14
fix(events): own the shared WebSocket connection attempt in the provider

### LINKED ISSUE #3424: Events: cancelling one subscriber can silently kill a shared WebSocket connection for others
### Summary

In `aws-sdk-appsync-events`, the shared WebSocket connection attempt is launched in the coroutine of whichever caller happens to arrive first. If that caller is cancelled while others are waiting on the same attempt, they are cancelled too — and because the failure arrives as a `CancellationException`, it is indistinguishable from those callers being cancelled themselves. No error surfaces.

I found this while writing a provider with the same shape elsewhere, and traced the pattern back here.

### Where

`appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt`

```kotlin
private suspend fun getConnectedWebSocketResult(): Result<EventsWebSocket> = coroutineScope {
    ...
    val newDeferredInProgressConnection = async { attemptConnection() }
    connectionInProgressReference.set(newDeferredInProgressConnection)
    val connectionResult = newDeferredInProgressConnection.await()
    ...
}
```

### Mechanism

`async { }` inherits the `coroutineScope { }` of the first caller, so the `Deferred` is that caller's child. It is then published to `connectionInProgressReference`, where later callers join it:

```kotlin
val deferredInProgressConnection = connectionInProgressReference.get()
if (deferredInProgressConnection != null && !deferredInProgressConnection.isCompleted) {
    return@coroutineScope deferredInProgressConnection.await()
}
```

So when the first caller is cancelled, its scope cancels the `async` child, and every joined caller's `await()` throws `CancellationException`. Structured concurrency then treats that as *the joiner* being cancelled, so the joiner stops without raising anything a caller can catch. The connection they were waiting for simply never arrives, and nothing explains why.

The attempt belongs to the provider, not to whichever caller reached it first — the lifetime is the thing that is wrong here, not the locking.

### Why this is reachable in normal use

`getConnectedWebSocket()` is called from inside `onStart` on the flow returned by `EventsWebSocketClient.subscribe(...)`:

```kotlin
return createSubscriptionEventDataFlow(subscriptionHolder)
    .onStart {
        val newWebSocket = eventsWebSocketProvider.getConnectedWebSocket()
        ...
    }
```

That places the connection attempt in the **collector's** coroutine, and a collector ending is entirely routine: `take(n)`, a timeout, `launchIn(viewModelScope)` on a screen that closes, or a `collect` inside a job that gets cancelled. Two subscriptions starting at roughly the same time are enough — if the first collector goes away during the handshake, the second silently gets nothing.

This is only a problem while an attempt is in flight, so it needs concurrent subscribers to bite. It is not a persistent broken state: a cancelled `Deferred` reports `isCompleted == true`, so the `!isCompleted` guard sends the *next* caller down the path of starting a fresh attempt. It is the callers already waiting who lose.

### Impact

- A subscription flow that stops before emitting, with no error delivered to the collector.
- Nothing distinguishes it from ordinary cancellation, so it is close to undebuggable from an app's perspective — there is no exception, no log, and the provider recovers by the next call.
- More likely under exactly the conditions apps create: several subscriptions started together during startup or a screen transition.

### Why the current tests do not catch it

`EventsWebSocketProviderTest.multiple calls return same instance when not closed` launches ten concurrent calls and `awaitAll()`s them. Nothing is ever cancelled, so the cross-caller cancellation path is never exercised. A concurrency test built on `awaitAll` structurally cannot catch this — every participant runs to completion by construction.

### Suggested fix

Give the provider its own scope so the attempt outlives any individual caller:

```kotlin
private val scope = CoroutineScope(ioDispatcher + SupervisorJob())
...
val newDeferredInProgressConnection = scope.async { attemptConnection() }
```

One wrinkle worth flagging rather than glossing over: `EventsWebSocketProvider` currently has no teardown, so a provider-owned scope needs somewhere to be cancelled — presumably from `EventsWebSocketClient.disconnect(...)`, which is the existing lifecycle boundary. That is the part of the change that needs a decision, not the `async` itself.

Adjacent, lower priority: `async` and its `await()` are both inside `mutex.withLock`, so the mutex is held for the whole network connect. The pre-lock `connectionInProgressReference` check covers the common case, but a caller arriving before that reference is published blocks on the mutex for the full connect rather than joining the attempt.

### Suggested regression test

Cancel the first caller specifically, rather than letting every participant finish:

```kotlin
@Test
fun `cancelling the first caller does not cancel the attempt others joined`() = runTest {
    val gate = CompletableDeferred<Unit>()
    every { anyConstructed<EventsWebSocket>().isClosed } returns false
    coEvery { anyConstructed<EventsWebSocket>().connect() } coAnswers { gate.await() }

    val first = launch { provider.getConnectedWebSocket() }
    val second = async { provider.getConnectedWebSocket() }
    runCurrent()

    first.cancel()
    gate.complete(Unit)

    // Fails before the fix: second is cancelled by first's cancellation.
    second.await()
    coVerify(exactly = 1) { anyConstructed<EventsWebSocket>().connect() }
}
```

### Environment

Observed by reading `main`; the code is unchanged since the module was added. Not reproduced against a live endpoint — the analysis is from the source and the call sites above.

mattcreaser:
## Issue

Fixes #3424.

## Problem

`EventsWebSocketProvider` launched the shared WebSocket connection attempt with `async` inside the `coroutineScope` of whichever caller reached it first, so the `Deferred` was that caller's child. Later callers joined the same attempt via `await()`.

When the first caller was cancelled — a routine event for a subscription collector (`take(n)`, a timeout, `launchIn` on a screen that closes, a cancelled enclosing job) — its scope cancelled the attempt, and every joined caller's `await()` threw `CancellationException`. Structured concurrency then treats that as *the joiner* being cancelled, so those subscriptions stopped with no error delivered to the collector: no exception, no log, nothing to debug. Two subscriptions starting at roughly the same time are enough to hit it.

## Fix

- Launch the connection attempt on a **provider-owned scope** (`SupervisorJob` + the IO dispatcher) instead of the caller's scope, so its lifetime is independent of any single caller. Cancelling one caller no longer cancels the attempt others joined.
- **Record the connection result from within the attempt** rather than from the triggering caller. Previously, if that caller was cancelled mid-`await()`, the successful websocket was never stored — future callers made a fresh connection and the connected socket was orphaned. The result is now stored before the attempt completes, so callers observing the attempt as completed also observe the stored result.
- **Await the attempt outside the mutex.** The lock is now held only long enough to decide which attempt to await, not for the whole network connect.

## Teardown decision

The issue flagged that a provider-owned scope needs somewhere to be cancelled. `EventsWebSocketClient.disconnect(...)` is the existing lifecycle boundary, so the scope is cancelled there (this also cancels an in-flight attempt whose socket would otherwise be orphaned by a disconnect). Because the client supports connecting again after `disconnect()`, the scope is recreated on the next connection attempt, so reconnect still works.

## Tests

- `cancelling the first caller does not cancel the attempt others joined` — the regression test from the issue; fails before the fix (the joiner is cancelled) and passes after.
- `close cancels in progress attempt and next call reconnects` — covers the teardown boundary and reconnect.
- Existing provider and client tests (including `disconnect with/without flush`) still pass; `ktlintCheck` and `apiCheck` are clean (no public API change — `close()` is `internal`).


appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt
```diff
@@ -40,55 +43,72 @@ internal class EventsWebSocketProvider(
...
+    /**
+     * Returns the established connection result if one is open, otherwise clears any closed connection and
+     * returns null. Must be called while holding [mutex] when a null result may lead to a new attempt.
+     */
+    private fun openConnectionResultOrNull(): Result<EventsWebSocket>? {
+        val existingResult = connectionResultReference.get()
+        val existingWebSocket = existingResult?.getOrNull() ?: return null
+        return if (existingWebSocket.isClosed) {
+            connectionResultReference.set(null)
+            null
```
mattcreaser: Stale closed-result read can erase a newer connection

This helper now runs on the fast path outside the mutex. One caller can read an old closed result, be suspended while another caller installs a fresh successful result, and then erase that fresh result with the unconditional set(null). A subsequent caller then starts a duplicate WebSocket even though an open socket exists. Use compareAndSet so only the exact stale result is cleared.

```suggestion
    private fun openConnectionResultOrNull(): Result<EventsWebSocket>? {
        val existingResult = connectionResultReference.get()
        val existingWebSocket = existingResult?.getOrNull() ?: return null
        return if (existingWebSocket.isClosed) {
            connectionResultReference.compareAndSet(existingResult, null)
            null
```

appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt
```diff
@@ -40,55 +43,72 @@ internal class EventsWebSocketProvider(
...
     }
 
+    /**
+     * Cancels any in-progress connection attempt and releases the provider's connection scope. A subsequent
+     * call to [getConnectedWebSocket] starts a fresh attempt on a new scope.
+     */
+    fun close() {
+        connectionScope.cancel()
+        connectionInProgressReference.set(null)
+    }
```
mattcreaser: close() is not serialized against starting and completing attempts

close() mutates the same state startConnection guards but does not take the mutex. A caller racing close can create a new scope after the old scope is cancelled and then have its Deferred reference cleared while the attempt is still running. A cancelled old attempt can also record its cancellation failure after a post-close attempt recorded success, discarding a usable socket and permitting another duplicate connection. Serialize close with the same mutex (or otherwise order it against attempt start and result publication).

appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt
```diff
@@ -40,55 +43,72 @@ internal class EventsWebSocketProvider(
     private val connectionResultReference = AtomicReference<Result<EventsWebSocket>?>(null)
     private val connectionInProgressReference = AtomicReference<Deferred<Result<EventsWebSocket>>?>(null)
 
+    // A connection attempt belongs to the provider, not to whichever caller happens to trigger it, so it is
+    // launched on this scope rather than the caller's. That way cancelling one caller cannot cancel an attempt
+    // that other callers are awaiting. The scope is cancelled by [close] and recreated for the next attempt.
```
mattcreaser: Suggestion: Condense the provider-scope rationale

Ownership, cancellation isolation, and recreation all fit in two sentences; the hypothetical-caller phrasing delays those facts.

```suggestion
    // Connection attempts use a provider-owned scope so cancelling one caller does not affect other awaiters.
    // [close] cancels the scope, which is recreated for the next attempt.
```

appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt
```diff
@@ -40,55 +43,72 @@ internal class EventsWebSocketProvider(
...
-            }
+    private suspend fun getConnectedWebSocketResult(): Result<EventsWebSocket> {
+        // If a connection is already established, return it.
+        openConnectionResultOrNull()?.let { return it }
+
+        // If an attempt is already in flight, join it without taking the lock or blocking it.
+        connectionInProgressReference.get()?.takeUnless { it.isCompleted }?.let { return it.await() }
+
+        // Otherwise resolve which attempt to await under the lock, then await it outside the lock so the
+        // network connect does not hold the mutex against other callers.
```
mattcreaser: Suggestion: Shorten the mutex rationale

The single constraint is: select under the mutex, await outside it. The transition and repeated lock wording obscure it.

```suggestion
        // Select an attempt under the mutex, then await outside it so connecting does not block other callers.
```

appsync/aws-sdk-appsync-events/src/main/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProvider.kt
```diff
@@ -40,55 +43,72 @@ internal class EventsWebSocketProvider(
...
+    }
+
+    private fun startConnection(): Deferred<Result<EventsWebSocket>> {
+        if (!connectionScope.isActive) {
+            connectionScope = newConnectionScope()
         }
+        return connectionScope.async {
+            // Record the result from within the attempt so it is stored even if the caller that triggered
+            // the attempt is cancelled while awaiting it. Completion happens after this runs, so callers that
+            // observe the attempt as completed also observe the stored result.
```
mattcreaser: Suggestion: Condense the result-publication explanation

Cancellation isolation and publication ordering both matter, but the repeated references to the attempt make this longer than needed.

```suggestion
            // Store the result before completion so caller cancellation cannot orphan a connection and completed attempts have a published result.
```

appsync/aws-sdk-appsync-events/src/test/java/com/amazonaws/sdk/appsync/events/EventsWebSocketProviderTest.kt
```diff
@@ -142,4 +149,62 @@ class EventsWebSocketProviderTest {
...
+            StandardTestDispatcher(testScheduler)
+        )
+        val firstConnectGate = CompletableDeferred<Unit>()
+        every { anyConstructed<EventsWebSocket>().isClosed } returns false
+        coEvery { anyConstructed<EventsWebSocket>().connect() } coAnswers { firstConnectGate.await() }
+
+        val firstAttempt = async { provider.getConnectedWebSocket() }
+        runCurrent()
+
+        // Closing the provider cancels the in-progress attempt.
```
mattcreaser: Suggestion: Remove the test-name repetition

The test name already states that close cancels the in-progress attempt, and the next line performs the close, so this comment adds no sequencing detail or rationale.

PR #3442, opened 2026-09-15
fix(api): echo a server-initiated close on the subscription socket

mattcreaser:
- [x] PR title and description conform to [Conventional Commits](https://www.conventionalcommits.org)

Stacked on #3434.

## What this changes

`AppSyncWebSocket` overrode `onClosed` but not `onClosing`, so when AppSync initiates the close the client never echoed the close frame. The socket stayed half-open, `onClosed` never arrived, and closure was only noticed when the keep-alive watchdog fired — up to the advertised `connectionTimeoutMs`, commonly five minutes, with every subscription on that connection waiting.

Measured on the new test: **300.158s before, 0.015s after.**

The fix is one commit (`f8b30bf2`) and is cleanly separable if you'd rather it went in on its own.

## Why the tests found it and the existing ones could not

`AppSyncWebSocketTest` drives `WebSocketListener` callbacks against a mocked `OkHttpClient`. A mock never sends a close frame, so `onClosing` was never invoked and the defect was invisible. The same blind spot let the missing connection-`host` entry reach a live endpoint last week.

So this adds coverage over a **real socket**:

`AppSyncWebSocketConnectionTest` — 8 unit tests against MockWebServer over a genuine `ws://` connection: handshake with a server-pushed `connection_ack`, **an assertion on the actual upgrade request's headers** (the class of bug that shipped), pushed keep-alive and data frames reaching `messages`, handshake timeout against a silent server, `connection_error`, and both closure directions.

A side finding: server-pushed WebSocket frames **do** work under Robolectric on MockWebServer 5.4.0. `AppSyncWebSocketTest`'s KDoc claimed the opposite; that claim predated #3426 and is corrected here. Real-socket coverage therefore needs no device.

## Instrumented tests — added, and NOT yet run

`AmplifyAppSyncClientInstrumentationTest` covers query, mutation, and a subscription observing a real server push, under API key and User Pools. It compiles. **It has never been executed** — there is no provisioned backend, so `amplify_outputs.json` is absent and every test skips via `assumeTrue`. I make no claim that these four pass against a real API. In particular the documents assume a `Todo { content }` schema and a specific `amplify_outputs.json` shape, both unverified.

To run them, place `amplify_outputs.json` (and `credentials.json` for the User Pools test) in `aws-appsync/src/androidTest/res/raw/`. Both are gitignored.

## Verification

303 unit tests, 0 failures (295 before). `testDebugUnitTest`, `compileDebugAndroidTestKotlin`, `apiCheck` and `ktlintCheck` all pass. No API delta — everything touched is `internal`.

## Possible overlap

#3313 and #3314 are also adding AppSync client e2e tests. I have not diffed against them; the instrumented tests here may duplicate that work.


aws-appsync/src/main/java/com/amazonaws/appsync/AppSyncWebSocket.kt
```diff
@@ -238,6 +238,14 @@ internal class AppSyncWebSocket(
         handleClosed()
     }
 
+    override fun onClosing(webSocket: WebSocket, code: Int, reason: String) {
+        // OkHttp does not answer a peer's close frame on its own. Without echoing it the socket stays
+        // half-open, onClosed never arrives, and closure is only noticed when the keep-alive watchdog
+        // fires — leaving every subscription on this connection waiting minutes for a close the service
+        // already sent.
+        webSocket.close(NORMAL_CLOSURE, null)
+    }
```
mattcreaser: Preserve abnormal peer-close failures

Only close code 1000 is a clean remote shutdown. Completing every peer-initiated close without recording its code or reason publishes an abnormal close such as 1011 as a normal closed state, so active subscription flows complete instead of reporting a connection failure. Record a pending AppSyncConnectionException for abnormal codes before acknowledging the close.

```suggestion
    override fun onClosing(webSocket: WebSocket, code: Int, reason: String) {
        if (pendingCloseCause == null && code != NORMAL_CLOSURE) {
            pendingCloseCause = AppSyncConnectionException(
                message = "The subscription connection closed with code $code: $reason"
            )
        }
        webSocket.close(NORMAL_CLOSURE, null)
    }
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/AmplifyAppSyncClientInstrumentationTest.kt
```diff
@@ -0,0 +1,280 @@
...
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpClass() {
+            val context = ApplicationProvider.getApplicationContext<Context>()
+            config = readAppSyncConfig(context)
+            credentials = readAppSyncCredentials(context)
+
+            val outputsId = context.resources.getIdentifier("amplify_outputs", "raw", context.packageName)
+            if (outputsId == 0) return
```
mattcreaser: Read androidTest resources from the instrumentation package

ApplicationProvider returns the target application context, but files under src/androidTest/res/raw ship in the instrumentation test APK. These lookups therefore never see the provisioned files, config stays null, and the new live-endpoint tests skip even on a configured lane. Resolve these resources through the instrumentation package context and keep the target application context where Amplify needs it.

aws-appsync/src/androidTest/java/com/amazonaws/appsync/testutils/AppSyncTestConfig.kt
```diff
@@ -0,0 +1,75 @@
...
+}
+
+private fun readRawJson(context: Context, name: String): JSONObject? {
+    val id = context.resources.getIdentifier(name, "raw", context.packageName)
+    if (id == 0) return null
+    return try {
+        val text = context.resources.openRawResource(id).bufferedReader().use { it.readText() }
+        JSONObject(text)
+    } catch (error: Exception) {
+        null
```
mattcreaser: Do not turn malformed provisioned configuration into a skipped suite

Once the resource ID resolves, an unreadable or malformed file means the configured test lane is broken and setup should fail. Swallowing every read and JSON exception returns null, which reaches the Assume checks and reports the live tests as skipped, so stale or corrupt CI provisioning stays green indefinitely.

```suggestion
private fun readRawJson(context: Context, name: String): JSONObject? {
    val id = context.resources.getIdentifier(name, "raw", context.packageName)
    if (id == 0) return null
    val text = context.resources.openRawResource(id).bufferedReader().use { it.readText() }
    return JSONObject(text)
}
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/AmplifyAppSyncClientInstrumentationTest.kt
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * Integration tests for [AmplifyAppSyncClient] against a real, deployed AppSync GraphQL API.
+ *
+ * Nothing here is mocked, deliberately. What these cover is what only a live endpoint can settle: the
+ * WebSocket upgrade AppSync actually accepts, the connection and per-subscription authorization
+ * objects it actually validates, and a genuine server-originated data frame. A mocked transport can
+ * confirm the client's own bookkeeping but not that the service agrees with it — a connection
+ * authorization object missing its `host` entry, for instance, satisfies every unit test and is
+ * rejected by every non-IAM auth mode on the real service.
```
mattcreaser: Do not claim the host omission passes every unit test

The new real-socket unit test asserts the host header, so a missing host no longer satisfies every unit test. State that live tests prove service compatibility instead. The comment is also considerably longer than the contract it documents.

```suggestion
 * These tests verify that a deployed endpoint accepts the WebSocket upgrade, connection and
 * subscription authorization, and a server-originated frame. Callback-driven tests cannot prove
 * service compatibility.
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/testutils/AppSyncTestConfig.kt
```diff
@@ -0,0 +1,75 @@
...
+    return AppSyncTestConfig(
+        endpoint = endpoint,
+        region = region,
+        apiKey = data.optString("api_key").takeIf { it.isNotBlank() }
+    )
+}
+
+/**
+ * Reads the last user in the `credentials` resource, or null when the resource is absent.
+ */
```
mattcreaser: Document every null return from the credentials loader

The helper also returns null for a missing or empty array, a non-object entry, and blank fields. Documenting only the absent-resource case leaves the nullable return contract materially incomplete.

```suggestion
/** Returns the last complete user in `credentials`, or null when the resource is absent or invalid. */
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncWebSocketConnectionTest.kt
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Tests [AppSyncWebSocket] over a **real** WebSocket connection to a local server.
+ *
+ * The distinction from [AppSyncWebSocketTest] is what is exercised, not what is asserted.
+ * [AppSyncWebSocketTest] invokes the [WebSocketListener] callbacks itself, so no socket is ever
+ * opened and no HTTP upgrade is ever performed — which means it cannot see the upgrade request the
+ * client actually sends. These tests complete a genuine HTTP upgrade and receive genuine
+ * server-pushed frames, so they cover the two things that mocking hides: the contents of the
+ * upgrade request, and whether an unprompted server frame is received at all.
```
mattcreaser: Distinguish request inspection from wire verification

AppSyncWebSocketTest captures the Request passed to newWebSocket and asserts its headers, so it can inspect the constructed upgrade request. What this class adds is server receipt of that request and real transport behavior.

```suggestion
 * [AppSyncWebSocketTest] scripts callbacks and inspects the request passed to OkHttp. These tests
 * verify the server receives that request and pushes frames over a real connection.
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncWebSocketConnectionTest.kt
```diff
@@ -0,0 +1,267 @@
...
+        socket.isClosed shouldBe false
+
+        socket.disconnect()
+    }
+
+    @Test
+    fun `the upgrade request carries the host entry the service authorizes against`() = runSocketTest {
+        // The regression this exists for: a missing `host` entry made every non-IAM auth mode fail at
+        // connection time. It is invisible to a mocked client, because nothing there ever inspects the
+        // request that would have gone out.
```
mattcreaser: Correct the mocked-request coverage claim

The callback-driven test captures the request passed to OkHttp and inspects its headers. It missed this regression because it did not assert host, not because a mocked client cannot inspect the request.

```suggestion
        // The mocked request assertion missed this regression because it did not check `host`; this
        // server verifies the value received over the wire.
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncWebSocketTest.kt
```diff
@@ -37,12 +37,10 @@ import org.robolectric.RobolectricTestRunner
...
- * This is the pattern the rest of the repo uses for WebSocket unit tests — `EventsWebSocketTest` mocks
- * `OkHttpClient` and `WebSocket` outright, and `LivenessWebSocketTest`'s server-side listener is
- * entirely empty because it invokes the client listener itself. Neither relies on MockWebServer
- * delivering server-to-client frames; I tried that first and the frames never arrive (the upgrade
- * succeeds with a 101, then the connection resets). Real-socket behaviour is covered by instrumented
- * tests instead.
+ * No socket is opened here, which is what lets these tests script exact callback orderings a real
+ * server would not reliably produce. It also means they cannot see the upgrade request the client
+ * actually sends, or whether a server-pushed frame arrives at all — [AppSyncWebSocketConnectionTest]
+ * covers those over a real connection.
```
mattcreaser: Correct the callback test's stated scope

This class captures the Request passed to newWebSocket and asserts its headers, so it can inspect the constructed upgrade request. Scope the stated difference to server receipt and server-pushed frames.

```suggestion
 * No socket is opened, allowing deterministic callback ordering. [AppSyncWebSocketConnectionTest]
 * verifies the server receives the upgrade request and can push frames over a real connection.
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/AmplifyAppSyncClientInstrumentationTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        )
+    }
+
+    /**
+     * Subscribes, then mutates, and asserts the subscriber observes the mutation.
+     *
+     * The broadest thing in the suite: reaching the final assertion requires the WebSocket upgrade to
+     * be accepted, the connection authorization object to be validated, the `start`/`start_ack` round
+     * trip to complete, and the service to originate a frame of its own accord.
+     */
```
mattcreaser: Suggestion: Remove the helper's narrated test steps

The summary restates the function name and implementation, and the useful coverage contract is buried in the second paragraph. Keep only a direct statement of what reaching the assertion verifies.

```suggestion
    /**
     * Reaching the final assertion verifies the WebSocket upgrade, connection authorization,
     * `start`/`start_ack` exchange, and a server-originated frame.
     */
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/testutils/AppSyncTestConfig.kt
```diff
@@ -0,0 +1,75 @@
...
+import android.content.Context
+import org.json.JSONObject
+
+/**
+ * The deployed API's coordinates, read from the `amplify_outputs` raw resource.
+ *
+ * @param endpoint The GraphQL endpoint URL.
+ * @param region The API's AWS region.
+ * @param apiKey The API key, absent when the deployed API does not issue one.
+ */
```
mattcreaser: Suggestion: Drop obvious parameter descriptions

The endpoint and region entries restate their names and types. One line can retain the resource source and the only non-obvious contract, API-key nullability.

```suggestion
/** Test API coordinates loaded from `amplify_outputs`; [apiKey] is null when unavailable. */
```

aws-appsync/src/androidTest/java/com/amazonaws/appsync/testutils/AppSyncTestConfig.kt
```diff
@@ -0,0 +1,75 @@
...
+internal data class AppSyncTestCredentials(val username: String, val password: String)
+
+/**
+ * Reads the deployed API's coordinates, or returns null when the resource is absent or does not
+ * describe a GraphQL API.
+ *
+ * Returning null rather than throwing is what lets the tests skip: the resource is gitignored, so it
+ * is missing on any machine that has not provisioned a backend, including CI.
+ */
```
mattcreaser: Suggestion: State the nullable configuration contract directly

The second paragraph restates the same nullable return contract and repeats setup context from the test class KDoc. One sentence preserves the absent-resource, invalid-API, and skip behavior.

```suggestion
/** Returns deployed API coordinates, or null so tests can skip when `amplify_outputs` is absent or lacks a GraphQL API. */
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncWebSocketConnectionTest.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Runs a test body on real threads.
+     *
+     * `runTest` is wrong here: its virtual clock advances as soon as every coroutine is idle, and a
+     * coroutine awaiting real network I/O looks idle — so the handshake timeout would fire while the
+     * socket was still connecting. Typed to Unit so a test body cannot accidentally return a value,
+     * which JUnit rejects.
+     */
```
mattcreaser: Suggestion: Condense the real-thread test rationale

The virtual-time warning and the Unit return constraint are worth keeping; the lead-in and the expanded scenario are not.

```suggestion
    /**
     * Uses real threads because `runTest` advances virtual time while network I/O is pending.
     * The Unit return type prevents invalid JUnit test results.
     */
```

aws-appsync/src/test/java/com/amazonaws/appsync/AppSyncWebSocketConnectionTest.kt
```diff
@@ -0,0 +1,267 @@
...
+            decorator = AppSyncRequestDecorator("us-east-1"),
+            client = OkHttpClient()
+        )
+
+    // ── Handshake ───────────────────────────────────────────────────────
+
+    @Test
+    fun `the handshake completes over a real socket`() = runSocketTest {
+        // The ack here is genuinely pushed by the server over the wire, unprompted by any read on the
+        // client's side. Nothing but a real connection demonstrates that path works.
```
mattcreaser: Suggestion: Remove the repeated server-push narration

The test name, the enqueueServer call, and the class KDoc already establish that the acknowledgement is server-pushed over a real socket. This comment adds no contract, caveat, or rationale.

PR #3450, opened 2026-09-18
chore: enable the Gradle configuration cache

mattcreaser:
- [x] PR title and description conform to [Conventional Commits](https://www.conventionalcommits.org)

Every Gradle invocation re-ran the configuration phase across the whole project before any task executed. This serializes the configured task graph and reuses it when the build scripts have not changed.

This is a different mechanism from `org.gradle.caching`, which is already enabled: the build cache reuses **task outputs**, the configuration cache reuses the **configured task graph**. They compose rather than overlap, which is why the flag sits next to it.

## Measured

Warm static analysis — the same command the static-analysis PR Verification runs:

| | time |
|---|---|
| before | 3.5s |
| after, entry reused | **1.2s** |

The saving applies to **every** invocation, including those where the build cache already makes all tasks `UP-TO-DATE`, because configuration happens before any task-level cache is consulted.

## Verified

`Configuration cache entry stored` with **no problems reported** across:

- `ktlintCheck checkstyle apiCheck` (the CI `static-analysis` job's command)
- `:core:compileDebugKotlin`, `:aws-api:compileDebugKotlin`
- `:aws-api:testDebugUnitTest`

## Not verified

I have not exercised `lint`, `connectedAndroidTest`, or the release/publishing tasks. Those are the likeliest places a latent incompatibility would surface, since the configuration cache is stricter about build logic that reads mutable state at execution time and this project has custom convention plugins in `build-logic/`.

Nor have I checked whether CI benefits. `gradle/actions/setup-gradle` caches by content hash, and the configuration cache is stored under `.gradle/configuration-cache`; whether that path is in the cached set needs confirming separately. This PR is worthwhile for local builds regardless.

Gradle reports an incompatibility as a build failure naming the offending task, so a problem here surfaces loudly rather than silently degrading.


thisisabhash (dismissed):

mattcreaser:
Need to look into why the scope verification failed here

